PR #4, opened 2016-01-31
newCachedThreadPool Perhaps more suitable  ExecutorEngine

Yaccc:


terrymanu:
目前可以根据连接数算出线程数，所以采用固定线程池更合适些

Yaccc:
I overlooked the fact that the timeout period of 60s is newCachedThreadPool. In the case of knowing that there are indeed several connections, the resources will be wasted.

PR #101, opened 2016-06-06
update OrVisitor

jinfei21:
fix bug for case：sql = "select * from order where (a=1 or b=2) and (c=3 or d=4)";


hanahmily:
@coveralls 谢谢你的提交。下次提交时，注意提交一个测试用例，否则会导致覆盖率不足哦

PR #397, opened 2017-10-04
bugfix: yaml can't resolve non-string type value

liolay:
when use yaml to config datasource, non-string type value can't resolve correctly


terrymanu:
请详细说明提交pr的目的，谢谢

PR #421, opened 2017-10-22
Fix incorrect usages of org.mockito.internal

TimvdLippe:
Hey! I saw you were depending on several `org.mockito.internal` classes. As we (the Mockito core developers) do not guarantee backwards compatibility on these classes, I converted them for you to the equivalent public API. Probably your IDE was suggesting the wrong import upon autocompletion.


terrymanu:
Thank you for reminder and  the pull request. We will pay attention to avoid to use mockito internal API in future.

PR #642, opened 2018-03-12
新测试工具Master

qiankunshe:
io.shardingjdbc.dbtest.StartTest 为用例启动类
integrate/env.properties 中存放数据库配置，系统是根据数据库配置来测试数据库的，如，现在配置了mysql，那么系统只测试mysql
asserts 下为用例xml

-
-
-


terrymanu:
Thank you for your contribution. I reviewed the codes and give some suggestion:

1. Please revise your code reference code convention: http://shardingjdbc.io/community/en/02-contribute/convention/
2. Please review your code reference checkstyle, the template is in sharding-jdbc/src/resources/dd_checks.xml.
3. Please remove all commits, javadoc(test case do not need javadoc)  and all Chinese words.
4. Please fill all SQL type for your assert.xml to make sure test engine is correct, for example: DQL,DQM,DDL,TCL,DAL
5. Missing some assert at parse context(for example missing condition assert here), please check SQLStatement and his child class to make sure all properties we need to assert.
6. Still need to consider about how to assert DML,DDL, TCL and DAL. expected-data-file only for DQL and DML.
7. Finally, fill all test case for xml (can do it after merged).

PR #649, opened 2018-03-16
fix 648

### LINKED ISSUE #648: SelectItem should be annotated by EqualsAndHashCode
I found that there are several subclasses implement  `SelectItem`. They are used as container elements in  `Set` collection(io.shardingjdbc.core.parsing.parser.sql.dql.select.SelectStatement#items), but I only see the `AggregationSelectItem` has `EqualsAndHashCode` annotation. the others are forgotten?

hongfuli:
Fixes #648


terrymanu:
Thank you for contribute codes

PR #698, opened 2018-04-03
支持直接通过注册中心读取配置

ma-xiao-guang-64:
fixed [537](https://github.com/shardingjdbc/sharding-jdbc/issues/537)


terrymanu:
Some suggestion for code reivew:

1.  For code conventions, break rule for `Indent (including blank lines) is consistent with the previous line.`

2. No test case, cloud you add some test case for them

3. We want to know how to use it for end-users. Could you also provide a pull request for example?  (https://github.com/shardingjdbc/sharding-jdbc-example/tree/dev/sharding-jdbc-orchestration-example & https://github.com/shardingjdbc/sharding-jdbc-example/tree/dev/sharding-jdbc-orchestration-spring-example)

siusn:
This function is on 3.x, but not on 2.x

PR #703, opened 2018-04-08
实现Or的准备工作

ma-xiao-guang-64:
[相关issue](https://github.com/shardingjdbc/sharding-jdbc/issues/501)
1、修改Conditions中conditions的类型
2、修改Conditions中add方法默认像第一个and conditions添加condition
3、修改Conditions中find方法默认获取第一个and conditions分片键相同的最后一个condition
4、修改测试用例

### DIFF
--- a/sharding-core/src/main/java/io/shardingjdbc/core/parsing/parser/context/condition/AndConditions.java
+++ b/sharding-core/src/main/java/io/shardingjdbc/core/parsing/parser/context/condition/AndConditions.java
@@ -1,0 +1,97 @@
+/*
+ * Copyright 1999-2015 dangdang.com.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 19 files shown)


terrymanu:
Code review suggestion:
in class `Conditions`, use a object to instead of 'List<List<Condition>> orConditions', such as `List<OrConditions>`, and rename `Condition` to `AndCondition` is better?

terrymanu:
Code review suggestions:

1. Class `OrConditions` and `AndConditions` should be final.
2. Line 63 of `OrConditions`, add final for the second parameter index.
3. Line 63 of `OrConditions`, maybe can simplify:

```java
public Optional<Condition> find(final Column column, final int index) {
        Optional<AndConditions> andConditions = get(index);
        return andConditions.isPresent() ? andConditions.get().find(column) : Optional.<Condition>absent();
    }
```

4. Line 78 of `OrConditions`, add final for  parameter index.
5. Line 87 and 89 of `OrConditions`, javadoc is incorrect, should be: Adjust and conditions is empty or not.
6. Line 31 of `Conditions`, use @NoArgsConstructor instead of @RequiredArgsConstructor.
7. Line 56 of `AndConditions`, cloud you just return if you find the result, maybe do not need loop all remain elements.
8. Line 72 of `AndConditions`, add final for  parameter index.
9. Line 81 and 83 of `AndConditions` , javadoc is incorrect, should be: Adjust conditions is empty or not.
10 . Line 61 of `ConditionAssert`, use `private` instead of `public`.
11. Line 56 and 65 of `ConditionAssert`,  before assert optional.get, use assertThat(isPresent) first.
12. Please use English for your pull request title, content and commit comments.

PR #726, opened 2018-04-14
Dev xiaoyu

yu199195:
Fixes #ISSUSE_ID.

-
-netty some optimize
-


terrymanu:
Code review suggestions:
1. Please follow code convenient, such as indent, for example: FrontendHandlerFactory line 33
2. Please make sure checkstyle can pass, for example: MySqlFrontendHandler line 41 and line 42, should only have one blank line.
3. Please don't rename ShardingProxy to ShardingNettyProxy
4. Please don't rename MySQLFrontendHandler to MySqlFrontendHandler
5. MySqlFrontendHandler line 48, use lombok @RequiredArgsConstructor instead of it.
6. authPluginData of MySqlFrontendHandler is an important field which should use for auth, should not move to a local variable at method handshake.

PR #799, opened 2018-05-05
modify strategy of loading and refresh sharding metadata

cherrylzhao:
Fixes sharding-jdbc-example #22

1. judge actual table exists when build shardingDatasource
2. refresh tableMetaData after sharding DDL operation


terrymanu:
Some suggestions:

1. Please make sure travis can pass all test cases before submit pull request.
2. Please don't comment any codes, just remove if code is useless. Comment codes is a temporary behaviour, should be forbided.

terrymanu:
Code review suggestions:
Please avoid use import *, such as:

```java
import java.sql.*;
```

Please use checkstyle check your code before submit pull request.

PR #820, opened 2018-05-10
Robustness for sharding-proxy

### LINKED ISSUE #818: Robustness for sharding-proxy
Make sharding-proxy a more robust system, and log errors with full stack trace.

tuohai666:
Fixes #818 .

1. remove dependency of sharding-jdbc-orchestration
2. any config faults will fire an exception at the startup phase, and lead to a shutdown of proxy
3. wrong SQL will get a prompt instead of being blocked


tuohai666:
I'm at home now, and I'll deal with this failure tomorrow

PR #856, opened 2018-05-21
 fix opentracing test sequence bug

beckhampu:


tuohai666:
Hi beckhampu, smart solution! However, I think you break the indent, not 4 white spaces no.

terrymanu:
Because we will release version now, so I will do it after merged this time. Please make sure all rule of checkstyle can be passed

beckhampu:
Ok, thanks for reminding me. I'll pay attention next time

PR #933, opened 2018-06-14
infinite looping is not necessary

tuohai666:


tuohai666:
tests:
1. example, CLI, GUI
2. performance

Issue #990, opened 2018-07-09
Introduce the SPI for the transaction manager in the sharding-sphere


zhfeng:
I want to make a proposal to introduce the SPI to allow the transaction manager integration to be pluggable that the different implementation could be used. It could be also used for the XA and the SAGA transaction.

For the XA transaction, it could be
```
void begin()             // start a XA transaction
void commit()          // commit to the TM
void rollback()         // rollback to the TM
```

For the SAGA transaction, it could be
```
void start()              // start a saga transaction
void close()            // close a saga transaction, that means all the sub transaction can be completed
void cancel()          // cancel a saga transaction, that means all the sub transaction can be cancelled
```
any thought ?

cherrylzhao:
that's a great suggest！
we can implement `Atomikos`, `narayana`, `SAGA` transaction manager based on the common SPI.
let's do the detail SPI design together.

cherrylzhao:
@zhfeng 
we have defined XA SPI with class `io.shardingsphere.transaction.api.TransactionManager`.
AtomikosTransactionManager have been refactored in this way.

now, we'd like to discuss how to integrate with `narayana` transaction manager.
Atomikos provided a wrapper datasource to enlist and delist resource into transaction manager while invoking jdbc interface.
I haven't found a equivalence implement in `narayana-jta`，would you please give me some hint？

zhfeng:
Narayana does not have the wrapper for the datasource but it has the transactional driver for doing the similar things. I will investigate and provide the implementation from the Narayana side.

Thanks for your contribution to introduce the XA SPI and it could be great work.

cherrylzhao:
@zhfeng 
thanks for your support, waiting for your goods news.
it looks like Narayana can be integrated with `ManagedDataSource` directly.

[DbcpXADataSourceWrapper](https://github.com/jbosstm/quickstart/blob/master/spring/narayana-spring-boot/src/main/java/org/springframework/boot/jta/narayana/DbcpXADataSourceWrapper.java)

zhfeng:
I just open https://github.com/sharding-sphere/sharding-sphere/pull/1137 to add the warpDataSource method in the XATransactionManager interface which is very similar with what we have done in the spring-boot/jta. Also I refactor the JDBCXABackendDataSourceFactory to build the data source from the wrapDataSource.

[...]

PR #1068, opened 2018-07-30
add sharding-transaction module, refactor transaction manager. for #990

cherrylzhao:
add sharding-transaction module, refactor transaction manager. for #990


cherrylzhao:
publish transaction modification in 3.1.0

PR #1069, opened 2018-07-30
 make PreparedStatementRegistry thread safe

tuohai666:
Changes:
1. make PreparedStatementRegistry thread safe.
2. combine SQL and parameters.
3. parameters repeatly updating will be prevented by jdbc mechanism automatically.

Tests:
example/CLI/GUI


terrymanu:
Name of SQLElement should be a better name, such as PrepareStatemntUnit etc

PR #1137, opened 2018-08-13
update to add the wrapDataSource in the XATransactionManager

zhfeng:


terrymanu:
Thank you, codes are great. I just merge them now.

PR #1149, opened 2018-08-15
Fixed AbstractStatementAdapter.getUpdateCount return -1 when use orac…

zhangkewei:
In oracle ojdbc driver, getUpdateCount can only be called once after a successful call.
Otherwise, this method return -1 by default.


terrymanu:
thank you very much

PR #1187, opened 2018-08-26
#1186, get connections synchronized in MEMORY_STRICTLY mode

### LINKED ISSUE #1186: Get connections synchronized in MEMORY_STRICTLY mode
In MEMORY_STRICTLY mode to avoid dead lock in connection pool, we must meet the condition:
**ds.maximumPoolSize >= acceptor.size * ds.table_count.**

If we get all connections needed synchronized between queries, we just need to meet the condition:
**ds.maximumPoolSize >= ds.table_count.**

According to the performance test, there's little degrade of the QPS.

tuohai666:
Fixes #1186 .


terrymanu:
just close it, we need separate create connection and create statement

PR #1362, opened 2018-10-22
Add the init and destroy method in the XATransactionManger

zhfeng:
Fixes https://github.com/sharding-sphere/sharding-sphere/issues/1357


sharding-transaction/src/main/java/io/shardingsphere/transaction/manager/xa/AbstractXATransactionManager.java
```diff
@@ -0,0 +1,79 @@
...
+import javax.transaction.HeuristicMixedException;
+import javax.transaction.HeuristicRollbackException;
+import javax.transaction.NotSupportedException;
+import javax.transaction.RollbackException;
+import javax.transaction.SystemException;
+import javax.transaction.UserTransaction;
+import java.sql.SQLException;
+
+@Slf4j
+public abstract class AbstractXATransactionManager implements XATransactionManager {
```
cherrylzhao: maybe AbstractXAtransactionManager only be used by AtomikosTransactionManager,
it think narayaTransactionManager was implemented in other repo and don't depend on SS,
so it seems that we should do init directly in AtomikosTransactionManager, and I think init() method can be removed.

sharding-transaction/src/main/java/io/shardingsphere/transaction/manager/xa/XATransactionManager.java
```diff
@@ -30,6 +30,17 @@
  * @author zhangliang
  */
 public interface XATransactionManager extends ShardingTransactionManager<XATransactionEvent> {
+
+    /**
+     * initialize the transaction manager
+     */
+    void init() throws Exception;
```
cherrylzhao: I think we can remove this init() method, and do initialize in constructor of AtomikosTransactionManager directly

sharding-transaction/src/main/java/io/shardingsphere/transaction/manager/xa/AbstractXATransactionManager.java
zhfeng: OK, it looks good to remove this abstract class.

sharding-transaction/src/main/java/io/shardingsphere/transaction/manager/xa/XATransactionManager.java
zhfeng: that makes sense.

cherrylzhao:
@terrymanu please merge this PR to dev-3.1.0

PR #1428, opened 2018-11-05
trim datasource names

dolores2018:
trim datasource names

Fixes #ISSUSE_ID.

-
-
-


terrymanu:
Can you provide the reason to trim this key? 
For my understanding, this key is from spring boot properties, between key and values is properties file, it don't permit space.

terrymanu:
ok, I get your point. thank you for contribute code

PR #1469, opened 2018-11-12
Support broadcast table

ma-xiao-guang-64:
fixed [1412](https://github.com/sharding-sphere/sharding-sphere/issues/1412)


terrymanu:
I want to discuss configuration of broadcast table. 

1. Consider about distribute PK, maybe is unnecessary. 
2. Consider about logic table name, broadcast table do not need shard tables, so maybe is unnecessary too.

ma-xiao-guang-64:
1.l use generator key to support insert on broadcast table,because database's auto increment strategy is not to ensure data consistency.
2.Logic table name may not be appropriate,it is used to record which tables are broadcast tables.

PR #1535, opened 2018-11-30
fix error mysql type mapping

dozer47528:
Based on [MySQL docs](https://dev.mysql.com/doc/internals/en/myisam-column-attributes.html),
The type code of Binary column should be `254`, (equals `-2`, `0xfe`).


sharding-proxy/src/test/java/io/shardingsphere/shardingproxy/transport/mysql/constant/ColumnTypeTest.java
```diff
@@ -25,7 +25,7 @@
 import static org.junit.Assert.assertThat;
 
 public final class ColumnTypeTest {
-    
+
```
tuohai666: The 4 white spaces in a empty line should not be removed.

dozer47528:
@tuohai666 
https://github.com/sharding-sphere/sharding-sphere/issues/1533
Related to this issue, there are two problems.

Compare to connecting MySQL directly, when I using `BINARY`, the result and MySQL type were different.

So I investigated the source code of sharding-proxy, after fixing this line, the MySQL type will match.
But I don't know how to fix the different result problem.

dozer47528:
Can not fix the problem by this way, will close this pull request.

PR #1542, opened 2018-12-01
use ANTLR parse PostgreSQL SET and RESET statment

loxp:
- add ANTLR PostgreSQL SET and RESET parsers.


loxp:
PR in #1543

PR #1544, opened 2018-12-02
add antlr sqlserver dcl test case

maxiaoguang64:
add antlr sqlserver dcl test case


codefairy08:
no HASHED Syntax  for (create | alter) user

Issue #1429, opened 2018-11-05
Performance enhancement of backend connect with database by NIO in sharding-proxy


solid210:
I have researched currently program in sharding-proxy where implements backend connection with database by NIO, and I have some advisement for performance enhancement.
To show my idea clearly, I draw up design plan.
### Legend
![Image of Legend](https://github.com/solid210/architecture/blob/master/sharding-sphere/nio/legend.png?raw=true)

### Current case
![Image of Current Case](https://github.com/solid210/architecture/blob/master/sharding-sphere/nio/current-case.png?raw=true)
Frontend and backend use different worker eventloop group(this means use different IO thread pool):

1. Application's request reaches worker eventloop group of frontend;
2. Then worker eventloop group writes data to user eventloop group, and the later processes protocol package; (Switch thread)
3. Then user eventloop group writes data to worker eventloop group of backend, and the later executes SQL(this means send protocol package to DB); (Switch thread)
4. Then query result from DB returned to work eventloop group of backend;
5. Then work eventloop group of backend writes data to user eventloop group; (Switch thread)
6. Then user eventloop group processes query result by blocking; 
7. Then user eventloop group writes the processed result to work eventloop group; (Switch thread)
8. Then work eventloop group writes back data to application.

Total 4 times switch thread in a query's life time.

### Case 1
![Image of Current Case](https://github.com/solid210/architecture/blob/master/sharding-sphere/nio/case-1.png?raw=true)
Frontend and backend use same worker eventloop group(this means use same IO thread pool):

1. Application's request reaches worker eventloop group;
2. Then worker eventloop group writes data to user eventloop group, and the later processes protocol package; (Switch thread)
3. Then user eventloop group writes data to worker eventloop group, and the later executes SQL(this means send protocol package to DB); (Switch thread)
4. Then query result from DB returned to work eventloop group;
5. Then work eventloop group of backend writes data to user eventloop group; (Switch thread)
6. Then user eventloop group processes query result by blocking; 
7. Then user eventloop group writes the processed result to work eventloop group; (Switch thread)
8. Then work eventloop group writes back data to application.

Advantages: Reduce a thread pool.
Maybe only a little performance enhancement, but still 4 times switch thread in a query's life time.

### Case 2
![Image of Current Case](https://github.com/solid210/architecture/blob/master/sharding-sphere/nio/case-2.png?raw=true)
[...]

PR #1545, opened 2018-12-02
Use same worker thread group for frontend connecting with client and …

solid210:
Use same worker thread group for frontend connecting with client and backend connecting with DB.
This PR is the first step for #1429.


solid210:
I submitted the pull request by choosing branch to branch, but this way is not good. 
Because after I merging code from upstream to the branch, github checks my code changing in the branch and adds the codes that are changed in the pull request automatically.

I close this pull request and will submit a new pull request by choosing commit id, this way will not affects me merging code.

PR #1551, opened 2018-12-03
Use same worker thread group to handle IO event for  frontend and backend.

solid210:
1. Use same worker thread group for frontend connecting with client and backend connecting with DB.

2. Replace userGroup(class is EventLoopGroup) with commandExecutorService (class is ExecutorService), because EventLoopGroup belongs to Netty, but CommandExecutor only runs in a thread pool, CommandExecutor has no relation with Netty.

### DIFF
--- a/sharding-proxy/src/main/java/io/shardingsphere/shardingproxy/frontend/ShardingProxy.java
+++ b/sharding-proxy/src/main/java/io/shardingsphere/shardingproxy/frontend/ShardingProxy.java
@@ -35,6 +35,14 @@ import io.shardingsphere.shardingproxy.backend.BackendExecutorContext;
 import io.shardingsphere.shardingproxy.backend.netty.client.BackendNettyClientManager;
 import io.shardingsphere.shardingproxy.frontend.common.netty.ServerHandlerInitializer;
 import io.shardingsphere.shardingproxy.runtime.GlobalRegistry;
+import lombok.AccessLevel;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
 
 /**
  * Sharding-Proxy.
@@ -44,7 +52,10 @@ import io.shardingsphere.shardingproxy.runtime.GlobalRegistry;
  * @author wangkai
  * @author panjuan
  */
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
 public final class ShardingProxy {
+    
+    private static final ShardingProxy INSTANCE = new ShardingProxy();
     
     private static final GlobalRegistry GLOBAL_REGISTRY = GlobalRegistry.getInstance();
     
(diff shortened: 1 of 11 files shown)


solid210:
First, in Netty, channel is completely asynchronous. One IO thread group can process thousands of channels' IO event. So there is not need for frontend's channels and backend's channels use different IO thread group, because one can do this enough.

Second, about case 2 of #1429, to reduce unnecessary thread switches. An inbound message is passed from a channel of frontend to a channel of backend will in the same IO thread, vice versa, if SQL result needn't to be merged. These IO threads will be contained in the thread pool which is combined of this PR. (I will submit another PR to implement the logic)

Thanks!

tuohai666:
Great! 
I have no more questions.

sharding-proxy/src/main/java/io/shardingsphere/shardingproxy/frontend/ShardingProxy.java
```diff
@@ -64,20 +85,21 @@
      */
     public void start(final int port) throws InterruptedException {
         try {
-            if (GLOBAL_REGISTRY.getShardingProperties().<Boolean>getValue(ShardingPropertiesConstant.PROXY_BACKEND_USE_NIO)) {
-                BackendNettyClientManager.getInstance().start();
-            }
             ServerBootstrap bootstrap = new ServerBootstrap();
             bossGroup = createEventLoopGroup();
+            commandExecutorService = createCommandExecutorService();
```
cherrylzhao: if we don't use userGroup, but define the customer ExecuteService in shardingProxy, it seems not a good idea. as we have discussed in wenxin chat, sharingProxy is only netty relevant variable.

sharding-proxy/src/main/java/io/shardingsphere/shardingproxy/frontend/ShardingProxy.java
```diff
@@ -87,6 +109,11 @@ public void start(final int port) throws InterruptedException {
         }
     }
     
+    private ExecutorService createCommandExecutorService() {
+        int nThreads = GLOBAL_REGISTRY.getShardingProperties().<Integer>getValue(ShardingPropertiesConstant.ACCEPTOR_SIZE);
+        return new ThreadPoolExecutor(nThreads, nThreads, 0, TimeUnit.MILLISECONDS, new LinkedBlockingQueue<Runnable>(16));
```
cherrylzhao: we should use excecuteService like ShardingExecuteEngine way, also make blockingQueue was fixed at 16 is not a good idea.

solid210: It's a good idea to use ShardingExecuteEngine, I will follow the suggestion.

sharding-proxy/src/main/java/io/shardingsphere/shardingproxy/frontend/ShardingProxy.java
solid210: Replace commandExecutorService with CommandExecutorContext which one likes BackendExecutorContext.

sharding-core/src/main/java/io/shardingsphere/core/executor/ShardingExecuteEngine.java
```diff
@@ -53,6 +53,15 @@ public ShardingExecuteEngine(final int executorSize) {
         MoreExecutors.addDelayedShutdownHook(executorService, 60, TimeUnit.SECONDS);
     }
     
+    /**
```
cherrylzhao: it seem that we should extract ListeningExecutorService to a Util class for common use, as CommandExecutorContext will not use other feature in ShardingExecuteEngine.

PR #1583, opened 2018-12-07
support java.sql.PreparedStatement.setXXX() methods whose argument do…

fifadxj:
…es not implement Serializable interface, such as setCharacterStream()

Fixes #ISSUSE_ID:1567

-
-update AbstractShardingPreparedStatementAdapter
-


terrymanu:

PR #1611, opened 2018-12-13
Revert "use maven shade"

terrymanu:
Reverts sharding-sphere/sharding-sphere#1605


terrymanu:
Use maven shade cannot compile with IDEA, need to release shaded jar to maven centre repo.
So we decide if user find guava conflict, should shade by themselves.

PR #1679, opened 2018-12-26
sharding config changes from v3.0.0 to v3.1.0

amanxu:
sharding.jdbc.config.sharding.props.sql.show= ...
sharding.jdbc.config.sharding.props.executor.size= ...
**Above config is valid but that is invalid in  v3.1.0.
When  i check the source code,I find the code have been changed and Official documents is not modified.**
##########################################################
V3.0.0：
**@ConfigurationProperties(prefix = "sharding.jdbc.config.sharding")**
public class SpringBootShardingRuleConfigurationProperties extends YamlShardingRuleConfiguration {
}

V3.1.0
**@ConfigurationProperties(prefix = "sharding.jdbc.config")**
public class SpringBootConfigMapConfigurationProperties {
}

### DIFF
--- a/sharding-core/src/main/antlr4/imports/BaseRule.g4
+++ b/sharding-core/src/main/antlr4/imports/BaseRule.g4
@@ -179,7 +179,7 @@ domainNames
     : domainName (COMMA domainName)*
     ;
     
-tableNamesWithParen
+tableList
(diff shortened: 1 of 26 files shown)


terrymanu:
Hi, why you submit pull request for other people?

PR #1938, opened 2019-02-27
fix return decimal column with an exponent field

avalon566:
- fix return decimal column with an exponent field

```
BigDecimal data = new BigDecimal(BigInteger.ZERO, 20);
System.out.println(data.toString()); // output 0E-20
System.out.println(data.toPlainString()); // output 0.00000000000000000000
```


tuohai666:
Thanks for your effort. I'm thinking deeply into this pr now.
I wonder what problem have you met? Do you have a test report?

tuohai666:
I noticed your merge branch operation, but the conflicts still exist.

I just finished a test:

There are two rows in MySQL server which values:
Row 1:     0.00000000000000000000
Row 2:     0E-20

Select from MySQL server:
0
0

Select from Sharding-Proxy:
0
0

As I asked just now, what problem have you met?

avalon566:
a column in table define as decimal(25,20) and store value 0.when i select that column,it returns 0E-20,mysql driver can not parse it.

tuohai666:
@avalon566 I got it.

PR #1945, opened 2019-02-28
support explain command

avalon566:
- support explain command,facilitate the users to locate problem


avalon566:
thanks,i will resolve these problem and create a new pr.

PR #1974, opened 2019-03-05
resolve resetLimitTokens removeLimitToken

acseed:
Fixes #ISSUSE_ID.

- delete list from foreach is not good
-
-

### DIFF
--- a/sharding-core/src/main/java/org/apache/shardingsphere/core/parsing/parser/sql/dql/select/SelectStatement.java
+++ b/sharding-core/src/main/java/org/apache/shardingsphere/core/parsing/parser/sql/dql/select/SelectStatement.java
@@ -38,6 +38,7 @@ import org.apache.shardingsphere.core.parsing.parser.token.SQLToken;
 import org.apache.shardingsphere.core.util.SQLUtil;
 
 import java.util.Collection;
+import java.util.Iterator;
 import java.util.LinkedHashSet;
 import java.util.LinkedList;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


acseed:
fix  在foreach 按照下标删除会导致越界

tuohai666:
1. Good idea to simplify the code, we decide to merge your pr. But the original code can not lead to a bug really, you can find out the reason by yourself.
2. Why not respond my comment before?

acseed:
> 1. Good idea to simplify the code, we decide to merge your pr. But the original code can not lead to a bug really, you can find out the reason by yourself.
> 2. Why not respond my comment before?
>    image

> it is not just simplify the code,  the original code do lead to a bug, the "toBeRemovedIndexes"  is the index to remove,  but after sqlTokens remove the first, the index in sqlTokens is changed, this may cause java.lang.IndexOutOfBoundsException,  when select sql has subQuery, the bug occurs, you try

acseed:
> > 1. Good idea to simplify the code, we decide to merge your pr. But the original code can not lead to a bug really, you can find out the reason by yourself.
> > 2. Why not respond my comment before?
> >    image
> 
> > it is not just simplify the code,  the original code do lead to a bug, the "toBeRemovedIndexes"  is the index to remove,  but after sqlTokens remove the first, the index in sqlTokens is changed, this may cause java.lang.IndexOutOfBoundsException,  when select sql has subQuery, the bug occurs, you try

it is my fault not to respond your comment before, sorry

PR #2057, opened 2019-03-19
#2009, remove sharding-proxy's dependency of mysql-connector-java

### LINKED ISSUE #2009: Apache release preparation
It seems like the todo list will increasing for several days. Thanks for @wu-sheng 's professional comments.
Code changes are in https://github.com/apache/incubator-shardingsphere/pull/2006

- [x] Add org.apache as parent in pom.xml.
- [x] Add apache-rat as the release audit tool.
- [x] Add Copyright for xml/yaml/properties/g4/... files.

tuohai666:
Fixes #2009.

### DIFF
--- a/sharding-proxy/sharding-proxy-backend/pom.xml
+++ b/sharding-proxy/sharding-proxy-backend/pom.xml
@@ -45,11 +45,6 @@
         </dependency>
         
         <dependency>
-            <groupId>mysql</groupId>
-            <artifactId>mysql-connector-java</artifactId>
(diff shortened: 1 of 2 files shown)


sharding-proxy/sharding-proxy-backend/pom.xml
```diff
@@ -44,11 +44,6 @@
             <scope>compile</scope>
         </dependency>
         
-        <dependency>
```
terrymanu: should not remove mysql, instead of provide is better

PR #2074, opened 2019-03-23
#1818 add sql.simple.length config

### LINKED ISSUE #1818: when the Logic SQL is very long, SQLLogger print too much. if the Logic SQL is Cross-sharding，that's worse
## Feature Request

when the Logic SQL is very long, SQLLogger print too much. if the Logic SQL is Cross-sharding，that's worse：
if there are 128 sharding table，there are 128 Actual SQL .
and SQLStatement and Logic SQL also will Output thousands of lines .

These outputs are of little value. So suggest that add a configuration(for example: **sharding.jdbc.config.props.sql.show.simple=true**) to simple it.

betterjava:
Fixes #1818 .

- add sql.simple.length for shard config
- In sharding mode, if the length of logic SQL longer than the number, log will display in simple style to avoid too much content


betterjava:
I will pull a request after fulfil requirement

PR #2078, opened 2019-03-25
Notice and license file are missing

peng-yongsheng:
- supplement groovy's notice file
- supplement postgresql's license file
- supplement transaction's license file


wu-sheng:
@terrymanu If this is checked, your release vote/test need to be canceled.

PR #2086, opened 2019-03-25
fix SQL rewrite issue

scuwuyu:
Fixes #ISSUSE_ID.

-I reported an issue 4 months ago,https://github.com/apache/incubator-shardingsphere/issues/1348.
-Loxp fixed it,https://github.com/apache/incubator-shardingsphere/pull/1499.
-But it's not perfect,it has another issue. If the SQL is single routing and the pagesize is bigger than one,then the query result will return all rows,bacause the parameters has been rewrited in ParsingSQLRouter.route().


tuohai666:
Accept. The https://github.com/apache/incubator-shardingsphere/issues/2062 is also caused by this rewrite issue.

PR #2088, opened 2019-03-25
Refactor: enhance the dockerfile

taojintianxia:
- replace ADD with wget, Dockerfile should be reproducible, people don't know how and where to retrieve the tar file . even if copy a local file we should use CPY  instead of ADD, ADD is not recommend
- add ENV in Dockerfile, just change the CURRENT_VERSION if PMC wanna release a new image
- not sure when release the 4.0.0-RC1, since it is so, 3.0.0 should be the up to date Dockerfile , change the CURRENT_VERSION if new version proxy released.


terrymanu:
This is a good idea.
For apache release, the distribution file is in `https://dist.apache.org/repos/dist/dev/incubator/shardingsphere/${CURRENT_VERSION}/apache-shardingsphere-incubating-${CURRENT_VERSION}-sharding-proxy.tar.gz` temporary, and will move to `https://dist.apache.org/repos/dist/release/incubator/shardingsphere/${CURRENT_VERSION}/apache-shardingsphere-incubating-${CURRENT_VERSION}-sharding-proxy.tar.gz`

So I suggest wget this file to instead of `https://raw.githubusercontent.com/apache/incubator-shardingsphere-doc/master/dist/${CURRENT_VERSION}.tar.gz`.

Another suggest, `${CURRENT_VERSION}.tar.gz` is a file, use `rm -f` is better than `rm -rf`

PR #2115, opened 2019-03-27
fix 2103

### LINKED ISSUE #2103: Using config center dynamic effectiveness has a bug with judging ShardingRule
image

I‘m using config center(zk). Then modify configurations on zk. Sharding-Jdbc will be pushed to the production environment and take effect immediately. But error.
Because only judge contain with "tables:\n". But it's actually "tables:\r\n"

zhaoyuguang:
Fixes #2103 .

- add judging code


terrymanu:
Thank you

PR #2120, opened 2019-03-28
Code optimization OrderItem.java

zhaoyuguang:
Code optimization OrderItem.java

- Override hashCode will missed at the class file with use @EqualsAndHashCode and Override equals() method.
- Judgment == null before instanceof maybe no need.


terrymanu:
This class use lombok to gen hashcode, It is unnecessary rewrite it

PR #2230, opened 2019-04-18
add check for update shard key

### LINKED ISSUE #2092: Update sharding key is allowed but not work
## Bug Report
Update sharding key is allowed in Sharding-Proxy. The new value may not be routed to the same backend as the old one. So this kind sql should not be allowed.

ShardingSphere 4.0.0RC1

Sharding-Proxy

### Expected behavior
Sharding table test1 by column ‘k’ % 2
1. insert into test1(k, c) values ( 1,  'a');
2. select * from test1 where k=1;
  1 rows in set
3. update test1 set k=2 where k=1;
4. select * from test1 where k=2;
  1 rows in set

### Actual behavior
4. select * from test1 where k=2;
  Empty set

Update sharding key should keep the route as old key but ShardingProxy doesn't check it. Just disallow that kind sql should be ok.

betterjava:
Fixes #2092 .

- add check for update shard key


terrymanu:
Hello, can you explain what's this pr want to do? any issue or mail list to discuss this pull requset?

betterjava:
see issue 2092.I should discuss this there.

terrymanu:
More discuss we can do on that pull request

betterjava:
OK.I do not know whether  Is it necessary to do check for update shard key, or maybe have some more better ways to do that.

Issue #2319, opened 2019-05-05
Insert error when field name was a keyword


AlbertTao:
### Which version of ShardingSphere did you use?

4.0.0-RC1

### Which project did you use? Sharding-JDBC or Sharding-Proxy?

Sharding-JDBC

### Expected behavior

insert success.

### Actual behavior

```
[main] INFO ShardingSphere-SQL - Logic SQL: INSERT INTO `ts_order` (`id`, `desc`) VALUES ('1', 'abc')
[main] INFO ShardingSphere-SQL - SQLStatement: InsertStatement(super=DMLStatement(super=AbstractSQLStatement(type=DML, tables=Tables(tables=[Table(name=ts_order, alias=Optional.absent())]), routeConditions=Conditions(orCondition=OrCondition(andConditions=[AndCondition(conditions=[Condition(column=Column(name=id, tableName=ts_order), operator=EQUAL, compareOperator=null, positionValueMap={0=1}, positionIndexMap={})])])), encryptConditions=Conditions(orCondition=OrCondition(andConditions=[])), sqlTokens=[TableToken(tableName=ts_order, quoteCharacter=BACK_QUOTE, schemaNameLength=0), SQLToken(startIndex=23)], parametersIndex=0, logicSQL=INSERT INTO `ts_order` (`id`, `desc`) VALUES ('1', 'abc')), deleteStatement=false, updateTableAlias={}, updateColumnValues={}, whereStartIndex=0, whereStopIndex=0, whereParameterStartIndex=0, whereParameterEndIndex=0), columnNames=[id, desc], values=[InsertValue(columnValues=[org.apache.shardingsphere.core.parse.old.parser.expression.SQLTextExpression@5a62b2a4, org.apache.shardingsphere.core.parse.old.parser.expression.SQLTextExpression@1051817b])])
[main] INFO ShardingSphere-SQL - Actual SQL: ds_0 ::: INSERT INTO `ts_order_0001`  (id, desc) VALUES ('1', 'abc')
Exception in thread "main" com.mysql.jdbc.exceptions.jdbc4.MySQLSyntaxErrorException: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'desc) VALUES ('1', 'abc')' at line 1
```

### Example codes for reproduce this issue (such as a github link).

terrymanu:
`DESC` is Reserved Words which cannot use to be a column name. 

[...]

PR #2338, opened 2019-05-07
#2319 bug fixes

codefairy08:
Fixes #ISSUSE_ID.

-
-
-


terrymanu:
`ColumnSegment` need to reference `TableSegment`, do not add a new filed name as `originalName`

terrymanu:
It is rewrite problem, we should not change parse module.
We should split insert column and insert values as 2 token .

Issue #2275, opened 2019-04-26
Reconstruct rewriting module


tristaZero:
**Reconstruct rewriting module**
The tasks of parsing module:
- [x] Make all of the tokens store startIndex parameter and stopIndex parameter.
- [x] Remove the invalid parameters for calculating the stopIndex.
- [x] Add different interface to distinguish the SQLTokens for startIndex and stopIndex.

The tasks of rewriteEngine module:
- [x] Use startIndex and stopIndex to rewrite SQLs.
- [x] Make all of the tokens rewritten in the consistent method.
- [x] Generate placeHolders to handle ItemsToken, LimitRowCountToken and so on.
- [x] Delete getLogicTableName() from ShardingPlaceholder.
- [x] Move encrypting processing part to ShardingEncryptorEngine.
- [x] Use pattern() to create constant placeHolders, variant placeHolders, original literal from SQLTokens.
- [x] Add interface for variant placeHolders to toString(parameters).
- [x] Use pattern() to make constant placeHolders become literals, and create SQLBuilder to store placeHolder-literals, variant placeHolders and original literals.
- [x] Use generateSQL() to generate SQLUnit.

The tasks of SQLBuilder module:
- [x] Store variant placeHolders and placeHolder-literals and original literals.
- [x] Use toSQL() to generate SQLUnit.

terrymanu:
Some code review suggestion:

1. XXXSQLRewriteEngine move to new package name as `engine`
2. XXXSQLRewriteEngine need implement a interface name as `SQLRewriteEngine`
3. MasterSlaveSQLRewriteEngine use @RequiredArgsConstructor
4. Constructor don't need java-doc
5. method of XXXSQLRewriteEngine.generateSQL is unnecessary, just call sqlBuilder.toSQL
6. Beside Alterable, add Addable interface
7. split sharding and encrypt placeholder in package `org.apache.shardingsphere.core.rewrite.placeholder`
8. EncryptXXXPlaceholder should not implement `ShardingPlaceholder`
9. Don't use @SuppressWarnings("all") on class EncryptUpdateItemColumnPlaceholder 
10. EncryptUpdateItemColumnPlaceholder has lots of field can be null, consider about the design, maybe split assistedColumn and not has assistedColumn situation
11. Question in EncryptWhereColumnPlaceholder, between should not supported
12. Should not call SQLUtil.getExactlyValue on rewrite module, current value should come from parse module. For example: IndexPlaceholder
[...]

PR #2373, opened 2019-05-13
Optimize all the rewrite engines

tristaZero:
#2275


tristaZero:
Some code review suggestion:

- [ ] XXXSQLRewriteEngine move to new package name as `engine`
- [x] XXXSQLRewriteEngine need implement a interface name as `SQLRewriteEngine`
- [x] MasterSlaveSQLRewriteEngine use @RequiredArgsConstructor
- [x] Constructor don't need java-doc
- [ ] method of XXXSQLRewriteEngine.generateSQL is unnecessary, just call sqlBuilder.toSQL
- [ ] Beside Alterable, add Addable interface
- [ ] split sharding and encrypt placeholder in package `org.apache.shardingsphere.core.rewrite.placeholder`
- [ ] EncryptXXXPlaceholder should not implement `ShardingPlaceholder`
- [x] Don't use @SuppressWarnings("all") on class EncryptUpdateItemColumnPlaceholder 
- [x] EncryptUpdateItemColumnPlaceholder has lots of field can be null, consider about the design, maybe split assistedColumn and not has assistedColumn situation
- [ ] Question in EncryptWhereColumnPlaceholder, between should not supported
- [ ] Should not call SQLUtil.getExactlyValue on rewrite module, current value should come from parse module. For example: IndexPlaceholder
- [ ] Rewrite of class InsertSetPlaceholder is out of scope, should not rewrite "SET"

PR #2425, opened 2019-05-22
use p3c check code

zkzong:
- add @override annotation
- use class name access static member
- resolve possible NPE


terrymanu:
Thank you for contribute codes.
But the codes cause travis failed.
The problem maybe in here:

```java
String text = duplicateSpecificationNode.get().getText();
```

You get optional directly, it will cause exception.

Can you fix this issue first?

Issue #2455, opened 2019-05-28
Support database persistence for atomikos transaction manager.


cherrylzhao:
in order to make Sharding-transaction-xa-manager running in the cloud native environment, we should extend atomikos transaction manager to support persistence log into database.

- [x] implement customize Assembler using atomikos SPI
- [x] implement database repository of atomikos
- [x] extend atomikos InMemoryRepository
- [x] extend atomikos CachedRepository
- [x] add unit test
- [x] extract persistence SPI

the whole image is like this
<img width="987" alt="image" src="https://user-images.githubusercontent.com/8317649/58444988-3c7f5500-812d-11e9-871d-14c86a5d34f0.png">

this issue is step 1~2, we'll create other issue for step 3~4

PR #2474, opened 2019-05-30
for #2455 implement database persistence for atomikos transaction manager.

cherrylzhao:


terrymanu:
We can simplify configuration:

1. remove app and group id
2. use yaml to config data source for xa persistence
3. hide other prop config for now

cherrylzhao:
> We can simplify configuration:
> 
> 1. remove app and group id
> 2. use yaml to config data source for xa persistence
> 3. hide other prop config for now

I have fixed it, please reviewed it again.

PR #2523, opened 2019-06-08
try commit a startup boosted mode and default set to false

evenardo:
When we used sharding-jdbc for a development environment , if we have many actual tables, like more than 200 total , it's cost more time to start up the sharding-jdbc .  This may cost several times for developers , especially non-elite programmers for debug process.

I try to found out why this happend , then I try to fix or improve some function especially for dev environment.

I used multi threads to get ready for some map data , really, it's cost more connection database , when sharding-jdbc start up . As @terrymanu  said , it may cause database resources exhausted.

Then I make threads in number of [ sqrt(total actual tables) ], assemby a arraylist to split the slice, into parts of alltableNames, to reduce the threads number , in hope to reduce connection to database.

I collected SQLExceptions , re throw them like origin code to upper logic, but can not throw new exceptions like InterruptedException to upper , because it will change the upper function's "throws" . so I droped them.

why I try to commit a unstable function (or it stable but we don't Long-term tested) , and make a switcher to close this function ?

Because , in our small business, or others like us . Our coder are not elited at all , they cost many many time to restart the local environment to debug ,to change propertitys. So sad , in the waiting time, they are very depressed and complaining complaining ... In our case, it's cost more than 4 minutes :(  in one time. And now it's reduce to 40 seconds , that's within temptation and boost morale !

So I hope is , may officer can make a global switcher in a suitable location to let us set this like a dev frendly mode , to let us boosted the startup process , especially  in a develop environment.

Fixes #ISSUSE_ID.

in class TableMetaDataInitializer

    //Here's a switch to open boosted startup mode;
    //before function integrated into base project,
    //you can rebuild org.apache.shardingsphere.core.execute module,
    //change this to true ,then "mvn clean install" to use this function in your local environment.
    private boolean startUpBoostedMode = false;

in function Map<String, TableMetaData> load(final ShardingRule shardingRule):

     if (startUpBoostedMode) {
            System.out.println("org.apache.shardingsphere.core.execute.TableMetaDataInitializer:\n" +
                    "    Beta function warn: Startup boosted mode is friendly for dev mode, " +
                    "but it may cause database resources exhausted. \n" +
                    "    Number of sqrt(number of all actual tables) threads created and finished in process. " +
                    "if error occurs, change to stable mode. \n"+
                    "    Or just use this in your local dev environment. ");
        }

### DIFF
--- a/sharding-core/sharding-core-execute/src/main/java/org/apache/shardingsphere/core/execute/metadata/TableMetaDataInitializer.java
+++ b/sharding-core/sharding-core-execute/src/main/java/org/apache/shardingsphere/core/execute/metadata/TableMetaDataInitializer.java
@@ -30,10 +30,8 @@ import java.sql.Connection;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.SQLFeatureNotSupportedException;
-import java.util.Collection;
-import java.util.HashMap;
-import java.util.LinkedHashSet;
-import java.util.Map;
+import java.util.*;
+import java.util.concurrent.ConcurrentHashMap;
 
 /**
  * Table meta data initializer.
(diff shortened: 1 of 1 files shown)


evenardo:
Local test passed in our case ,with rebuilded module core.execute . even startUpMode is true and false, it's worked fine . I think the coverage reduced , because of the switcher is turned off , need I recommit these and turn it ON ?  

I don't really want this submission to be used in a production environment, because this time is nothing. for several minutes . :)

terrymanu:
Hi, ref with #2522, the problems are still existed.

1. Travis still broken. Right now, although the travis are not broken because of you create a new boolean to switch off your new codes. If the boolean value is switch on, I think the travis still broken. Can you add test cases for your new codes?
2. The problem is still existed for unlimited threads cause resources exhausted if the boolean value is switch on.
3. The problem is still existed for refresh databases of orchestration module if boolean value is switch on.
4. `System.out.print` are not permitted.
5. We don't prefer include a temporary design for test mode only, we need to prove the new design are good enough for production code. 

If you just test in your own environment, all changes are fine if you think it is ok. 
But if you want merge your pr to master, could you solve the problems above?

evenardo:
Indeed, I haven't read all the code yet. This submission is a bit sloppy. Thank you for taking the time to read the submitted code and pointing out the problem.

1. Travis still broken. Right now, although the travis are not broken because of you create a new boolean to switch off your new codes. If the boolean value is switch on, I think the travis still broken. Can you add test cases for your new codes?

A: Travis broken early , because of I forgot change import packages .... , I'v got a long path problem stop me to commit the all modules project from local storage .  Then I just edit the file in the github. I'm already mvn installed the all sphere in local . and cost a long time to saw many ( hurge number )tests done well . 

2. The problem is still existed for unlimited threads cause resources exhausted if the boolean value is switch on.

A: Assume 10000 actual tables made , just used 100 threads connection to connect database , and just during the process assemby Map ( According to my understanding of these codes from origin ).And when the map result is ready it's just a result typed Map, and just a data , no threads exsist at all , nothing will run self , it's will as same result as the origin code, nothing different . When the project running smoothly, the database cost only by the CP pool from new connection , and the 100 connection created, ware already released after join() later. I don't agree that threads are made database exhausted when project running smoothly. because they already finished and joined , just left a Map data .

3. The problem is still existed for refresh databases of orchestration module if boolean value is switch on.

A: As No.2 . What every , I don't read all codes yet, sorry for that . But in my opinion .Map is already get ready ,  If there is any module will re-assemby all the Map , likely a 4 minute process "orchestration", it's sounds like a terrible process too . Will it suspend the program progress when the base metadata remaped ? If that needed to happen , duration also be shortly for better but minutes to suspend the program . If just need to change some of Map's element , may be there already had a function to point the changed entity. why use the TableMetaDataInitializer to initialize all of the map but not one element?  I don't believe TableMetaDataInitializer.load(ShardingRule) will be use for.

4. We don't prefer include a temporary design for test mode only, we need to prove the new design are good enough for production code.

A: ok and accept.  But , how to make sure all of them ? 
In my case used spring , and several entity , several sharding tables. I'm difficult to write the unit test to assmby the environment , like database , sping runner , api controller . how to test the connection to the database ? connect to what , and how make sure the connection in unitTest run well after database not online ? uhhhh...   : )

5 .System.out.print are not permitted.
A: ok . 

After all , This submission is a bit sloppy. And change core is really dangerous .
 
: ) 

Thank you for your replay , thank you for your encourage.

evenardo:
In any case, ensuring code quality is a serious process, and NO stability No faster. Maybe my needs are just individual case.  Keep up harder . :)

zhang6:
TableMetaData不遍历，对一般的增删改查也没影响吧

PR #2533, opened 2019-06-11
add config to check actual data node in runtime or not

zhang98722:
in some  case there is lots of table need to add in run time
and cant save them to the config or somewhere before sharding sphere init

in v4.0.0 if you insert data to a new table not configured in table rule actual data node
sharding sphere wont execute because it will do the check in org.apache.shardingsphere.core.route.type.standard.StandardRoutingEngine#removeNonExistNodes, return -1 as the result
so i added a configuration to tell sharding sphere not check the actual data node, just execute the sql,and throw exception if it involved

- table rule config add property _checkActualDataNode_
- table rule add perperty _checkActualDataNode_ , default value is true
- org.apache.shardingsphere.core.route.type.standard.StandardRoutingEngine#removeNonExistNodes : is not _checkActualDataNode_ then dont execute this remove


zhang98722:
revert this pr
in lots of case the sharding sphere need to know the actual data nodes belong to the logic table
run time table add/remove not fit sharding sphere

tristaZero:
@zhang98722 Thanks for your pr anyway.

PR #2668, opened 2019-07-07
add `getValueByColumnType` method

sunbufu:
Fixes https://github.com/apache/incubator-shardingsphere/issues/2604.

- add `getValueByColumnType` method
- add some unit test

### DIFF
--- a/sharding-core/sharding-core-execute/src/main/java/org/apache/shardingsphere/core/execute/sql/execute/result/StreamQueryResult.java
+++ b/sharding-core/sharding-core-execute/src/main/java/org/apache/shardingsphere/core/execute/sql/execute/result/StreamQueryResult.java
@@ -68,12 +68,12 @@ public final class StreamQueryResult implements QueryResult {
     
     @Override
     public Object getValue(final int columnIndex, final Class<?> type) throws SQLException {
-        return decrypt(columnIndex, QueryResultUtil.getValue(resultSet, columnIndex));
+        return decrypt(columnIndex, QueryResultUtil.getValue(resultSet, columnIndex, type));
     }
     
     @Override
     public Object getValue(final String columnLabel, final Class<?> type) throws SQLException {
-        return decrypt(columnLabel, QueryResultUtil.getValue(resultSet, metaData.getColumnIndex(columnLabel)));
+        return decrypt(columnLabel, QueryResultUtil.getValue(resultSet, metaData.getColumnIndex(columnLabel), type));
     }
     
     @Override
--- a/sharding-core/sharding-core-execute/src/main/java/org/apache/shardingsphere/core/execute/sql/execute/result/QueryResultUtil.java
+++ b/sharding-core/sharding-core-execute/src/main/java/org/apache/shardingsphere/core/execute/sql/execute/result/QueryResultUtil.java
@@ -17,18 +17,25 @@
 
 package org.apache.shardingsphere.core.execute.sql.execute.result;
 
+import java.math.BigDecimal;
+import java.sql.Blob;
+import java.sql.Clob;
+import java.sql.Date;
 import java.sql.ResultSet;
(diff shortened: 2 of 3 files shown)


sharding-core/sharding-core-execute/src/main/java/org/apache/shardingsphere/core/execute/sql/execute/result/StreamQueryResult.java
```diff
@@ -68,12 +68,12 @@ public boolean next() throws SQLException {
     
     @Override
     public Object getValue(final int columnIndex, final Class<?> type) throws SQLException {
-        return decrypt(columnIndex, QueryResultUtil.getValue(resultSet, columnIndex));
+        return decrypt(columnIndex, QueryResultUtil.getValue(resultSet, columnIndex, type));
```
tuohai666: StreamQueryResult pass the type to getValue(), but what about MemoryQueryResult?

sunbufu: > StreamQueryResult pass the type to getValue(), but what about MemoryQueryResult?

This is why i doesn't deleted `QueryResultUtil.getValueByColumnType()`, I have overload a `getValue()` method with `Class<?> type` in `QueryResultUtil`,  `StreamQueryResult` will call the new `getValue` with `Class<?> type`, and `MemoryQueryResult ` will continue to call the old `getValue()`.
image
image

sunbufu: > StreamQueryResult pass the type to getValue(), but what about MemoryQueryResult?

This is why I doesn't deleted `QueryResultUtil.getValueByColumnType()`, I just have overload a `getValue()` method with `Class<?> type` in `QueryResultUtil`,  `StreamQueryResult` will call the new `getValue()` with `Class<?> type`, and `MemoryQueryResult ` will continue to call the old `getValue()`.
StreamQueryResult call the new getValue
MemoryQueryResult call the old getValue

tuohai666: If so, the problem was not resolved. Two modes get two different value is even worse, isn't?

sunbufu: It's my fault, I overlooked this.
And I have a question. Why do we need to create a QueryRow Iterator and manage it ? Why not do just like [druid's DruidPooledResultSet](https://github.com/alibaba/druid/blob/master/src/main/java/com/alibaba/druid/pool/DruidPooledResultSet.java) ？

tuohai666: It's OK. We have noticed your work and expect more contribution from you. I suggest you become a contributor first, then try to become a committer.

> And I have a question. Why do we need to create a QueryRow Iterator in MemoryQueryResult and manage it ? Why not do just like [druid's DruidPooledResultSet](https://github.com/alibaba/druid/blob/master/src/main/java/com/alibaba/druid/pool/DruidPooledResultSet.java) ？

Sometimes we need to get all data to memory:
1. Merge engine:
https://shardingsphere.apache.org/document/current/en/features/sharding/principle/merge/
2. Execute engine:
https://shardingsphere.apache.org/document/current/en/features/sharding/principle/execute/

sunbufu: I'll read it first, thank you.

PR #2701, opened 2019-07-12
Log configuration from reg center into log files

sunbufu:
Fixes https://github.com/apache/incubator-shardingsphere/issues/1907.

- add `ConfigurationPrinter` class for configuration log.
- print configuration on `EncryptRule`, `MasterSlaveRule`, `ShardingRule`
- add `setAuthentication`, `setProperties` on `ShardingProxyContext` for configuration print
- add unit test


sunbufu:
> I don't think simply printing the logs in unit tests (without assertions) is a good practice. Your tests pass even when the printer prints wrong logs, which makes the tests useless.
> 
> I'd suggest mocking the `log` in class `ConfigurationPrinter` and `verify` the `log.info` method with expected parameters

You are right, and I will mock a `log` on `ConfigurationPrinterTest`.

PR #2737, opened 2019-07-22
Modify ‘addUnit’ function in 'ShardingInsertOptimizedStatement.java' and related tests.

### LINKED ISSUE #2693: The method "ShardingInsertOptimizedStatement.addUnit()" has done more than one thing
## The method "ShardingInsertOptimizedStatement.addUnit()" has done more than one thing

The method "ShardingInsertOptimizedStatement.addUnit()" has done more than one thing:create InsertOptimizeResult Object and add it to the member 'units'. I think it is better to split the method into two parts. One method creates InsertOptimizeResult and Another use this InsertOptimizeResult Object as input to add it in the member 'units'.

wgy8283335:
Fixes #2693.

- modify 'addUnit' function.
- add ''createUnit' function.
- modify test cases.


terrymanu:
The conflict  need to be resolved.

Issue #2746, opened 2019-07-23
add Java Microbenchmark Harness to our project


CalvinKirs:
## Feature Request

**For English only**, other languages will not accept.

Please pay attention on issues you submitted, because we maybe need more details. 
If no response **more than 7 days** and we cannot make decision by current information, we will **close it**.

Please answer these questions before submitting your issue. Thanks!

### Describe the feature you would like.
add Java Microbenchmark Harness to our project，In the future, our code commit optimization will be more convincing.

terrymanu:
Yes, it is a good idea, any suggestion to do it?

CalvinKirs:
> Yes, it is a good idea, any suggestion to do it?
#2747
Create a model separately here, or we can create the ShardingSphere-benchmark directory under the Apache community. I prefer the former, so it is more clearly visible to all committers, and does not need to re-clone project submission project. For example, netty is doing this.

terrymanu:
@CalvinKirs Just as @nancyzrh said, this PR only add a maven dependency for jmh, but does not add any test cases yet, so I think you need add test cases for that.

As my understanding, jmh only can print benchmark info to console which is not very make sense, we have already use jmeter to test benchmark.

The precondition for introduce jmh to ShardingSphere is we need to auto assert if benchmark is under threshold.

terrymanu:
It is duplicated with #2747, I just close this issue

PR #2747, opened 2019-07-23
add Java Microbenchmark Harness

CalvinKirs:
Fixes #ISSUSE_ID.
#2746 
-
-add Java Microbenchmark Harness
-

### DIFF
--- a/sharding-benchmark/pom.xml
+++ b/sharding-benchmark/pom.xml
@@ -1,0 +1,120 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Licensed to the Apache Software Foundation (ASF) under one or more
+  ~ contributor license agreements.  See the NOTICE file distributed with
+  ~ this work for additional information regarding copyright ownership.
+  ~ The ASF licenses this file to You under the Apache License, Version 2.0
+  ~ (the "License"); you may not use this file except in compliance with
+  ~ the License.  You may obtain a copy of the License at
+  ~
+  ~     http://www.apache.org/licenses/LICENSE-2.0
+  ~
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.apache.shardingsphere</groupId>
+        <artifactId>shardingsphere</artifactId>
+        <version>4.0.0-RC2-SNAPSHOT</version>
+    </parent>
+    <artifactId>sharding-benchmark</artifactId>
+    <packaging>pom</packaging>
+    <name>${project.artifactId}</name>
+    <properties>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <jmh.version>1.9</jmh.version>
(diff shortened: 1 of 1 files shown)


CalvinKirs:
> Can one of the admins verify this patch?
of course

CalvinKirs:
@tristaZero can you help me review

nancyzrh:
is there any way to run jmh compared with its last results automatically or just run as junit cases?

CalvinKirs:
> is there any way to run jmh compared with its last results automatically or just run as junit cases?



> is there any way to run jmh compared with its last results automatically or just run as junit cases?



> is there any way to run jmh compared with its last results automatically or just run as junit cases?

Wait a moment, I will finish it.

CalvinKirs:
> is there any way to run jmh compared with its last results automatically or just run as junit cases?

Wait a moment, I will finish it.

nancyzrh:
> > is there any way to run jmh compared with its last results automatically or just run as junit cases?
> 
> Wait a moment, I will finish it.

good~

CalvinKirs:
> > > is there any way to run jmh compared with its last results automatically or just run as junit cases?
> > 
> > 
> > Wait a moment, I will finish it.
> 
> good~

I think I understand what you mean wrong. This is a beautiful misunderstanding. JMH is just a quantifiable data report to reflect your code's performance. It is mainly used for our general core logic optimization code (basic code or algorithm level), and we usually use the Test package to indicate the accuracy of your code. The focus of the two is different.

CalvinKirs:
@terrymanu Hello, can you help me review my code?

terrymanu:
@CalvinKirs  Just as @nancyzrh said, this pr just add a maven dependency for jmh, but does not include any test cases yet, so I think you need add test cases first.

As my understanding, jmh only can print benchmark info to console which is not very make sense, we have already use jmeter to test benchmark.

The precondition for introduce jmh to ShardingSphere is we need to auto assert if benchmark is under threshold.

CalvinKirs:
> @CalvinKirs Just as @nancyzrh said, this PR only add a maven dependency for jmh, but does not add any test cases yet, so I think you need add test cases for that.
> 
> As my understanding, jmh only can print benchmark info to console which is not very make sense, we have already use jmeter to test benchmark.
> 
> The precondition for introduce jmh to ShardingSphere is we need to auto assert if benchmark is under threshold.

I don't quite agree with your point of view. The existence is reasonable. The microcode test I think needs more lightweight JMH, not Jmeter. The level of action of the two is different. Using JMH will make our committers more convincing when submitting certain code. Of course, it may be my fault. I still want to hear your suggestions or suggestions from others.

CalvinKirs:
> > @CalvinKirs Just as @nancyzrh said, this PR only add a maven dependency for jmh, but does not add any test cases yet, so I think you need add test cases for that.
> > As my understanding, jmh only can print benchmark info to console which is not very make sense, we have already use jmeter to test benchmark.
> > The precondition for introduce jmh to ShardingSphere is we need to auto assert if benchmark is under threshold.
> 
> I don't quite agree with your point of view. The existence is reasonable. The microcode test I think needs more lightweight JMH, not Jmeter. The level of action of the two is different. Using JMH will make our committers more convincing when submitting certain code. Of course, it may be my fault. I still want to hear your suggestions or suggestions from others.

It is absolutely necessary to use Jmeter for large-scale stress testing, which allows us to implement the project with data results as a support. But the microcode test uses Jmeter just like Take not a musket to kill a butterfly.

terrymanu:
Can jmh auto assert if benchmark under threshold?

CalvinKirs:
> Unit tests need assert automatically, otherwise nobody will notice performance is slow at the moment.
> 
> Can jmh auto assert if benchmark under threshold?

of course

terrymanu:
How to do it, could you add more jmh test case on this pr?

CalvinKirs:
> How to do it, could you add more jmh test cases on this pr?

Sorry, I am busy recently, I will finish it later.

terrymanu:
@CalvinKirs Hi, it is not a good practice to close pull request without any reason.
Can you explain the reason?

CalvinKirs:
> @CalvinKirs Hi, it is not a good practice to close pull request without any reason.
> Can you explain the reason?

So sorry, my understanding has a deviation，so this may be a wrong decision

PR #2755, opened 2019-07-24
Modify addUnit() and add createUnit() functions

### LINKED ISSUE #2693: The method "ShardingInsertOptimizedStatement.addUnit()" has done more than one thing

wgy8283335:
Modify addUnit() and add createUnit() functions in ShardingInsertOptimizedStatement.java. And modify related functions in caller.

Fixes #2693.

- modify 'addUnit' function.
- add ''createUnit' function.
- modify test cases.


terrymanu:
What is the relationship with #2737?

tristaZero:
Good job
@wgy8283335 ^_^

PR #2798, opened 2019-08-01
[INFRA-18812] Switch to .mvn/jvm.config ...

rfscholte:
... so MAVEN_OPTS won't get polluted.

Recently while running integration tests of Apache Maven we noticed our builds becoming more and more unstable. These tests all were testing MAVEN_OPTS.
We discovered that this project was generating .mavenrc files that stayed on the server, meaning that every new build was using this file and the new MAVEN_OPTS.
Since Maven 3.3.1 it is possible to define jvm arguments per project. This PR contains those changes and that will restore the predictable behaviour of Jenkins.

### DIFF
--- a/.mvn/jvm.config
+++ b/.mvn/jvm.config
@@ -1,0 +1,1 @@
+-Xmx1024m -XX:MaxPermSize=256m
--- a/jenkinsfile
+++ b/jenkinsfile
@@ -39,7 +39,8 @@ pipeline {
         }
         stage('Build & Test') {
             tools {
-                maven 'maven-3.2.1'
+			    // use at least Apache Maven 3.3.1 to have .mvn/jvm.config support
+                maven 'maven-3.3.1'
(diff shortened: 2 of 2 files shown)


hboutemy:
notice: what is polluted first is `.mavenrc` file that is added to Jenkins node filesystem...
and yes, this script pollutes MAVEN_OPTS...

.mvn/jvm.config
```diff
@@ -0,0 +1 @@
+-Xmx1024m -XX:MaxPermSize=256m
```
hboutemy: @rfscholte need to add this `.mvn/jvm.config` to rat excludes in root `pom.xml`

terrymanu:
Got it. Thank you very much for the pull request, and sorry about for polluted for other projects.

rfscholte:
I'm pretty sure that `jvm.config` should not have a copyright header and should be added to the rat excludes, because its content is added directly to the `java` executable as is.

PR #2805, opened 2019-08-02
add unreserve keyword to mysql grammar

jingshanglu:
- add unreserve keyword 'NAME' for mysql


terrymanu:
I notice this pr only add some new keywords but not handle with  srs and trigger, do you miss commit some codes?

PR #2830, opened 2019-08-08
Added support for no sharding column in sql

### LINKED ISSUE #2826: Ability to define behaviour when sharding column is not present in the SQL
https://github.com/apache/incubator-shardingsphere/issues/2815

There needs to be a way to define what to be done (which target names to be chosen) when sharding column is not in the SQL. Currently, as per `StandardRoutingEngine`, all available target names are returned. 

As a suggestion, we can also think of 

- moving this logic to the sharding algorithm itself. That means, let the sharding algorithm decide which target names to select in case there is no sharding column in the SQL 

- and may be have a default behaviour, in case the sharding algorithm also does not return any target name.

rimal:
Fixes #2826

- Empty `shardingValues` to be handled in the `ShardingStategy`.
- Currently, `ShardingStategy` is not invoked if `shardingValues` are empty.

### DIFF
--- a/sharding-core/sharding-core-common/src/main/java/org/apache/shardingsphere/core/strategy/route/complex/ComplexShardingStrategy.java
+++ b/sharding-core/sharding-core-common/src/main/java/org/apache/shardingsphere/core/strategy/route/complex/ComplexShardingStrategy.java
@@ -71,7 +71,9 @@ public final class ComplexShardingStrategy implements ShardingStrategy {
         }
         Collection<String> shardingResult = shardingAlgorithm.doSharding(availableTargetNames, new ComplexKeysShardingValue(logicTableName, columnShardingValues, columnRangeValues));
         Collection<String> result = new TreeSet<>(String.CASE_INSENSITIVE_ORDER);
-        result.addAll(shardingResult);
+        if (null != shardingResult) {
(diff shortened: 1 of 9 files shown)


rimal:
@terrymanu @KomachiSion Can you please review this.

KomachiSion:
Thanks for your pr, I will review as soon as possible.

KomachiSion:
Hi @rimal, I've already finished reviewing your PR, I think the implementation is contrary to our design and may cause some problems for other users.

1.  We design ShardingSphere should be a complete database for application. The SQL is the command and basis which let database execute. When SQL miss sharding column, the sharding values should be empty and shardingSphere should do full-routing when sharding values is empty. No database will callback a custom algorithm when sharding values is empty.

2. For almost users, maybe 99%, they implement their algorithm with a premise that `shardingValues` is not empty. This change may cause program exceptions and users' confusion.

So we can't merge this PR to main branch.

I think `SQL examine` @terrymanu said in #2826 may be a better way to implement this. We look forward to your more PR on `SQL examine` features.

BTW, If you really want to achieve like this, you can fork a branch to do. Thanks.

rimal:
@KomachiSion Thanks for the quick response here :)

I will then carry on further discussion on #2826 . Closing this PR.

Issue #2152, opened 2019-04-02
MemoryQueryResult getColumnIndex throw exception when using Oracle database


KomachiSion:
## Bug Report

when execute Oracle's DQL, getting the result throws an exception
```
java.lang.IndexOutOfBoundsException: Index: 0, Size: 0
	at java.util.ArrayList.rangeCheck(ArrayList.java:657)
	at java.util.ArrayList.get(ArrayList.java:433)
	at org.apache.shardingsphere.core.execute.sql.execute.result.QueryResultMetaData.getColumnIndex(QueryResultMetaData.java:140)
	at org.apache.shardingsphere.core.execute.sql.execute.result.MemoryQueryResult.getValue(MemoryQueryResult.java:95)
	at org.apache.shardingsphere.core.merge.dql.common.StreamMergedResult.getValue(StreamMergedResult.java:56)
	at org.apache.shardingsphere.core.merge.dql.common.DecoratorMergedResult.getValue(DecoratorMergedResult.java:46)
	at org.apache.shardingsphere.shardingjdbc.jdbc.core.resultset.ShardingResultSet.getString(ShardingResultSet.java:141)
	at org.hibernate.type.descriptor.sql.VarcharTypeDescriptor$2.doExtract(VarcharTypeDescriptor.java:62)
	at org.hibernate.type.descriptor.sql.BasicExtractor.extract(BasicExtractor.java:47)
	at org.hibernate.type.AbstractStandardBasicType.nullSafeGet(AbstractStandardBasicType.java:261)
	at org.hibernate.type.AbstractStandardBasicType.nullSafeGet(AbstractStandardBasicType.java:257)
	at org.hibernate.type.AbstractStandardBasicType.nullSafeGet(AbstractStandardBasicType.java:247)
	at org.hibernate.type.AbstractStandardBasicType.hydrate(AbstractStandardBasicType.java:333)
	at org.hibernate.loader.Loader.extractKeysFromResultSet(Loader.java:794)
	at org.hibernate.loader.Loader.getRowFromResultSet(Loader.java:719)
	at org.hibernate.loader.Loader.processResultSet(Loader.java:991)
	at org.hibernate.loader.Loader.doQuery(Loader.java:949)
	at org.hibernate.loader.Loader.doQueryAndInitializeNonLazyCollections(Loader.java:341)
	at org.hibernate.loader.Loader.doList(Loader.java:2692)
	at org.hibernate.loader.Loader.doList(Loader.java:2675)
[...]

PR #2848, opened 2019-08-10
Feature#2715

sunbufu:
Fixes https://github.com/apache/incubator-shardingsphere/issues/2715.

1. add getQueryResultMetaData() method on QueryResult.
2. add QueryResultMetaData on StreamQueryResult class.
3. modify AggregationDistinctQueryResult constructor to use queryResultMetaData instead of columnLabelAndIndexMap.
4. use HashMultimap instead of TreeMap on QueryResultMetaData.
5. modify unit test class.
6. add integrated test.


sharding-core/sharding-core-execute/src/main/java/org/apache/shardingsphere/core/execute/sql/execute/result/QueryResultMetaData.java
```diff
@@ -70,8 +71,8 @@ public QueryResultMetaData(final ResultSetMetaData resultSetMetaData) {
     }
     
     @SneakyThrows
-    private Map<String, Integer> getColumnLabelAndIndexMap() {
-        Map<String, Integer> result = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
+    private Multimap<String, Integer> getColumnLabelAndIndexMap() {
+        Multimap<String, Integer> result = HashMultimap.create();
```
KomachiSion: The Mulitmap is `CASE_SENSITIVE` may be cause some bug like #2152 .

PR #2854, opened 2019-08-12
modify the enlistment to support the transaction propagation REQUIRES_NEW

liuyangming:
XAShardingTransactionManager should not maintain the enlisted resources, because it cannot control the association/desociation of transactions. Consider the following scenario( service **A**{REQUIRED} -> service **B**{REQUIRES_NEW} ): 
0. TM begins a xa transaction(txA) for A;
1. A performs an insert operation in DataSource demo_ds;
2. A invokes service B, the propagation of B is REQUIRES_NEW, so TM begins a new xa transaction(txB) for B;
3. B performs an insert operation in the same DataSource demo_ds too;
in this scenario, the XAResource should be enlisted to both txA and txB. In the existing implementation, the XAResource is only enlisted to txA, because when B get the connection, there is a dataSourceName associated with current thread.

Since the XAShardingTransactionManager does not participate in the thread control of transaction, it should always dispatch the enlistment to the TM, the TM will decide if and how to enlist these xa resources.

### DIFF
--- a/sharding-transaction/sharding-transaction-2pc/sharding-transaction-xa/sharding-transaction-xa-core/src/main/java/org/apache/shardingsphere/transaction/xa/XAShardingTransactionManager.java
+++ b/sharding-transaction/sharding-transaction-2pc/sharding-transaction-xa/sharding-transaction-xa-core/src/main/java/org/apache/shardingsphere/transaction/xa/XAShardingTransactionManager.java
@@ -33,9 +33,9 @@ import javax.transaction.Status;
 import java.sql.Connection;
 import java.util.Collection;
 import java.util.HashMap;
-import java.util.LinkedList;
-import java.util.List;
 import java.util.Map;
+import javax.transaction.TransactionManager;
+import javax.transaction.Transaction;
 
 /**
  * Sharding transaction manager for XA.
@@ -47,14 +47,14 @@ public final class XAShardingTransactionManager implements ShardingTransactionMa
     private final Map<String, SingleXADataSource> singleXADataSourceMap = new HashMap<>();
     
     private final XATransactionManager xaTransactionManager = XATransactionManagerLoader.getInstance().getTransactionManager();
-    
-    private ThreadLocal<List<String>> enlistedXAResource = new ThreadLocal<List<String>>() {
-        @Override
-        public List<String> initialValue() {
-            return new LinkedList<>();
-        }
-    };
-    
+
+	// private ThreadLocal<List<String>> enlistedXAResource = new ThreadLocal<List<String>>() {
+	// @Override
+	// public List<String> initialValue() {
+	// return new LinkedList<>();
+	// }
+	// };
+
     @Override
     public void init(final DatabaseType databaseType, final Collection<ResourceDataSource> resourceDataSources) {
         for (ResourceDataSource each : resourceDataSources) {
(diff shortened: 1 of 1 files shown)


liuyangming:
> I have understood the nested transaction issue.
> but for XA TM implement (such as atomikos), they don't support enlist same resource multiple times in one XA transaction. you should make sure the modification runnable in our transaction-example at first.

This scenario involves two XA transactions, not just one, but both XA transactions need to access the same resource. Since the transaction propagation level of the second transaction is REQUIRES_NEW, it requiures the TM performs the suspend/resume operation on the XAResource. 

TM like Atomikos support suspend/resume operation, but currently mysql/postgresql as an RM does not supported (include join), other databases such as oracle and db2 are supported, the database h2 used in my current test case also supports too.

The sample provided by shardingsphere is not a typical usage of a spring declarative transaction, so I created a new sample project based on this, hope can help to reproduce this problem. 

github: ([https://github.com/liuyangming/shardingjdbc-sample](https://github.com/liuyangming/shardingjdbc-sample))

cherrylzhao:
thanks for your reply, it looks like you have a great experiment about usaging XA.
in fact our sharding-transaction-xa integrated with ShardingConnection, you should try to make ShardingConnection support nested transaction at first. see `ShardingConnection#setAutoCommit`

liuyangming:
> thanks for your reply, it looks like you have a great experiment about usaging XA.
> in fact our sharding-transaction-xa integrated with ShardingConnection, you should try to make ShardingConnection support nested transaction at first. see `ShardingConnection#setAutoCommit`

In my understanding, although the commit/rollback of the RMLT(resource manager local transaction) depends on java.sql.Connection to complete, the XA transaction is completely independent of the java.sql.Connection. It may be not appropriate to control the XA transaction through the java.sql.Connection, i think the TM is more suitable. 

Furthermore, this scenario involves two different xa transactions, they get two different Connections, except share the same thread context, there is no association between them. To support the REQUIRES_NEW transaction propagation, simply forwarding the enlistResource operation to the TM is sufficient.

cherrylzhao:
I see, REQUIRES_NEW works on different connection to support new transaction. forward the enlistResource to TM can indeed solve this problem. but in ShardingSphere there also have some scenarios that getting multiple connections from one datasource, we should avoid to enlist same XAResource(same datasourceName) more than once.

liuyangming:
> I see, REQUIRES_NEW works on different connection to support new transaction. forward the enlistResource to TM can indeed solve this problem. but in ShardingSphere there also have some scenarios that getting multiple connections from one datasource, we should avoid to enlist same XAResource(same datasourceName) more than once.

Under normal circumstances, this will not be a problem, because the connection sharing mechanism of the connection pool component will ensure that the same physical connection will be used for multiple getConnections within a transaction, and will not cause multiple enlistments.

However, I read the connection management implementation of shardingsphere, Since it encapsulates connection pool, the connection pool can't interact with TM, so it really needs to solve this problem by itself, and it also causes shardingsphere cannot support REQUIRES_NEW transaction propagation level. It seems that this problem cannot be resolved by forwarding the enlistment.

Btw, closing the ShardingConnection will cause the TransactionType changed, this is really unacceptable. It is impossible for the business system to manually change the TransactionType every time. It is expected that the implementation of subsequent versions will be more standardized, otherwise integration with business systems will be difficult.

PR #2871, opened 2019-08-14
Make sure SpringBootConfiguration.class load before DataSourceAutoConfiguration.class

ChenTT5:
…ourceAutoConfiguration.class

Fixes #ISSUSE_ID.

-add @AutoConfigureBefore(DataSourceAutoConfiguration.class
-
-


ctruk:
When upgrading from 4.0.0-RC1 to 4.0.0-RC2 or above, the error "URL is not specified" is reported.

After debugging, I discovered that it is because 'com.alibab.druid.spring.boot.autoconfigure.DruidDataSourceAutoConfigure. was loaded in AbstractBeanFactory.doGetBean instead of this class "org.apache.shardingsphere.shardingjdbc.spring.boot.SpringBootConfiguration". As a result, the code you submitted does not take effect.

So I used "@SpringBootApplication (exclude = DruidDataSourceAutoConfigure.class)" to solve it temporarily. Is there any other solution?

PR #2951, opened 2019-08-29
fix issue 2902: Fix bugs in support for spring declarative transactions

liuyangming:
Fixes 2902.

- Fix the problem with the supports for spring declarative transactions, and move the enlistment into the upderling TransactionManager. However, with the addition of a constraint, you can only use the dbcp connection pool management component when using XA transactions. For now, in the open source connection pool management component, only dbcp can support XADataSource.

Please refer to [shardingjdbc-sample](https://github.com/liuyangming/shardingjdbc-sample/tree/issue-2902) for modified usage examples.

### DIFF
--- a/sharding-jdbc/sharding-jdbc-core/src/main/java/org/apache/shardingsphere/shardingjdbc/api/yaml/YamlShardingDataSourceFactory.java
+++ b/sharding-jdbc/sharding-jdbc-core/src/main/java/org/apache/shardingsphere/shardingjdbc/api/yaml/YamlShardingDataSourceFactory.java
@@ -54,6 +54,27 @@ public final class YamlShardingDataSourceFactory {
     /**
      * Create sharding data source.
      *
+     * @param yamlFile YAML file for rule configuration of databases and tables sharding with data sources
+     * @param configClass target class that the YAML configuration needs to be converted
+     * @return sharding data source
+     * @throws SQLException SQL exception
+     * @throws IOException IO exception
+     */
+    public static DataSource createDataSource(final File yamlFile, Class<? extends YamlRootShardingConfiguration> configClass) throws SQLException, IOException {
+		YamlRootShardingConfiguration config = YamlEngine.unmarshal(yamlFile, configClass);
+		try {
+			config.afterPropertiesSet();
+		} catch (RuntimeException error) {
+			throw error;
+		} catch (Exception error) {
+			throw new RuntimeException(error);
+		}
+        return ShardingDataSourceFactory.createDataSource(config.getDataSources(), new ShardingRuleConfigurationYamlSwapper().swap(config.getShardingRule()), config.getProps());
+    }
+    
+    /**
+     * Create sharding data source.
+     *
      * @param yamlBytes YAML bytes for rule configuration of databases and tables sharding with data sources
      * @return sharding data source
      * @throws SQLException SQL exception
@@ -61,6 +82,27 @@ public final class YamlShardingDataSourceFactory {
      */
     public static DataSource createDataSource(final byte[] yamlBytes) throws SQLException, IOException {
         YamlRootShardingConfiguration config = YamlEngine.unmarshal(yamlBytes, YamlRootShardingConfiguration.class);
+        return ShardingDataSourceFactory.createDataSource(config.getDataSources(), new ShardingRuleConfigurationYamlSwapper().swap(config.getShardingRule()), config.getProps());
+    }
+    
+    /**
+     * Create sharding data source.
+     *
+     * @param yamlBytes YAML bytes for rule configuration of databases and tables sharding with data sources
+     * @param configClass target class that the YAML configuration needs to be converted
+     * @return sharding data source
+     * @throws SQLException SQL exception
+     * @throws IOException IO exception
+     */
+    public static DataSource createDataSource(final byte[] yamlBytes, Class<? extends YamlRootShardingConfiguration> configClass) throws SQLException, IOException {
+        YamlRootShardingConfiguration config = YamlEngine.unmarshal(yamlBytes, configClass);
+		try {
+			config.afterPropertiesSet();
+		} catch (RuntimeException error) {
+			throw error;
+		} catch (Exception error) {
+			throw new RuntimeException(error);
+		}
         return ShardingDataSourceFactory.createDataSource(config.getDataSources(), new ShardingRuleConfigurationYamlSwapper().swap(config.getShardingRule()), config.getProps());
     }
     
--- a/sharding-jdbc/sharding-jdbc-core/src/main/java/org/apache/shardingsphere/shardingjdbc/jdbc/adapter/AbstractConnectionAdapter.java
+++ b/sharding-jdbc/sharding-jdbc-core/src/main/java/org/apache/shardingsphere/shardingjdbc/jdbc/adapter/AbstractConnectionAdapter.java
@@ -28,7 +28,6 @@ import org.apache.shardingsphere.core.route.router.masterslave.MasterVisitedMana
 import org.apache.shardingsphere.shardingjdbc.jdbc.adapter.executor.ForceExecuteCallback;
 import org.apache.shardingsphere.shardingjdbc.jdbc.adapter.executor.ForceExecuteTemplate;
 import org.apache.shardingsphere.shardingjdbc.jdbc.unsupported.AbstractUnsupportedOperationConnection;
-import org.apache.shardingsphere.transaction.core.TransactionTypeHolder;
 
 import javax.sql.DataSource;
 import java.sql.Connection;
@@ -204,7 +203,7 @@ public abstract class AbstractConnectionAdapter extends AbstractUnsupportedOpera
     public final void close() throws SQLException {
         closed = true;
         MasterVisitedManager.clear();
-        TransactionTypeHolder.clear();
+        // TransactionTypeHolder.clear();
         int connectionSize = cachedConnections.size();
         try {
             forceExecuteTemplateForClose.execute(cachedConnections.entries(), new ForceExecuteCallback<Entry<String, Connection>>() {
--- a/sharding-jdbc/sharding-jdbc-core/src/main/java/org/apache/shardingsphere/shardingjdbc/jdbc/core/connection/ShardingConnection.java
+++ b/sharding-jdbc/sharding-jdbc-core/src/main/java/org/apache/shardingsphere/shardingjdbc/jdbc/core/connection/ShardingConnection.java
@@ -17,7 +17,15 @@
 
 package org.apache.shardingsphere.shardingjdbc.jdbc.core.connection;
 
-import lombok.Getter;
+import java.sql.Connection;
+import java.sql.DatabaseMetaData;
+import java.sql.PreparedStatement;
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.util.Map;
+
+import javax.sql.DataSource;
+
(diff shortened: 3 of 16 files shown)


cherrylzhao:
@liuyangming thanks for your contribution.
when I see your example [RequiresNewOrderServiceImpl.java](https://github.com/liuyangming/shardingjdbc-sample/blob/issue-2902/src/main/java/com/bytesvc/shardingjdbc/sample/service/impl/RequiresNewOrderServiceImpl.java), I want to clarify our current transaction scope.

1.support distribute transaction in one ShardingConnection. 
2.support distribute transaction using local transaction api. 
3.not support distribute transaction in multiple sharding connections. 
4.not support distribute transaction in JTA way.

it seems that you're trying to do some work on `4.` point, in my opinion, the example code can be only supported by JTA, but this is not what we want to integrate currently.

liuyangming:
> @liuyangming thanks for your contribution.
> when I see your example [RequiresNewOrderServiceImpl.java](https://github.com/liuyangming/shardingjdbc-sample/blob/issue-2902/src/main/java/com/bytesvc/shardingjdbc/sample/service/impl/RequiresNewOrderServiceImpl.java), I want to clarify our current transaction scope.
> 
> 1.support distribute transaction in one ShardingConnection.
> 2.support distribute transaction using local transaction api.
> 3.not support distribute transaction in multiple sharding connections.
> 4.not support distribute transaction in JTA way.
> 
> it seems that you're trying to do some work on `4.` point, in my opinion, the example code can be only supported by JTA, but this is not what we want to integrate currently.

This is a standard usage scenario, regardless of whether or not JTA is used. There is unnecessary to require the business system to abandon the spring declarative transaction. Although such a constraint can be made, in general it is not reasonable. sharding-jdbc should consider this requirement as much as possible.

cherrylzhao:
from my side the example is not the standard local transaction usage, you can try to modify the example to use local transaction only, the transaction will be not effective.

liuyangming:
> from my side the example is not the standard local transaction usage, you can try to modify the example to use local transaction only, the transaction will be not effective.

[Shardingjdbc-sample](https://github.com/liuyangming/shardingjdbc-sample/tree/issue-2902) contains both XA transactions and local transactions, because I verified the XA transaction scenario at the end, so the default configuration is an example of an XA transaction instead of a Local transaction. To change to a Local transaction, just adjust @Import(AtomikosConfiguration.class) to @Import(ShardingConfiguration.class) in the ShardingJdbcSampleBootstrap class. The ShardingConfiguration class explicitly defines a DataSourceTransactionManager that manages local transactions.

liuyangming:
> > from my side the example is not the standard local transaction usage, you can try to modify the example to use local transaction only, the transaction will be not effective.
> 
> [Shardingjdbc-sample](https://github.com/liuyangming/shardingjdbc-sample/tree/issue-2902) contains both XA transactions and local transactions, because I verified the XA transaction scenario at the end, so the default configuration is an example of an XA transaction instead of a Local transaction. To change to a Local transaction, just adjust @import(AtomikosConfiguration.class) to @import(ShardingConfiguration.class) in the ShardingJdbcSampleBootstrap class. The ShardingConfiguration class explicitly defines a DataSourceTransactionManager that manages local transactions.

Sorry, you may also need to exclude the sharding-transaction-xa-atomikos-manager dependency from pom.xml.

cherrylzhao:
let me try to modify your code of example later.

cherrylzhao:
there exist some compile error in your example, I think you should create another branch in your example before you have not modified the source.

cherrylzhao:
I have modified the code in my fork repo, please see [example](https://github.com/cherrylzhao/shardingjdbc-sample), the raw local transaction case is `curl -X POST http://127.0.0.1:8080/order/createThree`,  transaction in this way not supported.

liuyangming:
> I have modified the code in my fork repo, please see [example](https://github.com/cherrylzhao/shardingjdbc-sample), the raw local transaction case is `curl -X POST http://127.0.0.1:8080/order/createThree`, transaction in this way not supported.

Yes, this pr is mainly to adjust the data source from the DataSource to XADataSource. Therefore, the DataSource you mentioned is definitely not supported, and the existing sharding-jdbc implementation also cannot support this. Since sharding-jdbc encapsulates the connection pool component, it must use the XADataSource to make the connection pool component access the TransactionManager to perform the enlistmenet/delistment.

cherrylzhao:
since raw datasourceTransactionManager don’t support to use @transaction like example, it is not necessary to support it in sharingjdbc. This pr have one important point that let user can only select dbcp if they want to use XA in shardingsphere, this is not what we want to see.


2.

sharding-jdbc/sharding-jdbc-core/src/main/java/org/apache/shardingsphere/shardingjdbc/api/yaml/YamlShardingDataSourceFactory.java
```diff
@@ -51,6 +51,27 @@ public static DataSource createDataSource(final File yamlFile) throws SQLExcepti
...
+    /**
+     * Create sharding data source.
+     *
+     * @param yamlFile YAML file for rule configuration of databases and tables sharding with data sources
+     * @param configClass target class that the YAML configuration needs to be converted
+     * @return sharding data source
+     * @throws SQLException SQL exception
+     * @throws IOException IO exception
+     */
+    public static DataSource createDataSource(final File yamlFile, Class<? extends YamlRootShardingConfiguration> configClass) throws SQLException, IOException {
```
cherrylzhao: I think this is only for your example code use, in fact yaml is only one of our configuration, we should not change the api easily.

sharding-jdbc/sharding-jdbc-core/src/main/java/org/apache/shardingsphere/shardingjdbc/jdbc/adapter/AbstractConnectionAdapter.java
```diff
@@ -204,7 +203,7 @@ public void execute(final Connection connection) throws SQLException {
     public final void close() throws SQLException {
         closed = true;
         MasterVisitedManager.clear();
-        TransactionTypeHolder.clear();
+        // TransactionTypeHolder.clear();
```
cherrylzhao: TransactionTypeHolder should always clear after shardingConnection was closed by user.

sharding-jdbc/sharding-jdbc-core/src/main/java/org/apache/shardingsphere/shardingjdbc/jdbc/core/connection/ShardingConnection.java
```diff
@@ -129,42 +134,64 @@ public Statement createStatement(final int resultSetType, final int resultSetCon
...
-                connection.close();
-            }
-        });
-        getCachedConnections().clear();
-    }
+	public void setAutoCommit(final boolean autoCommit) throws SQLException {
+		if (this.originAutoCommit != null && this.originAutoCommit == autoCommit) {
+			return;
+		}
+
```
cherrylzhao: we should always delete redundant blank line for checkstyle, and it seems your incident is wrong.

sharding-jdbc/sharding-jdbc-core/src/main/java/org/apache/shardingsphere/shardingjdbc/jdbc/core/connection/ShardingConnection.java
```diff
@@ -129,42 +134,64 @@ public Statement createStatement(final int resultSetType, final int resultSetCon
...
-            
-            @Override
-            public void execute(final Connection connection) throws SQLException {
-                connection.close();
-            }
-        });
-        getCachedConnections().clear();
-    }
+	public void setAutoCommit(final boolean autoCommit) throws SQLException {
+		if (this.originAutoCommit != null && this.originAutoCommit == autoCommit) {
```
cherrylzhao: we should not skip the setAutoCommit

sharding-jdbc/sharding-jdbc-core/src/main/java/org/apache/shardingsphere/shardingjdbc/jdbc/core/connection/ShardingConnection.java
```diff
@@ -129,42 +134,64 @@ public Statement createStatement(final int resultSetType, final int resultSetCon
...
+		}
+
+		this.transaction = this.shardingTransactionManager.suspend();
+		shardingTransactionManager.begin();
+		this.originAutoCommit = autoCommit;
+		this.closeCachedConnections();
+
+		// if (autoCommit && !shardingTransactionManager.isInTransaction() || !autoCommit &&
+		// shardingTransactionManager.isInTransaction()) {
+		// return;
```
cherrylzhao: template code should not always appear.

sharding-jdbc/sharding-jdbc-core/src/main/java/org/apache/shardingsphere/shardingjdbc/jdbc/core/connection/ShardingConnection.java
```diff
@@ -129,42 +134,64 @@ public Statement createStatement(final int resultSetType, final int resultSetCon
...
+		getCachedConnections().clear();
+	}
     
     @Override
     public void commit() throws SQLException {
         if (TransactionType.LOCAL == transactionType) {
             super.commit();
         } else {
             shardingTransactionManager.commit();
+            if (this.transaction != null) {
```
cherrylzhao: both transaction and shardingTransacitonMananger appear on ShardingConnection, it is very Confused. transaction should managered by TransactionManager.

sharding-jdbc/sharding-jdbc-core/src/main/java/org/apache/shardingsphere/shardingjdbc/jdbc/core/connection/ShardingConnection.java
```diff
@@ -52,6 +55,8 @@
     private final TransactionType transactionType;
     
     private final ShardingTransactionManager shardingTransactionManager;
+    private transient Object transaction = null;
+    private transient Boolean originAutoCommit = null;
```
cherrylzhao: ShardingConnection is not pooledConnection, it will new from datasource everytime, so we should not keep transaction and originalAutoCommit status here.

sharding-transaction/sharding-transaction-2pc/sharding-transaction-xa/sharding-transaction-xa-core/src/main/java/org/apache/shardingsphere/transaction/xa/XAShardingTransactionManager.java
```diff
@@ -84,10 +86,10 @@ public boolean isInTransaction() {
     @Override
     public Connection getConnection(final String dataSourceName) throws SQLException {
         SingleXAConnection singleXAConnection = singleXADataSourceMap.get(dataSourceName).getXAConnection();
-        if (!enlistedXAResource.get().contains(dataSourceName)) {
-            xaTransactionManager.enlistResource(singleXAConnection.getXAResource());
-            enlistedXAResource.get().add(dataSourceName);
-        }
+		// if (!enlistedXAResource.get().contains(dataSourceName)) {
+		// xaTransactionManager.enlistResource(singleXAConnection.getXAResource());
```
cherrylzhao: we should not make XA only work on dbcp

PR #2978, opened 2019-09-05
Filter out incorrect table which with only alias.

sunbufu:
Fixe https://github.com/apache/incubator-shardingsphere/issues/2961.

1. filter out incorrect table which with only alias.
2. modify UT.


terrymanu:
Can you add test case for sql `delete xxx from tbl AS xxx WHERE xxx`?

PR #2996, opened 2019-09-07
Feature#2926 replace int with long for limit parameter.

sunbufu:
Fixes https://github.com/apache/incubator-shardingsphere/issues/2926.

- replace int with long for limit parameter.
- modify some UT classes.


terrymanu:
Wonderful pull request

PR #2999, opened 2019-09-07
Replay for ShardingPreparedStatement when execute.

sunbufu:
Fixes https://github.com/apache/incubator-shardingsphere/issues/2981.

1. add replayMethodForStatements for `ShardingPreparedStatement`.
2. add UT.


tristaZero:
@sunbufu Not only the bugfix but also the test was added, thanks.

PR #3054, opened 2019-09-18
Replace localhost to real ip address

yanyzy:
Complete TODO of replace localhost to real ip address for MySQLAuthenticationEngine


tuohai666:
What happened?

yanyzy:
@tuohai666 
I get an error when I execute the command `mvn clean install`, i think this may be caused by my error submission. So I want to solve it and submit again.

PR #3056, opened 2019-09-18
Replace localhost to real ip address

yanyzy:
Complete TODO of replace localhost to real ip address for MySQLAuthenticationEngine

### DIFF
--- a/sharding-proxy/sharding-proxy-frontend/sharding-proxy-frontend-mysql/src/main/java/org/apache/shardingsphere/shardingproxy/frontend/mysql/auth/MySQLAuthenticationEngine.java
+++ b/sharding-proxy/sharding-proxy-frontend/sharding-proxy-frontend-mysql/src/main/java/org/apache/shardingsphere/shardingproxy/frontend/mysql/auth/MySQLAuthenticationEngine.java
@@ -32,6 +32,9 @@ import org.apache.shardingsphere.shardingproxy.transport.mysql.packet.handshake.
 import org.apache.shardingsphere.shardingproxy.transport.mysql.payload.MySQLPacketPayload;
 import org.apache.shardingsphere.shardingproxy.transport.payload.PacketPayload;
 
+import java.net.InetSocketAddress;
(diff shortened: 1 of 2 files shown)


sharding-proxy/sharding-proxy-frontend/sharding-proxy-frontend-mysql/src/main/java/org/apache/shardingsphere/shardingproxy/frontend/mysql/auth/MySQLAuthenticationEngine.java
```diff
@@ -60,9 +63,10 @@ public boolean auth(final ChannelHandlerContext context, final PacketPayload pay
             backendConnection.setUserName(response41.getUsername());
             context.writeAndFlush(new MySQLOKPacket(response41.getSequenceId() + 1));
         } else {
-            // TODO localhost should replace to real ip address
+            SocketAddress socketAddress = context.channel().remoteAddress();
+            String hostAddress = socketAddress instanceof InetSocketAddress ? ((InetSocketAddress) socketAddress).getAddress().getHostAddress() : socketAddress.toString();
```
tuohai666: What if the socketAddress is null? You can find the null value possibility in the comment of API.

yanyzy: I find “null if this channel is not connected” in the API comment, but this method is called by auth method, i think the channel is connected

tuohai666: Yes, you are right.

terrymanu:
Thanks for contribution

grissomsh:
Great!

PR #3079, opened 2019-09-21
Fixes #3068 Add a judgment for user's database privileges.

### LINKED ISSUE #3068: Sharding-Proxy: Configuring user access to the database is invalid
dev

Sharding-Proxy

### Expected behavior
According to the configuration file, users are prohibited from accessing databases that are not authorized for them.

### Actual behavior
User does not have access to a database，but this user can access it.

I have two config yaml to configure two logic datasource, user access is configured in the sever.yaml, this user has only one of the database permissions, but he can access both database.

config-sharding.yaml：
```yaml
schemaName: sharding_db

dataSources:
  ds0:
   ···
  ds1:
   ···
```

config-sharding-remote.yaml：
```yaml
schemaName: sharding_db_remote

dataSources:
  ds0:
    ···
  ds1:
    ···
```

server.yaml：
```yaml
authentication:
  users:
    root:
      password: root
    sharding:
      password: sharding
      authorizedSchemas: sharding_db
    zhyee:
      password: zhyee
      authorizedSchemas: sharding_db_remote
```

my Springboot demo configuration is:
```yaml
spring:
  datasource:
    url: jdbc:mysql://localhost:3307/sharding_db
    driver-class-name: com.mysql.jdbc.Driver
    username: zhyee
    password: zhyee
```
and it works.

sunbufu:
Fixes https://github.com/apache/incubator-shardingsphere/issues/3068.

1. modify MySQLAuthenticationHandler's login method to return Optional<ProxyUser>.
2. check response41 database with ProxyUser's database.
3. add UT.


sharding-proxy/sharding-proxy-frontend/sharding-proxy-frontend-mysql/src/main/java/org/apache/shardingsphere/shardingproxy/frontend/mysql/auth/MySQLAuthenticationEngine.java
```diff
@@ -54,20 +56,31 @@ public void handshake(final ChannelHandlerContext context, final BackendConnecti
     @Override
     public boolean auth(final ChannelHandlerContext context, final PacketPayload payload, final BackendConnection backendConnection) {
         MySQLHandshakeResponse41Packet response41 = new MySQLHandshakeResponse41Packet((MySQLPacketPayload) payload);
-        if (authenticationHandler.login(response41.getUsername(), response41.getAuthResponse())) {
-            if (!Strings.isNullOrEmpty(response41.getDatabase()) && !LogicSchemas.getInstance().schemaExists(response41.getDatabase())) {
+        final Optional<ProxyUser> user = authenticationHandler.login(response41.getUsername(), response41.getAuthResponse());
+        if (!user.isPresent()) {
```
tristaZero: **Hi, Haisheng, i appreciate your bugfix and  unit tests!**
Here is my suggestion, can you consider it is better or not?

> How about move the judegment of user existence and its authorized schemas to `authenticationHandler.login()`, therefore `AuthenticationHandler` will just tell `MySQLAuthenticationEngine` that the login information is true or not. And the task of `MySQLAuthenticationEngine` is only to set properties for `backendConnection` or call `context.writeAndFlush(getAccessDeniedError(context, response41))` by login information.

sunbufu: OK 👍

sharding-proxy/sharding-proxy-frontend/sharding-proxy-frontend-mysql/src/main/java/org/apache/shardingsphere/shardingproxy/frontend/mysql/auth/MySQLAuthenticationEngine.java
```diff
@@ -54,11 +54,11 @@ public void handshake(final ChannelHandlerContext context, final BackendConnecti
...
-        if (authenticationHandler.login(response41.getUsername(), response41.getAuthResponse())) {
-            if (!Strings.isNullOrEmpty(response41.getDatabase()) && !LogicSchemas.getInstance().schemaExists(response41.getDatabase())) {
-                context.writeAndFlush(new MySQLErrPacket(response41.getSequenceId() + 1, MySQLServerErrorCode.ER_BAD_DB_ERROR, response41.getDatabase()));
-                return true;
-            }
+        if (!Strings.isNullOrEmpty(response41.getDatabase()) && !LogicSchemas.getInstance().schemaExists(response41.getDatabase())) {
+            context.writeAndFlush(new MySQLErrPacket(response41.getSequenceId() + 1, MySQLServerErrorCode.ER_BAD_DB_ERROR, response41.getDatabase()));
+            return true;
+        }
+        if (authenticationHandler.login(response41.getUsername(), response41.getAuthResponse(), response41.getDatabase())) {
```
tristaZero: Will it be better if we pass `response41` to `authenticationHandler.login()` instead of `username, pwd, db`?

sharding-proxy/sharding-proxy-transport/sharding-proxy-transport-mysql/src/main/java/org/apache/shardingsphere/shardingproxy/transport/mysql/packet/handshake/MySQLAuthenticationHandler.java
```diff
@@ -44,11 +45,17 @@
...
-    public boolean login(final String username, final byte[] authResponse) {
+    public boolean login(final String username, final byte[] authResponse, final String schema) {
         Optional<ProxyUser> user = getUser(username);
-        return user.isPresent() && (Strings.isNullOrEmpty(user.get().getPassword()) || Arrays.equals(getAuthCipherBytes(user.get().getPassword()), authResponse));
+        if (!user.isPresent()) {
+            return false;
+        }
+        boolean isPasswordRight = Strings.isNullOrEmpty(user.get().getPassword()) || Arrays.equals(getAuthCipherBytes(user.get().getPassword()), authResponse);
+        boolean isAuthorizedSchema = null == schema || CollectionUtils.isEmpty(user.get().getAuthorizedSchemas()) || user.get().getAuthorizedSchemas().contains(schema);
+        return isPasswordRight && isAuthorizedSchema;
```
tristaZero: Oh, i like this expression which makes code more clear. 👍

sharding-proxy/sharding-proxy-frontend/sharding-proxy-frontend-mysql/src/main/java/org/apache/shardingsphere/shardingproxy/frontend/mysql/auth/MySQLAuthenticationEngine.java
sunbufu: Good idea.

yanyzy:
Hi, i have a suggestion. i check the [Server Error Message Reference](https://dev.mysql.com/doc/refman/5.7/en/server-error-reference.html), i think the error message of access denied for user should be different from access denied for user  to database.

sunbufu:
> Hi, i have a suggestion. i check the [Server Error Message Reference](https://dev.mysql.com/doc/refman/5.7/en/server-error-reference.html), i think the error message of access denied for user should be different from access denied for user to database.

Good job! I think we should consider this.

tristaZero:
@yanyzy yanyzy  Hi, thanks your issue and suggestion, and @sunbufu Your pr is great. 👍

PR #3093, opened 2019-09-23
Fix SCTL statement execute error with comments

### LINKED ISSUE #3092: SCTL statement get error when sql contains comments

Dev

Sharding-Proxy

### Expected behavior

SCTL statement can get correct results

### Actual behavior

When sql contains comments, the stcl statement will report an error.

First i use DataGrip to execute command, and it get an error

image

so, i use command line tool to execute it, it works fine. but if sql has comments, will get error.
image

yanyzy:
Fixes #3092.

add trim to support sctl statement with comments

### DIFF
--- a/sharding-proxy/sharding-proxy-backend/src/main/java/org/apache/shardingsphere/shardingproxy/backend/text/TextProtocolBackendHandlerFactory.java
+++ b/sharding-proxy/sharding-proxy-backend/src/main/java/org/apache/shardingsphere/shardingproxy/backend/text/TextProtocolBackendHandlerFactory.java
@@ -50,6 +50,12 @@ import org.apache.shardingsphere.transaction.core.TransactionOperationType;
  */
 @NoArgsConstructor(access = AccessLevel.PRIVATE)
 public final class TextProtocolBackendHandlerFactory {
+
+    private static final String COMMENT_PREFIX = "/*";
+
+    private static final String COMMENT_SUFFIX = "*/";
+
+    private static final String SQL_END = ";";
     
     /**
      * Create new instance of text protocol backend handler.
@@ -60,8 +66,9 @@ public final class TextProtocolBackendHandlerFactory {
      * @return instance of text protocol backend handler
      */
     public static TextProtocolBackendHandler newInstance(final DatabaseType databaseType, final String sql, final BackendConnection backendConnection) {
-        if (sql.toUpperCase().startsWith(ShardingCTLBackendHandlerFactory.SCTL)) {
-            return ShardingCTLBackendHandlerFactory.newInstance(sql, backendConnection);
+        final String trimCommentSQL = trimSqlComment(sql);
+        if (trimCommentSQL.toUpperCase().startsWith(ShardingCTLBackendHandlerFactory.SCTL)) {
+            return ShardingCTLBackendHandlerFactory.newInstance(trimCommentSQL, backendConnection);
         }
         SQLStatement sqlStatement = new SQLParseKernel(ParseRuleRegistry.getInstance(), databaseType, sql).parse();
         if (sqlStatement instanceof TCLStatement) {
@@ -103,5 +110,16 @@ public final class TextProtocolBackendHandlerFactory {
             return new BroadcastBackendHandler(sql, backendConnection);
         }
         return new UnicastBackendHandler(sql, backendConnection);
+    }
+
+    private static String trimSqlComment(final String sql) {
+        String result = sql;
+        if (sql.startsWith(COMMENT_PREFIX)) {
+            result = sql.substring(sql.indexOf(COMMENT_SUFFIX) + 2).trim();
+        }
+        if (sql.endsWith(SQL_END)) {
+            result = result.substring(0, result.length() - 1);
+        }
+        return result;
     }
 }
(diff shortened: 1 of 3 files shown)


sharding-proxy/sharding-proxy-backend/src/main/java/org/apache/shardingsphere/shardingproxy/backend/text/TextProtocolBackendHandlerFactory.java
```diff
@@ -60,8 +66,9 @@
      * @return instance of text protocol backend handler
      */
     public static TextProtocolBackendHandler newInstance(final DatabaseType databaseType, final String sql, final BackendConnection backendConnection) {
-        if (sql.toUpperCase().startsWith(ShardingCTLBackendHandlerFactory.SCTL)) {
-            return ShardingCTLBackendHandlerFactory.newInstance(sql, backendConnection);
+        final String trimCommentSQL = trimSqlComment(sql);
```
cherrylzhao: it's better to move trimComentSQL logic to `ShardingCTLBackendHandlerFactory`

sharding-proxy/sharding-proxy-backend/src/main/java/org/apache/shardingsphere/shardingproxy/backend/text/TextProtocolBackendHandlerFactory.java
```diff
@@ -60,8 +66,9 @@
...
+        if (trimCommentSQL.toUpperCase().startsWith(ShardingCTLBackendHandlerFactory.SCTL)) {
+            return ShardingCTLBackendHandlerFactory.newInstance(trimCommentSQL, backendConnection);
```
cherrylzhao: so here should use contains method, then trim in ShardingCTLBackendHandlerFactory

sharding-proxy/sharding-proxy-backend/src/main/java/org/apache/shardingsphere/shardingproxy/backend/text/TextProtocolBackendHandlerFactory.java
```diff
@@ -104,4 +111,15 @@ private static TextProtocolBackendHandler createDALBackendHandler(final DALState
         }
         return new UnicastBackendHandler(sql, backendConnection);
     }
+
+    private static String trimSqlComment(final String sql) {
+        String result = sql;
+        if (sql.startsWith(COMMENT_PREFIX)) {
+            result = sql.substring(sql.indexOf(COMMENT_SUFFIX) + 2).trim();
+        }
+        if (sql.endsWith(SQL_END)) {
```
cherrylzhao: `;` will not be passed to proxy, please verify it again.

sharding-proxy/sharding-proxy-backend/src/main/java/org/apache/shardingsphere/shardingproxy/backend/text/TextProtocolBackendHandlerFactory.java
yanyzy: Yes, it looks clearer. but if sql is like 
```bash
select x from table where z="sctl:"
```
this situation is rare but also possible.

sharding-proxy/sharding-proxy-backend/src/main/java/org/apache/shardingsphere/shardingproxy/backend/text/TextProtocolBackendHandlerFactory.java
yanyzy: image
I think it may be that the client is unable to recognize the statement, resulting in no processing.

sharding-proxy/sharding-proxy-backend/src/main/java/org/apache/shardingsphere/shardingproxy/backend/text/TextProtocolBackendHandlerFactory.java
cherrylzhao: I have understood it. so I suggest you add a SCTLUtils class, it should include 2 method.

- isSCTL(final string sql )
   use regx to judge it 
- trimComment

are you agree with it?

sharding-proxy/sharding-proxy-backend/src/main/java/org/apache/shardingsphere/shardingproxy/backend/text/TextProtocolBackendHandlerFactory.java
cherrylzhao: ok, it beyond client, let's retain this logic

sharding-proxy/sharding-proxy-backend/src/main/java/org/apache/shardingsphere/shardingproxy/backend/text/TextProtocolBackendHandlerFactory.java
yanyzy: Add a SCTLUtils class is a good idea! i will modify it. but i think use regx will get complicated, we need to consider sql comments, it is outside the control. on the other hand, "SCTL:" may exist in the conditional  statement. so my advice is remove comments first, and then use startWith method to judge it. what do you think of it?

cherrylzhao: if we remove comment first, it may effect non-sctl SQL, our parse engine can recognize these 
hint syntax. so I think we should just remove comment of sctl

yanyzy: Yes, so the trim method just for sctl statement,  will not affect other statements. i modified the code this way. what do you think of it?

cherrylzhao: ok, I have got it, it's no problem.

PR #3199, opened 2019-10-09
improved coverage for package of preprocessor

sluk3r:
testcase code added into classes of ProjectionsContextEngineTest and ProjectionEngineTest


sharding-core/sharding-core-preprocessor/src/test/java/org/apache/shardingsphere/core/optimize/segment/select/projection/engine/ProjectionEngineTest.java
```diff
@@ -45,4 +55,38 @@ public void assertProjectionCreatedWhenSelectItemSegmentInstanceOfShorthandSelec
...
         assertTrue(projection.get() instanceof ShorthandProjection);
     }
+    
+    @Test
+    public void assertProjectionCreatedWhenSelectItemSegmentInstanceOfColumnSelectItemSegment() {
+        ColumnSelectItemSegment columnSelectItemSegment = new ColumnSelectItemSegment("text", new ColumnSegment(0, 10, "name"));
+        columnSelectItemSegment.setAlias("alias");
+        Optional<Projection> projection = new ProjectionEngine().createProjection(null, columnSelectItemSegment);
+        assertTrue(projection.isPresent());
+        assertTrue(projection.get() instanceof ColumnProjection);
```
terrymanu: Should use assertThat and instanceOf to instead of assertTrue

PR #3200, opened 2019-10-09
improved coverage for package of transaction

sluk3r:
coverage improved for ResourceIDGenerator and XAShardingTransactionManager


sharding-transaction/sharding-transaction-core/src/test/java/org/apache/shardingsphere/transaction/core/ResourceIDGeneratorTest.java
```diff
@@ -0,0 +1,30 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.shardingsphere.transaction.core;
+
+import org.junit.Test;
+
+import static org.junit.Assert.assertTrue;
+
+public class ResourceIDGeneratorTest {
```
terrymanu: Should use final for non-extended class

PR #3249, opened 2019-10-14
#3240, the OR keyword should not appear in simpleExpr

### LINKED ISSUE #3240: SQL92 parse or expression have problem
When using sql92 to parse sql, the expression is error, but MySQL dialect is ok.

```sql
DELETE FROM t_cipher_assisted_query_plain WHERE encrypt_1 = ? or encrypt_2 IN (?,?,?)
```

It maybe cause by `OR`, we need to investigate it.

tuohai666:
Fixes #3240.


terrymanu:
Please create a new branch to submit pull request to avoid too more old comments

Issue #3213, opened 2019-10-10
There are two methods which are very like each other in DerivedColumn.class


wgy8283335:
## There are two methods which are very like each other in DerivedColumn.class

org.apache.shardingsphere.core.optimize.segment.select.projection.DerivedColumn

```
    /**
     * Judge is derived column name or not.
     *
     * @param columnName column name to be judged
     * @return is derived column name or not
     */
    public static boolean isDerivedColumnName(final String columnName) {
        for (DerivedColumn each : DerivedColumn.values()) {
            if (columnName.startsWith(each.pattern)) {
                return true;
            }
        }
        return false;
    }
    
    /**
     * Judge is derived column or not.
     *
     * @param columnName column name to be judged
     * @return is derived column or not
     */
    public static boolean isDerivedColumn(final String columnName) {
        for (DerivedColumn each : DerivedColumn.getValues()) {
            if (columnName.startsWith(each.pattern)) {
                return true;
            }
        }
        return false;
    }

```

terrymanu:
It maybe cause by merge issue, we can keep the method `isDerivedColumn` and remove other one.
Does some one want to submit a pull request?

wgy8283335:
@terrymanu  I'd like to make this pull request.

terrymanu:
I check the code carefully, they are different with that 2 methods.

In method `isDerivedColumnName`, it use `DerivedColumn.values()`;
In method `isDerivedColumn`, it use `DerivedColumn.getValues()`;

The different of `DerivedColumn.values()` and `DerivedColumn.getValues()` is deal with AGGREGATION_DISTINCT_DERIVED.

So we cannot just merge those 2 methods, need refactor further. So I just close this pr, please keep focus is you want to continue refactor this method, thank you.

PR #3263, opened 2019-10-14
Delete duplicate method: isDerivedColumnName

tcmtang:
Fix the problem #3213

Delete duplicate method: isDerivedColumnName


terrymanu:
I just carry on the comment from #3213 


I check the code carefully, they are different with that 2 methods.

In method `isDerivedColumnName`, it use `DerivedColumn.values()`;
In method `isDerivedColumn`, it use `DerivedColumn.getValues()`;

The different of `DerivedColumn.values()` and `DerivedColumn.getValues()` is deal with AGGREGATION_DISTINCT_DERIVED.

So we cannot just merge those 2 methods, need refactor further. So I just close this pr, please keep focus is you want to continue refactor this method, thank you.

PR #3268, opened 2019-10-15
refine DerviedColumn method which is dependent on the core path of query

gufengwyx8:
- refine method isDerivedColumn that it will reuse a same object for relative collection


gufengwyx8:
> The code readability may not increase during this pull request, cloud you consider about to redesign it?

This pr is intent to improve the performance of GC, consider that the original code will create ArrayList everytime.

PR #3276, opened 2019-10-15
Add table owner for logic encrypt column

### LINKED ISSUE #3262: Table owner is lost for logic encrypt column
## Bug Report

### Bug
When logic encrypt column has owner in Query SQL, its assistedQueryColumn, cipherColumn and plainColumn can not be rewrite in actual SQLs correctly.

**4.0.0-RC2**

### Description
Considering Encrypt Rule is,
```
encryptRule:
  tables:
    t_account_bak:
      columns:
        password:
          cipherColumn: cipher_password
          assistedQueryColumn: assisted_query_password
          plainColumn: plain_password
          encryptor: assisted_query
        amount:
          cipherColumn: cipher_amount
          plainColumn: plain_amount
          encryptor: normal
props:
  query.with.cipher.column: true
```
And SQL is,
> SELECT a.account_id, a.password, a.amount AS a, a.status AS s FROM t_account_bak AS a WHERE a.account_id = 1 AND a.password = 'aaa' AND a.amount = 1000 AND a.status = 'OK'

The correct rewritten SQL is,
> SELECT a.account_id, a.cipher_password, a.cipher_amount AS a, a.status AS s FROM t_account_bak AS a WHERE a.account_id = 1 AND a.assisted_query_password = 'assisted_query_aaa' AND a.cipher_amount = 'encrypt_1000' AND a.status = 'OK'

But currently it is rewritten as,
>SELECT a.account_id, a.cipher_password, a.cipher_amount AS a, a.status AS s FROM t_account_bak AS a WHERE a.account_id = 1 AND assisted_query_password = 'assisted_query_aaa' AND cipher_amount = 'encrypt_1000' AND a.status = 'OK'

**As you can see, the table owner,`a` of `assisted_query_password` and `cipher_amount` is lost.** 😢 

### How to make it recurred?
1. Modify [select_for_query_with_cipher.xml](https://github.com/apache/incubator-shardingsphere/blob/dev/sharding-core/sharding-core-rewrite/src/test/resources/encrypt/select_for_query_with_cipher.xml) in ShardingSphere project, and make `the correct ouput` valid.
Here it is,
```
 <!--Here is the correct rewritten SQL. If you want to test and fix me, please note the above output, and make me valid.-->
 <!--<output sql="SELECT a.account_id, a.cipher_password, a.cipher_amount AS a, a.status AS s FROM t_account_bak AS a WHERE a.account_id = 1 AND a.assisted_query_password = 'assisted_query_aaa' AND a.cipher_amount = 'encrypt_1000' AND a.status = 'OK'" />-->
```
2.Run [Related tests](https://github.com/apache/incubator-shardingsphere/blob/dev/sharding-core/sharding-core-rewrite/src/test/java/org/apache/shardingsphere/core/rewrite/parameterized/engine/EncryptSQLRewriteEngineParameterizedTest.java)
3.AssertionError brings up.

❗️ **Therefore our task is to make this AssertionError disappear.**

### Why does it happen?
1.After parsing, the objects of cipherColumn, assistedQueryColumn, plainColumn will be stored as 
`EncryptPredicateToken`, which does not keep `table owner`. 
2.When rewriting SQL, rewriter will call `toString()` of `EncryptPredicateToken`, but there is no `table owner`, so the result of `toString()` is `assisted_query_password` and `cipher_amount`, not `a.assisted_query_password` and `a.cipher_amount`.

### How to fix it?
1. To make `EncryptPredicateToken` keep `table owner`, instead of putting `columnName` in it, you should make `ColumnSegment` as its member.
2. `Table owner` and `column name` are the members of `ColumnSegment`.
3.Modify toString() of `EncryptPredicateToken` to append `tableName` from `ColumnSegment` to result.

#### Looking forward to your participation.

gufengwyx8:
Fixes #3262.

- Add table owner for logic encrypt column
- Fix unit test

### DIFF
--- a/sharding-core/sharding-core-parse/sharding-core-parse-common/src/main/java/org/apache/shardingsphere/core/parse/sql/segment/dml/column/ColumnSegment.java
+++ b/sharding-core/sharding-core-parse/sharding-core-parse-common/src/main/java/org/apache/shardingsphere/core/parse/sql/segment/dml/column/ColumnSegment.java
@@ -66,15 +66,6 @@ public class ColumnSegment implements SQLSegment, PredicateRightValue, OwnerAvai
         return null == owner ? name : owner.getTableName() + "." + name;
     }
 
-    /**
-     * Get owner name.
-     *
-     * @return owner name
-     */
-    public final String getOwnerName() {
-        return null == owner ? null : owner.getTableName();
-    }
-    
     @Override
     public final Optional<TableSegment> getOwner() {
         return Optional.fromNullable(owner);
--- a/sharding-core/sharding-core-parse/sharding-core-parse-common/src/main/java/org/apache/shardingsphere/core/parse/sql/segment/dml/column/ColumnSegment.java
+++ b/sharding-core/sharding-core-parse/sharding-core-parse-common/src/main/java/org/apache/shardingsphere/core/parse/sql/segment/dml/column/ColumnSegment.java
@@ -65,6 +65,15 @@ public class ColumnSegment implements SQLSegment, PredicateRightValue, OwnerAvai
     public final String getQualifiedName() {
         return null == owner ? name : owner.getTableName() + "." + name;
     }
+
+    /**
+     * Get owner name.
+     *
+     * @return owner name
+     */
+    public final String getOwnerName() {
+        return null == owner ? null : owner.getTableName();
+    }
     
     @Override
     public final Optional<TableSegment> getOwner() {
--- a/sharding-core/sharding-core-rewrite/src/main/java/org/apache/shardingsphere/core/rewrite/feature/encrypt/EncryptCondition.java
+++ b/sharding-core/sharding-core-rewrite/src/main/java/org/apache/shardingsphere/core/rewrite/feature/encrypt/EncryptCondition.java
@@ -46,7 +46,7 @@ public final class EncryptCondition {
     
     private final String tableName;
 
-    private final String tableOwner;
+    private final String columnOwner;
     
     private final int startIndex;
     
@@ -58,20 +58,20 @@ public final class EncryptCondition {
     
     private final Map<Integer, Object> positionValueMap = new LinkedHashMap<>();
     
-    public EncryptCondition(final String columnName, final String tableName, final String tableOwner, final int startIndex, final int stopIndex, final ExpressionSegment expressionSegment) {
+    public EncryptCondition(final String columnName, final String tableName, final String columnOwner, final int startIndex, final int stopIndex, final ExpressionSegment expressionSegment) {
         this.columnName = columnName;
         this.tableName = tableName;
-        this.tableOwner = tableOwner;
+        this.columnOwner = columnOwner;
         this.startIndex = startIndex;
         this.stopIndex = stopIndex;
         operator = ShardingOperator.EQUAL;
         putPositionMap(0, expressionSegment);
     }
     
-    public EncryptCondition(final String columnName, final String tableName, final String tableOwner, final int startIndex, final int stopIndex, final List<ExpressionSegment> expressionSegments) {
+    public EncryptCondition(final String columnName, final String tableName, final String columnOwner, final int startIndex, final int stopIndex, final List<ExpressionSegment> expressionSegments) {
         this.columnName = columnName;
         this.tableName = tableName;
-        this.tableOwner = tableOwner;
+        this.columnOwner = columnOwner;
         this.startIndex = startIndex;
         this.stopIndex = stopIndex;
         operator = ShardingOperator.IN;
--- a/sharding-core/sharding-core-rewrite/src/main/java/org/apache/shardingsphere/core/rewrite/feature/encrypt/EncryptCondition.java
+++ b/sharding-core/sharding-core-rewrite/src/main/java/org/apache/shardingsphere/core/rewrite/feature/encrypt/EncryptCondition.java
@@ -45,6 +45,8 @@ public final class EncryptCondition {
     private final String columnName;
     
     private final String tableName;
+
+    private final String tableOwner;
     
     private final int startIndex;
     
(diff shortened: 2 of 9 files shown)


tristaZero:
@gufengwyx8 Hi, So great and so fast!
The solution in your PR is correct, but to make the members of `EncryptCondition` and `EncryptPredicateToken` less, can you keep `ColumnSegment` in `EncryptPredicateToken` and delete `columnName`? Through that, we can get `columnName` and `its owner` from `ColumnSegment` in `ColumnSegment`.

gufengwyx8:
> @gufengwyx8 Hi, So great and so fast!
> The solution in your PR is correct, but to make the members of `EncryptCondition` and `EncryptPredicateToken` less, can you keep `ColumnSegment` in `EncryptPredicateToken` and delete `columnName`? Through that, we can get `columnName` and `its owner` from `ColumnSegment` in `ColumnSegment`.

sure, I can refine it.

tristaZero:
@gufengwyx8 Hi, you are right, and please review my comments. Waiting for your reply.

sharding-core/sharding-core-rewrite/src/main/java/org/apache/shardingsphere/core/rewrite/feature/encrypt/EncryptCondition.java
```diff
@@ -56,18 +58,20 @@
...
         this.stopIndex = stopIndex;
         operator = ShardingOperator.EQUAL;
         putPositionMap(0, expressionSegment);
     }
     
-    public EncryptCondition(final String columnName, final String tableName, final int startIndex, final int stopIndex, final List<ExpressionSegment> expressionSegments) {
+    public EncryptCondition(final String columnName, final String tableName, final String tableOwner, final int startIndex, final int stopIndex, final List<ExpressionSegment> expressionSegments) {
         this.columnName = columnName;
         this.tableName = tableName;
+        this.tableOwner = tableOwner;
```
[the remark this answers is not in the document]
gufengwyx8: Actually tableName and tableOwner are diffierent, tableName is the original name, tableOwner is the alias instead.

sharding-core/sharding-core-parse/sharding-core-parse-common/src/main/java/org/apache/shardingsphere/core/parse/sql/segment/dml/column/ColumnSegment.java
```diff
@@ -65,6 +65,15 @@ public ColumnSegment(final int startIndex, final int stopIndex, final String nam
...
         return null == owner ? name : owner.getTableName() + "." + name;
     }
+
+    /**
+     * Get owner name.
+     *
+     * @return owner name
+     */
+    public final String getOwnerName() {
+        return null == owner ? null : owner.getTableName();
```
tristaZero: If owner is a alias, is it possible this function returns its alias？

sharding-core/sharding-core-rewrite/src/main/java/org/apache/shardingsphere/core/rewrite/feature/encrypt/EncryptCondition.java
```diff
@@ -56,18 +58,20 @@
...
         this.startIndex = startIndex;
         this.stopIndex = stopIndex;
         operator = ShardingOperator.EQUAL;
         putPositionMap(0, expressionSegment);
     }
     
-    public EncryptCondition(final String columnName, final String tableName, final int startIndex, final int stopIndex, final List<ExpressionSegment> expressionSegments) {
```
tristaZero: Do you think it is better to store tableAlias instead of tableOwner, for people will puzzle by tableName and tableOwner from this list of parameters. Therefore if tableAlias is nulll, we can pass tableName to `EncryptPredicateToken` as a tableOwner, otherwise tableAlias will be tableOwner. What' your opinion?

sharding-core/sharding-core-parse/sharding-core-parse-common/src/main/java/org/apache/shardingsphere/core/parse/sql/segment/dml/column/ColumnSegment.java
gufengwyx8: This func returns the alias truly, you can check the unit test `EncryptSQLRewriteEngineParameterizedTest`

sharding-core/sharding-core-rewrite/src/main/java/org/apache/shardingsphere/core/rewrite/feature/encrypt/EncryptCondition.java
gufengwyx8: Actually the variable `tableOwner` and `tableAlias` have the same meaning, I can rename `tableOwner` to `tableAlias`.  But if I change the logic as you said, 
the SQL
`select * from table where pwd = 1` 
will be rewritten as 
`select * from table where table.cipher  = xxx`
This result SQL is right?

sharding-core/sharding-core-parse/sharding-core-parse-common/src/main/java/org/apache/shardingsphere/core/parse/sql/segment/dml/column/ColumnSegment.java
tristaZero: Thanks for your reply, there are tableName and alias in `TableSegment`, when getTableName(), it should not return alias actually. But you are right, getTableName() return alias in `EncryptSQLRewriteEngineParameterizedTest` , which is an another issue.

tristaZero:
Thanks for your reply, there are tableName and alias in TableSegment, when getTableName(), it should not return alias actually. But you are right, getTableName() return alias in EncryptSQLRewriteEngineParameterizedTest , which is an another issue.
@gufengwyx8 

I approve of your pr, and here are just  two items we can consider,
1. Rename `tableOwner` to `columnOwner`, which means the owner of colum.
2. Instead of adding a new function `getOwnerName()`, can we reuse `getOwner()` ?
That' all, waiting for your suggetions.

gufengwyx8:
> Thanks for your reply, there are tableName and alias in TableSegment, when getTableName(), it should not return alias actually. But you are right, getTableName() return alias in EncryptSQLRewriteEngineParameterizedTest , which is an another issue.
> @gufengwyx8
> 
> I approve of your pr, and here are just two items we can consider,
> 
> 1. Rename `tableOwner` to `columnOwner`, which means the owner of colum.
> 2. Instead of adding a new function `getOwnerName()`, can we reuse `getOwner()` ?
>    That' all, waiting for your suggetions.

I just follow your suggestion

sharding-core/sharding-core-rewrite/src/main/java/org/apache/shardingsphere/core/rewrite/feature/encrypt/token/pojo/EncryptPredicateToken.java
```diff
@@ -32,22 +32,26 @@
  * @author panjuan
  */
 public final class EncryptPredicateToken extends SQLToken implements Substitutable {
-    
+
     @Getter
     private final int stopIndex;
-    
+
+    private final String tableOwner;
```
tristaZero: Hi..maybe it could be columnOwner?

gufengwyx8: sry.. I forgot to rename this field.

tristaZero:
@gufengwyx8 Hi, good job. :)

PR #3284, opened 2019-10-16
Fix assignmentValueNode get error index

### LINKED ISSUE #3279: Shardingjdbc execute update exception
I configured read-write separation to make an error while executing the update statement, which contains case when.

For example, SQL statement: UPDATE OTS_ORDER SET UPDATE_DATE = now (),STATE = (CASE WHEN (STATE is null) THEN 1 ELSE STATE+1 END)

The error message is as follows:
  Org.antlr.v4.runtime.tree.TerminalNodeImpl cannot be cast to org.antlr.v4.runtime.ParserRuleContext

yanickxia:
Fixes #3279

- Fix get error index for nest Assignment Segment

Case 1：
```sql
update OTS_ORDER set UPDATE_DATE = now() where ID = 1;
```
then assignmentNode child size 3,  a -> CloumName; b -> = ; c -> now(), should get index 2

Case 2:
```sql
update OTS_ORDER set UPDATE_DATE = (now()+1) where ID = 1;
update OTS_ORDER set UPDATE_DATE = (now()) where ID = 1;
```

then assignmentNode child size 5,  a -> CloumName; b -> = ; c -> (;  d-> now()+1; e -> ), should get index 3

---
This change is [Reviewable](https://reviewable.io/reviews/apache/incubator-shardingsphere/3284)

### DIFF
--- a/sharding-core/sharding-core-parse/sharding-core-parse-common/src/main/java/org/apache/shardingsphere/core/parse/core/extractor/impl/dml/update/AssignmentExtractor.java
+++ b/sharding-core/sharding-core-parse/sharding-core-parse-common/src/main/java/org/apache/shardingsphere/core/parse/core/extractor/impl/dml/update/AssignmentExtractor.java
@@ -55,7 +55,7 @@ public final class AssignmentExtractor implements OptionalSQLSegmentExtractor {
     }
     
     private ExpressionSegment getAssignmentValue(final Map<ParserRuleContext, Integer> parameterMarkerIndexes, final ParserRuleContext assignmentNode) {
-        ParserRuleContext assignmentValueNode = 3 == assignmentNode.getChildCount() ? (ParserRuleContext) assignmentNode.getChild(2) : (ParserRuleContext) assignmentNode.getChild(4);
+        ParserRuleContext assignmentValueNode = 3 == assignmentNode.getChildCount() ? (ParserRuleContext) assignmentNode.getChild(2) : (ParserRuleContext) assignmentNode.getChild(3);
         Optional<? extends ExpressionSegment> result = expressionExtractor.extract(assignmentValueNode, parameterMarkerIndexes);
         Preconditions.checkState(result.isPresent());
         return result.get();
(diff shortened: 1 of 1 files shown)


jingshanglu:
> The modification may cause problem.
> Right now we just can recognize 2 model:
> 
>     1. xxx = xxx
> 
>     2. xxx = values(xxx)
> 
> 
> So the value we need to fetch is index `2` or index `4`.
> 
> I know we cannot get all scenarios, but if we change index to `3` on this pr, the scenario `xxx = values(xxx)` will be broken.
> 
> The problem of this issue is AST. AST should make consist of set assignment as `3` children, the right value should in one node and other more expression should in the next level of AST.
> We need to modify g4 file to change the structure of AST.

g4 like this will be ok?
```
assignment
    : columnName EQ_ assignmentRight_
    ;

assignmentRight_
    : VALUES LP_ RP_ | assignmentValue
    ;

```

jingshanglu:
#3290

yanickxia:
use  always get index 2， base on new g4

terrymanu:
> > The modification may cause problem.
> > Right now we just can recognize 2 model:
> > ```
> > 1. xxx = xxx
> > 
> > 2. xxx = values(xxx)
> > ```
> > 
> > 
> > So the value we need to fetch is index `2` or index `4`.
> > I know we cannot get all scenarios, but if we change index to `3` on this pr, the scenario `xxx = values(xxx)` will be broken.
> > The problem of this issue is AST. AST should make consist of set assignment as `3` children, the right value should in one node and other more expression should in the next level of AST.
> > We need to modify g4 file to change the structure of AST.
> 
> g4 like this will be ok?
> 
> ```
> assignment
>     : columnName EQ_ assignmentRight_
>     ;
> 
> assignmentRight_
>     : VALUES LP_ RP_ | assignmentValue
>     ;
> ```

Maybe not enough ,because they may have calculate expression like `values(xx) + values(yy)`

terrymanu:
Fixed with #3290 together

PR #3290, opened 2019-10-16
fix g4 for assignmentValueNode get error by index

### LINKED ISSUE #3279: Shardingjdbc execute update exception

jingshanglu:
Fixes #3279 .

- fix g4 for assignmentValueNode get error by index

### DIFF
--- a/sharding-core/sharding-core-parse/sharding-core-parse-mysql/src/main/antlr4/imports/mysql/BaseRule.g4
+++ b/sharding-core/sharding-core-parse/sharding-core-parse-mysql/src/main/antlr4/imports/mysql/BaseRule.g4
@@ -377,7 +377,7 @@ frameBetween_
 
 specialFunction_
     : groupConcatFunction_ | windowFunction_ | castFunction_ | convertFunction_ | positionFunction_ | substringFunction_ | extractFunction_ 
-    | charFunction_ | trimFunction_ | weightStringFunction_
+    | charFunction_ | trimFunction_ | weightStringFunction_ | valuesFunction_
     ;
 
 groupConcatFunction_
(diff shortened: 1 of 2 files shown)


jingshanglu:
@terrymanu

jingshanglu:
> That maybe not enough because the expression may `values(xx) + values(xx)`.
> So `VALUES` should be a part of expr

```
assignmentRight_
    : VALUES LP_ assignmentValue RP_ | assignmentValue
    ;

setAssignmentsClause
    : SET assignment (COMMA_ assignment)*
    ;

assignmentValues
    : LP_ assignmentValue (COMMA_ assignmentValue)* RP_
    | LP_ RP_
    ;

assignmentValue
    : expr | DEFAULT | blobValue
    ;
```
assignmentValue may be expr

jingshanglu:
> Are you sure the keyword `VALUES` in the rule expr? I cannot found it, if remove `VALUES` from set assignment and do not process it in expr, the SQL will parse error

values() is a function ,I add the func in expr

jingshanglu:
@terrymanu a test relate to zk failed 
>  https://builds.apache.org/job/shardingsphere-ci/666/testReport/junit/org.apache.shardingsphere.orchestration.config.zookeeper.curator.test/CuratorZookeeperConfigCenterTest/org_apache_shardingsphere_orchestration_config_zookeeper_curator_test_CuratorZookeeperConfigCenterTest/ 

 what's error?

PR #3322, opened 2019-10-17
#3138, improved coverage for sharding-core-preprocessor

### LINKED ISSUE #3138: Add more unit test cases for sharding-core-preprocessor module
sharding-core-preprocessor module is decouple with Sharding & encrypt. The codes has already stable.
Should Add more unit test cases for this module.

sluk3r:
For #3138


sluk3r:
> Hi, please do not use `Fixes xxx` it the issue is not finished, otherwise the issue will be closed.
> Please use `for xxx` if this pull request want related some issue but not finish.

Ok, Tks. 
And changed

PR #3331, opened 2019-10-19
MySQL parse insert statement function value with parameter

### LINKED ISSUE #3136: java.sql.SQLException: No value specified for parameter
version:shardingJDBC(4.0.0-RC2)  
execute SQL with function(UNIX_TIMESTAMP) 
insert into tb_user (id, name, add_time_unix)  values (?, ?,UNIX_TIMESTAMP(?));
java.sql.SQLException: No value specified for parameter 3
set sql.show configuration is true. I see the Actual SQL 
insert into tb_user (id, login_name, add_time_unix)  values (?, ?);

SteNicholas:
MySQLParser function that parses insert statement was unhandled for the condition of insert statement function value with parameter.See
Fixes #3136.

- I modify the expr expression by add predicate and add MySQL UNIX_TIMESTAMP and LOWER regular function.

### DIFF
--- a/sharding-core/sharding-core-parse/sharding-core-parse-mysql/src/main/antlr4/imports/mysql/BaseRule.g4
+++ b/sharding-core/sharding-core-parse/sharding-core-parse-mysql/src/main/antlr4/imports/mysql/BaseRule.g4
@@ -262,7 +262,6 @@ expr
     | notOperator_ expr
     | LP_ expr RP_
     | booleanPrimary_
-    | predicate
     ;
 
 logicalOperator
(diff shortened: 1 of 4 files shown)


sharding-core/sharding-core-parse/sharding-core-parse-mysql/src/main/antlr4/imports/mysql/BaseRule.g4
```diff
@@ -262,6 +262,7 @@ expr
     | notOperator_ expr
     | LP_ expr RP_
     | booleanPrimary_
+    | predicate
```
terrymanu: Why need add predicate here? It seems different with MySQL document.

SteNicholas: @terrymanu Just put predicate into above level.

PR #3334, opened 2019-10-20
Implement MySQL TimeService

### LINKED ISSUE #3326: Implement TimeService for MySQL
## Feature Request

If no response **more than 7 days** and we cannot make decision by current information, we will **close it**.

Yes
Using class Implementing TimeService to execute `select now()`  for MySQL, determine time service DataSource through configuration
but I have two problems
1. Which module should this class be placed on?
2. How to give this class the ability to execute

Pramy:
Change-Id: Ie00304d62a54506c449a773b52f4a0dcdfd3eac2

Fixes #3326 

-
-
-

### DIFF
--- a/sharding-jdbc/sharding-jdbc-core/src/main/java/org/apache/shardingsphere/shardingjdbc/jdbc/core/context/RuntimeContextHolder.java
+++ b/sharding-jdbc/sharding-jdbc-core/src/main/java/org/apache/shardingsphere/shardingjdbc/jdbc/core/context/RuntimeContextHolder.java
@@ -1,0 +1,118 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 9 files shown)


sharding-core/sharding-core-api/src/main/java/org/apache/shardingsphere/api/config/sharding/ShardingRuleConfiguration.java
```diff
@@ -45,6 +45,8 @@
     private Collection<String> broadcastTables = new LinkedList<>();
     
     private String defaultDataSourceName;
+
+    private String timeServiceDataSourceName;
```
terrymanu: Please do not modify sharding configuration. `timeServiceDataSourceName` should use properties from SPI

sharding-jdbc/sharding-jdbc-core/src/main/java/org/apache/shardingsphere/shardingjdbc/jdbc/core/context/RuntimeContextWrapper.java
```diff
@@ -0,0 +1,58 @@
...
+import javax.sql.DataSource;
+import java.util.Map;
+
+/**
+ *  Sharding context wrapper.
+ *
+ * @author chenchuangliu
+ */
+@RequiredArgsConstructor
+public final class RuntimeContextWrapper<T extends RuntimeContext> {
```
terrymanu: This RuntimeContextWrapper and RuntimeContext maybe to complicated, please do it simple if we can use SPI only can do fetch from MySQL.

PR #3364, opened 2019-10-24
[LICENSE FIX] Update the license file for maven wrapper

wu-sheng:
This is the additional fix for #3356. Maven wrapper is in Apache 2.0 license now, and as it will release with the source code, the source code LICENSE file should be fixed.

@terrymanu Is my modified file as the source release LICENSE file in the official release process?

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -199,3 +199,20 @@
    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
    See the License for the specific language governing permissions and
    limitations under the License.
+
+=======================================================================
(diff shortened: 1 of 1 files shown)


JaredTan95:
https://github.com/apache/incubator-shardingsphere/pull/3360 Aha, I closed it by mistake. :P

terrymanu:
@wu-sheng Thank you for update the license.
Yes, the LICENSE file in root folder is in the official `source` release process, but not in `binary` release process. 
Because of the source code reference of mvnw is only in source release, so I think change this LICENSE file is enough.

PR #3385, opened 2019-10-28
fix #3273, route to defaultDataSource or master when the sql without tableName

### LINKED ISSUE #3273: Default database doesn't work on sql without tables
My case is like this:
I have two db `db1` and `db2`.  `db1` ows table `t1`,`t2`,`t3`, function `f1`, and `db2` only onws `t2`. Only table `t2` was in partition, so I want sql related to `t1` , `t3`, `f1` to be executed on `db1`, only `t2` be executed on `db1` and `db2` according to it's condition value. I use the configuration `spring.shardingsphere.sharding.default-data-source-name`, now sql related to `t1` or `t3` was executed on `d1` as expected, but `f1`'s execution database was in random.  

The result I **expected** is like this,
```select 1 from t1;``` on db1
```select 1 from t3;``` on db1
**``` select f1;``` on db1**
```select 1 from t2;``` on db1,db2

The **actual** result is like this:
```select 1 from t1;``` on db1
```select 1 from t3;``` on db1
**``` select f1;``` on (db1 or db2 in random)**
```select 1 from t2;``` on db1,db2

RaigorJiang:
fix #3273, route to defaultDataSource or master when the sql without tableName

Fixes #3273.

sharding:
- With default data source, route to default data source
- Without default data source, unicast route(stay the same)
Master-Slave:
- route to master database

### DIFF
--- a/sharding-core/sharding-core-route/src/main/java/org/apache/shardingsphere/core/route/router/masterslave/MasterSlaveRouter.java
+++ b/sharding-core/sharding-core-route/src/main/java/org/apache/shardingsphere/core/route/router/masterslave/MasterSlaveRouter.java
@@ -70,6 +70,12 @@ public final class MasterSlaveRouter {
     }
     
     private boolean isMasterRoute(final SQLStatement sqlStatement) {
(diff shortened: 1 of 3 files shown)


RaigorJiang:
Sorry,  CI FAILURE, I'll amend it.

PR #3390, opened 2019-10-29
support select for update

### LINKED ISSUE #1496: Read write splitter for `Select ... for update` should route to master data node
### Which version of Sharding-Sphere did you use?

3.0.0

All

### Expected behavior

Read write splitter for `Select ... for update` should route to master data node

### Actual behavior

Read write splitter for `Select ... for update` route to slave data node now

associated with #767

Tasks here:

- [ ] Parse `Select ... for update`
- [ ] Modify master-slave route engine

yanyzy:
Fixes #1496.

make select for update route to master node.

### DIFF
--- a/sharding-core/sharding-core-route/src/main/java/org/apache/shardingsphere/core/route/router/masterslave/MasterSlaveRouter.java
+++ b/sharding-core/sharding-core-route/src/main/java/org/apache/shardingsphere/core/route/router/masterslave/MasterSlaveRouter.java
@@ -17,11 +17,9 @@
 
 package org.apache.shardingsphere.core.route.router.masterslave;
 
-import com.google.common.base.Optional;
 import lombok.RequiredArgsConstructor;
 import org.apache.shardingsphere.api.hint.HintManager;
 import org.apache.shardingsphere.core.parse.SQLParseEngine;
-import org.apache.shardingsphere.core.parse.sql.segment.dml.predicate.LockSegment;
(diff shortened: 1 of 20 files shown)


sharding-core/sharding-core-route/src/main/java/org/apache/shardingsphere/core/route/router/masterslave/MasterSlaveRouter.java
```diff
@@ -70,6 +72,14 @@
     }
     
     private boolean isMasterRoute(final SQLStatement sqlStatement) {
-        return !(sqlStatement instanceof SelectStatement) || MasterVisitedManager.isMasterVisited() || HintManager.isMasterRouteOnly();
+        return isLockSegment(sqlStatement) || !(sqlStatement instanceof SelectStatement) || MasterVisitedManager.isMasterVisited() || HintManager.isMasterRouteOnly();
+    }
+
+    private boolean isLockSegment(final SQLStatement sqlStatement) {
```
terrymanu: maybe we can simplify the code as:

`return sqlStatement instanceof SelectStatement && ((SelectStatement) sqlStatement).getLock().isPresent()`

terrymanu: The method name should be `containsLockSegment `

Issue #3370, opened 2019-10-25
oracle load meta data get index info user schema error


54chenjian:
### Which version of ShardingSphere did you use?

4.0.0-RC2

### Which project did you use? Sharding-JDBC or Sharding-Proxy?

Sharding-JDBC

### Reason analyze (If you can)
```java
 private Collection<String> getLogicIndexes(final Connection connection, final String catalog, final String actualTableName) throws SQLException {
        Collection<String> result = new HashSet<>();
        try (ResultSet resultSet = connection.getMetaData().getIndexInfo(catalog, catalog, actualTableName, false, false)) {
            while (resultSet.next()) {
                Optional<String> logicIndex = getLogicIndex(resultSet.getString("INDEX_NAME"), actualTableName);
                if (logicIndex.isPresent()) {
                    result.add(logicIndex.get());
                }
            }
        }
        return result;
    }
```
```java
   ResultSet getIndexInfo(String catalog, String schema, String table, boolean unique, boolean approximate)
        throws SQLException;
```
oracle catalog is sid ，the second param is scheam，is not sid

tristaZero:
Hi, it is a easy issue, if anyone is interested in it, please raise a PR. Thanks
### How to solve it?
If null == resultSet.getString(INDEX_NAME), then return Optional.<String>absent() in `getLogicIndex(final String actualIndexName, final String actualTableName)`.

beijing-penguin:
I read the pssql、mysql and oracle jdbc source code，about catalog and schema method parameter，the result is :
1、All database's catalog  equal database name
2、oracle's schema equal username ,but other database's schema equal null.

tristaZero:
> I read the pgsql、mysql and oracle jdbc source code，about catalog and schema method parameter，the result is :
> 1、All database's catalog equal database name
> 2、oracle's schema equal username ,but other database's schema equal null.

Hi penguin, Appreciated your investigation, if available, we welcome your contribution. :)

tristaZero:
> I read the pgsql、mysql and oracle jdbc source code，about catalog and schema method parameter，the result is :
> 1、All database's catalog equal database name
> 2、oracle's schema equal username ,but other database's schema equal null.

It is great. Your detail investigation inspired us how to solve this issue!

PR #3414, opened 2019-10-30
Fix jdbc getIndexInfo method and NullPointerException on oracle

beijing-penguin:
Fixes #3370.

- getIndexInfo mothed second parameter 
- judge not null

### DIFF
--- a/sharding-core/sharding-core-execute/src/main/java/org/apache/shardingsphere/core/execute/metadata/TableMetaDataLoader.java
+++ b/sharding-core/sharding-core-execute/src/main/java/org/apache/shardingsphere/core/execute/metadata/TableMetaDataLoader.java
@@ -199,7 +199,7 @@ public final class TableMetaDataLoader {
     
     private Collection<String> getLogicIndexes(final Connection connection, final String catalog, final String actualTableName) throws SQLException {
         Collection<String> result = new HashSet<>();
-        try (ResultSet resultSet = connection.getMetaData().getIndexInfo(catalog, catalog, actualTableName, false, false)) {
+        try (ResultSet resultSet = connection.getMetaData().getIndexInfo(catalog, null, actualTableName, false, false)) {
             while (resultSet.next()) {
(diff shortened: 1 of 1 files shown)


sharding-core/sharding-core-execute/src/main/java/org/apache/shardingsphere/core/execute/metadata/TableMetaDataLoader.java
```diff
@@ -199,7 +199,7 @@ private boolean isTableExist(final Connection connection, final String catalog,
     
     private Collection<String> getLogicIndexes(final Connection connection, final String catalog, final String actualTableName) throws SQLException {
         Collection<String> result = new HashSet<>();
-        try (ResultSet resultSet = connection.getMetaData().getIndexInfo(catalog, catalog, actualTableName, false, false)) {
+        try (ResultSet resultSet = connection.getMetaData().getIndexInfo(catalog, null, actualTableName, false, false)) {
```
terrymanu: Any performance issue here if use `null` to instead of catalog name if database is not Oracle?

dongzl: I think it will make a mistake, for some database, e.g PostgreSQL, the second param is useful.

beijing-penguin: I read mysql-connector-java-5.1.48.java and ojdbc8.jar and postresql-42.2.8.jar,the result is
1.mysql is not using the  getIndexInfo's second parameter
2.oracle is using second parameter as owner,like this "select * from all_tables where owner=?"
3.postresql is using second parameter as tablespace_name,like this
```
if (schema != null && !schema.isEmpty()) {
        sql += " AND n.nspname = " + escapeQuotes(schema);
}
```
so i think that 
1.if use mysql ,the second parameter is null
2.if use oracle that the second parameter is username
3.if use postresql that the second parameter is nspname (tablespace_name)

PR #3494, opened 2019-11-10
Fix three mistakes

beijing-penguin:
Fixes #3370 .

- Fix jdbc getIndexInfo method and NullPointerException.
- Fix catalog and schema method parameter error.
- Fix resultSet.getString(IS_NULLABLE) and
resultSet.getString(IS_AUTOINCREMENT) return type error.


sharding-core/sharding-core-execute/src/main/java/org/apache/shardingsphere/core/execute/metadata/TableMetaDataLoader.java
```diff
@@ -165,8 +174,9 @@ private boolean isTableExist(final Connection connection, final String catalog,
                 String columnName = resultSet.getString(COLUMN_NAME);
                 String columnType = resultSet.getString(TYPE_NAME);
                 boolean isPrimaryKey = primaryKeys.contains(columnName);
-                boolean isNotNull = isPrimaryKey || !resultSet.getBoolean(IS_NULLABLE);
```
tristaZero: Can `` not work well? I did a following test, and it returns true.
```
ResultSet resultSet = statement.executeQuery("select 'YES' ");
        resultSet.next();
        System.out.println(resultSet.getBoolean(1));
```
If resultSet.getString(IS_NULLABLE) == 'YES', resultSet.getBoolean(IS_NULLABLE) will be true. Is it right?

sharding-core/sharding-core-common/src/main/java/org/apache/shardingsphere/core/util/DataSourceUtil.java
```diff
@@ -0,0 +1,61 @@
...
+                if (!Modifier.isStatic(field.getModifiers())) {
+                    if (field.getName().toLowerCase().contains("url")) {
+                        field.setAccessible(true);
+                        String jdbcurl = field.get(dataSource) == null ? null : field.get(dataSource).toString();
+                        sourceInfo.setUrl(jdbcurl);
+                    }
+                    if (field.getName().toLowerCase().contains("username")) {
+                        field.setAccessible(true);
+                        username = field.get(dataSource) == null ? null : field.get(dataSource).toString();
+                        sourceInfo.setUsername(username);
```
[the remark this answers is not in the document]
tristaZero: Therefore we don't need this class anymore.

tristaZero:
Hi, @beijing-penguin Appreciated your contrition. The point need to disscuss i think is that can we use jdbc interface to get url and username? Therefore we will not care about which connection pool user chooses. When get connection for `jdbc url`, `username` can be got as well. Do you think it is available？

sharding-core/sharding-core-common/src/main/java/org/apache/shardingsphere/core/util/DataSourceUtil.java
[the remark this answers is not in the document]
beijing-penguin: > Therefore we don't need this class anymore.
1、you way is right result in oracle,but runing  the code will get username="root@localhost" in mysql.
the result is not right,only work right....so,now we can don't need this class,but suggested first support the class.
2、if use getConn().getMeta().getUserName,the code performance will be low...not only the code line will run getConn()....so the getConn should be avoided.

sharding-core/sharding-core-execute/src/main/java/org/apache/shardingsphere/core/execute/metadata/TableMetaDataLoader.java
beijing-penguin: I connect database to test resultSet.getBoolean(IS_NULLABLE) is not work.

image
as image,only return "YES" OR "NO",if  getBoolean will throw exception

beijing-penguin:
reset pr

sharding-core/sharding-core-execute/src/main/java/org/apache/shardingsphere/core/execute/metadata/TableMetaDataLoader.java
junwangit: MySQL Connector/J 8.0 + not work

Issue #3185, opened 2019-10-08
The configuration of RegistryCenter, ConfigCenter,LeafCenter should be split apart or not?


wgy8283335:
## The configuration of RegistryCenter, ConfigCenter,LeafCenter should be split apart or not?
As we plan to split the Resgitry Center Interface into RegistryCenter, ConfigCenter,LeafCenter. So,configuration of RegistryCenter and ConfigCenter should be split apart or not? For example:
```
orchestration:
  name: test
  overwrite: true
  registry:
    type: TestRegistryCenter
    namespace: test_encrypt
    serverLists: localhost:3181
  config:
    type: TestRegistryCenter
    namespace: test_encrypt
    serverLists: localhost:3181
  leaf:
    type: TestRegistryCenter
    namespace: test_encrypt
    serverLists: localhost:3181
```

terrymanu:
I prefer do not split reg center and config center if user use same component (eg. zookeeper). 
But we'd better support user use different components for reg center and config center (eg. eureka for reg center and apollo for config center).

So we need to talk about how to deal with the configuration of these 2 different scenarios.

wgy8283335:
Below is the current configuration 
```
spring.shardingsphere.orchestration.name= #Orchestration instance name
spring.shardingsphere.orchestration.overwrite= #Whether to overwrite local configurations with registry center configurations; if it can, each initialization should refer to local configurations
spring.shardingsphere.orchestration.registry.type= #Registry center type. Example:zookeeper
spring.shardingsphere.orchestration.registry.server-lists= #The list of servers that connect to registry center, including IP and port number; use commas to separate
spring.shardingsphere.orchestration.registry.namespace= #Registry center namespace
spring.shardingsphere.orchestration.registry.digest= #The token that connects to the registry center; default means there is no need for authentication
spring.shardingsphere.orchestration.registry.operation-timeout-milliseconds= #The millisecond number for operation timeout; default value: 500 milliseconds
spring.shardingsphere.orchestration.registry.max-retries= #Maximum retry time after failing; default value: 3 times
spring.shardingsphere.orchestration.registry.retry-interval-milliseconds= #Interval time to retry; default value: 500 milliseconds
spring.shardingsphere.orchestration.registry.time-to-live-seconds= #Living time of temporary nodes; default value: 60 seconds
spring.shardingsphere.orchestration.registry.props= #Customize registry center props.
```
Below is the modified
```
[...]

PR #3495, opened 2019-11-10
Modify pom.xml in sharding-orchestration-core module

wgy8283335:
Fixes #3185.

- modify pom.xml
-
-


terrymanu:
Please rename the title of this pr, we should let it for search friendly

PR #3519, opened 2019-11-13
Fixes #3370 .and change some code structure for compatibility

beijing-penguin:
Fixes #3370 .

- use jdbc api to fix oracle difference and fix other jdbc compatibility
- Fix jdbc getIndexInfo method and NullPointerException.
- Fix and change catalog and schema method parameter error.
- Fix and change resultSet.getString(IS_NULLABLE) and
resultSet.getString(IS_AUTOINCREMENT) return type error.(at MySQL Connector/J 8.0 + not work)
- change some code structure for current compatibility


sharding-core/sharding-core-common/src/main/java/org/apache/shardingsphere/core/metadata/datasource/dialect/H2DataSourceMetaData.java
```diff
@@ -31,24 +37,43 @@
...
             throw new UnrecognizedDatabaseURLException(url, pattern.pattern());
         }
         hostName = "";
         port = DEFAULT_PORT;
-        schemaName = matcher.group(2);
+        catalog = matcher.group(2);
+        schemaName = null;
+    }
+
+    public H2DataSourceMetaData(final DataSource dataSource) {
```
tristaZero: Why do we need this funciton? How about passing `dataSourceMetadata` containing `jdbcUrl` and `username` to constructor above?

beijing-penguin: At present we only need jdbcurl and username ,but at future ,maybe we need more datasouce infomation,so,it is possibility that will use this datasource to get more infomation.

beijing-penguin:
1.add DataSourceInfo class
2.add and change TEST unit
3.fix and change HashMap to LinkedHashMap

PR #3539, opened 2019-11-17
the fix present table not exist when start application

### LINKED ISSUE #3446: present table not exist when start application
org.apache.shardingsphere sharding-jdbc-core 4.0.0-RC2

Sharding-JDBC

### Oracle info
Oracle 11g

### DataSource
HikariDataSource com.zaxxer.HikariCP 3.4.1

### Expected behavior
find table info

### Actual behavior
ORA-00942: Table or view does not exist

I view source code and know this issue reason. Shardingsphere will get table info with schema.table in db when start application.The schema originate from setup dataSource url.The url as below is my project:
jdbc:oracle:thin:@//localhost:1522/XE ,

I find OracleDataSourceMetaData.class will parse oracle configuration and the code is get schema code as below:
`this.schemaName = matcher.group(5);`

Finally, this will be combined with the table name to query the table information in the database.
Like this: XE.table.
I don't set a schema for the . My login user name is regarded as default schema.
I try to use this method,

`HikariDataSource.setSchema`
,but i got an exception.
Caused by: java.lang.AbstractMethodError: oracle.jdbc.driver.T4CConnection.setSchema(Ljava/lang/String;)V

So i think it should be username.table or provide a method to set schema.

BR!

zjcnb:
Fixes #3446

The fix present table not exist when start application

### DIFF
--- a/sharding-core/sharding-core-execute/src/main/java/org/apache/shardingsphere/core/execute/metadata/TableMetaDataLoader.java
+++ b/sharding-core/sharding-core-execute/src/main/java/org/apache/shardingsphere/core/execute/metadata/TableMetaDataLoader.java
@@ -142,7 +142,7 @@ public final class TableMetaDataLoader {
     
     private TableMetaData createTableMetaData(final Connection connection, final String catalog, 
                                               final String logicTableName, final String actualTableName,
-                                              final String generateKeyColumnName, final EncryptRule encryptRule,final String dataSourceName) throws SQLException {
+                                              final String generateKeyColumnName, final EncryptRule encryptRule, final String dataSourceName) throws SQLException {
         if (isTableExist(connection, catalog, actualTableName)) {
             return new TableMetaData(
                     getColumnMetaDataList(connection, catalog, logicTableName, actualTableName, generateKeyColumnName, encryptRule),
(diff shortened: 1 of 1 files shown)


sharding-core/sharding-core-execute/src/main/java/org/apache/shardingsphere/core/execute/metadata/TableMetaDataLoader.java
```diff
@@ -205,9 +207,9 @@ private boolean isTableExist(final Connection connection, final String catalog,
         return Optional.of(new ColumnMetaData(columnName, columnType, isPrimaryKey, isNotNull, isAutoIncrement));
     }
     
-    private Collection<String> getLogicIndexes(final Connection connection, final String catalog, final String actualTableName) throws SQLException {
+    private Collection<String> getLogicIndexes(final Connection connection, final String catalog, final String actualTableName, final String dataSourceName) throws SQLException {
         Collection<String> result = new HashSet<>();
-        try (ResultSet resultSet = connection.getMetaData().getIndexInfo(catalog, catalog, actualTableName, false, false)) {
+        try (ResultSet resultSet = connection.getMetaData().getIndexInfo(catalog, dataSourceName, actualTableName, false, false)) {
```
tristaZero: Hi, why do you use datasource instead of catalog?
dataSourceName, which is configured by user, may be different with catalog.

zjcnb: emmm, That's true. the schemaName is allow Null, I didn't think about it  may be different with catalog. And this problem is fix. I will close this pr

tristaZero:
Please wait for #3541

tristaZero:
@zjcnb Thanks for your pr,

PR #3585, opened 2019-11-25
Modify MasterSlaveLoadBalanceAlgorithm "getDataSource()" method and r…

wgy8283335:
…elated functions.

Fixes ##3227.

- MasterSlaveLoadBalanceAlgorithm.getDataSource()
- Implementation classes
- Unit Tests.


terrymanu:
Maybe we need to discuss this pr.
User maybe need to know `masterDataSourceName` on the API, so we need to keep this parameter.
By the way, we may supported multiple master data sources in future.

PR #3597, opened 2019-11-26
Add SyncTaskExecuteEngine

avalon566:


sharding-scaling/sharding-scaling-core/src/main/java/org/apache/shardingsphere/shardingscaling/core/execute/engine/DefaultSyncTaskExecuteEngine.java
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * Default implement for sync task execute engine.
+ *
+ * @author avalon566
+ */
+public class DefaultSyncTaskExecuteEngine implements SyncTaskExecuteEngine {
+
+    private final ListeningExecutorService executorService;
+
+    private int availableWorkerThread;
```
KomachiSion: I think there should be AtomicInteger, because ExecuteEngine may be used by multiple thread.

avalon566: I had use synchronized method.

PR #3611, opened 2019-11-27
Fix completed mysql jdbc url parameters

avalon566:


avalon566:
One of the scenes is parameter yearIsDateType. The storage value is integer, but jdbc default return return date value, it can not  insert into destination column.

PR #3668, opened 2019-12-05
Impl Heartbeat

avalon566:
For #3638 .


sharding-scaling/sharding-scaling-core/src/main/java/org/apache/shardingsphere/shardingscaling/core/synctask/realtime/RealtimeSyncChannel.java
```diff
@@ -15,8 +15,11 @@
  * limitations under the License.
  */
 
-package org.apache.shardingsphere.shardingscaling.core.execute.executor.channel;
+package org.apache.shardingsphere.shardingscaling.core.synctask.realtime;
```
KomachiSion: I think all Chanel should in Channel package.

KomachiSion:
What's more, If use heartbeat to monitor delay, we need users give us bigger privileges. It may be confused for some users.

PR #3702, opened 2019-12-10
[Shadow] add yaml configuration & rule

yanickxia:
- add shadow rule & config yaml loader

---
This change is [Reviewable](https://reviewable.io/reviews/apache/incubator-shardingsphere/3702)

### DIFF
--- a/sharding-core/sharding-core-api/src/main/java/org/apache/shardingsphere/api/config/shadow/ShadowRuleConfiguration.java
+++ b/sharding-core/sharding-core-api/src/main/java/org/apache/shardingsphere/api/config/shadow/ShadowRuleConfiguration.java
@@ -15,32 +15,30 @@
  * limitations under the License.
  */
 
-package org.apache.shardingsphere.core.rule;
+package org.apache.shardingsphere.api.config.shadow;
 
+import com.google.common.base.Preconditions;
 import lombok.Getter;
-import org.apache.shardingsphere.api.config.shadow.ShadowRuleConfiguration;
+import lombok.RequiredArgsConstructor;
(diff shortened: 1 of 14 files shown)


sharding-core/sharding-core-common/src/main/java/org/apache/shardingsphere/core/rule/ShadowRule.java
```diff
@@ -18,18 +18,30 @@
...
+ * @author xiayan
  */
 @Getter
 public class ShadowRule implements BaseRule {
-
-    @Override
-    public final RuleConfiguration getRuleConfiguration() {
-        return null;
+    
+    @Getter
```
yanyzy: This class already has the @Getter , is that a repetition ?

sharding-core/sharding-core-api/src/main/java/org/apache/shardingsphere/api/config/shadow/ShadowRuleConfiguration.java
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * Shadow rule configuration.
+ *
+ * @author xiayan
+ */
+@RequiredArgsConstructor
+@Getter
+public final class ShadowRuleConfiguration implements RuleConfiguration {
+    
+    private final String column;
```
yanyzy: Will it be better to named it 'shadowCloumn' ?

yanickxia: in shadowRule class, it's clear, i think

PR #3768, opened 2019-12-19
fix for #3750

### LINKED ISSUE #3750: ResultSet miss the plain column  because of using contains to get the logicColumn when using  data mask.
sharding-jdbc-4.0.0-RC3
sharding-jdbc
### Expected behavior
return the all of table's columns.
### Actual behavior
resultSet miss the logic column  because of using contains to get the logicColumn

org.apache.shardingsphere.shardingjdbc.jdbc.core.resultset.EncryptResultSetMetaData#getLogicColumn
image

rules are:
image

The boss_real_name contains real_name cause the resultSet missing.
I think that it should use equals  insteaded

xiyelife:
Fixes #3750 
use equals instead of contains to get the plain column.


sharding-jdbc/sharding-jdbc-core/src/main/java/org/apache/shardingsphere/shardingjdbc/jdbc/core/resultset/EncryptResultSetMetaData.java
```diff
@@ -126,7 +126,7 @@ public String getColumnName(final int column) throws SQLException {
     
     private String getLogicColumn(final String actualColumn) throws SQLException {
         for (Entry<String, String> entry : logicAndActualColumns.entrySet()) {
-            if (entry.getValue().contains(actualColumn)) {
+            if (entry.getValue().equals(actualColumn)) {
```
kimmking: you should add a test case for this corner case to make sure that this modification works well as we expect.

SteNicholas: @xiyelife You could add unit test in `EncryptStatementTest`.

kimmking: and look over this modification, you can use map.values().contains instead of for statement.

xiyelife: > and look over this modification, you can use map.values().contains instead of for statement.
Here the code want to get the matched value's key, so must use the for each.

xiyelife: > @xiyelife You could add unit test in `EncryptStatementTest`.

Ok  I will.

xiyelife: > @xiyelife You could add unit test in `EncryptStatementTest`.

OK, I will.

tristaZero:
@xiyelife Hi Thans for your revise, welcome more contributions from you and enjoy it.

tristaZero:
[#3809](https://github.com/apache/incubator-shardingsphere/pull/3809) answears the related issue.

Issue #3602, opened 2019-11-26
Refactor and enhance sharding-scaling-core


KomachiSion:
- [x] Refactor SyncExecutor to SyncExecuteEngine
- [x] Refactor DataSyncTask
- [x] Implement SyncTaskProgress
- [x] Use Callback to report completed sync task

avalon566:
I am working on task one, assignees this to me?

PR #3771, opened 2019-12-19
For #3602, refactor channel

KomachiSion:
For #3602 .


KomachiSion:
OK, I will refactor and fix it.

PR #3802, opened 2019-12-25
[Shadow] add more code to make the feature more complete

yanyzy:

### DIFF
--- a/sharding-core/sharding-core-rewrite/src/main/java/org/apache/shardingsphere/core/rewrite/feature/shadow/SimpleJudgementEngine.java
+++ b/sharding-core/sharding-core-rewrite/src/main/java/org/apache/shardingsphere/core/rewrite/feature/shadow/SimpleJudgementEngine.java
@@ -29,7 +29,6 @@ import org.apache.shardingsphere.sql.parser.sql.statement.generic.WhereSegmentAv
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
-import java.util.Map;
 
 /**
  * Simple shadow judgement engine.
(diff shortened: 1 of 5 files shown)


sharding-core/sharding-core-rewrite/src/main/java/org/apache/shardingsphere/core/rewrite/feature/shadow/SimpleJudgementEngine.java
```diff
@@ -59,10 +58,7 @@ public boolean isShadowSql() {
                 return false;
             }
             List<Object> values = shadowCondition.get().getValues(Collections.emptyList());
-            for (Map.Entry<Integer, Integer> entry : shadowCondition.get().getPositionIndexMap().entrySet()) {
-                Object value = values.get(entry.getValue());
-                return value instanceof Boolean && (Boolean) value;
-            }
+            return "TRUE".equals(((String) values.get(0)).toUpperCase());
```
SteNicholas: Does `value.get(0)` return null ?

sharding-core/sharding-core-rewrite/src/main/java/org/apache/shardingsphere/core/rewrite/feature/shadow/SimpleJudgementEngine.java
```diff
@@ -74,7 +70,7 @@ private boolean judgeShadowSqlForInsert(final InsertValueContext insertValueCont
             if (shadowRule.getColumn().equals(columnName)) {
                 int columnIndex = insertSQLStatementContext.getColumnNames().indexOf(columnName);
                 Object value = insertValueContext.getValue(columnIndex);
-                return value instanceof Boolean && (Boolean) value;
+                return "TRUE".equals(((String) value).toUpperCase());
```
SteNicholas: 同上。

sharding-core/sharding-core-rewrite/src/main/java/org/apache/shardingsphere/core/rewrite/feature/shadow/SimpleJudgementEngine.java
yanyzy: I will add a judgement for it, thanks

sharding-core/sharding-core-rewrite/src/main/java/org/apache/shardingsphere/core/rewrite/feature/shadow/SimpleJudgementEngine.java
yanyzy: I think use column index to get value should not be null, or it will throw runtime exception.

PR #3811, opened 2019-12-26
Solve the problem when project's default charset is not utf-8

### LINKED ISSUE #3810: When the default character code of my project is not UTF-8, decryption fails due to scrambling
When the default character code of my project is not UTF-8, decryption fails due to scrambling

minghu-zhang:
Fixes #3810.

- modified：AESShardingEncryptor.java
-
-


terrymanu:
How about if the original string is not UTF8?

minghu-zhang:
> How about if the original string is not UTF8?

It will cause the result of decryption to be disorderly

minghu-zhang:
> How about if the original string is not UTF8?

The encryption uses utf-8 encoding, but the decryption does not specify the encoding, and the default system encoding is used. When our system defaults encoding is GBK , the correct result cannot be obtained

terrymanu:
Can `UTF8` solve all charset problems?
Do we need add a property key to let user provide the charset, just like `aes.key.value`?

PR #3820, opened 2019-12-27
use utf8 charset.

tristaZero:


sharding-core/sharding-core-common/src/main/java/org/apache/shardingsphere/core/strategy/encrypt/impl/AESShardingEncryptor.java
```diff
@@ -71,7 +71,7 @@ public Object decrypt(final String ciphertext) {
             return null;
         }
         byte[] result = getCipher(Cipher.DECRYPT_MODE).doFinal(Base64.decodeBase64(String.valueOf(ciphertext)));
-        return new String(result, Charsets.UTF_8);
+        return new String(result, StandardCharsets.UTF_8);
```
SteNicholas: Why use `java.nio.charset.StandardCharsets`  not `org.apache.commons.codec.Charsets`?

terrymanu: Charsets is deprecated

PR #3868, opened 2020-01-04
Part of the shard key in multiple sharding policies can be null

### LINKED ISSUE #3644: part of sharding key in multiple sharding strategy  can not null
sharding column value cannot insert  null ,but '' is ok?

if my sharding column is id_number
insert into a(id,id_number)values(1,'');   √
insert into a(id,id_number)values(1,null);  ×

lyzhb9014:
Fixes #3644 .

- Part of the shard key in multiple sharding policies can be null
-
-


sharding-core/sharding-core-route/src/main/java/org/apache/shardingsphere/core/route/router/sharding/condition/engine/InsertClauseShardingConditionEngine.java
```diff
@@ -84,27 +84,40 @@ private ShardingCondition createShardingCondition(final String tableName, final
         for (ExpressionSegment each : insertValueContext.getValueExpressions()) {
             String columnName = columnNames.next();
             if (shardingRule.isShardingColumn(columnName, tableName)) {
-                if (each instanceof SimpleExpressionSegment) {
-                    result.getRouteValues().add(new ListRouteValue<>(columnName, tableName, Collections.singletonList(getRouteValue((SimpleExpressionSegment) each, parameters))));
-                } else if (ExpressionConditionUtils.isNowExpression(each)) {
-                    result.getRouteValues().add(new ListRouteValue<>(columnName, tableName, Collections.singletonList(timeService.getTime())));
+                Comparable<?> routeValue = getRouteValue(each, parameters, timeService);
+                if (null == routeValue) {
```
SteNicholas: This could judge routeValue is not null, and `result.getRouteValues().add(new ListRouteValue<>(columnName, tableName, Collections.singletonList(routeValue)))` add to if logic, remove the continue.

kimmking:
I think a git squash is necessary.

sharding-core/sharding-core-route/src/main/java/org/apache/shardingsphere/core/route/router/sharding/condition/engine/InsertClauseShardingConditionEngine.java
```diff
@@ -84,27 +84,39 @@ private ShardingCondition createShardingCondition(final String tableName, final
...
             result = ((LiteralExpressionSegment) expressionSegment).getLiterals();
         }
+        if (null == result) {
+            return null;
+        }
         Preconditions.checkArgument(result instanceof Comparable, "Sharding value must implements Comparable.");
         return (Comparable) result;
     }
-    
+
```
KomachiSion: Please do not change the indents

lyzhb9014:
> The implement has no problem, but has one code style problem.
> Can you fix it?
> 
> What's more, Can you add unit test case for this PR?

ok, I will fix it, and add unit test case.

KomachiSion:
@lyzhb9014 , Sorry, My mistake. I read the issue again, and I found there are some error  in this implementation.

1. this implementation will cause insert with null value full-routing for one record.
2. Even just for ComplexShardingStrategy, the null value also can't be input to sharding algorithm.

So we can't accept this PR.

PR #3879, opened 2020-01-06
fix proxy hang up while throwing a customize SQLException

cherrylzhao:
After proxy throw a customize SQLException, proxy will hang up with error packet since SQLSate is null


sharding-proxy/sharding-proxy-frontend/sharding-proxy-frontend-mysql/src/main/java/org/apache/shardingsphere/shardingproxy/frontend/mysql/MySQLErrPacketFactory.java
```diff
@@ -48,7 +48,7 @@
     public static MySQLErrPacket newInstance(final int sequenceId, final Exception cause) {
         if (cause instanceof SQLException) {
             SQLException sqlException = (SQLException) cause;
-            return new MySQLErrPacket(sequenceId, sqlException.getErrorCode(), sqlException.getSQLState(), sqlException.getMessage());
+            return new MySQLErrPacket(sequenceId, sqlException.getErrorCode(), null != sqlException.getSQLState() ? sqlException.getSQLState() : "", sqlException.getMessage());
```
terrymanu: Please use a SQL State instead of empty string if we can find a suitable one.

PR #3883, opened 2020-01-06
Code style revise for shadow feature

terrymanu:
Revise #3859.


shadow-core/shadow-core-rewrite/src/main/java/org/apache/shardingsphere/shadow/rewrite/PreparedJudgementEngine.java
```diff
@@ -41,16 +41,16 @@
...
     private final SQLStatementContext sqlStatementContext;
-
+    
     private final List<Object> parameters;
-
+    
     @Override
-    public boolean isShadowSql() {
+    public boolean isShadowSQL() {
         if (sqlStatementContext.getSqlStatement() instanceof InsertStatement) {
```
taojintianxia: why you make SQL in upper case in isShadowSQL but not in getSqlStatement() ?

PR #3907, opened 2020-01-08
Encrypt the field, and take the value of string type directly. For example, after digital encryption, the database can be saved as varchar, and the corresponding field of bean can still be number

longjy:
Fixes #ISSUSE_ID.

-
-
-


kimmking:
you should add a test case for it.

terrymanu:
Please use English for git comment only

Issue #3914, opened 2020-01-09
Optimization for Sharing Parser with ANTLR Visitor for MySQL


tristaZero:
Hi folks, we would like to refactor existing parser engine with **ANTLR visitor** for better performance. It is scheduled as a long-term task and a big change for our 5.x release, **we would like to welcome anyone who is interested in it to join us for this wonderful sharding parser**! 😃 

> Take a look at the task detail below, you would learn it more. ⬇️ 

**Please follow up with your comment 📝 , if you want to join**!  What’s more, welcome your any comment and suggestion!

tristaZero:
## Here is instruction

### Tasks:
1. Add rule visitors in `MySQLVisitor.java` based on g4 files[1].
2. Check whether rule name in g4 files is appropriate or not.
3. Add parsed integrated test for your **new main rule visitor**.

### Task detail:
1. Add rule visitors

1.1 Since `MySQLVisitor.java` extends `MySQLStatementBaseVisitor<ASTNode>.java`, you should overwrite `visitRuleName(ctx)` from `MySQLStatementBaseVisitor<ASTNode>` referring to rules in g4 files.
1.2 When adding `visitRuleName(ctx)`, you should put it below the corresponding note generally named `xxx.g4` and keep its order same with where it is in `xxx.g4` file.
1.3 What we need after parsing is those classes implement `SQLStatemnt` interface, so you should parse all necessary rules (add necessary rule visitors) making up `the main rules` in `MySQLStatement.g4`, and then fill necessary `segments` to `xxxSQLStatemnt`.


2.Check rule name

2.1 If we do not care one rule (do not add visitor for it), its name should end up with `_` in g4 file. For instance, we do not add visitor for `IDENTIFIER_` in `baseRule.g4`.
2.2 Otherwise this rule name should not contain `_` in the end.
2.3 If existing rule name does not accord with the above rules, please fix it.


3. Add parsed integrated test

When you successfully add `one main rule visitor`, adding parsed integrated test for it is necessary.

3.1 Copy sql case you need from `./incubator-shardingsphere/sharding-sql-test/src/main/resources/sql/sharding/` to `./incubator-shardingsphere/sharding-sql-test/src/main/resources/sql/visitor/` (`Please note to keep the same directory`). 
[...]

PR #3915, opened 2020-01-09
Add interface for antlr visitors

tristaZero:
Refer to #3914.


shardingsphere-sql-parser/shardingsphere-sql-parser-mysql/pom.xml
```diff
@@ -49,8 +49,8 @@
                         <id>antlr</id>
                         <configuration>
                             <libDirectory>src/main/antlr4/imports/mysql/</libDirectory>
-                            <listener>false</listener>
-                            <visitor>false</visitor>
+                            <listener>true</listener>
```
terrymanu: Do we need generate listener too?

tristaZero: Listeners are used for checking later and not necessary now.

PR #3934, opened 2020-01-11
Get ciphertext column as a string

longjy:
Encrypt the field, and take the value of string type directly. For example, after digital encryption, the database can be saved as varchar, and the corresponding field of bean can still be numbe

### DIFF
--- a/encrypt-core/encrypt-core-merge/src/main/java/org/apache/shardingsphere/encrypt/merge/dql/EncryptMergedResult.java
+++ b/encrypt-core/encrypt-core-merge/src/main/java/org/apache/shardingsphere/encrypt/merge/dql/EncryptMergedResult.java
@@ -44,15 +44,15 @@ public final class EncryptMergedResult implements MergedResult {
     public boolean next() throws SQLException {
         return mergedResult.next();
     }
-    
+
     @Override
     public Object getValue(final int columnIndex, final Class<?> type) throws SQLException {
-        Object value = mergedResult.getValue(columnIndex, type);
-        if (null == value || !queryWithCipherColumn) {
-            return value;
+        Optional<Encryptor> encryptor;
+        if (!queryWithCipherColumn || !(encryptor = metaData.findEncryptor(columnIndex)).isPresent()) {
+            return mergedResult.getValue(columnIndex, type);
         }
-        Optional<Encryptor> encryptor = metaData.findEncryptor(columnIndex);
-        return encryptor.isPresent() ? encryptor.get().decrypt(value.toString()) : value;
+        String ciphertext = (String) mergedResult.getValue(columnIndex, String.class);
+        return ciphertext == null ? null : encryptor.get().decrypt(ciphertext);
     }
     
     @Override


terrymanu:
I have a little confuse, why force get value from ResultSet by String type?
I can't get the point, what is the issue in original code?

encrypt-core/encrypt-core-merge/src/main/java/org/apache/shardingsphere/encrypt/merge/dql/EncryptMergedResult.java
```diff
@@ -44,15 +44,15 @@
...
-        if (null == value || !queryWithCipherColumn) {
-            return value;
+        Optional<Encryptor> encryptor;
+        if (!queryWithCipherColumn || !(encryptor = metaData.findEncryptor(columnIndex)).isPresent()) {
+            return mergedResult.getValue(columnIndex, type);
         }
-        Optional<Encryptor> encryptor = metaData.findEncryptor(columnIndex);
-        return encryptor.isPresent() ? encryptor.get().decrypt(value.toString()) : value;
+        String ciphertext = (String) mergedResult.getValue(columnIndex, String.class);
+        return ciphertext == null ? null : encryptor.get().decrypt(ciphertext);
```
SteNicholas: This should be `null == ciphertext` according to code style.

encrypt-core/encrypt-core-merge/src/main/java/org/apache/shardingsphere/encrypt/merge/dql/EncryptMergedResult.java
```diff
@@ -44,15 +44,15 @@
...
-        Object value = mergedResult.getValue(columnIndex, type);
-        if (null == value || !queryWithCipherColumn) {
-            return value;
+        Optional<Encryptor> encryptor;
+        if (!queryWithCipherColumn || !(encryptor = metaData.findEncryptor(columnIndex)).isPresent()) {
+            return mergedResult.getValue(columnIndex, type);
         }
-        Optional<Encryptor> encryptor = metaData.findEncryptor(columnIndex);
-        return encryptor.isPresent() ? encryptor.get().decrypt(value.toString()) : value;
+        String ciphertext = (String) mergedResult.getValue(columnIndex, String.class);
```
SteNicholas: Does the `mergeResult` include result type except String? This transformation could cause exception.

longjy: After verification, mysql is not a varchar field type, getString will not have a type exception. And encrypted field storage is generally varchar, Encryptor class
String encrypt (Object plaintext) also returns String

encrypt-core/encrypt-core-merge/src/main/java/org/apache/shardingsphere/encrypt/merge/dql/EncryptMergedResult.java
longjy: Coding style I will fix

terrymanu:
Why close this pr?

longjy:
> Why close this pr?

Misoperation, please help me reopen PR

PR #3972, opened 2020-01-14
Determine if there is an encryption requirement according to all SQL tables

longjy:
Only modified here, but it is not clear whether only the first table was judged for other reasons before

see    https://github.com/apache/incubator-shardingsphere/issues/3971


tristaZero:
@longjy  Hi sir, thanks for your PR.
what you modified appears reasonable, however could you add one unit test to verify our conclusion? My suggestion is to new one unit test or add test for `EncryptStatementTest` or `EncryptPreparedStatementTest`. :-)
Look forward your reply, thx.

terrymanu:
Closed because we change the default branch from dev to master.
Please re-summit pr to master.

PR #3996, opened 2020-01-16
carry pr#3969 from dev to 4.0.1

### LINKED ISSUE #3968: Ambiguous description of COM_STMT_EXECUTE in MySQL protocol
The description of COM_STMT_EXECUTE command like this:

image

What if the **num-params** equals 0? 

Implementation of Connector/J like this:

image

While Implementation of sysbench like this:

image

The former has 1 more byte than latter:

image

Sharding-Proxy's command handler only support Connector/J's implementation for now, so access by sysbench will be failed. We need to support all two implementations.

tuohai666:
Fixes #3968.


sharding-proxy/sharding-proxy-transport/sharding-proxy-transport-mysql/src/main/java/org/apache/shardingsphere/shardingproxy/transport/mysql/packet/command/query/binary/execute/MySQLComStmtExecutePacket.java
```diff
@@ -37,7 +37,7 @@
 
 /**
  * COM_STMT_EXECUTE command packet for MySQL.
- * 
+ *
```
SteNicholas: Recover to origin style.

PR #3999, opened 2020-01-16
#3968, Ambiguous description of COM_STMT_EXECUTE in MySQL protocol

### LINKED ISSUE #3968: Ambiguous description of COM_STMT_EXECUTE in MySQL protocol

tuohai666:
Fixes #3968.

### DIFF
--- a/sharding-proxy/sharding-proxy-transport/sharding-proxy-transport-mysql/src/main/java/org/apache/shardingsphere/shardingproxy/transport/mysql/packet/command/query/binary/execute/MySQLComStmtExecutePacket.java
+++ b/sharding-proxy/sharding-proxy-transport/sharding-proxy-transport-mysql/src/main/java/org/apache/shardingsphere/shardingproxy/transport/mysql/packet/command/query/binary/execute/MySQLComStmtExecutePacket.java
@@ -33,6 +33,7 @@ import org.apache.shardingsphere.shardingproxy.transport.mysql.payload.MySQLPack
 
 import java.sql.SQLException;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 
 /**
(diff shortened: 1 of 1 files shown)


sharding-proxy/sharding-proxy-transport/sharding-proxy-transport-mysql/src/main/java/org/apache/shardingsphere/shardingproxy/transport/mysql/packet/command/query/binary/execute/MySQLComStmtExecutePacket.java
```diff
@@ -86,7 +86,7 @@ public MySQLComStmtExecutePacket(final MySQLPacketPayload payload) throws SQLExc
         } else {
             nullBitmap = null;
             newParametersBoundFlag = null;
-            parameters = null;
+            parameters = new ArrayList<>(0);;
```
terrymanu: How about use Collections.emptyList()?

terrymanu:
Please set milestone and lable

sharding-proxy/sharding-proxy-transport/sharding-proxy-transport-mysql/src/main/java/org/apache/shardingsphere/shardingproxy/transport/mysql/packet/command/query/binary/execute/MySQLComStmtExecutePacket.java
tuohai666: Yes, it's better.

PR #4045, opened 2020-01-20
#4042, Use the two-stage parsing strategy

### LINKED ISSUE #4042: Use the two-stage parsing strategy
The first step uses a slightly weaker parsing strategy, SLL(\*), that almost always works. If the first parsing step fails, try the full LL(\*) parse.

After failing the first step, we don't know whether it's a true syntax error or whether it's because the SLL(\*) strategy wasn't strong enough.  Input that passes the SLL(\*) step is guaranteed to pass the full LL(\*) , so there's no point in trying out that more expensive stratety. Input fails the second step is truly syntactically invalid.

tuohai666:
Fixes #4042.

### DIFF
--- a/shardingsphere-sql-parser/shardingsphere-sql-parser-engine/src/main/java/org/apache/shardingsphere/sql/parser/core/parser/SQLParserEngine.java
+++ b/shardingsphere-sql-parser/shardingsphere-sql-parser-engine/src/main/java/org/apache/shardingsphere/sql/parser/core/parser/SQLParserEngine.java
@@ -63,7 +63,7 @@ public final class SQLParserEngine {
             ((Parser) sqlParser).setErrorHandler(new BailErrorStrategy());
             ((Parser) sqlParser).getInterpreter().setPredictionMode(PredictionMode.SLL);
             parseTree = sqlParser.execute().getChild(0);
-        } catch (ParseCancellationException ex) {
+        } catch (final ParseCancellationException ex) {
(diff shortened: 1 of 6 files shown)


shardingsphere-sql-parser/shardingsphere-sql-parser-engine/src/main/java/org/apache/shardingsphere/sql/parser/core/parser/SQLParserEngine.java
```diff
@@ -51,7 +57,18 @@
...
      */
     public SQLAST parse() {
-        ParseTree parseTree = SQLParserFactory.newInstance(databaseTypeName, sql).execute().getChild(0);
+        SQLParser sqlParser = SQLParserFactory.newInstance(databaseTypeName, sql);
+        ParseTree parseTree;
+        try {
+            ((Parser) sqlParser).setErrorHandler(new BailErrorStrategy());
+            ((Parser) sqlParser).getInterpreter().setPredictionMode(PredictionMode.SLL);
+            parseTree = sqlParser.execute().getChild(0);
+        } catch (ParseCancellationException ex) {
```
terrymanu: should add final for exception catch

tuohai666: fixed

PR #4067, opened 2020-01-22
Using multithreading make TableMetaData faster

### LINKED ISSUE #3681: tomcat start slowly with default datasource setting
use shardingjdbc 4.0 version.
i config shardingjdbc with spring-namespace.  and there are about two hundred tables that do not  participate in split database and table.  i use default-data-source configuration to set it. but i found that tomcat start is very slowly for about two minutes。i checked the shardingjdbc source code and found the class "TableMetaDataInitializer.java" "load" method cost a lot of time . here is the source code：
image。
the loadDefaultTables method will  load all the tables  info in default-data-source .but there are about two hundred tables,  that will caused the tomcat start slowly.   any idea?

beijing-penguin:
Fixes #3681.

- Overload ExecutorEngine execute() method
- load() method add a serial parameter
- use maxConnectionsSizePerQuery to control init load speed


tristaZero:
Please refer to #4091

tristaZero:
There is a better solution, so it was closed. please refer to  #4091

PR #4068, opened 2020-01-22
Encryptor encrypt plaintext with null object

### LINKED ISSUE #3924: the method get a 'null' string value when the encryption field value is null.
the method get a 'null' string value when the encryption field value is null.

SteNicholas:
Fixes #3924.

`Encryptor` encrypts plaintext with null object could cause exception, because `String.valueOf` method with null object return "null" string, which should be regarded as null object.  Therefore, this should add null object judgement. Similar to [3925](https://github.com/apache/incubator-shardingsphere/pull/3925/files).

- Method `encrypt` of   dd null object judgement.

### DIFF
--- a/encrypt-core/encrypt-core-common/src/main/java/org/apache/shardingsphere/encrypt/strategy/impl/AESEncryptor.java
+++ b/encrypt-core/encrypt-core-common/src/main/java/org/apache/shardingsphere/encrypt/strategy/impl/AESEncryptor.java
@@ -60,6 +60,9 @@ public final class AESEncryptor implements Encryptor {
     @Override
     @SneakyThrows
     public String encrypt(final Object plaintext) {
+        if (null == plaintext) {
+            return null;
+        }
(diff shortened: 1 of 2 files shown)


encrypt-core/encrypt-core-common/src/main/java/org/apache/shardingsphere/encrypt/strategy/impl/AESEncryptor.java
```diff
@@ -70,7 +73,7 @@ public Object decrypt(final String ciphertext) {
         if (null == ciphertext) {
             return null;
         }
-        byte[] result = getCipher(Cipher.DECRYPT_MODE).doFinal(Base64.decodeBase64(String.valueOf(ciphertext)));
+        byte[] result = getCipher(Cipher.DECRYPT_MODE).doFinal(Base64.decodeBase64(ciphertext));
```
tristaZero: Why do you want to delete `String.valueOf()`?

PR #4076, opened 2020-01-24
persist SchemaNames for getChildren.

dongzl:
For #3185 .

persist SchemaNames for getChildren.


dongzl:
closed, reference #4109 .

PR #4091, opened 2020-01-27
Using multithread to load tablemetadata from default data source

### LINKED ISSUE #3681: tomcat start slowly with default datasource setting

tristaZero:
Fixes #3681.


zebei:
I am using version 4.1.0, but found that the effect of multithread to load tablemetadata is not obvious. 
When I used it, there were only 200 tables in the database, setting parameter 'max.connections.size.per.query=30' , console print 'Meta data load finished, cost 142547 milliseconds.'
Personal advice will load the data table to the developer optional configuration items.
 @tristaZero

Issue #3947, opened 2020-01-13
Improve integrate SQL parser test engine


terrymanu:
The current integrate SQL parser test engine is created by old SQL parser engine which parsed from sharding rule and encrypt rule.
The new SQL parser engine is decoupled with rule which only focus SQL itself, so we need to improve the test engine.
The tasks are:

- [x] Redesign SQL parser test engine.
  - [x] Redesign jaxb domain.
  - [x] Redesign assert engine.
- [x] Remove useless remain assertions.
  - [x] Remove SQL cases for sharding, encrypt and shadow, just need a unify SQL cases.
  - [x] Remove assertions for sharding condition which has already move to route module.
  - [x] Remove assertions for token which has already move to rewrite module.
- [ ] Finish accurate assert for all test cases.
  - [x] Finish accurate assert for DML.
    - [x] Finish accurate assert for select.
    - [x] Finish accurate assert for insert.
    - [x] Finish accurate assert for update.
    - [x] Finish accurate assert for delete.
  - [x] Finish accurate assert for DDL.
    - [x] Finish accurate assert for create table.
    - [x] Finish accurate assert for alter table.
    - [x] Finish accurate assert for drop table.
    - [x] Finish accurate assert for truncate table. 
    - [x] Finish accurate assert for create index.
    - [x] Finish accurate assert for alter index. 
    - [x] Finish accurate assert for drop index.
  - [ ] Finish accurate assert for DCL.
    - [x] Finish accurate assert for grant.
    - [x] Finish accurate assert for revoke.
    - [x] Finish accurate assert for create-user.
    - [ ] Finish accurate assert for create-role.
    - [ ] Finish accurate assert for create-login.
    - [ ] Finish accurate assert for alter-user.
    - [ ] Finish accurate assert for alter-role.
    - [ ] Finish accurate assert for alter-login.
    - [ ] Finish accurate assert for drop-user.
    - [ ] Finish accurate assert for drop-role.
    - [ ] Finish accurate assert for drop-login.
    - [ ] Finish accurate assert for deny-user.
    - [ ] Finish accurate assert for rename-user.
    - [ ] Finish accurate assert for set-role.
    - [ ] Finish accurate assert for set-password.
  - [x] Finish accurate assert for TCL.
    - [x] Finish accurate assert for set transaction.
    - [x] Finish accurate assert for set auto commit.
    - [x] Finish accurate assert for begin transaction.
    - [x] Finish accurate assert for commit.
[...]

PR #4100, opened 2020-01-29
Finish accurate assert for alter table

terrymanu:
For #3947.


terrymanu:
Invalid because version control error

PR #4223, opened 2020-02-10
add revoke DCL SQL visitor parse and test unit

beijing-penguin:
- add revoke.xml
- add revoke visitor parse code


tristaZero:
Hi @beijing-penguin It is a great PR, which i can see your understanding of Visitor mechanism.
Congrats for your grow-up!

Other than that, the rule name of `onObjectClause_` should be renamed a `onObjectClause` in its g4 file. You could fix it in another PR.

PR #4295, opened 2020-02-13
#4278 Add log details for metadata load when app startup

### LINKED ISSUE #4278: Add log details for metadata load when app startup
ShardingSphere need database and table metadata when app startup.
It will spend lots time if there are lots of amount databases and tables.
It is better to log more details about metadata load when app startup.

The log may include:

- [x] Start to load metadata.
- [x] Number of tables and details of each table name.
- [x] Finish to load metadata and log the total loading time.

yuzel:
Fixes #4278

- Add log details for metadata load when app startup
-
-


terrymanu:
A new suggestion:
Can we consider about log the table one by one, it makes the loading looks like active. So the log format should be: 

```
Loading metadata, total tables: %s
Loading table %s 1
Loading table % s2
...
Loading table %s n
Loading metadata finished, cost %s miles
```

PR #4312, opened 2020-02-14
support parsing multiple parameter-marker from one expr

tristaZero:


sharding-sql-test/src/main/resources/sql/supported/dml/update.xml
```diff
@@ -20,7 +20,7 @@
     <sql-case id="update_without_alias" value="UPDATE t_order SET status = ? WHERE order_id = ? AND user_id = ?" />
     <sql-case id="update_with_alias" value="UPDATE t_order AS o SET o.status = ? WHERE o.order_id = ? AND o.user_id = ?" db-types="MySQL,H2" />
     <!-- FIXME cannot parse multiple parameter-marker from one expr -->
```
terrymanu: We should remove FIXME if the case fixed.

PR #4324, opened 2020-02-15
add downloadRoot for frontend-maven-plugin

cherrylzhao:
`frontend-maven-plugin` will download node and npm from `https://nodejs.org/dist` by default, this is very slow for china's network. we can use a mirror site(`http://npm.taobao.org/mirrors/node/`) to speed up the build of ui-frontend module.


cherrylzhao:
just close, it's bettor to  prepare node_modules in local side

PR #4328, opened 2020-02-16
add profile releaseCN to use local node&cnpm

kimmking:
- add profile releaseCN to use local node&cnpm

### DIFF
--- a/shardingsphere-ui/shardingsphere-ui-frontend/pom.xml
+++ b/shardingsphere-ui/shardingsphere-ui-frontend/pom.xml
@@ -159,5 +159,4 @@
             </build>
         </profile>
     </profiles>
-
 </project>
--- a/shardingsphere-ui/shardingsphere-ui-frontend/pom.xml
+++ b/shardingsphere-ui/shardingsphere-ui-frontend/pom.xml
@@ -16,7 +16,8 @@
   ~ limitations under the License.
   -->
 
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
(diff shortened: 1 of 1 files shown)


shardingsphere-ui/shardingsphere-ui-distribution/shardingsphere-ui-bin-distribution/pom.xml
```diff
@@ -62,6 +62,34 @@
     <profiles>
         <profile>
             <id>release</id>
```
cherrylzhao: release and releaseCN is same, shall we keep only one?

shardingsphere-ui/shardingsphere-ui-distribution/shardingsphere-ui-src-distribution/pom.xml
```diff
@@ -44,6 +44,9 @@
     <profiles>
         <profile>
             <id>release</id>
+            <activation>
```
cherrylzhao: if we make activation=true distribution will happed every time

cherrylzhao: if we make activation=true distribution will happed every time, I think we should delete it

shardingsphere-ui/shardingsphere-ui-frontend/pom.xml
```diff
@@ -31,63 +31,120 @@
...
+                </goals>
+                <configuration>
+                  <nodeVersion>${node.version}</nodeVersion>
+                  <npmVersion>${npm.version}</npmVersion>
+                </configuration>
+              </execution>
+              <execution>
+                <id>npm run rimraf</id>
+                <goals>
+                  <goal>npm</goal>
```
terrymanu: The indent of xml is incorrect.

kimmking: done

kimmking:
only add profile in shardingsphere-ui-frontend pom.

Issue #4335, opened 2020-02-17
Switch SQL parser engine to new one


tristaZero:
We plan to present a better performance SQL parser for 5.x release. Basically all SQL visitors for different DBs, including MySQL, PostgreSQL, Oracle, SQLServer has finished already, and here is the current to-do list,
- [x] Switch SQL parser engine to new one.
- [x] Delete stale SQL parser engine.

PR #4384, opened 2020-02-20
add table segment for projection and where

terrymanu:
For #4335.


dongzl:
There are some duplicate codes, why don't we create a parent class and put the duplicate codes?

PR #4399, opened 2020-02-21
sharding-scaling-ui fix details

### LINKED ISSUE #4234: The enhancement for the first version sharding-scaling-ui
Scaling-ui has almost done.

But when I test, I found there are some problem need to fix and enhance.

problem:

1. scaling-ui can't configure the host of scaling-server.
2. UI has cross-domain problem so that web browser which do not support cross-domain can't use scaling.

enhancement:

1. History Task total progress only update when history task finished.
2. The detail progress of history task can't refresh automatically. 

For solve these problems, we should do follow thinks:

- [x] Add configure way for the host of scaling-service in `data-scaling` page.
- [x] Add gateway forwarding in shardingsphere-ui-backend to solve cross-domain problem.
- [x] Enhance progress show.
    - [x] Update history task total progress.
    - [x] Update detail progresses.
    - [x] Scaling service name will disappear when `data-scaling` page flush.

wqzwh:
Fixes #4234 .

- 1. Refresh the data scaling page after the service is configured. The service is not displayed
- 2. Entering data scaling for the first time is to report an error rather than pop-up the window for filling
- 3. Progress details page, progress will not be automatically updated, the window needs to be closed and reopened


shardingsphere-ui/shardingsphere-ui-frontend/src/views/data-scaling/module/index.vue
```diff
@@ -24,7 +24,7 @@
         icon="el-icon-plus"
         @click="add">{{ $t('dataScaling.btnTxt') }}</el-button>
       <span style="margin-left: 20px;">
-        server: <samp style="color: #E17425;">{{ serviceForm.serviceName }}</samp>
+        server: <samp style="color: #E17425;">{{ serviceForm.serviceUrl }}</samp>
```
KomachiSion: I think there should show the serviceName rather than url

PR #4475, opened 2020-02-25
rename sharding-orchestration-center-zookeeper-curator

### LINKED ISSUE #4437: Create Alias segment
🗒 :
- [x] Create Alias segment.
- [x] Replace Alias member of TableSegment with Alias segment.
- [x] Replace Alias member of ColumnProjectionSegment with Alias segment.
- [x] Replace Alias member of ExpressionProjectionSegment and AggregationProjectionSegment with Alias segment.
- [x] Refactor the interface of `AliasAvailable`

### LINKED ISSUE #4473: Rename project name `sharding-orchestration-center-zookeeper-curator`
Hi, community.

We should rename project name `sharding-orchestration-center-zookeeper-curator` to `sharding-orchestration-center-zookeeper`, cause curator is the client library for zookeeper. It shouldn't be in the project name.

kimmking:
Fixes #4473.

- Rename project name `sharding-orchestration-center-zookeeper-curator`


kimmking:
> The comment `Fixes #4437.` maybe incorrect. The issue NO is not match for this PR.

correct to 4473

kimmking:
> Maybe we need to consider about this PR.
> If using curator is named as `sharding-orchestration-center-zookeeper`, how to deal with using native zookeeper API only?

We only need a zookeeper implement, so if a native zookeeper API is better than curator, it should be instead of this one in the same sub project.

PR #4476, opened 2020-02-25
fix sharding-orchestration-center-zookeeper-curator references

kimmking:
- fix sharding-orchestration-center-zookeeper-curator references


kimmking:
merged into #4475, it hasn't merged, but I forget.

PR #4500, opened 2020-02-27
try to optimize ci

nancyzrh:
test for optimize ci


kimmking:
good.

nancyzrh:
try to resolve block due to scm checkout error

Issue #4510, opened 2020-02-28
Preparing to merge sharding-scaling branch into master branch


KomachiSion:
ShardingScaling has finish some basic function for migrating and scaling out, detail see #3598 .

So I'd like to open a discussion and do some checks for merging sharding-scaling/sharding-scaling-ui into master branch.

Following is the check list:

- [x] Double check the License and dependency
- [x] Double check ASF headers
- [x] Check spelling and naming

Any more check item?

PR #4543, opened 2020-03-01
UI details modification

wqzwh:
- Handling of display problems of bread crumb components
- syncStatic replace syncStatus


KomachiSion:
For #4510

PR #4545, opened 2020-03-01
Issue2010

### LINKED ISSUE #2010: Do not support connectionInitSqls config of druid connection while using datasource orchestration
## Bug Report

- Searched open and closed [GitHub issues](https://github.com/sharding-sphere/sharding-sphere/issues).
- Read documentation: [ShardingSphere Doc](http://shardingsphere.io/document/current/en/overview/).

3.1.0
Sharding-JDBC
### Expected behavior
support     connectionInitSqls
### Actual behavior
do not
connectionInitSqls config is a collection:

com.alibaba.druid.pool.DruidAbstractDataSource
```
    public Collection<String> getConnectionInitSqls() {...}
```
while uploading, only general types will be serialized, which do not include ```Collection```

io.shardingsphere.core.config.DataSourceConfiguration
```
        GENERAL_CLASS_TYPE = Sets.<Class<?>>newHashSet(boolean.class, Boolean.class, int.class, Integer.class, long.class, Long.class, String.class);
```

menghaoranss:
Fixes #2010  .

- support connectionInitSqls config of dbcp and druid connection pool via spring-boot 
- support connectionInitSqls config of dbcp and druid while using orchestration

The following configurations of `connectionInitSqls` has been verified using dbcp and druid connection pools:
+ springboot
```
spring.shardingsphere.datasource.ds_0.connectionInitSqls=set names utf8mb4;,set serverTimezone=UTC;
```
+ yaml
```
connectionInitSqls:
      - set names utf8mb4;
      - set serverTimezone=UTC;
```
+ namespace
```
<property name="connectionInitSqls">
            <list>
                <value>set names utf8mb4;</value>
                <value>set serverTimezone=UTC;</value>
            </list>
</property>
```
+ Using the local and zookeeper config center, the connectionInitSqls is uploaded to the config center as: 
```
connectionInitSqls:
      - set names utf8mb4;
      - set serverTimezone=UTC;
```
+ Configure connectionInitSqls using only Configuration Center
+ Configure connectionInitSqls with UI


shardingsphere-underlying/shardingsphere-common/src/test/java/org/apache/shardingsphere/underlying/common/config/DataSourceConfigurationTest.java
```diff
@@ -83,4 +88,25 @@ public void assertAddAlias() {
         assertThat(actual.getProperties().get("jdbcUrl"), is(actual.getProperties().get("url")));
         assertThat(actual.getProperties().get("username"), is(actual.getProperties().get("user")));
     }
+    
+    @Test
+    public void assertGetDataSourceConfigurationWithConnectionInitSqls() throws SQLException {
```
kimmking: Here is a integration test.
I suggestion:
1. if do this way, it's neccesary to use HikariDataSource instead of BasicDatasource.
2. creat a table in h2 db, and set an insert sql to initSql, then check it work well by a select and compare.

Or anothor approach: 
just mock a jdbc driver behind HikariDataSource, check hikariDataSource send the initSql to mock jdbc connection/statement.

cherrylzhao: connectionInitSqls in HikariDataSource is a String type, but we want to verify a collection type for this issue. also I found that collection type property not existed in in HikariDataSource, it seems that DBCP2 is ok for that

kimmking:
@menghaoranss  need some help？

menghaoranss:
> I want to know have you verify the spring-boot, spring-namespace, yaml case in orchestration mode.

@cherrylzhao , yes , i configured connectionInitSqls list via spring-boot, spring-namespace and yaml 
in orchestration,  the connectionInitSqls list uploaded to configCenter as yaml, and these sqls executed successfully. I also verified only use orchestration without local configuration, it worked well too.

shardingsphere-underlying/shardingsphere-common/src/test/java/org/apache/shardingsphere/underlying/common/config/DataSourceConfigurationTest.java
menghaoranss: > connectionInitSqls in HikariDataSource is a String type, but we want to verify a collection type for this issue. also I found that collection type property not existed in in HikariDataSource, it seems that DBCP2 is ok for that

@kimmking @cherrylzhao In HikariDataSource, the configuratoin is connectionInitSql, and it is a String type, i also veryfied HikariDataSource , it's ok. DBCP2 and Druid support Connection type,  the only difference is the return parameter of `getConnectionInitsqls` method, DBCP2 return List type, Druid return Connection type, I have done compatibility processing in the pr.

PR #4563, opened 2020-03-02
fix a wrong place license

kimmking:
- fix a wrong place license


kimmking:
Two tips:
1. If we use IDEA to update copyright, it will process the files type (java/xml/sql/properties).
Others type files (sh/yaml/yml) , manual processing required.
2. File `examples/src/resources/manual_schema.sql` may added a license in the middle of lines by IDEA. Need correct it manually too.

Issue #4452, opened 2020-02-24
I want try to sharding-orchestration-core module


zjcnb:
Detail task in #4439

terrymanu:
Welcome, I just assign the issue to you

PR #4581, opened 2020-03-03
add ut for  for sharding-orchestration-core module

zjcnb:
For #4452


menghaoranss:
@zjcnb please fix ci problem: 
```
[ERROR] src/test/java/org/apache/shardingsphere/orchestration/internal/registry/listener/PostShardingConfigCenterRepositoryEventListenerTest.java:[58,83] (misc) FinalParameters: Parameter event should be final.
```

PR #4601, opened 2020-03-04
Replace mocks with instances in Unit Tests

### LINKED ISSUE #4540: Do not use mock on regular test cases
Mock should only use to building difficult or slow objects.
We'd better avoid using mock on regular test cases, it is not a good way for guarantee for test.

The test cases need to rewrite are:

- [ ] ProjectionEngineTest
- [ ] ProjectionsContextEngineTest
- [ ] GroupByContextEngineTest
- [ ] OrderByContextEngineTest
- [ ] PaginationContextTest
- [ ] PaginationContextEngineTest
- [ ] RowNumberPaginationContextEngineTest
- [ ] TopPaginationContextEngineTest
- [ ] SQLStatementContextFactoryTest
- [ ] SelectStatementContextTest

rohithbalaji123:
Fixes #4540.

The test cases rewritten are as follows:
- [x] ProjectionEngineTest
- [x] ProjectionsContextEngineTest
- [x] GroupByContextEngineTest
- [x] OrderByContextEngineTest
- [x] PaginationContextTest
- [x] PaginationContextEngineTest
- [x] RowNumberPaginationContextEngineTest
- [x] TopPaginationContextEngineTest
- [ ] SQLStatementContextFactoryTest
- [x] SelectStatementContextTest

#### Further Comments
- There is one reference in `SQLStatementContextFactoryTest` which mocks an interface. I couldn't find a suitable implementation to replace it with. There are few `TCL` type concerete classes implementing to `SQLStatement`, but it looks too out of place to be included in the particular test case. So, I am leaving the mock for it as such.
- Few assert conditions in `GroupByContextEngineTest.assertCreateGroupByContextWithGroupBy` and `OrderByContextEngineTest.assertCreateOrderByWithOrderBy` seemed redundant as a previous assert condition verifying the list of items returned, already asserts the matching of the indexes.
- Apart from the above two variations, everything else are straight-forward changes.

### DIFF
--- a/shardingsphere-sql-parser/shardingsphere-sql-parser-relation/src/test/java/org/apache/shardingsphere/sql/parser/relation/segment/select/groupby/engine/GroupByContextEngineTest.java
+++ b/shardingsphere-sql-parser/shardingsphere-sql-parser-relation/src/test/java/org/apache/shardingsphere/sql/parser/relation/segment/select/groupby/engine/GroupByContextEngineTest.java
@@ -30,7 +30,6 @@ import org.apache.shardingsphere.sql.parser.sql.value.identifier.IdentifierValue
 import org.junit.Test;
 
 import java.util.Arrays;
-import java.util.Collection;
 
 import static org.hamcrest.CoreMatchers.is;
 import static org.junit.Assert.assertThat;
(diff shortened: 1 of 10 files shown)


terrymanu:
Wonderful pull request, thank you very much.

PR #4663, opened 2020-03-08
add metadata load time log

kimmking:
For #4278.

- add metadata load time log


terrymanu:
Any different with #4295?

Issue #4482, opened 2020-02-26
Correct the `DataTypeName` for each DataBase


tristaZero:
The definition of `DataTypeName` in `RuleName.g4`(Antlr grammer file) of each DataBase is not correct so far, therefore this issue is created to correct them. The sub-task list is below,
- [x] Correct the `DataTypeName` for `MySQL`.
- [x] Correct the `DataTypeName` for `PostgreSQL`. 
- [x] Correct the `DataTypeName` for `SQLServer`. 
- [x] Correct the `DataTypeName` for `Oracle`. 
- [x] Correct the `DataTypeName` for `SQL92`. 

Q1: How to learn about the definition of `DataTypeName`?
A:
 - Official document of each DataBase, i.g [MySQL](https://dev.mysql.com/doc/refman/8.0/en/data-types.html).
- [Antlr SQL grammer](https://github.com/antlr/grammars-v4/tree/master/sql)
- Other related Website. (I am not sure. 😃 )

Q2: Which file of ShardingSphere does `DataTypeName` exist?
A: All files named with `RuleName.g4`

yuzel:
Ok, i can support mysql and postgresql first.

tristaZero:
Hi @yuzel You made a good start for this issue, good job. 👍

yuzel:
> Hi @yuzel You made a good start for this issue, good job. 👍
> BTW, are you willing to go on with PostgreSQL or other ones?

Yes, I am going to continue to modify PostgreSQL.

tristaZero:
@yuzel Your continual contribution helps this ISSUE complete. Very appreciated your effort. Hope see your interaction in community more. :)

PR #4714, opened 2020-03-11
 #4482 Correct the DataTypeName for SQL92

yuzel:
Ref #4482 .

reference link
- [SQL92](http://www.contrib.andrew.cmu.edu/~shadow/sql/sql1992.txt) Directory 6.1

note:
I didn't remove the identifier because I'm not sure if there are other types.


yuzel:
Package structure changes, resubmit

PR #4718, opened 2020-03-12
Update faq

### LINKED ISSUE #4604: Too slow to load meta data when start up
At present, I have more than 3000 sub tables. It takes too long to load sharding at the start of the project. I have set the consistency check metadata to "false" and it takes effect, but the time is still very long. I have located the code part that takes a long time

private Map<String, TableMetaData> loadDefaultTables(final ShardingRule shardingRule) throws SQLException {
        Map<String, TableMetaData> result = new HashMap<>(shardingRule.getTableRules().size(), 1);
        Optional<String> actualDefaultDataSourceName = shardingRule.findActualDefaultDataSourceName();
        if (actualDefaultDataSourceName.isPresent()) {
            for (String each : getAllTableNames(actualDefaultDataSourceName.get())) {
                result.put(each, tableMetaDataLoader.load(each, shardingRule));
            }
        }
        return result;
    }
tableMetaDataLoader.load(each, shardingRule)
*This part is the main reason

geomonlin:
Fixes #4604 .


tristaZero:
@geomonlin Hi
Could you add `(Version >= 3.0.0.M3)` to the second item? For instance,
> Configure `max.connections.size.per.query`(Default value is 1) higher referring to connection pool you adopt (Version >= 3.0.0.M3).

PR #4747, opened 2020-03-13
Optimize seata at transaction manager

cherrylzhao:
- inline xid broadcast
- clear xid of sub thread when SQL executed
- revise unit test


sharding-transaction/sharding-transaction-base/sharding-transaction-base-seata-at/src/main/java/org/apache/shardingsphere/transaction/base/seata/at/Constant.java
```diff
@@ -17,31 +17,10 @@
...
-import java.util.Map;
-
 /**
- * Seata transaction broadcaster.
+ * Constant.
  */
-class SeataTransactionBroadcaster {
-    
-    private static final String SEATA_TX_XID = "SEATA_TX_XID";
+final class Constant {
```
terrymanu: Please do not use `Constant` as a class name, the scope of this name is too big, it is very easy conflict and misunderstanding with other modules

terrymanu:
Please set the label as `bug` or `enhancement`

sharding-transaction/sharding-transaction-base/sharding-transaction-base-seata-at/src/main/java/org/apache/shardingsphere/transaction/base/seata/at/Constant.java
cherrylzhao: ok, I have nine it to TransactionSQLExecuteHook

Issue #4671, opened 2020-03-09
Parsing more segments


tristaZero:
#### TODO
- RenameColumnSegment
    - [x] Instead of `columnName`, parse `ColumnNameSegment` as its member.

- ColumnAfterPositionSegment
    - [x] Instead of `columnName`, parse `ColumnNameSegment` as its member.

- ColumnPositionSegment
    - [x] Instead of `columnName`, parse `ColumnNameSegment` as its member.

- ColumnDefinitionSegment
    - [x] Instead of `columnName`, parse `ColumnNameSegment` as its member.

- ColumnDefinitionSegment
    - [x] Instead of `columnName`, parse `ColumnNameSegment` as its member.
    - [x] Instead of `dataType`, parse `DataTypeSegment` as its member.

### Notes
- Refer to `TableNameSegment` to create `ColumnNameSegment`.
- `ComparisonOperatorType`  is expected to be `enum`.

jingshanglu:
I'll support

jingshanglu:
There is ColumnSegment. ColumnSegment is refer by orthers, and is similar to ColumnNameSegment.I want  move ColumnSegment to  generic, and use ColumnSegment instead of ColumnNameSegment .How do you think? @tristaZero

tristaZero:
Hi, The last one seems optional, and the second last is our end subtask.

tristaZero:
All done, good job.

PR #4751, opened 2020-03-13
Fix 4671

jingshanglu:
ref #4671 .

- delete afterColumName from ColumnAfterPositionSegment
- fix testCase about position column


jingshanglu:
@ tristaZero

shardingsphere-sql-parser/shardingsphere-sql-parser-test/src/test/java/org/apache/shardingsphere/sql/parser/integrate/asserts/segment/definition/ColumnPositionAssert.java
```diff
@@ -47,15 +47,19 @@ public static void assertIs(final SQLCaseAssertContext assertContext, final Colu
         if (null != actual.getColumnName()) {
             columnname = actual.getColumnName().getQualifiedName();
         }
```
tristaZero: It is better to extract a function.

shardingsphere-sql-parser/shardingsphere-sql-parser-test/src/test/java/org/apache/shardingsphere/sql/parser/integrate/asserts/segment/definition/ColumnPositionAssert.java
```diff
@@ -47,15 +47,19 @@ public static void assertIs(final SQLCaseAssertContext assertContext, final Colu
...
-        assertThat(assertContext.getText("Column change position name assertion error: "), columnname, is(expected.getColumn().getName()));
+        String expectColumnName = null;
+        if (null != expected.getColumn()) {
+            expectColumnName = expected.getColumn().getName();
+        }
```
tristaZero: This situation is similar to the above one.

tristaZero:
Hi@jingshanglu, Your PR is great, but could you look at my comment which can make this PR perfect, I think. Also, welcome your input. :)

jingshanglu:
> Hi@jingshanglu, Your PR is great, but could you look at my comment which can make this PR perfect, I think. Also, welcome your input. :)

These codes are only used in this place, there is no need to draw a method

tristaZero:
@jingshanglu Ok, it is not a big deal, thanks for your feedback.

PR #4771, opened 2020-03-16
add assert index

### LINKED ISSUE #4690: Add start-index and stop-index assertion for ColumnOrderByItemSegment
This issue is to count the `start-index` and `stop-index` to assert `ColumnOrderByItemSegment `

1. Open class `OrderByItemAssert` and clear the comment of `SQLSegmentAssert.assertIs(assertContext, actual, expected)` in `assertColumnOrderByItem`.

3. Calculate the correct `start-index` and `stop-index` of `order-by` segment of this `SQL`.
4. Add the calculated `start-index` and `stop-index` labels of `order-by` segment to `parsing result` of this `SQL Case ID`, like
```
<order-by>
            <column-item name="user_id" start-index="10" stop-index="11" />
            <column-item name="order_id" start-index="20" stop-index="21" />
</order-by>
```
5. Add the correct `literal-start-index` and `literal-stop-index` of `order-by` segment to `parsing result` of this `SQL Case ID` when it is necessary, like
```
<order-by>
            <column-item name="user_id" literal-start-index="10" literal-stop-index="11" />
            <column-item name="order_id" literal-start-index="20" literal-stop-index="21" />
</order-by>
```
6. Run `SQLParserParameterizedTest` again to check whether this `SQL Case ID` does not stay in the list of `SQL` with  `bad assertion result`.

### Note
1. What is the difference between `literal-start-index` and `start-index`?

When a SQL does not contain parameter maker, i.e.,`?`, this SQL only has one form; Otherwise, you will see two forms of this SQL, i.e, a SQL form filled with parameter and a SQL form only with parameter marks. For instance,
```
Case1 One SQL without parameter mark,
select * from table where id = 1;

Case2 One SQL with parameter mark,
Original SQL: select * from table where id = ?;
SQL filled with Parameter: select * from table where id = 1;
```
Replace `the parameter mark` with `the actual parameter` will change the position of the segment. Therefore we need to add `start-index` to represent the start position of a segment for `Original SQL`, and `literal-start-index` is used to represent the start position of a segment for `SQL filled with Parameter`.

geomonlin:
Fixes #4690 .

### DIFF
--- a/shardingsphere-sql-parser/shardingsphere-sql-parser-test/src/test/java/org/apache/shardingsphere/sql/parser/integrate/asserts/segment/orderby/OrderByItemAssert.java
+++ b/shardingsphere-sql-parser/shardingsphere-sql-parser-test/src/test/java/org/apache/shardingsphere/sql/parser/integrate/asserts/segment/orderby/OrderByItemAssert.java
@@ -20,6 +20,7 @@ package org.apache.shardingsphere.sql.parser.integrate.asserts.segment.orderby;
 import lombok.AccessLevel;
 import lombok.NoArgsConstructor;
 import org.apache.shardingsphere.sql.parser.integrate.asserts.SQLCaseAssertContext;
+import org.apache.shardingsphere.sql.parser.integrate.asserts.segment.SQLSegmentAssert;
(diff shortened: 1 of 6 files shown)


tristaZero:
@geomonlin You did what we want, great job!

PR #4789, opened 2020-03-17
ref 4671

jingshanglu:
ref #4671 .

- fix mysql ColumnDefinitionSegment,change type of dataType from String to DataTypeSegment


jingshanglu:
@terrymanu CI failed for the same reason: "t_log not found",help take a look.

tristaZero:
Hi why is `CreateTableStatement` completely created in this PR? The same one has already existed in ShardingSphere, which makes me confused. Could you give me some hints?
@jingshanglu

tristaZero:
It contains much workload in this PR, yet i want to know why you want to add the duplicated class named `AlterTableStatement`?

jingshanglu:
> It contains much workload in this PR, yet i want to know why you want to add the duplicated class named `AlterTableStatement`?

The purpose is to modify the mysql code independently without affecting the code of other db

shardingsphere-sql-parser/shardingsphere-sql-parser-test/src/test/java/org/apache/shardingsphere/sql/parser/integrate/engine/SQLParserParameterizedTest.java
```diff
@@ -140,6 +140,9 @@ private static boolean isPlaceholderWithoutParameter(final Object[] sqlTestParam
     
     @Test
     public void assertSupportedSQL() {
+        if ("select_sub_query_with_table".equals(sqlCaseId)) {
+            System.out.println("adfasf");
+        }
```
tristaZero: Please remove them.

sharding-sql-test/src/main/resources/sql/supported/ddl/truncate.xml
```diff
@@ -17,7 +17,7 @@
   -->
 
 <sql-cases>
-    <sql-case id="truncate_table" value="TRUNCATE TABLE t_log" />
+    <sql-case id="truncate_table" value="TRUNCATE TABLE t_log" db-types="MySQL,Oracle,PostgreSQL"/>
```
tristaZero: Hi, why did you want to limit to those databases?

shardingsphere-sql-parser/shardingsphere-sql-parser-statement/src/main/java/org/apache/shardingsphere/sql/parser/sql/segment/generic/DataTypeLengthSegment.java
```diff
@@ -0,0 +1,37 @@
...
+@Setter
+public final class DataTypeLengthSegment implements SQLSegment {
+    
+    private int startIndex;
+    
+    private int stopIndex;
+    
+    private int length;
+    
+    private LinkedList<Integer> lengthPair;
```
tristaZero: Consider the samples of length may be (1) or (2, 1), I suggest the members of `DataTypeLengthSegment ` to be,
```
 private int startIndex;

     private int stopIndex;

     private int firstNumber;
 
     private int secondNumber;

     public Optional<int> getSecondNumber(){}
```

tristaZero:
Hi Thanks for your effort, and I appended my opinions for this PR, please review them.

sharding-sql-test/src/main/resources/sql/supported/ddl/truncate.xml
jingshanglu: This testcase for H2 failed, so remove it. I just change mysql parser . I think this change will not affect others.

shardingsphere-sql-parser/shardingsphere-sql-parser-statement/src/main/java/org/apache/shardingsphere/sql/parser/sql/segment/generic/DataTypeLengthSegment.java
jingshanglu: I think it is better as a whole. When using the whole length, users can use it according to their needs.

shardingsphere-sql-parser/shardingsphere-sql-parser-test/src/test/java/org/apache/shardingsphere/sql/parser/integrate/engine/SQLParserParameterizedTest.java
jingshanglu: I'll check.

sharding-sql-test/src/main/resources/sql/supported/ddl/truncate.xml
tristaZero: Why did H2 fail? AFAIK, when db is h2, parser test engine will use MySQL to parse SQL.

jingshanglu: Not parser test,it is jdbc-test.

tristaZero:
Here is my comment, please give it a look. @jingshanglu

shardingsphere-sql-parser/shardingsphere-sql-parser-statement/src/main/java/org/apache/shardingsphere/sql/parser/sql/segment/generic/DataTypeLengthSegment.java
tristaZero: What I concerned about is that `length` and `lengthPair`will confused user who considers `length` and `lengthPair` can have value at the same time. But the fact is that either `length` or `lengthPairan' can have the value at one time. If the user calls `getLength()` and `getLengthPair()` together, what will happen?

jingshanglu: If just have one num,lengthPair is null, if have two, length is null.

jingshanglu:
I have reverted to H2 coverage from previous test cases.

shardingsphere-sql-parser/shardingsphere-sql-parser-statement/src/main/java/org/apache/shardingsphere/sql/parser/sql/segment/generic/DataTypeLengthSegment.java
```diff
@@ -0,0 +1,45 @@
...
+@Setter
+public final class DataTypeLengthSegment implements SQLSegment {
+    
+    private int startIndex;
+    
+    private int stopIndex;
+    
+    private int firstNumber;
+    
+    private int secondNumber;
```
tristaZero: As I do not think `firstNumber` and `secondNumber` are proper names, i dug into many doc, and found `firstNumber` is supposed to be `precision`; On the other hand, `secondNumber` means `scale`[1]. Can you give them other expressive names?
[1] https://dev.mysql.com/doc/refman/5.7/en/precision-math-decimal-characteristics.html

jingshanglu: firstLength and secondLength ,is it ok?

tristaZero:
Hi @jingshanglu 
I appreciated your modification, and current PR is clear enough. Good job,   only one point here is the naming, could you look at my comment?

shardingsphere-sql-parser/shardingsphere-sql-parser-statement/src/main/java/org/apache/shardingsphere/sql/parser/sql/segment/generic/DataTypeLengthSegment.java
tristaZero: Actually, `firstxx` and `secondxx` have  no meaning. It is like your name is `Jingshang`,  rather than `firstPerson`. My suggestion is `precision` and `scale`, or `precisionLength` and `scaleLength`. How about you?

tristaZero:
Actually, `firstxx` and `secondxx` have  no meaning. It is like your name is `Jingshang`,  rather than `firstPerson`. My suggestion is `precision` and `scale`, or `precisionLength` and `scaleLength`. How about you?
@jingshanglu

jingshanglu:
> Actually, `firstxx` and `secondxx` have no meaning. It is like your name is `Jingshang`, rather than `firstPerson`. My suggestion is `precision` and `scale`, or `precisionLength` and `scaleLength`. How about you?
> @jingshanglu

ok

sharding-sql-test/src/main/resources/sql/supported/ddl/create.xml
```diff
@@ -24,8 +24,8 @@
     <sql-case id="create_global_temporary_table" value="CREATE GLOBAL TEMPORARY TABLE t_temp_log(id int, status varchar(10))" db-types="Oracle,PostgreSQL" />
     <sql-case id="create_local_temp_table" value="CREATE LOCAL TEMP TABLE t_temp_log(id int, status varchar(10))" db-types="PostgreSQL" />
     <sql-case id="create_unlogged_table" value="CREATE UNLOGGED TABLE t_log(id int, status varchar(10))" db-types="PostgreSQL" />
-    <sql-case id="create_index" value="CREATE INDEX t_log_index ON t_log (id)" />
-    <sql-case id="create_unique_index" value="CREATE UNIQUE INDEX t_log_index ON t_log (id)" />
+    <sql-case id="create_index" value="CREATE INDEX t_log_index ON t_log (id)"/>
+    <sql-case id="create_unique_index" value="CREATE UNIQUE INDEX t_log_index ON t_log (id)"/>
```
tristaZero: Hi You should reverse those changes, for spaces is needed.

tristaZero:
It seems correct, clear, and in line with the code of conduct currently. It is time to merge it. 
@jingshanglu

PR #4791, opened 2020-03-17
Fix 4782

yu199195:
Fixes 
https://github.com/apache/incubator-shardingsphere/issues/4782

another Fix Solve the problem of Chinese file path chaos when loading the file.


terrymanu:
I just close this pr.

Issue #4696, opened 2020-03-10
Make subquery parsing test work well


tristaZero:
To make the following subquery parsing test work well, we need to review the parsing process for subquery SQL.

- [x] Subquery in `./incubator-shardingsphere/sharding-sql-test/src/main/resources/sql/supported/dql/select.xml`
- [x] Subquery in `./incubator-shardingsphere/sharding-sql-test/src/main/resources/sql/supported/dql/select-sub-query.xml`
- [x] Subquery in `./incubator-shardingsphere/sharding-sql-test/src/main/resources/sql/supported/dql/select-pagination.xml`
- [x] Subquery in `./incubator-shardingsphere/sharding-sql-test/src/main/resources/sql/supported/dql/select-pagination-group-by-order-by.xml`
- [ ] `Subquery` in SubqueryRouteTest

tristaZero:
Hi @strongduanmu Recently I looked over those stale issues and found this one is related to #6286, so could you give the last item a look?

I mean, when we remove  `@Ignore("Can not support subquery at current")` in`SubqueryRouteTest`, some of the cases could not run well. Especially, some cases expecting `IllegalStateException.class`. I guess maybe we lack some judgments about `subquery routing`.

strongduanmu:
> Hi @strongduanmu Recently I looked over those stale issues and found this one is related to #6286, so could you give the last item a look?
> 
> I mean, when we remove `@Ignore("Can not support subquery at current")` in`SubqueryRouteTest`, some of the cases could not run well. Especially, some cases expecting `IllegalStateException.class`. I guess maybe we lack some judgments about `subquery routing`.

@tristaZero I will investigate the cause of this problem as soon as possible.

PR #4801, opened 2020-03-17
test sql with sub query table

tristaZero:
Ref #4696


tristaZero:
As all integration tests are passed, does any committer help merge this one?

PR #4812, opened 2020-03-17
Fix 4805.

yu199195:
Fixes https://github.com/apache/incubator-shardingsphere/issues/4805

### DIFF
--- a/sharding-sql-test/src/main/java/org/apache/shardingsphere/test/sql/loader/SQLCasesLoader.java
+++ b/sharding-sql-test/src/main/java/org/apache/shardingsphere/test/sql/loader/SQLCasesLoader.java
@@ -20,6 +20,7 @@ package org.apache.shardingsphere.test.sql.loader;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Splitter;
 import com.google.common.base.Strings;
+import com.google.common.collect.Lists;
 import lombok.SneakyThrows;
 import org.apache.shardingsphere.test.sql.SQLCase;
 import org.apache.shardingsphere.test.sql.SQLCaseType;
@@ -32,6 +33,7 @@ import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URL;
+import java.net.URLDecoder;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Enumeration;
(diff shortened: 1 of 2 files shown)


sharding-sql-test/src/main/java/org/apache/shardingsphere/test/sql/loader/SQLCasesLoader.java
```diff
@@ -190,7 +193,7 @@ private String getLiteralSQL(final String sql, final List<?> parameters) {
     }
     
     private static Collection<String> getDatabaseTypes(final String databaseTypes) {
-        return Strings.isNullOrEmpty(databaseTypes) ? getALlDatabaseTypes() : Splitter.on(',').trimResults().splitToList(databaseTypes);
+        return Strings.isNullOrEmpty(databaseTypes) ? getALlDatabaseTypes() : Lists.newArrayList(Splitter.on(',').trimResults().split(databaseTypes));
```
terrymanu: What is the different with origin code?

sharding-sql-test/src/main/java/org/apache/shardingsphere/test/sql/loader/SQLCasesLoader.java
```diff
@@ -73,13 +75,14 @@ public SQLCasesLoader(final String rootDirection) {
...
     
+    @SneakyThrows
     private static Map<String, SQLCase> loadSQLCasesFromTargetDirectory(final String path) {
         Map<String, SQLCase> result = new TreeMap<>();
         URL url = SQLCasesLoader.class.getClassLoader().getResource(path);
         if (null == url) {
             return result;
         }
-        File filePath = new File(url.getPath());
+        File filePath = new File(URLDecoder.decode(url.getPath(), "UTF-8"));
```
terrymanu: How about if the url is not for UTF-8 encoding? Use default encoding may better.

sharding-sql-test/src/main/java/org/apache/shardingsphere/test/sql/loader/SQLCasesLoader.java
yu199195: 1. splitToList  retrun readOnly List ,if  use add, remove ,can throws exception.
2. init list No default capacity is specified，so this method is @beta。

sharding-sql-test/src/main/java/org/apache/shardingsphere/test/sql/loader/SQLCasesLoader.java
yu199195: URLDecoder.decode() Without loss of performance, but also to prevent Chinese garbled code，

sharding-sql-test/src/main/java/org/apache/shardingsphere/test/sql/loader/SQLCasesLoader.java
terrymanu: Why getDatabaseTypes need to modify by user?

sharding-sql-test/src/main/java/org/apache/shardingsphere/test/sql/loader/SQLCasesLoader.java
terrymanu: How about if the url is not for UTF-8 encoding?

kimmking: url.getPath() is string, if its charset is wrong, decode is unuseful.

Issue #4778, opened 2020-03-16
 Optimize configuration structure of config&registry center 


menghaoranss:
For #4522
Based on the completed split of the config center and registry center, the related configuration structure will be adjusted as follows:

#### Current structure
```
/
├─orchestration-namespace
  ├─orchestration-name
  │  ├─config
  │  │  ├─authentication
  │  │  ├─props
  │  │  ├─schema
  │  │  │  ├─sharding-schema-name
  │  │  │  │  ├─datasource
  │  │  │  │  └─rule
  │  │  │  ├─encrypt-schema-name
  │  │  │  │  ├─datasource
  │  │  │  │  └─rule
  │  │  │  └─master_slave-schema-name
  │  │  │  │  ├─datasource
  │  │  │  │  └─rule
  │  ├─state
  │  │  ├─datasources
  │  │  └─instances
```
##### config center
```
/
├─orchestration-namespace
  ├─orchestration-name
  │  ├─config
  │  │  ├─authentication
  │  │  ├─props
  │  │  ├─schema
  │  │  │  ├─sharding-schema-name
  │  │  │  │  ├─datasource
  │  │  │  │  └─rule
  │  │  │  ├─encrypt-schema-name
  │  │  │  │  ├─datasource
  │  │  │  │  └─rule
  │  │  │  └─master_slave-schema-name
  │  │  │  │  ├─datasource
  │  │  │  │  └─rule
```
##### registry center
```
/
├─orchestration-namespace
  ├─orchestration-name
  │  ├─registry
  │  │  ├─datasources
  │  │  └─instances
```
Note:

+ Split the registry and modify state node to registry node to match the registry
+ Modify the datasources node of registry center to states, currently used to record the status of slave database

Task lists:
 - [x] 1. Refactor persist and get configuration from registry center
 - [x] 2. Refactor unit tests
 - [x] 3. Refactor UI

Any suggestions， please reply directly to this issue.

terrymanu:
It is better if change `states` to `status`

terrymanu:
Add `rule` type maybe not good because the type can be extension by user via SPI. It is better do not add a static field here.

menghaoranss:
> Add `type` maybe not good because the type can be extension by user via SPI. It is better do not add a static field here.

ok, I will reconsider the design of this part, thanks.

menghaoranss:
> It is better if change `states` to `status`

[...]

PR #4818, opened 2020-03-18
For issue4778, modify registry center root node

menghaoranss:
For #4778 .

- change registry center root node from `state` to `registry`
- modify registry center listener manager class name
- refactor unit tests


menghaoranss:
I will resubmit after 4.1.0 released

PR #4823, opened 2020-03-18
Fix #4785 check uniformed with all actual tables' meta data in once when loading them

### LINKED ISSUE #4785: Check uniformed with all actual tables' meta data in once when loading them
ShardingSphere will check uniformed for all actual tables' meta data if user config to enable meta data checking.

The current `checkUniformed()` method can only find the first violation and halt the check by throw an exception. The message of exception is included the first violated actual table only.

It will be better to check all actual tables and record all violations. After check finished, the exception can include all violations information.

The code is in `org.apache.shardingsphere.core.metadata.ShardingMetaDataLoader`, please search `// TODO check all meta data in once` to get the check point.

JasonKing168:
Fixes #4785.

### DIFF
--- a/sharding-core/sharding-core-common/src/main/java/org/apache/shardingsphere/core/metadata/ShardingMetaDataLoader.java
+++ b/sharding-core/sharding-core-common/src/main/java/org/apache/shardingsphere/core/metadata/ShardingMetaDataLoader.java
@@ -113,7 +113,7 @@ public final class ShardingMetaDataLoader {
 
         if (!errorMetaDataList.isEmpty()) {
             StringBuilder exceptionMessageBuilder = new StringBuilder("Cannot get uniformed table structure for logic table `%s`"
-                    + ", it has different meta data of actual tables are as follows:");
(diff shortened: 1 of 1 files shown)


sharding-core/sharding-core-common/src/main/java/org/apache/shardingsphere/core/metadata/ShardingMetaDataLoader.java
```diff
@@ -30,12 +30,8 @@
 
 import javax.sql.DataSource;
 import java.sql.SQLException;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
kimmking: star import is not allowed.
you can use `mvn install` to check all code style.

JasonKing168: Ok, got it, thanks. I fix this problem.

PR #4840, opened 2020-03-19
Parameter marker#4776

### LINKED ISSUE #4776: Add start-index and stop-index assertion for ParameterMarkerExpression
This issue is to count the `start-index` and `stop-index` to assert `ParameterMarkerExpression `

1. Open class `ExpressionAssert` and clear the comment of `SQLSegmentAssert.assertIs(assertContext, actual, expected)` in `assertParameterMarkerExpression`.

3. Calculate the correct `start-index` and `stop-index` of `parameter-marker-expression` segment (i.e., `?`) of this `SQL`.
4. Add the calculated `start-index` and `stop-index` labels of `parameter-marker-expression` segment to `parsing result` of this `SQL Case ID`, like
```
            <parameter-marker-expression value="2" start-index="10" stop-index="11" />
```
5. Run `SQLParserParameterizedTest` again to check whether this `SQL Case ID` does not stay in the list of `SQL` with  `bad assertion result`.

geomonlin:
Fixes #4776 .


tristaZero:
@geomonlin Hi Merged, clear and hard-work. Thanks

PR #4848, opened 2020-03-19
Add parsing result for test case

tristaZero:


tristaZero:
Ref #4696 .

tristaZero:
1. Open class `SQLParserParameterizedTest` and delete the following lines,
```
        // TODO Correct for new parser, please remove them after using new parser
        sqlCases.add("insert_on_duplicate_key_update_with_base64_aes_encrypt");
        sqlCases.add("insert_with_one_auto_increment_column");
        sqlCases.add("insert_on_duplicate_key_update_with_complicated_expression");
        sqlCases.add("insert_without_columns_and_with_generate_key_column");
        sqlCases.add("insert_without_columns_and_without_generate_key_column");
        sqlCases.add("insert_without_columns_with_all_placeholders");
```

3. Calculate the correct `start-index` and `stop-index` of `parameter-marker-expression` segment (i.e., `?`) of this `SQL`.
4. Add the calculated `start-index` and `stop-index` labels of `parameter-marker-expression` segment to `parsing result` of this `SQL Case ID`, like
```
            <parameter-marker-expression value="2" start-index="10" stop-index="11" />
```
5. Run `SQLParserParameterizedTest` again to check whether this `SQL Case ID` does not stay in the list of `SQL` with  `bad assertion result`.

PR #4877, opened 2020-03-21
Logic table name and logic column name toLowerCase() for Encrypt.

### LINKED ISSUE #4770: Case insensitive for DESC table for encrypt
DESC table is case sensitive now, We should change it to case insensitive.

dongzl:
Fixes #4770.

- Logic table name to lower case, for user sql's table name is inconsistent with real table name.
- Logic column name to lower case, for user sql's column name is inconsistent with real column name.
- add unit test.

### DIFF
--- a/encrypt-core/encrypt-core-common/src/main/java/org/apache/shardingsphere/encrypt/rule/EncryptRule.java
+++ b/encrypt-core/encrypt-core-common/src/main/java/org/apache/shardingsphere/encrypt/rule/EncryptRule.java
@@ -97,18 +97,18 @@ public final class EncryptRule implements BaseRule {
     
     private void initTables(final Map<String, EncryptTableRuleConfiguration> tables) {
         for (Entry<String, EncryptTableRuleConfiguration> entry : tables.entrySet()) {
-            this.tables.put(entry.getKey(), new EncryptTable(entry.getValue()));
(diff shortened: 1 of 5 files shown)


encrypt-core/encrypt-core-common/src/main/java/org/apache/shardingsphere/encrypt/rule/EncryptRule.java
```diff
@@ -210,7 +210,7 @@ public boolean isCipherColumn(final String tableName, final String columnName) {
      * @return logic and cipher columns
      */
     public Map<String, String> getLogicAndCipherColumns(final String logicTable) {
-        return tables.containsKey(logicTable) ? tables.get(logicTable).getLogicAndCipherColumns() : Collections.emptyMap();
+        return tables.containsKey(logicTable.toLowerCase()) ? tables.get(logicTable.toLowerCase()).getLogicAndCipherColumns() : Collections.emptyMap();
```
kimmking: return Optional.ofNullable(tables.get(logicTable)).map(EncryptTable::getLogicAndPlainColumns).orElse(Collections.emptyMap());

should be better.

dongzl: OK, I fix it.

dongzl:
This pr is not suitable, lots of code is not well considered.

PR #4900, opened 2020-03-23
Load ShardingMetaDataLoader metadata for different data sources concu…

### LINKED ISSUE #4786: Use multiple threads to load meta data for different data sources
The current meta data loading process are using single thread, it is better to load them concurrency with different data sources.

The code is in class `org.apache.shardingsphere.core.metadata.ShardingMetaDataLoader`, please search `// TODO use multiple threads to load meta data for different data sources` to find check point.

ranjeet-floyd:
Fixes #4786.

- Load ShardingMetaDataLoader metadata for different data sources concurrently


ranjeet-floyd:
@terrymanu Please review.

PR #4964, opened 2020-03-26
Fixes insensitive for DESC table for encrypt.

### LINKED ISSUE #4770: Case insensitive for DESC table for encrypt

dongzl:
Fixes #4770 .

- Get column encrypt field from SchemaMetaData instead of EncryptRule.

### DIFF
--- a/encrypt-core/encrypt-core-merge/src/main/java/org/apache/shardingsphere/encrypt/merge/dal/impl/EncryptColumnsMergedResult.java
+++ b/encrypt-core/encrypt-core-merge/src/main/java/org/apache/shardingsphere/encrypt/merge/dal/impl/EncryptColumnsMergedResult.java
@@ -31,7 +31,7 @@ import java.sql.SQLFeatureNotSupportedException;
 import java.util.Calendar;
 import java.util.Collection;
 import java.util.LinkedList;
-import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Optional;
 import java.util.stream.Collectors;
 
@@ -82,11 +82,10 @@ public abstract class EncryptColumnsMergedResult implements MergedResult {
     
     private Collection<EncryptColumnMetaData> getTableEncryptColumnMetaDatas() {
         Collection<EncryptColumnMetaData> result = new LinkedList<>();
-        for (Map.Entry<String, ColumnMetaData> entry : schemaMetaData.get(tableName).getColumns().entrySet()) {
-            if (!(entry.getValue() instanceof EncryptColumnMetaData)) {
-                continue;
+        for (Entry<String, ColumnMetaData> entry : schemaMetaData.get(tableName).getColumns().entrySet()) {
+            if (entry.getValue() instanceof EncryptColumnMetaData) {
+                result.add((EncryptColumnMetaData) entry.getValue());
             }
-            result.add((EncryptColumnMetaData) entry.getValue());
         }
         return result;
     }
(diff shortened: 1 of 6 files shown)


encrypt-core/encrypt-core-merge/src/main/java/org/apache/shardingsphere/encrypt/merge/dal/impl/EncryptColumnsMergedResult.java
```diff
@@ -66,19 +70,50 @@ public final boolean next() throws SQLException {
...
+    }
+    
+    private Collection<String> getPlainColumns() {
+        return getTableEncryptColumnMetaDatas().stream().map(EncryptColumnMetaData::getPlainColumnName)
+                .collect(Collectors.toList());
+    }
+    
+    private Collection<EncryptColumnMetaData> getTableEncryptColumnMetaDatas() {
+        Collection<EncryptColumnMetaData> result = new LinkedList<>();
+        for (Map.Entry<String, ColumnMetaData> entry : schemaMetaData.get(tableName).getColumns().entrySet()) {
```
terrymanu: Please do not use `Map.Entry`, just use `Entry`

encrypt-core/encrypt-core-merge/src/main/java/org/apache/shardingsphere/encrypt/merge/dal/impl/EncryptColumnsMergedResult.java
```diff
@@ -66,19 +70,50 @@ public final boolean next() throws SQLException {
...
+            if (!(entry.getValue() instanceof EncryptColumnMetaData)) {
+                continue;
+            }
+            result.add((EncryptColumnMetaData) entry.getValue());
```
terrymanu: It is better to change to: 

```java
if (entry.getValue() instanceof EncryptColumnMetaData) {
    result.add((EncryptColumnMetaData) entry.getValue());
}
```

encrypt-core/encrypt-core-merge/src/main/java/org/apache/shardingsphere/encrypt/merge/dal/impl/EncryptColumnsMergedResult.java
```diff
@@ -66,19 +70,50 @@ public final boolean next() throws SQLException {
...
-            }
-            return columnName;
+            Optional<String> logicColumn = getLogicColumnOfCipher(columnName);
+            return logicColumn.isPresent() ? logicColumn.get() : columnName;
         }
         return getOriginalValue(columnIndex, type);
     }
     
+    private Optional<String> getLogicColumnOfCipher(final String cipherColumn) {
+        for (Map.Entry<String, ColumnMetaData> entry : schemaMetaData.get(tableName).getColumns().entrySet()) {
```
terrymanu: Please do not use `Map.Entry`, just use `Entry`

encrypt-core/encrypt-core-merge/src/main/java/org/apache/shardingsphere/encrypt/merge/dal/impl/EncryptColumnsMergedResult.java
dongzl: OK, I fix it. Thanks.

encrypt-core/encrypt-core-merge/src/main/java/org/apache/shardingsphere/encrypt/merge/dal/impl/EncryptColumnsMergedResult.java
dongzl: OK, I fix it. Thanks.

PR #4996, opened 2020-03-29
add sharding.transaction.seata.at.enable config to compatible with se…

cherrylzhao:
when `sharding.transaction.seata.at.enable = false`,  `SeataATShardingTransactionManager` will be disabled, in this way original seata global transaction across services will compatible with shardingsphere since `TransactionalSQLExecutionHook` can transport global txId in multi-threads.

user should pay attention that `@ShardingTransactionType(TransactionType.BASE)` should not been introduced in original seata global transaction across services

### DIFF
--- a/sharding-transaction/sharding-transaction-base/sharding-transaction-base-seata-at/src/main/java/org/apache/shardingsphere/transaction/base/seata/at/SeataATShardingTransactionManager.java
+++ b/sharding-transaction/sharding-transaction-base/sharding-transaction-base-seata-at/src/main/java/org/apache/shardingsphere/transaction/base/seata/at/SeataATShardingTransactionManager.java
@@ -47,14 +47,33 @@ public final class SeataATShardingTransactionManager implements ShardingTransact
     
     private final Map<String, DataSource> dataSourceMap = new HashMap<>();
     
-    private final FileConfiguration configuration = new FileConfiguration("seata.conf");
+    private final String applicationId;
+    
+    private final String transactionServiceGroup;
+    
+    private final boolean enableSeataAT;
+    
+    public SeataATShardingTransactionManager() {
+        FileConfiguration configuration = new FileConfiguration("seata.conf");
+        enableSeataAT = configuration.getBoolean("sharding.transaction.seata.at.enable", true);
+        applicationId = configuration.getConfig("client.application.id");
+        transactionServiceGroup = configuration.getConfig("client.transaction.service.group", "default");
+    }
     
     @Override
     public void init(final DatabaseType databaseType, final Collection<ResourceDataSource> resourceDataSources) {
-        initSeataRPCClient();
-        for (ResourceDataSource each : resourceDataSources) {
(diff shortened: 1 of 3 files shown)


sharding-transaction/sharding-transaction-base/sharding-transaction-base-seata-at/src/main/java/org/apache/shardingsphere/transaction/base/seata/at/SeataATShardingTransactionManager.java
```diff
@@ -47,34 +47,56 @@
...
-    private final FileConfiguration configuration = new FileConfiguration("seata.conf");
+    private final String applicationId;
+    
+    private final String transactionServiceGroup;
+    
+    private final boolean enableSeataAT;
+    
+    public SeataATShardingTransactionManager() {
+        FileConfiguration configuration = new FileConfiguration("seata.conf");
+        enableSeataAT = configuration.getBoolean("sharding.transaction.seata.at.enable", true);
```
kimmking: Constant for literal？

sharding-transaction/sharding-transaction-base/sharding-transaction-base-seata-at/src/main/java/org/apache/shardingsphere/transaction/base/seata/at/SeataATShardingTransactionManager.java
```diff
@@ -47,34 +47,56 @@
...
+        applicationId = configuration.getConfig("client.application.id");
+        transactionServiceGroup = configuration.getConfig("client.transaction.service.group", "default");
+    }
     
     @Override
     public void init(final DatabaseType databaseType, final Collection<ResourceDataSource> resourceDataSources) {
-        initSeataRPCClient();
-        for (ResourceDataSource each : resourceDataSources) {
-            dataSourceMap.put(each.getOriginalName(), new DataSourceProxy(each.getDataSource()));
+        if (enableSeataAT) {
```
kimmking: If turn off this flag, seata client is not initialized and seat-at is disabled?

sharding-transaction/sharding-transaction-base/sharding-transaction-base-seata-at/src/main/java/org/apache/shardingsphere/transaction/base/seata/at/SeataATShardingTransactionManager.java
cherrylzhao: I'll fix it in next PR

sharding-transaction/sharding-transaction-base/sharding-transaction-base-seata-at/src/main/java/org/apache/shardingsphere/transaction/base/seata/at/SeataATShardingTransactionManager.java
cherrylzhao: yes it is，meantime physical datasource will not be wrapped to seata DataSourceProxy

Issue #4885, opened 2020-03-22
Parse `Join table` for supporting `Join table` query SQL


tristaZero:
Content: Parsing `Join table` is first and vital for supporting `Join table` query SQL.
Difficulty Rating: ★★★☆☆

### How to handle this issue?
- [ ] Create `JoinedTableSegment`, `JoinSpecificationSegment`,  and `TableReferenceSegment` like
```
class JoinSpecificationSegment {
PredicateSegment
Collection<columnSegment>
}
class JoinedTableSegment {
tableSegment
joinSpecificationSegment
}
class TableReferenceSegment {
tableSegment
Collection<JoinedTableSegment>
}
```
- [ ] Add `Collection< TableReferenceSegment>` to `SelectStatement`
- [ ] Parse `JoinedTableSegment`, `joinSpecificationSegment` and `TableReferenceSegment ` by visitXXX() 
- [ ] Modify `visitTableReference()`
- [ ] In order to make sure `getAllTables()` in `SelectStatementContext` return correct result, it is needed to move `getTableSegments(result.getValue(), each)` in `XXXDMLVisitor` to `SelectStatementContext`
- [ ] Create `ExpectedJoinedTableSegment `, `ExpectedJoinSpecificationSegment `, `Expected TableReferenceSegment ` and `TableReferenceSegments` in `./incubator-shardingsphere/shardingsphere-sql-parser/shardingsphere-sql-parser-test/src/test/java/org/apache/shardingsphere/sql/parser/integrate/jaxb/domain/segment/impl` for the preparation of Parse integrated tests.
- [ ] Add `TableReferenceSegments ` to `SelectStatementTestCase`
- [ ] Modify `assertTable()` in `SelectStatementAssert`
- [ ] Add `assertJoinTable()` in `SelectStatementAssert `
- [ ] Modify the parsing test results of all the `Select` SQLs in `./incubator-shardingsphere/shardingsphere-sql-parser/shardingsphere-sql-parser-test/src/test/resources/sql/dml/`
- [ ] Modify the parsing test result of all the `Select join` SQLs ./incubator-shardingsphere/shardingsphere-sql-parser/shardingsphere-sql-parser-test/src/test/resources/sql/dml/`
- [ ] Run `SQLParserParameterizedTest` well
- [ ] Run `ShardingSQLRewriterParameterizedTest` well

tristaZero:
[...]

PR #5056, opened 2020-04-03
replace getTables() with getSimpleTableSegments()

jingshanglu:
ref #4885 .

- replace getTables() with getSimpleTableSegments()


shardingsphere-sql-parser/shardingsphere-sql-parser-test/src/test/java/org/apache/shardingsphere/sql/parser/integrate/jaxb/domain/segment/impl/table/ExpectedTableFactor.java
```diff
@@ -39,6 +41,9 @@
     @XmlElement
     private ExpectedSubquery subquery;
     
+    @XmlElement
+    private List<ExpectedColumn> columns;
```
tristaZero, line 45: You had better add at least one SQL case for `columns`.

jingshanglu: Columns have been deleted, and they will be added later with the test case

tristaZero:
@jingshanglu I found you are familiar with PR rules gradually. It is preferred to remove `columns` and fix it later. I like it,  clever boy, aha.

PR #5060, opened 2020-04-03
fix connection.getSchema error

cherrylzhao:
add CatalogSchemaPair to fix connection.getSchema error


cherrylzhao:
it's better to create another connection adapter to override the getCatalog and getSchema method

PR #5085, opened 2020-04-07
sharding-jdbc and sharding-proxy use same metaData refresh strategy.

yu199195:
i find sharding-jdbc and sharding-proxy have same code for metaDataRefresh after execute ddl.

so i remove MetaDataRefreshStrategy from Sharding-jdbc-core to ShardingSphere-common.

then sharding-jdbc ,sharding-proxy can use same code.


shardingsphere-underlying/shardingsphere-common/src/test/java/org/apache/shardingsphere/underlying/common/metadata/refresh/AlterTableStatementMetaDataRefreshStrategyTest.java
```diff
@@ -0,0 +1,52 @@
...
+import org.apache.shardingsphere.underlying.common.metadata.refresh.impl.AlterTableStatementMetaDataRefreshStrategy;
+import org.junit.Test;
+
+import java.util.Collections;
+import java.util.Optional;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assert.assertThat;
+
+public class AlterTableStatementMetaDataRefreshStrategyTest extends AbstractMetaDataRefreshStrategyTest {
```
terrymanu, line 38: Please add final if the class is not for extend

yu199195: ok，i fix it

Issue #5065, opened 2020-04-03
Support remove schema dynamically


menghaoranss:
For the second module `Remove schemas in sharding-ui` of issue #5043 , there are 3 steps: 

- [x] 1. Add `delete(key)` interface in `CenterRepository`  and implement it with zookeeper&etcd&nacos&apollo(#5101)
- [x] 2. Add unit tests(#5101)
- [x] 3. Add a button to remove schemas in sharding-ui(#5112)

kimmking:
A tip: need check only one schema left, can't allow to remove it.

menghaoranss:
> A tip: need check only one schema left, can't allow to remove it.

Got it, great advice!

ssxlulu:
Some questions, remove the schema will also delete the actual schemas and the tables in the backend database?

menghaoranss:
> Some questions, remove the schema will also delete the actual schemas and the tables in the backend database?

Just remove schema configuration in config center.

PR #5101, opened 2020-04-08
Support remove schema dynamically#step 1&2

menghaoranss:
For #5065  .

- Add delete(key) interface in CenterRepository and implement it with zookeeper&etcd&nacos&apollo
- Add unit tests


sharding-orchestration/sharding-orchestration-center/sharding-orchestration-center-apollo/src/main/java/org/apache/shardingsphere/orchestration/center/instance/ApolloCenterRepository.java
```diff
@@ -103,6 +103,11 @@ public void watch(final String key, final DataChangedEventListener dataChangedEv
         configWrapper.addChangeListener(listener, Collections.singleton(apolloKey), Collections.singleton(apolloKey));
     }
     
+    @Override
+    public void delete(final String key) {
+        openApiWrapper.remove(ConfigKeyUtils.path2Key(key));
```
terrymanu, line 108: Please do not use abbreviation, `pathToKey` is better

PR #5102, opened 2020-04-08
if more rule decorate metaData bug fix.

yu199195:


shardingsphere-underlying/shardingsphere-common/src/main/java/org/apache/shardingsphere/underlying/common/metadata/schema/RuleSchemaMetaDataLoader.java
```diff
@@ -146,8 +146,10 @@ private SchemaMetaData decorate(final SchemaMetaData schemaMetaData) {
         Map<String, TableMetaData> result = new HashMap<>(schemaMetaData.getAllTableNames().size(), 1);
         Map<BaseRule, RuleMetaDataDecorator> decorators = OrderedSPIRegistry.getRegisteredServices(rules, RuleMetaDataDecorator.class);
         for (String each : schemaMetaData.getAllTableNames()) {
+            TableMetaData decorate = null;
```
terrymanu, line 149: Please do not use verb as a variable name

terrymanu, line 149: Use result to get meta data is better than use null

yu199195: fix it

PR #5104, opened 2020-04-08
Fix yaml meta data marshal

kimmking:
- Fix yaml meta data marshal


sharding-orchestration/sharding-orchestration-core/sharding-orchestration-core-metadatacenter/src/main/java/org/apache/shardingsphere/orchestration/core/metadatacenter/yaml/YamlTableMetaData.java
```diff
@@ -0,0 +1,37 @@
...
+import org.apache.shardingsphere.underlying.common.yaml.config.YamlConfiguration;
+
+import java.util.Map;
+
+/**
+ * Table Metadata configuration for yaml.
+ */
+@Getter
+@Setter
+public class YamlTableMetaData implements YamlConfiguration {
```
terrymanu, line 31: Please add final if class do not for extend

PR #5112, opened 2020-04-09
Support remove schema dynamically#step 3

menghaoranss:
For #5065 .

- Add a button to remove schemas in sharding-ui


shardingsphere-ui/shardingsphere-ui-backend/src/main/java/org/apache/shardingsphere/ui/servcie/ShardingSchemaService.java
```diff
@@ -71,4 +71,11 @@
      * @param dataSourceConfiguration data source configuration
      */
     void addSchemaConfiguration(String schemaName, String ruleConfiguration, String dataSourceConfiguration);
+    
+    /**
+     * Delete schema configuration.
+     *
+     * @param schemaName schema name
+     */
+    void deleteSchemaConfiguration(String schemaName);
```
kimmking, line 80: deleteSchema is ok, 
AddSchemaConfiguration means add schema from configuration parameter.

PR #5138, opened 2020-04-10
add refresh metadata event after executing ddlstatement.

yu199195:
Fixes https://github.com/apache/incubator-shardingsphere/issues/5128


sharding-orchestration/sharding-orchestration-core/sharding-orchestration-core-facade/src/main/java/org/apache/shardingsphere/orchestration/core/facade/listener/ShardingOrchestrationListenerManager.java
```diff
@@ -33,11 +34,14 @@
     
     private final StateChangedListenerManager stateChangedListenerManager;
     
+    private final MetaDataListenerManager metaDataListenerManager;
+    
     public ShardingOrchestrationListenerManager(final String registryCenterRepositoryName, final RegistryCenterRepository registryCenterRepository,
```
kimmking, line 39: add metadata center repository name here.

PR #5139, opened 2020-04-10
if config same centerRepository,will repetition init.

yu199195:


kimmking:
This pr is unavailable now.
When the second newInstance for this class, the object and fields are both null/empty.
Only breaking init methods lead to the repository is not available.

PR #5165, opened 2020-04-13
metadata refresh exclude myself.

yu199195:


kimmking:
This pr is invalid, cause of the new event does not take effect with original event.

Issue #5017, opened 2020-03-31
[DISCUSS] Brand new sharding configuration API of Release 5.x


tristaZero:
Hi community,

Our 4.1.0 release is on the way. Moreover, I received plenty of feedback and complaints that our configuration is complicated and tedious. When trying it by myself, I convinced it is time to make some changes.
I propose to simplify the configuration API for sharding rule since release 5.x.
#### Pros
1. Through much Straightforward sharding configuration, users do not need to understand too many sharding strategies[1] for their scenarios since sharding strategies decreases from 5 to 2. The next comment will show you the initial design for new API.

2. ShardingSphere will provide many default sharding strategies for users, which means 80% of users could adopt default sharding strategies rather than write by themselves. No code, no compile, no redundant setting. 

#### Cons
1. New configuration API does not have a great compatibility across 4.x and 5.x, i.e., it is a brand new API since release 5.x (I guess it will become our argument).

Therefore, I sincerely welcome your opinion and comment to develop this proposal better and better.

[1] https://shardingsphere.apache.org/document/current/en/manual/sharding-jdbc/configuration/config-yaml/

Cheers,
Trista

tristaZero:
Looking forward to your suggestions and comments for brand new sharding configuration API. :)
Here is the initial design for API.

**The old one**
```
shardingRule:
  tables: 
    t_order:  
      databaseStrategy: 
        standard:  
          shardingColumn:  
            preciseAlgorithmClassName: 
            rangeAlgorithmClassName: 
          complex:  
            shardingColumns:  
            algorithmClassName: 
          inline:  
            shardingColumn:  
            algorithmInlineExpression:  
          hint:  
            algorithmClassName: 
           none:  
      tableStrategy:
```
**The new one**
```
shardingRule:
  tables: 
    t_order: 
      databaseStrategy: 
        standard: 
          shardingColumn: order_id
          shardingAlgorithm:
            type: Inline
            props:
              ShardingDBVolume: 2
              ShardingTBVolume: 4
        complex: 
          shardingColumns: year, month
          shardingAlgorithm:
            type: complex
      tableStrategy:
```

tristaZero:
### TODO
- [ ] Optimize metadata module to route none-sharding tables correctly
- [ ] New API for standard sharding strategy
- [ ] New API for complex sharding strategy
- [ ] Implement Inline algorithm for standard sharding strategy
- [ ] Implement range algorithm for standard sharding strategy
[...]

PR #5176, opened 2020-04-14
Add standard algorithm into configuration

tristaZero:
Ref #5017


tristaZero:
This PR needs merging...

Issue #5283, opened 2020-04-23
[Call For Idea] `Any other` Standard Sharding Algorithm


tristaZero:
Hi guys, welcome! 😊 

### Background
ShardingSphere 5.x will present itself with new valuable features and simple configuration API  to our community. 

**We will provide many of build-in sharding algorithms to users**, which means uses can choose a proper one from sharding-algorithm selections before implementing the interface of the sharding algorithm themselves.

**Therefore we call for your idea for sharding algorithm selections!** 🎉 

### Why
ShardingSphere has an active community, so I believe you have your sharding algorithm to meet your scenarios, right? If so, why do not you try to contribute them to the Apache community? 

Only a few of the effort, you can make some contribution to the community, **which will become a part of sharding algorithm selections and is likely to be adopted by people all over the world.** ✨

What's more, if you have another sharding scenario one day, then this selection will be the help for you in turn!

### What is `any other sharding algorithm`
Any sharding algorithm which is outside the range of `Range`, `Hash`, `Date time` sharding algorithms is welcome! 👏 **(Since they all have their own issues.)**

You can light up your imagination to give a better design! 🎯 

Welcome your comments and look forward to your PR! 🤚

tristaZero:
### How
1. Fork this project and clone your fork one to local.
2. Install this project.
3. Create your sharding algorithm in `./incubator-shardingsphere/sharding-core/sharding-core-common/src/main/java/org/apache/shardingsphere/core/strategy/algorithm/sharding/`
4. Implement `StandardShardingAlgorithm` interface.
5. Add unit test to check what you coded in `./incubator-shardingsphere/sharding-core/sharding-core-common/src/test/java/org/apache/shardingsphere/core/strategy/algorithm/sharding/inline`

### Note
1. This issue is to call for `any other` sharding algorithm. You should look at other issues for your concern.
2. Please comment `Ref #[This ISSUE ID]` and give some explanation when you raize a PR.
3. Please make sure your sharding algorithm does not include your business logic. A sharding algorithm likely to be used universally is the best.
4. `InlineShardingAlgorithm` and `InlineShardingAlgorithmTest` give you good examples of implementation and its unit test.
5. `getType()` of  `ShardingAlgorithm` is expected to return the algorithm ID with capital letters.

PR #5318, opened 2020-04-26
Create MODULO sharding algorithm

tristaZero:
Ref #5283 

Modulo Sharding Algorithm is online.


sharding-core/sharding-core-common/src/main/java/org/apache/shardingsphere/core/strategy/algorithm/sharding/inline/ModuloShardingAlgorithm.java
```diff
@@ -0,0 +1,94 @@
...
+/**
+ * Modulo sharding algorithm.
+ * 
+ * <p>Shard with `y = MOD(x)` algorithm.</p>
+ */
+public final class ModuloShardingAlgorithm implements StandardShardingAlgorithm<Long> {
+    
+    private static final String MODULO_VALUE = "modulo.value";
+    
+    private Properties properties = new Properties();
```
terrymanu, line 38: Please use lombok

terrymanu, line 38: Properties is sync one, we can not use like this way, it will cause all concurrency threads as single synchronized thread.

PR #5327, opened 2020-04-26
create RANGE sharding algorithm

### LINKED ISSUE #5280: [Call For Idea] `Range` Standard Sharding Algorithm
Hi guys, welcome! 😊 

### Background
ShardingSphere 5.x will present itself with new valuable features and simple configuration API  to our community. 

**We will provide many of build-in sharding algorithms to users**, which means uses can choose a proper one from sharding-algorithm selections before implementing the interface of the sharding algorithm themselves.

**Therefore we call for your idea for sharding algorithm selections!** 🎉 

### Why
ShardingSphere has an active community, so I believe you have your sharding algorithm to meet your scenarios, right? If so, why do not you try to contribute them to the Apache community? 

Only a few of the effort, you can make some contribution to the community, **which will become a part of sharding algorithm selections and is likely to be adopted by people all over the world.** ✨

What's more, if you have another sharding scenario one day, then this selection will be the help for you in turn!

### What is `Range sharding algorithm`
**The key point of this algorithm** is that a group of sequential numbers (sharding values) is considered as a sharding unit (partition table).

For instance, if users want to accumulate five rows as a sharding table, then the first table will contain ID IN (1, 2, 3, 4, 5), and the second one will include rows with ID numbers from 6 to 10...

You can light up your imagination to give a better design! 🎯 

Welcome your comments and look forward to your PR! 🤚

strongduanmu:
Ref #5280 .

- create RANGE sharding algorithm.

### DIFF
--- a/sharding-core/sharding-core-common/src/main/java/org/apache/shardingsphere/core/strategy/algorithm/sharding/RangeShardingAlgorithm.java
+++ b/sharding-core/sharding-core-common/src/main/java/org/apache/shardingsphere/core/strategy/algorithm/sharding/RangeShardingAlgorithm.java
@@ -22,6 +22,8 @@ import com.google.common.base.Splitter;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Range;
 import com.google.common.primitives.Longs;
+import lombok.Getter;
+import lombok.Setter;
 import org.apache.commons.collections4.CollectionUtils;
 import org.apache.shardingsphere.api.sharding.standard.PreciseShardingValue;
 import org.apache.shardingsphere.api.sharding.standard.RangeShardingValue;
@@ -53,6 +55,8 @@ public final class RangeShardingAlgorithm implements StandardShardingAlgorithm<L
 
     private static final String RANGE_PARTITION_SPLIT_VALUE = "range.partition.split.value";
 
+    @Getter
+    @Setter
     private Properties properties = new Properties();
 
     @Override
@@ -60,7 +64,7 @@ public final class RangeShardingAlgorithm implements StandardShardingAlgorithm<L
         Preconditions.checkNotNull(properties.get(RANGE_PARTITION_SPLIT_VALUE), "Range sharding algorithm range partition split value cannot be null.");
         Map<Integer, Range<Long>> partitionRangeMap = getPartitionRangeMap();
         for (String each : availableTargetNames) {
-            if (each.endsWith(getPartition(partitionRangeMap, shardingValue.getValue()))) {
+            if (each.endsWith(getPartition(partitionRangeMap, shardingValue.getValue()) + "")) {
                 return each;
             }
         }
@@ -72,9 +76,11 @@ public final class RangeShardingAlgorithm implements StandardShardingAlgorithm<L
         Preconditions.checkNotNull(properties.get(RANGE_PARTITION_SPLIT_VALUE), "Range sharding algorithm range partition split value cannot be null.");
         Map<Integer, Range<Long>> partitionRangeMap = getPartitionRangeMap();
         Collection<String> result = new LinkedHashSet<>(availableTargetNames.size());
-        for (long value = shardingValue.getValueRange().lowerEndpoint(); value <= shardingValue.getValueRange().upperEndpoint(); value++) {
+        int lowerEndpointPartition = getPartition(partitionRangeMap, shardingValue.getValueRange().lowerEndpoint());
+        int upperEndpointPartition = getPartition(partitionRangeMap, shardingValue.getValueRange().upperEndpoint());
+        for (int partition = lowerEndpointPartition; partition <= upperEndpointPartition; partition++) {
             for (String each : availableTargetNames) {
-                if (each.endsWith(getPartition(partitionRangeMap, value))) {
+                if (each.endsWith(partition + "")) {
                     result.add(each);
                 }
             }
@@ -102,27 +108,17 @@ public final class RangeShardingAlgorithm implements StandardShardingAlgorithm<L
         return partitionRangeMap;
     }
 
-    private String getPartition(final Map<Integer, Range<Long>> partitionRangeMap, final Long value) {
+    private Integer getPartition(final Map<Integer, Range<Long>> partitionRangeMap, final Long value) {
         for (Map.Entry<Integer, Range<Long>> entry : partitionRangeMap.entrySet()) {
             if (entry.getValue().contains(value)) {
-                return entry.getKey().toString();
+                return entry.getKey();
             }
         }
-        return partitionRangeMap.keySet().stream().mapToInt(Integer::valueOf).max().toString();
+        throw new UnsupportedOperationException();
     }
 
     @Override
     public String getType() {
         return "RANGE";
-    }
-
-    @Override
-    public Properties getProperties() {
-        return properties;
-    }
-
-    @Override
-    public void setProperties(final Properties properties) {
-        this.properties = properties;
     }
 }
--- a/sharding-core/sharding-core-common/src/main/java/org/apache/shardingsphere/core/strategy/algorithm/sharding/RangeShardingAlgorithm.java
+++ b/sharding-core/sharding-core-common/src/main/java/org/apache/shardingsphere/core/strategy/algorithm/sharding/RangeShardingAlgorithm.java
@@ -1,0 +1,128 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.shardingsphere.core.strategy.algorithm.sharding;
+
+import com.google.common.base.Preconditions;
+import com.google.common.base.Splitter;
+import com.google.common.collect.Maps;
+import com.google.common.collect.Range;
+import com.google.common.primitives.Longs;
+import org.apache.commons.collections4.CollectionUtils;
+import org.apache.shardingsphere.api.sharding.standard.PreciseShardingValue;
+import org.apache.shardingsphere.api.sharding.standard.RangeShardingValue;
+import org.apache.shardingsphere.api.sharding.standard.StandardShardingAlgorithm;
+
+import java.util.Collection;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Properties;
+import java.util.stream.Collectors;
+
+/**
+ * Range sharding algorithm.
+ * <p>
+ * Range sharding algorithm is similar to the rule of partition table.
+ * User can specify the range by setting the `range.partition.split.value` parameter.
+ * The `range.partition.split.value` parameter is an ordered list of numbers, separated by commas.
+ * </p>
+ * <p>
+ * For example: If the `range.partition.split.value` parameter is set to `1,5,10`,
+ * the parameter will split all values into four intervals——(-∞, 1), [1,5), [5,10), [10, +∞),
+ * which corresponding to partition_0, partition_1, partition_2, partition_3.
+ * The sharding values will be divided into different partition by its value.
+ * </p>
+ */
+public final class RangeShardingAlgorithm implements StandardShardingAlgorithm<Long> {
+
+    private static final String RANGE_PARTITION_SPLIT_VALUE = "range.partition.split.value";
+
+    private Properties properties = new Properties();
+
+    @Override
+    public String doSharding(final Collection<String> availableTargetNames, final PreciseShardingValue<Long> shardingValue) {
+        Preconditions.checkNotNull(properties.get(RANGE_PARTITION_SPLIT_VALUE), "Range sharding algorithm range partition split value cannot be null.");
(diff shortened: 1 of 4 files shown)


tristaZero:
@strongduanmu 

Hi, I am glad to read your organized and straightforward code. :)
There is a concern about `range.partition.split.value` which is the key point of this algorithm.

Users need to list all endpoints of each range partitions, i.e, 10,20,30,40...
Besides, it generates complicated code handling. Do you think it is better to prefer the partition volume than listing all endpoints? That is, in this case: instead of `range.partition.split.value=10,20,30,40`, maybe we can consider `partition.volume = 10`. I am not sure, so I'd like to listen to your voice. 😀

tristaZero:
BTW, instead of `Fixes #5280 .` which makes this issue closed after merging, `Ref #5280` is suggested, as we still welcome any other great ideas. :)

sharding-core/sharding-core-common/src/main/java/org/apache/shardingsphere/core/strategy/algorithm/sharding/RangeShardingAlgorithm.java
```diff
@@ -0,0 +1,128 @@
...
+                return each;
+            }
+        }
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public Collection<String> doSharding(final Collection<String> availableTargetNames, final RangeShardingValue<Long> shardingValue) {
+        Preconditions.checkNotNull(properties.get(RANGE_PARTITION_SPLIT_VALUE), "Range sharding algorithm range partition split value cannot be null.");
+        Map<Integer, Range<Long>> partitionRangeMap = getPartitionRangeMap();
```
SanmuWangZJU, line 73: for line 61 and 73: 
`doSharding()` method will be called while route value, it may be not efficient to build partitionRangeMap in `doSharding` method, I think it can be used as a static field and  built once while initing instance.

strongduanmu:
> @strongduanmu
> 
> Hi, I am glad to read your organized and straightforward code. :)
> There is my concern about `range.partition.split.value` which is the key point of this algorithm.
> 
> Users need to list all endpoints of each range partitions, i.e, 10,20,30,40...
> Besides, it generates complicated code handling. Do you think it is better to prefer the partition volume than listing all endpoints? That is, in this case: instead of `range.partition.split.value=10,20,30,40`, maybe we can consider `partition.volume = 10`. I am not sure, so I'd like to listen to your voice. 😀

@tristaZero Partition volume looks more concise, but it can only split all values into same interval partition except the first and last partition, this way may not meet the scene of custom partition range.

Users use partition volume to split all values, and may encounter some scenarios where the amount of partition data is small. At this time, it may be more desirable to be able to merge multiple partition ranges, that is, custom partition ranges.

Is it better to provide two Range algorithms? One meets the requirements of flexible scenarios, and the other is more suitable for general scenarios.

sharding-core/sharding-core-common/src/main/java/org/apache/shardingsphere/core/strategy/algorithm/sharding/RangeShardingAlgorithm.java
strongduanmu: Good suggestion, I will optimize this question.

strongduanmu: > for line 61 and 73:
> `doSharding()` method will be called while route value, it may be not efficient to build partitionRangeMap in `doSharding` method, I think it can be used as a static field and built once while initing instance.

Good suggestion, I will optimize this question.

strongduanmu:
> BTW, instead of `Fixes #5280 .` which makes this issue closed after merging, `Ref #5280` is suggested, as we still welcome any other great ideas. :)

Okay, I will modify it.

sharding-core/sharding-core-common/src/main/java/org/apache/shardingsphere/core/strategy/algorithm/sharding/RangeShardingAlgorithm.java
tristaZero: @SanmuWangZJU Thanks. :)

sharding-core/sharding-core-common/src/main/java/org/apache/shardingsphere/core/strategy/algorithm/sharding/RangeShardingAlgorithm.java
```diff
@@ -0,0 +1,128 @@
...
+        }
+        return partitionRangeMap.keySet().stream().mapToInt(Integer::valueOf).max().toString();
+    }
+
+    @Override
+    public String getType() {
+        return "RANGE";
+    }
+
+    @Override
```
tristaZero, line 119: @Setter and @Getter are recommended.

sharding-core/sharding-core-common/src/test/java/org/apache/shardingsphere/core/strategy/algorithm/sharding/range/RangeShardingAlgorithmTest.java
```diff
@@ -0,0 +1,66 @@
...
+        StandardShardingStrategyConfiguration shardingStrategyConfig = new StandardShardingStrategyConfiguration("order_id", shardingAlgorithm);
+        shardingStrategy = new StandardShardingStrategy(shardingStrategyConfig);
+    }
+
+    @Test
+    public void assertPreciseDoSharding() {
+        List<String> availableTargetNames = Lists.newArrayList("t_order_0", "t_order_1", "t_order_2", "t_order_3");
+        List<RouteValue> shardingValues = Lists.newArrayList(new ListRouteValue<>("order_id", "t_order", Lists.newArrayList(0L, 1L, 2L, 4L, 7L)));
+        Collection<String> actual = shardingStrategy.doSharding(availableTargetNames, shardingValues, new ConfigurationProperties(new Properties()));
+        assertThat(actual.size(), is(3));
```
tristaZero, line 56: It is suggested to assert some of the specific targets, not only `size()`.

sharding-core/sharding-core-common/src/main/java/org/apache/shardingsphere/core/strategy/algorithm/sharding/RangeShardingAlgorithm.java
```diff
@@ -0,0 +1,128 @@
...
+        }
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public Collection<String> doSharding(final Collection<String> availableTargetNames, final RangeShardingValue<Long> shardingValue) {
+        Preconditions.checkNotNull(properties.get(RANGE_PARTITION_SPLIT_VALUE), "Range sharding algorithm range partition split value cannot be null.");
+        Map<Integer, Range<Long>> partitionRangeMap = getPartitionRangeMap();
+        Collection<String> result = new LinkedHashSet<>(availableTargetNames.size());
+        for (long value = shardingValue.getValueRange().lowerEndpoint(); value <= shardingValue.getValueRange().upperEndpoint(); value++) {
```
tristaZero, line 75: It is likely that`value++` is time-consuming if there is a vast range.
Here are some of my ideas.
For instance,the shardings are as follows,
1,2 | 3,4,5 | 6,7,8,9 | 10,11
So the `range.partition.split.value` is `2,5,9,11`
Imaging the query condition is `2 =< x <= 9`, then we just need to calcute 2==2 and 9<11, therefore the results is `0,1,2`.
That is, we can get the critical values from `range.partition.split.value`, so we only need to compare the bottom point and the top point with those critical values.

tristaZero:
@strongduanmu 
I got your point. This RANGE algorithm is flexible for sharding, which is supposed to be the right choice for users, so please review my comment to perfect this PR. :) 
Besides, we can consider another type of RANGE algorithm. 😀

sharding-core/sharding-core-common/src/main/java/org/apache/shardingsphere/core/strategy/algorithm/sharding/RangeShardingAlgorithm.java
strongduanmu: Thank you, i will modify it.

sharding-core/sharding-core-common/src/test/java/org/apache/shardingsphere/core/strategy/algorithm/sharding/range/RangeShardingAlgorithmTest.java
strongduanmu: I got it, thanks!

sharding-core/sharding-core-common/src/main/java/org/apache/shardingsphere/core/strategy/algorithm/sharding/RangeShardingAlgorithm.java
strongduanmu: > It is likely that`value++` is time-consuming if there is a vast range.
> Here are some of my ideas.
> For instance,the shardings are as follows,
> 1,2 | 3,4,5 | 6,7,8,9 | 10,11
> So the `range.partition.split.value` is `2,5,9,11`
> Imaging the query condition is `2 =< x <= 9`, then we just need to calcute 2==2 and 9<11, therefore the results is `0,1,2`.
> That is, we can get the critical values from `range.partition.split.value`, so we only need to compare the bottom point and the top point with those critical values.

This suggestion is great, and I will optimize it.

sharding-core/sharding-core-common/src/main/java/org/apache/shardingsphere/core/strategy/algorithm/sharding/RangeShardingAlgorithm.java
strongduanmu: > for line 61 and 73:
> `doSharding()` method will be called while route value, it may be not efficient to build partitionRangeMap in `doSharding` method, I think it can be used as a static field and built once while initing instance.

@SanmuWangZJU @tristaZero 
Hi, I have tried to use static field to define `partitionRangeMap`, but I found that it is not suitable. Users may instantiate multiple different range partition rules in the same application instance to correspond to different requirements. Static initialization means same range partition rule. Perhaps the instance field is more suitable.

tristaZero: @strongduanmu Got it. Maybe we need a `init()` interface function to initialize some local members so as to avoid calculating some local values like `partitionRangeMap`

tristaZero:
@strongduanmu Got it. Maybe we need a init() interface function to initialize some local members so as to avoid calculating some local values like partitionRangeMap

sharding-core/sharding-core-common/src/main/java/org/apache/shardingsphere/core/strategy/algorithm/sharding/RangeShardingAlgorithm.java
```diff
@@ -0,0 +1,124 @@
...
+ * <p>
+ * For example: If the `range.partition.split.value` parameter is set to `1,5,10`,
+ * the parameter will split all values into four intervals——(-∞, 1), [1,5), [5,10), [10, +∞),
+ * which corresponding to partition_0, partition_1, partition_2, partition_3.
+ * The sharding values will be divided into different partition by its value.
+ * </p>
+ */
+public final class RangeShardingAlgorithm implements StandardShardingAlgorithm<Long> {
+
+    private static final String RANGE_PARTITION_SPLIT_VALUE = "range.partition.split.value";
```
tristaZero, line 56: Do you think there is any better names for `range.partition.split.value` to make users easy get its meaning. Like `range.partition.upper.points` or `range.partition.boundary.points`?

tristaZero:
Hi @strongduanmu It seems this PR is excellent, so I can not wait to merge it, thanks for your effort! However, there are still some questions that remained for us,
1. Are there any better names for range.partition.split.value to make users easily get its meaning. Like range.partition.upper.points or range.partition.boundary.points?
2.

tristaZero:
@strongduanmu Hi
@terrymanu gave another option, i.e, `partition.range`. What do you think?

tristaZero:
After more consideration, the second one `init()` is not necessary, maybe.
At first, I considered to initialize `partitionRangeMap` in init(). However, that means `partitionRangeMap` will not be a final member. If so, at the multi-thread condition, it is not safe, which is more important than the performance of `doSharding()`.
What's your idea about this one? Wait for your kind reply. 

Trista

strongduanmu:
> @strongduanmu Hi
> @terrymanu gave another option, i.e, `partition.range`. What do you think?

The parameter `partition.range` is better and more easier for users to understand!👍

tristaZero:
How about `range.partitions`?
@strongduanmu

strongduanmu:
> How about `range.partitions`?
> @strongduanmu



> How about `range.partitions`?
> @strongduanmu

Hi, I found that the `partition.seconds` parameter is defined in the DatetimeShardingAlgorithm class,
Can we define it as `partition.ranges` in the RangeShardingAlgorithm class? 
Ranges can remind users that this algorithm supports configuring multiple ranges for sharding.

strongduanmu:
> After more consideration, the second one `init()` is not necessary, maybe.
> At first, I considered to initialize `partitionRangeMap` in init(). However, that means `partitionRangeMap` will not be a final member. If so, at the multi-thread condition, it is not safe, which is more important than the performance of `doSharding()`.
> What's your idea about this one? Wait for your kind reply.
> 
> Trista

Hi, @tristaZero , I agree with you that thread safety is more important. We can use another way to improve performance by providing a `useCache` parameter. 
If users configured fewer partitions, the performance should be able to meet the requirements. But if the num of partition ranges is very large, 
such as 100 partitions, users can set `useCache` parameter to `true`, use the cache to improve performance, just like SQLParseResultCache.

tristaZero:
@strongduanmu That is exactly what I am thinking, maybe another PR to rename it as `partition.ranges` is expected. :)

strongduanmu:
> @strongduanmu That is exactly what I am thinking, maybe another PR to rename it as `partition.ranges` is expected. :)



> @strongduanmu That is exactly what I am thinking, maybe another PR to rename it as `partition.ranges` is expected. :)

I have modified it, help me review it, thanks!

PR #5333, opened 2020-04-27
Create Datetime sharding algorithm

tristaZero:
Ref #5282

### DIFF
--- a/sharding-core/sharding-core-common/src/main/java/org/apache/shardingsphere/core/strategy/algorithm/sharding/DatetimeShardingAlgorithm.java
+++ b/sharding-core/sharding-core-common/src/main/java/org/apache/shardingsphere/core/strategy/algorithm/sharding/DatetimeShardingAlgorithm.java
@@ -35,14 +35,14 @@ import java.util.Properties;
  * Datetime sharding algorithm.
  * 
  * <p>Shard by `y = floor(x/v)` algorithm, which means y begins from 0.
- * v is `PARTITION_VOLUME`, and the minimum time unit is 1 sec.
+ * v is `PARTITION_SECONDS`, and the minimum time unit is 1 sec.
  * `SINCE_DATETIME` decides the beginning datetime to shard. </p>
  */
 public final class DatetimeShardingAlgorithm implements StandardShardingAlgorithm<Comparable<?>> {
     
-    private static final String PARTITION_VOLUME = "partition.volume";
+    private static final String PARTITION_SECONDS = "partition.seconds";
     
-    private static final String SINCE_DATETIME = "since.datetime";
+    private static final String EPOCH = "epoch";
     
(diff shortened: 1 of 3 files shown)


sharding-core/sharding-core-common/src/main/java/org/apache/shardingsphere/core/strategy/algorithm/sharding/DatetimeShardingAlgorithm.java
```diff
@@ -0,0 +1,128 @@
...
+        for (long i = parseDate(shardingValue.getValueRange().lowerEndpoint()); i <= parseDate(shardingValue.getValueRange().upperEndpoint()); i++) {
+            for (String each : availableTargetNames) {
+                if (each.endsWith(doSharding(i))) {
+                    result.add(each);
+                }
+            }
+            if (result.size() == availableTargetNames.size()) {
+                return result;
+            }
+        }
```
terrymanu, line 79: It is pretty slow if loop for every second, could we consider about increase for one partition range once?

tristaZero: I will consider it in my next PR

sharding-core/sharding-core-common/src/main/java/org/apache/shardingsphere/core/strategy/algorithm/sharding/DatetimeShardingAlgorithm.java
```diff
@@ -0,0 +1,128 @@
...
+ * v is `PARTITION_VOLUME`, and the minimum time unit is 1 sec.
+ * `SINCE_DATETIME` decides the beginning datetime to shard. </p>
+ */
+public final class DatetimeShardingAlgorithm implements StandardShardingAlgorithm<Comparable<?>> {
+    
+    private static final String PARTITION_VOLUME = "partition.volume";
+    
+    private static final String SINCE_DATETIME = "since.datetime";
+    
+    private static final SimpleDateFormat DATE_FORMAT = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss");
```
SanmuWangZJU, line 47: `SimpleDateFormat` is not thread safe, is it possible to call `SimpleDateFormat.parse()` concurrently? if so, perhaps use `LocalDateTime` and `DateTimeFormatter` is better

tristaZero: @ SanmuWangZJU

> `SimpleDateFormat` is not thread safe, is it possible to call `SimpleDateFormat.parse()` concurrently? if so, perhaps use `LocalDateTime` and `DateTimeFormatter` is better

Hi, Thanks for your suggestion which helps raize #5341, welcome your review and comments.

PR #5369, opened 2020-04-28
simplify integrate test 

cherrylzhao:
- add literal sql in xml driver file
- remove sqlCaseId and SQLCaseLoader
- decouple with DatabaseTypeEnvironment, only init available database case

after refactor, integrate-test-case look like this
image


cherrylzhao:
I have create a issue #5377 for that, it is better to summary the disabled and enable cases in the form of summary report.

PR #5363, opened 2020-04-28
a custom datetime shard algorithm used to shard tables by days/weeks/months/QUARTER_OF_YEAR

SanmuWangZJU:
Ref #5282.

Changes proposed in this pull request:
- supply a sharding algorithm used to shard tables by days/weeks/months/QUARTER_OF_YEAR/...
- usage examples
  - shard by QUARTER_OF_YEAR (按季度分表)
  ``` yaml 
      actualDataNodes: ds.t_order_20${10..20}0${1..4}
      tableStrategy:
        standard:
          shardingColumn: created_at
          shardingAlgorithm:
            type: CUSTOM_DATE_TIME
            props:
              datetime.format: yyyy-MM-dd HH:mm:ss
              table.suffix.format: yyyyQQ
              datetime.lower: "2010-01-01 00:00:00.000"
              datetime.upper: "2020-12-31 00:00:00.000"
              datetime.step.unit: Months
              datetime.step.amount: 3
  ```
- config detail can refer to java doc of `org.apache.shardingsphere.core.strategy.algorithm.sharding.CustomDateTimeShardingAlgorithmTest`

PR #5371, opened 2020-04-28
rename properties in RangeShardingAlgorithm to partition.ranges

strongduanmu:
Ref #5280 .

- rename properties in RangeShardingAlgorithm to partition.ranges


tristaZero:
A better variable name for users. 😀 

BTY, do you think `getPartitionRangeMap()` is time-consuming? If so, there is [a good idea](https://github.com/apache/shardingsphere/pull/5363/files#diff-2194e93614984507fecf4f07962a5beaR139) for reference.

strongduanmu:
> A better variable name for users. 
> 
> BTY, do you think `getPartitionRangeMap()` is time-consuming? If so, there is [a good idea](https://github.com/apache/shardingsphere/pull/5363/files#diff-2194e93614984507fecf4f07962a5beaR139) for reference.

Thank you, I will refer to this PR to improve the performance of RangeShardingAlgorithm.😃

PR #5372, opened 2020-04-28
fix for mysql tcl

jingshanglu:
Fixes #ISSUSE_ID.

- fix for mysql tcl
-  ref mysql dml


shardingsphere-sql-parser/shardingsphere-sql-parser-test/src/test/java/org/apache/shardingsphere/sql/parser/integrate/engine/SQLParserParameterizedTest.java
```diff
@@ -127,9 +127,6 @@ private static boolean isPlaceholderWithoutParameter(final Object[] sqlTestParam
     
     @Test
     public void assertSupportedSQL() {
-        if ("select_pagination_with_row_number_for_greater_than_and_equal".equals(sqlCaseId)) {
```
tristaZero, line 130: aha, self-righting? Like!

PR #5436, opened 2020-05-06
fix TableMetaData load error when use oracle.

### LINKED ISSUE #5421: The keyGenerator doesn't work in transaction-2pc-xa-raw-jdbc-example module when a oracle database used
I have reproduced the behavior as @lvye351 says in #5383. The keyGenerator doesn't work in the transaction-2pc-xa-raw-jdbc sample module when a oracle database 12c and ojbdc8.jar used.

The log is as follows,
image

It seems that when table t_order created, its tableMetaData doesn't put into schemaMetaData.
I will try to find out the reason and fix it.

TaoZhiMLND:
Fixes #5421.


shardingsphere-sql-parser/shardingsphere-sql-parser-binder/src/main/java/org/apache/shardingsphere/sql/parser/binder/metadata/index/IndexMetaDataLoader.java
```diff
@@ -47,6 +47,9 @@
         try (ResultSet resultSet = connection.getMetaData().getIndexInfo(connection.getCatalog(), connection.getSchema(), table, false, false)) {
             while (resultSet.next()) {
                 String indexName = resultSet.getString(INDEX_NAME);
+                if (null == indexName) {
+                    continue;
+                }
                 result.add(new IndexMetaData(indexName));
```
terrymanu, line 53: It is better to:

```java
if (null != indexName) {
    result.add(new IndexMetaData(indexName));
}
```

PR #5440, opened 2020-05-06
fix for setVariable of mysql

### LINKED ISSUE #4453: Completion of parse SET statement
We just parse SET statement simplify, it is not enough.

The SET statement is different with each database dialect.

Syntax of MySQL: 

```
SET variable = expr [, variable = expr] ...

variable: {
    user_var_name
  | param_name
  | local_var_name
  | {GLOBAL | @@GLOBAL.} system_var_name
  | {PERSIST | @@PERSIST.} system_var_name
  | {PERSIST_ONLY | @@PERSIST_ONLY.} system_var_name
  | [SESSION | @@SESSION. | @@] system_var_name
}
```

Syntax of PostgreSQL: 

```
SET [ SESSION | LOCAL ] configuration_parameter { TO | = } { value | 'value' | DEFAULT }
SET [ SESSION | LOCAL ] TIME ZONE { timezone | LOCAL | DEFAULT }
```

Even they are different totally, but we still can find the same abstraction of them.

The tasks of SET statement:

- [x] Parse variable name
- [ ] Parse variable value
- [ ] Parse scope
- [ ] Can support set multiple variables
- [ ] Consider about Oracle and SQLServer
- [ ] Consider about use sub-class to split different of SET dialect

jingshanglu:
Fixes #4453.

- fix for setVariable of mysql

### DIFF
--- a/shardingsphere-sql-parser/shardingsphere-sql-parser-dialect/shardingsphere-sql-parser-mysql/src/main/antlr4/imports/mysql/BaseRule.g4
+++ b/shardingsphere-sql-parser/shardingsphere-sql-parser-dialect/shardingsphere-sql-parser-mysql/src/main/antlr4/imports/mysql/BaseRule.g4
@@ -119,7 +119,11 @@ unreservedWord
     ;
 
 variable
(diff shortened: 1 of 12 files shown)


jingshanglu:
@terrymanu @tristaZero Please review.

PR #5605, opened 2020-05-14
Fix orchestration create datasource error when datasource map is empty

yanyzy:


sharding-jdbc/sharding-jdbc-orchestration/src/main/java/org/apache/shardingsphere/shardingjdbc/orchestration/api/yaml/YamlOrchestrationShardingSphereDataSourceFactory.java
```diff
@@ -100,7 +100,7 @@ public static DataSource createDataSource(final Map<String, DataSource> dataSour
     
     private static DataSource createDataSource(final Map<String, DataSource> dataSourceMap, final YamlOrchestrationRootRuleConfigurations configurations, 
                                                final Properties props, final Map<String, YamlCenterRepositoryConfiguration> yamlInstanceConfigurationMap) throws SQLException {
-        if (null == configurations) {
+        if (configurations.getRules().size() == 0 || dataSourceMap.size() == 0) {
```
terrymanu, line 103: please use `isEmpty()`

PR #5634, opened 2020-05-16
Defect more when concurrently registering services

kezhenxu94:
This pull request follows up #5618.

`ConcurrentHashMap` is thread safe, but `get-and-check-and-put` are composite operations and are thus not atomic, this patch replaces the former with `computeIfAbsent`, which is atomic now.

### DIFF
--- a/shardingsphere-spi/src/main/java/org/apache/shardingsphere/sharding/spi/ShardingSphereServiceLoader.java
+++ b/shardingsphere-spi/src/main/java/org/apache/shardingsphere/sharding/spi/ShardingSphereServiceLoader.java
@@ -34,9 +34,9 @@ import java.util.stream.Collectors;
  */
 @NoArgsConstructor(access = AccessLevel.PRIVATE)
 public final class ShardingSphereServiceLoader {
-    
-    private static final Map<Class, Collection<Class<?>>> SERVICE_MAP = new ConcurrentHashMap<>();
-    
+
+    private static final Map<Class<?>, Collection<Class<?>>> SERVICE_MAP = new ConcurrentHashMap<>();
+
     /**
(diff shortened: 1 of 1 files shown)


shardingsphere-spi/src/main/java/org/apache/shardingsphere/sharding/spi/ShardingSphereServiceLoader.java
```diff
@@ -51,17 +51,12 @@
...
-        Collection<Class<?>> serviceClasses = SERVICE_MAP.get(service);
-        if (null == serviceClasses) {
-            serviceClasses = new LinkedHashSet<>();
-        }
+        Collection<Class<?>> serviceClasses = SERVICE_MAP.computeIfAbsent(service, unused -> new LinkedHashSet<>());
         serviceClasses.add(instance.getClass());
-        SERVICE_MAP.put(service, serviceClasses);
     }
-    
+
```
cherrylzhao, line 59: should keep incident in new line

shardingsphere-spi/src/main/java/org/apache/shardingsphere/sharding/spi/ShardingSphereServiceLoader.java
```diff
@@ -73,7 +68,7 @@
     public static <T> Collection<T> newServiceInstances(final Class<T> service) {
         return SERVICE_MAP.containsKey(service) ? SERVICE_MAP.get(service).stream().map(each -> (T) newServiceInstance(each)).collect(Collectors.toList()) : Collections.emptyList();
     }
-    
+
```
cherrylzhao, line 71: should keep incident in new line

Issue #5593, opened 2020-05-14
[Refactoring] New architecture of runtime context for Sharding Proxy and JDBC


tristaZero:
Hi folks,

There is a significant change for release 5.x need to be public to our community. 
To get a consistent and straightforward orchestration function for both of Sharding Proxy and Sharding JDBC, we extracted the joint members from them, which will be passed down to `shardingsphere-underlying` module. Based on that, orchestration engine could function simply and uniformly.

A todo list is appended here.
- [x] Remove cache from Sharding Parser
- [x] Call the `clear()` of Sharding Parser when needed
- [x] Exact joint members from Sharding Proxy and JDBC to group into a new new `SchemaContext`
- [x] Pass a list of `SchemaContext` named `SchemaContexts` to `shardingsphere-underlying`
- [ ] Refactor the behaviors of `orchestration`

Here is the view of the new structure and relationships of related classes.
![image](https://user-images.githubusercontent.com/27757146/81891558-4c60a500-95db-11ea-8654-1c6a2a191cff.png)

kimmking:
Based on this new architecture, shardingsphere will go further and evolve to a new epoch.

PR #5668, opened 2020-05-18
create shardingsphere-kernal module

tristaZero:
Ref #5593


ningpp:
rename module  shardingsphere-kernal  to shardingsphere-kernel 
rename module  shardingsphere-kernal-context  to shardingsphere-kernel-context
rename package org.apache.shardingsphere.kernal to org.apache.shardingsphere.kernel

PR #5679, opened 2020-05-19
Fixes sharding-scaling NPE in mysql incremental task update case

### LINKED ISSUE #5390: sharding-scaling nullpoint
4.1.0
sharding-scaling

### Actual behavior
I use sharding-scaling and sharding-proxy to move data from old cluster to the new cluster.
Below exception occured:
`[ERROR] 09:53:53.645 [pool-1-thread-21] o.a.s.s.c.e.e.SyncTaskExecuteCallback - RealtimeDataSyncTask realtime-hc_hiwork_suite_new execute exception exit
java.lang.NullPointerException: null
	at org.apache.shardingsphere.shardingscaling.mysql.MySQLBinlogReader.handleUpdateRowsEvent(MySQLBinlogReader.java:137)
	at org.apache.shardingsphere.shardingscaling.mysql.MySQLBinlogReader.read(MySQLBinlogReader.java:97)
	at org.apache.shardingsphere.shardingscaling.mysql.MySQLBinlogReader.run(MySQLBinlogReader.java:75)
`
the following code caused a nullpointexception:
`!oldValue.equals(newValue)`

The reason is that there is a field, it's oldValue is null.

Use scaling to sync database, and update a row which has null value. 

The correct logic :
`boolean updated = false;
                if(null == oldValue){
                    updated = null != newValue;
                }else{
                    updated = !oldValue.equals(newValue);
                }`

Lucas-307:
Fixes #5390.

### DIFF
--- a/sharding-scaling/sharding-scaling-mysql/src/main/java/org/apache/shardingsphere/shardingscaling/mysql/MySQLBinlogReader.java
+++ b/sharding-scaling/sharding-scaling-mysql/src/main/java/org/apache/shardingsphere/shardingscaling/mysql/MySQLBinlogReader.java
@@ -44,6 +44,7 @@ import org.apache.shardingsphere.sql.parser.binder.metadata.table.TableMetaData;
 import lombok.Setter;
 import lombok.extern.slf4j.Slf4j;
 import java.io.Serializable;
+import java.util.Objects;
 
 /**
  * MySQL binlog reader.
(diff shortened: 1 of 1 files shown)


dongzl:
Hi @Lucas-307 , is This the same with #5670 ?

Lucas-307:
@dongzl This one is commit into apache:dev-4.x, that one is commits into apache:master.

PR #5759, opened 2020-05-22
Optimized aes encryption implementation

klboke:
1、Modified the implementation of AESEncryptor, moved the instantiation and initialization of AES encryption and decryption to init, without having to initialize it every time


klboke:
have conflict

PR #5765, opened 2020-05-23
download the wait-for-it.sh from github

cherrylzhao:
for license issue, remove the wait-for-it.sh, download it from GitHub in dockfile


tristaZero:
Hi @kezhenxu94  
I contacted @cherrylzhao just now and he was kind to fix it here. If available, could you give it a quick look? Thanks. 

BTW, @cherrylzhao do you think it is required to wget this third part file? What concerns me a bit is that we can not ensure this file exists there all the time, right?

shardingsphere-integration-test/shardingsphere-proxy-docker-build/Dockerfile
```diff
@@ -20,6 +20,6 @@ FROM java:8
 
 ARG APP_NAME
 ADD target/${APP_NAME}.tar.gz /opt
-ADD src/main/resources/tool/wait-for-it.sh /opt
+ADD https://github.com/vishnubob/wait-for-it/blob/master/wait-for-it.sh /opt
```
kezhenxu94, line 23: This link seems to be in html format, you may want to use https://raw.githubusercontent.com/vishnubob/wait-for-it/master/wait-for-it.sh ?

kezhenxu94:
> BTW, @cherrylzhao do you think it is required to wget this third part file? What concerns me a bit is that we can not ensure this file exists there all the time, right?

Yes agree, downloading from third-party repo may break someday if they're gone, or even when they have breaking changes, since you're always downloading the latest (master branch) codes, but for test purpose, I think it's OK to do so, yet it's better to lock the version, say https://raw.githubusercontent.com/vishnubob/wait-for-it/c096cface5fbd9f2d6b037391dfecae6fde1362e/wait-for-it.sh for example

cherrylzhao:
> BTW, @cherrylzhao do you think it is required to wget this third part file? What concerns me a bit is that we can not ensure this file exists there all the time, right?

I have no better idea currently, it need this shell to control the order of docker container

kezhenxu94:
> > BTW, @cherrylzhao do you think it is required to wget this third part file? What concerns me a bit is that we can not ensure this file exists there all the time, right?
> 
> I have no better idea currently, it need this shell to control the order of docker container

@cherrylzhao after checking the previous pull request, if I understand correctly, you're trying to determine the startup order and wait for some services to be ready, we have the same requirement in SkyWalking too, and we finally did it by using docker-compose version 2 format, where you can define the health check and dependencies between services, take this as an example

https://github.com/apache/skywalking/blob/a12ecea4494ff9e00f81d82b4ba9119039d0e011/test/e2e/e2e-test/docker/base-compose.yml#L61-L65

https://github.com/apache/skywalking/blob/a12ecea4494ff9e00f81d82b4ba9119039d0e011/test/e2e/e2e-test/docker/simple/jdk/docker-compose.yml#L36-L38

hope it helps

cherrylzhao:
> @cherrylzhao after checking the previous pull request, if I understand correctly, you're trying to determine the startup order and wait for some services to be ready, we have the same requirement in SkyWalking too, and we finally did it by using docker-compose version 2 format, where you can define the health check and dependencies between services, take this as an example
> 
> https://github.com/apache/skywalking/blob/a12ecea4494ff9e00f81d82b4ba9119039d0e011/test/e2e/e2e-test/docker/base-compose.yml#L61-L65
> 
> https://github.com/apache/skywalking/blob/a12ecea4494ff9e00f81d82b4ba9119039d0e011/test/e2e/e2e-test/docker/simple/jdk/docker-compose.yml#L36-L38
> 
> hope it helps

LGTM, I will consider about downgrading the docker-compose if possible

cherrylzhao:
> Yes agree, downloading from third-party repo may break someday if they're gone, or even when they have breaking changes, since you're always downloading the latest (master branch) codes, but for test purpose, I think it's OK to do so, yet it's better to lock the version, say https://raw.githubusercontent.com/vishnubob/wait-for-it/c096cface5fbd9f2d6b037391dfecae6fde1362e/wait-for-it.sh for example

agree with it, raw.github file with version make sense to me

kimmking:
Another choice is post to offical website and refered url like:
> http://shardingsphere.apache.org/tools/xx.sh

kezhenxu94:
The commit is reverted and this could be closed now

PR #5766, opened 2020-05-23
 Optimized aes encryption implementation

klboke:
Migrating the initialization of the encryptor to the init method，Avoid reloading the encryptor every time

### DIFF
--- a/shardingsphere-features/shardingsphere-encrypt/shardingsphere-encrypt-common/src/main/java/org/apache/shardingsphere/encrypt/strategy/impl/AESEncryptor.java
+++ b/shardingsphere-features/shardingsphere-encrypt/shardingsphere-encrypt-common/src/main/java/org/apache/shardingsphere/encrypt/strategy/impl/AESEncryptor.java
@@ -27,11 +27,8 @@ import org.apache.commons.codec.digest.DigestUtils;
 import org.apache.shardingsphere.encrypt.strategy.spi.Encryptor;
 
 import javax.crypto.Cipher;
-import javax.crypto.NoSuchPaddingException;
 import javax.crypto.spec.SecretKeySpec;
 import java.nio.charset.StandardCharsets;
-import java.security.InvalidKeyException;
-import java.security.NoSuchAlgorithmException;
 import java.util.Arrays;
 import java.util.Properties;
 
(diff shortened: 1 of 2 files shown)


klboke:
> I don't think it's a good idea to lift the initialization into `init`, `javax.crypto.Cipher` is not thread-safe and the method `org.apache.shardingsphere.encrypt.strategy.impl.AESEncryptor#encrypt` is definitely possible to be called by multiple threads, you'd better keep it as is, or at least put it into `ThreadLocal`

After the thread shares the instance, there is indeed a thread safety problem. Do you have more detailed documentation on the thread safety of javax.crypto.Cipher?

kezhenxu94:
> > I don't think it's a good idea to lift the initialization into `init`, `javax.crypto.Cipher` is not thread-safe and the method `org.apache.shardingsphere.encrypt.strategy.impl.AESEncryptor#encrypt` is definitely possible to be called by multiple threads, you'd better keep it as is, or at least put it into `ThreadLocal`
> 
> After the thread shares the instance, there is indeed a thread safety problem. Do you have more detailed documentation on the thread safety of javax.crypto.Cipher?

I don't have actually, but by tradition, the classes in JDK usually indicate the thread-safety in the JavaDoc, if it's not documented explicitly, it's not thread-safe, you can also simply check the source codes of `javax.crypto.Cipher`, it's stateful

klboke:
> > > I don't think it's a good idea to lift the initialization into `init`, `javax.crypto.Cipher` is not thread-safe and the method `org.apache.shardingsphere.encrypt.strategy.impl.AESEncryptor#encrypt` is definitely possible to be called by multiple threads, you'd better keep it as is, or at least put it into `ThreadLocal`
> > 
> > 
> > After the thread shares the instance, there is indeed a thread safety problem. Do you have more detailed documentation on the thread safety of javax.crypto.Cipher?
> 
> I don't have actually, but by tradition, the classes in JDK usually indicate the thread-safety in the JavaDoc, if it's not documented explicitly, it's not thread-safe, you can also simply check the source codes of `javax.crypto.Cipher`, it's stateful

I found an answer similar to yours on stackoverflow, https://stackoverflow.com/questions/6957406/is-cipher-thread-safe. In the official jdk documentation, it does not show the thread safety of this class: https://docs.oracle.com/javase/8/docs/api/javax/crypto/Cipher.html#init-int-java. security.cert.Certificate-. I decided to close this merge request first, thank you for your reminder, it may have saved our online application

PR #5768, opened 2020-05-23
fix #4784

### LINKED ISSUE #4784: Throw NPE when INSERT column and values mismatch
Throw NullPointException when INSERT column and values mismatch.
It is better to throw suitable SQLException for JDBC-driver and proxy.

The tasks are:
- [x] Throw suitable SQLException for JDBC-driver.
- [x] Report suitable error packet with MySQL proxy.
- [x] Report suitable error packet with PostgreSQL proxy.

hitnewbee:
Fixes #4784 .

judgement for insertstatement whether column and value match before route.
covered:
1.insert into t_user set xxx=xxx;
2.insert into t_user values(1,2,xxx),(1,2,xxx);
3.insert into t_user(username) values(xxx);

### DIFF
--- a/shardingsphere-infra/shardingsphere-infra-route/src/main/java/org/apache/shardingsphere/infra/route/DataNodeRouter.java
+++ b/shardingsphere-infra/shardingsphere-infra-route/src/main/java/org/apache/shardingsphere/infra/route/DataNodeRouter.java
@@ -17,11 +17,14 @@
 
 package org.apache.shardingsphere.infra.route;
 
+import org.apache.shardingsphere.infra.exception.ShardingSphereException;
 import org.apache.shardingsphere.infra.spi.ShardingSphereServiceLoader;
 import org.apache.shardingsphere.infra.spi.order.OrderedSPIRegistry;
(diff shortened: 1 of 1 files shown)


hitnewbee:
> This pr maybe misunderstand the original meaning.
> We need to throw suitable SQLException for JDBC-driver and return suitable error packet with db proxy.
> ShardingSphereException is not a suitable SQLException, and this pr missing to return suitable error packet with db proxy.

Still a little confuse.
Base on the comment for this issue, i think it is about to block sql that not qualified in shardingsphere.
Do you mean we should let sql execute and get error message from backend database then process it with suitable message and tell user ?
And error packet with db proxy
where i can find an example for this?

terrymanu:
> Do you mean we should let sql execute and get error message from backend database then
> process it with suitable message and tell user ?
> And error packet with db proxy

Yes, just like regular JDBC and database report the error.

> where i can find an example for this?
There is no example for now. Maybe you need to find out which SQLException will throw if SQL is wrong for jdbc.

hitnewbee:
Got it, I will reopen PR util finish it.

Issue #5666, opened 2020-05-18
Task list about clusterstate feature


menghaoranss:
We will start the `control-panel-cluster` module for clusterstate feature through the following task list: 

 - [x] Initialize module basic structure, design API(#5671 ,#5691)
 - [x] Design cluster module configuration structure
 - [x] Proxy read configuration, initialize control-panel-cluster(#5783)
 - [x] Write datasource node and data to registry center(#5691) 
 - [x] Save datasource node after proxy started(#5926)
 - [x] HeartBeat task API design and implementation(#5713)
 - [x] ClusterState API design and implementation（#5691）
 - [x] Unit tests (#5801, #5930)
 - [x] Display topology in UI (#5812, #5845 )

Please refer to #5627 to talk about it.

kimmking:
good job

PR #5812, opened 2020-05-27
Add load all instance states api for UI # control-panel-cluster

menghaoranss:
For #5666 .

- add load all instance states api for UI

### DIFF
--- a/shardingsphere-ui/shardingsphere-ui-backend/src/main/java/org/apache/shardingsphere/ui/servcie/impl/ClusterServiceImpl.java
+++ b/shardingsphere-ui/shardingsphere-ui-backend/src/main/java/org/apache/shardingsphere/ui/servcie/impl/ClusterServiceImpl.java
@@ -1,0 +1,64 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.shardingsphere.ui.servcie.impl;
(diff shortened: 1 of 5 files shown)


shardingsphere-ui/shardingsphere-ui-backend/src/main/java/org/apache/shardingsphere/ui/servcie/impl/ClusterServiceImpl.java
```diff
@@ -0,0 +1,64 @@
...
+        List<String> instanceIds = registryCenterService.getActivatedRegistryCenter()
+                .getChildrenKeys(registryCenterService.getActivatedStateNode().getInstanceNodeRootPath());
+        Map<String, InstanceState> instanceStateMap = new HashMap<>();
+        try {
+            instanceIds.forEach(each -> {
+                instanceStateMap.put(each, loadInstanceState(each));
+            });
+        } catch (Exception ex) {
+            log.error("Load all instance states error", ex);
+        }
```
kezhenxu94, line 54: 1. `try-catch` outside of the `forEach` is considered as a bad idea, if an item throws, all the following items are skipped
2. Simply swallowing the exception is also a bad idea IMO, if I were you, I'll consider the instance state as something like "Unknown" if it throws

shardingsphere-ui/shardingsphere-ui-backend/src/main/java/org/apache/shardingsphere/ui/servcie/impl/ClusterServiceImpl.java
```diff
@@ -0,0 +1,64 @@
...
+    private RegistryCenterService registryCenterService;
+    
+    @Override
+    public Map<String, InstanceState> loadAllInstanceStates() {
+        List<String> instanceIds = registryCenterService.getActivatedRegistryCenter()
+                .getChildrenKeys(registryCenterService.getActivatedStateNode().getInstanceNodeRootPath());
+        Map<String, InstanceState> instanceStateMap = new HashMap<>();
+        try {
+            instanceIds.forEach(each -> {
+                instanceStateMap.put(each, loadInstanceState(each));
```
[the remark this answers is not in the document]
menghaoranss: i will fix.

shardingsphere-ui/shardingsphere-ui-backend/src/main/java/org/apache/shardingsphere/ui/servcie/impl/ClusterServiceImpl.java
menghaoranss: Thanks for your comment, my consideration before was if one instance load failed, the wrong data should not be returned . I agree with your suggestion，‘Unknown’ state is a good idea. 
Thanks again.

Issue #5302, opened 2020-04-24
Enhance Bitronix and Narayana of sharding-transaction


cherrylzhao:
- [x] add unit test for sharding-transaction-xa-bitronix-mananger
- [ ] add unit test for sharding-transaction-xa-narayana-manager
- [x] add example for sharding-transaction-xa-bitronix-mananger
- [ ] add example for sharding-transaction-xa-narayana-manager

TaoZhiMLND:
I want to solve this issue.

PR #5862, opened 2020-05-31
add unit test for sharding-transaction-xa-bitronix-mananger

TaoZhiMLND:
For #5302.


TaoZhiMLND:
After submitting this pull request, there are three reduplicative `ReflectiveUtil` class in test packages of shardingsphere-transaction-xa module.  Should we try to make it better?

cherrylzhao:
> After submitting this pull request, there are three reduplicative `ReflectiveUtil` class in test packages of shardingsphere-transaction-xa module. Should we try to make it better?

we can just inline the reflect call in Test code like other module if necessary

Issue #5446, opened 2020-05-06
Improve scaling abnormal recovery ability


avalon5666:
I think the scaling job always successes in the end, we should try our best to make it complete. The network exception and task execute node crash are the two main reasons which synchronization is failure.
- We can re run executor when network exception.
- For executor node crash, we can dependency scheduling framework rescheduling mechanism, such as elastic job and kubernetes.
- For other reasons, we need admin to fix the problems, but do not shutdown other tasks.

## Breakpoint resume ability
This is necessary for abnormal recovery ability.
- For inventory, we read data order by primary key order, it make us can mark it and re run by this. If there no primary key, we only can read full inventory data.
- For incremental, there is a log position, and we just need to persistence the position.
- Abount the position persistence, i think we can asynchronous persistence to zk.

KomachiSion:
Good suggestion.

There are one question for 
> We can re run executor when network exception.

Whether re-run the executor or just re-try the command which network excepted?

For example, the dumper query with SQL `select` and caused exception by network. Can we just retry this SQL rather than restart this dumper?

avalon5666:
If we want the breakpoint resume in inventory dumper, I think restart the dumper rather than retry SQL. Otherwise just like you say.

avalon5666:
## Task
- [ ] Persistence task position info
- [ ] Importer retry insert when network exception
- [ ] Dumper restart when network exception
- [ ] Start with task position info when dumper start

PR #5894, opened 2020-06-02
Implement persistence task position info

avalon5666:
For #5446.


shardingsphere-scaling/shardingsphere-scaling-core/src/main/java/org/apache/shardingsphere/scaling/core/job/task/incremental/LogPositionManager.java
```diff
@@ -15,25 +15,20 @@
  * limitations under the License.
  */
 
-package org.apache.shardingsphere.scaling.core.job.position;
```
KomachiSion, line 18: Position is also used in job preparer, so I think it should not be move to incremental task package.

avalon5666: The LogPosition manager is the util to get current binlog(wal) position from database. I haven't found the suitable naming for it yet, just adjuest the package.
Now the task position metadata manager is different class, the manager will refactor soon.

PR #5976, opened 2020-06-10
Fix: fix the type transfer issue (#5957)

### LINKED ISSUE #5957: sharding scaling didn't sync the increment data
## Bug Report

4.1.0 or 5.0.0 SNAPSHOT , MySQL 5.7.28

sharding scaling

### Expected behavior

Sync increment data

### Actual behavior

Does'n work

Maybe the `handleUpdateRowsEvent` doesn't work well 

before start this issue , I already ensure the binlog sync mode is ROW , and test table has PK

1. insert some test data into  table `source_schema.test`
2. start sharding scaling, and post the `scaling/job/start` request
3. scaling start to sync the inventory data , this works well
4. during scaling sync the inventory data , try to update one of the data in `source_schema.test` , such as `update test set amount = amount +1 where id = 123`
5. wait for a moment , all inventory data have been transtered to new sharded database through proxy, but the update change doesn't work

I grant the root user all privileges , and config root user in proxy.I'm not sure if there is any other configuration for database or proxy I should set up ?

I tried insert data ,during the inventory sync , if insert some new data , these could be sync to sharded part , but update data does't work

taojintianxia:
Fixes #5957.

- whenever the BIGINT in this situation will transfer to String, so set the jdbc type explicitly

### DIFF
--- a/shardingsphere-scaling/shardingsphere-scaling-core/src/main/java/org/apache/shardingsphere/scaling/core/execute/executor/importer/AbstractJDBCImporter.java
+++ b/shardingsphere-scaling/shardingsphere-scaling-core/src/main/java/org/apache/shardingsphere/scaling/core/execute/executor/importer/AbstractJDBCImporter.java
@@ -127,7 +127,7 @@ public abstract class AbstractJDBCImporter extends AbstractShardingScalingExecut
         try {
             for (int i = 0; i < record.getColumnCount(); i++) {
                 SQLType sqlType = extractTargetSQLType(record.getColumn(i).getValue());
-                if (sqlType != null) {
+                if (null != sqlType) {
                     ps.setObject(i + 1, record.getColumn(i).getValue(), sqlType);
                 } else {
                     ps.setObject(i + 1, record.getColumn(i).getValue());
--- a/shardingsphere-scaling/shardingsphere-scaling-core/src/main/java/org/apache/shardingsphere/scaling/core/execute/executor/importer/AbstractJDBCImporter.java
+++ b/shardingsphere-scaling/shardingsphere-scaling-core/src/main/java/org/apache/shardingsphere/scaling/core/execute/executor/importer/AbstractJDBCImporter.java
@@ -20,6 +20,7 @@ package org.apache.shardingsphere.scaling.core.execute.executor.importer;
 import lombok.Setter;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.shardingsphere.scaling.core.config.RdbmsConfiguration;
+import org.apache.shardingsphere.scaling.core.datasource.DataSourceManager;
 import org.apache.shardingsphere.scaling.core.exception.SyncTaskExecuteException;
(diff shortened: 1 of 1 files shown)


shardingsphere-scaling/shardingsphere-scaling-core/src/main/java/org/apache/shardingsphere/scaling/core/execute/executor/importer/AbstractJDBCImporter.java
```diff
@@ -123,12 +126,24 @@ private void executeInsert(final Connection connection, final DataRecord record)
         ps.setQueryTimeout(30);
         try {
             for (int i = 0; i < record.getColumnCount(); i++) {
-                ps.setObject(i + 1, record.getColumn(i).getValue());
+                SQLType sqlType = extractTargetSQLType(record.getColumn(i).getValue());
+                if (sqlType != null) {
```
terrymanu, line 130: we prefer use `null != xxx` to instead of `xxx != null`

shardingsphere-scaling/shardingsphere-scaling-core/src/main/java/org/apache/shardingsphere/scaling/core/execute/executor/importer/AbstractJDBCImporter.java
```diff
@@ -123,12 +126,24 @@ private void executeInsert(final Connection connection, final DataRecord record)
...
         } catch (SQLIntegrityConstraintViolationException ignored) {
         }
     }
+
+    private SQLType extractTargetSQLType(final Object recordValue) {
+        if (recordValue instanceof BigInteger) {
+            return JDBCType.BIGINT;
+        }
+        return null;
+    }
```
Lucas-307, line 146: Only process Bigint is not eough, we'd better cover all type.

taojintianxia: I tried some types , currently just Bigint will cause that issue.I will test all types later , so that's why I make that check as a method, may add more types in `extractTargetSQLType `

KomachiSion:
I think we should discuss more to solve the problem. From the issue, I think scaling parses the unsigned bigint as `BigInteger`, and `BigInteger` will be transfer as `String` in PreparedStatement as default.

Or Scaling parses unsigned bigint as `String` directly, not `BigInteger`.

avalon5666:
This will implicit conversion in mysql, should the proxy support implicit conversion?

taojintianxia:
> This will implicit conversion in mysql, should the proxy support implicit conversion?

Integer or Long or other will be transfered by mysql , but not BigInteger. and JDBC driver 8.x seems have this implicit conversion

taojintianxia:
this PR is not elegent, I will close this PR

PR #5992, opened 2020-06-11
fix issue#5893:

### LINKED ISSUE #5893: EncryptRule use equalsIgnoreCase method to match encrypt column, but use equals method to match decrypt column.
## Bug Report

ShardingSphere  4.0.1

ShardingSphere-JDBC

### Expected behavior
With the same encrypt-rule, decrypt mode work as same as encrypt mode.
### Actual behavior
Encrypt mode works, but decrypt mode not.

EncryptRule use equalsIgnoreCase method to match encrypt column, but use equals method to match decrypt column.

---------------------------------------------------------------
 sharding rule configuration: 
```
spring:
  shardingsphere:
    props:
      sql.show: true
      query.with.cipher.column: true
    sharding:
      encrypt-rule:
        encryptors:
          encryptor_aes_salt:
            type: aes-salt
            props:
              aes.key.value: 123456
        tables:
          t_account_salt:
            columns:
              mobile:
                cipherColumn: mobile_cipher
                assistedQueryColumn: mobile_assisted
                encryptor: encryptor_aes_salt
```
---------------------------------------------------------------------------------------
Tip: Use mysql config with "lower_case_table_names=1" , it will makes mysql don't care about the column is lower case or upper case. In sharding encrypt table,  use upper case columns to design. In application config, use lower case to config the cipherColumn and assistedQueryColumn. Then insert sql works correctly, but select sql not work.

----------------------------------------------------------------------------------------
With this problem, I find the code in 'encrypt-core-common' module :
```
EncryptTable.java : 
    /**
     * Find encryptor.
     *
     * @param logicColumn column name
     * @return encryptor
     */
    public Optional<String> findEncryptor(final String logicColumn) {
        Optional<String> originLogicColumnName = findOriginLogicColumnName(logicColumn);
        return originLogicColumnName.isPresent() && columns.containsKey(originLogicColumnName.get())
                ? Optional.of(columns.get(originLogicColumnName.get()).getEncryptor()) : Optional.empty();
    }
    
    private Optional<String> findOriginLogicColumnName(final String logicColumn) {
        for (String each : columns.keySet()) {
            if (logicColumn.equalsIgnoreCase(each)) {
                return Optional.of(each);
            }
        }
        return Optional.empty();
    }

```
note: Here use String's equalsIgnoreCase method to match. If true sharding-jdbc may invoke ShardingEncryptor's encrypt method.
```
EncryptRule.java : 

    /**
     * Is cipher column or not.
     *
     * @param tableName table name
     * @param columnName column name
     * @return cipher column or not
     */
    public boolean isCipherColumn(final String tableName, final String columnName) {
        return tables.containsKey(tableName) && tables.get(tableName).getCipherColumns().contains(columnName);
    }

```
note: tables.get(tableName).getCipherColumns() retuen a linkedlist , LinkedList's contains method use equals to match.  If true sharding-jdbc may invoke ShardingEncryptor's decrypt method.

-----------------------------------------------------------------------

see previous.

big-mountain-z:
EncryptRule use equalsIgnoreCase method to match encrypt column, but use equals method to match decrypt column. #5893

Fixes #5893.

- In decrypt mode, use method of String#equalsIgnoreCase to match columnName is cipher column or not.
-
-

### DIFF
--- a/shardingsphere-features/shardingsphere-encrypt/shardingsphere-encrypt-common/src/main/java/org/apache/shardingsphere/encrypt/rule/EncryptRule.java
+++ b/shardingsphere-features/shardingsphere-encrypt/shardingsphere-encrypt-common/src/main/java/org/apache/shardingsphere/encrypt/rule/EncryptRule.java
@@ -167,23 +167,7 @@ public final class EncryptRule implements ShardingSphereRule {
      * @return cipher column or not
      */
     public boolean isCipherColumn(final String tableName, final String columnName) {
-        return tables.containsKey(tableName) && isCipherColumn(tables.get(tableName).getCipherColumns(), columnName);
-    }
-
-    /**
-     * Is cipher column or not.
-     *
-     * @param cipherColumns cipher columns
-     * @param columnName column name
-     * @return cipher column or not
-     */
-    private boolean isCipherColumn(final Collection<String> cipherColumns, final String columnName) {
-        for (String each : cipherColumns) {
-            if (each.equalsIgnoreCase(columnName)) {
-                return true;
-            }
-        }
-        return false;
+        return tables.containsKey(tableName) && tables.get(tableName).getCipherColumns().stream().anyMatch(each -> each.equalsIgnoreCase(columnName));
     }
     
     /**
--- a/shardingsphere-features/shardingsphere-encrypt/shardingsphere-encrypt-common/src/main/java/org/apache/shardingsphere/encrypt/rule/EncryptRule.java
+++ b/shardingsphere-features/shardingsphere-encrypt/shardingsphere-encrypt-common/src/main/java/org/apache/shardingsphere/encrypt/rule/EncryptRule.java
@@ -167,7 +167,23 @@ public final class EncryptRule implements ShardingSphereRule {
      * @return cipher column or not
      */
     public boolean isCipherColumn(final String tableName, final String columnName) {
-        return tables.containsKey(tableName) && tables.get(tableName).getCipherColumns().contains(columnName);
+        return tables.containsKey(tableName) && isCipherColumn(tables.get(tableName).getCipherColumns(), columnName);
+    }
+
+    /**
+     * Is cipher column or not.
+     *
+     * @param cipherColumns cipher columns
+     * @param columnName column name
+     * @return cipher column or not
+     */
+    private boolean isCipherColumn(final Collection<String> cipherColumns, final String columnName) {
+        for (String each : cipherColumns) {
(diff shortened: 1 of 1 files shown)


shardingsphere-features/shardingsphere-encrypt/shardingsphere-encrypt-common/src/main/java/org/apache/shardingsphere/encrypt/rule/EncryptRule.java
```diff
@@ -167,7 +167,23 @@ public String getCipherColumn(final String logicTable, final String logicColumn)
...
+        return tables.containsKey(tableName) && isCipherColumn(tables.get(tableName).getCipherColumns(), columnName);
+    }
+
+    /**
+     * Is cipher column or not.
+     *
+     * @param cipherColumns cipher columns
+     * @param columnName column name
+     * @return cipher column or not
+     */
```
terrymanu, line 179: private method do not need javadoc

shardingsphere-features/shardingsphere-encrypt/shardingsphere-encrypt-common/src/main/java/org/apache/shardingsphere/encrypt/rule/EncryptRule.java
```diff
@@ -167,7 +167,23 @@ public String getCipherColumn(final String logicTable, final String logicColumn)
...
+    private boolean isCipherColumn(final Collection<String> cipherColumns, final String columnName) {
+        for (String each : cipherColumns) {
```
terrymanu, line 181: It is better to use java8 lambda instead of for each

shardingsphere-features/shardingsphere-encrypt/shardingsphere-encrypt-common/src/main/java/org/apache/shardingsphere/encrypt/rule/EncryptRule.java
big-mountain-z: ok, I will make a change.

shardingsphere-features/shardingsphere-encrypt/shardingsphere-encrypt-common/src/main/java/org/apache/shardingsphere/encrypt/rule/EncryptRule.java
big-mountain-z: ok, I will make a change.

PR #6002, opened 2020-06-11
AbstractConnectionAdapter implement Connection#isValid(int) method

### LINKED ISSUE #5882: Sharding 4.1.0 and Spring-boot 2.3.0 incompatible? SQLFeatureNotSupportedException: isValid
When I use sharding-jdbc-spring-boot-starter:4.1.0 and spring-boot:2.2.7 , it works well .
After upgrade spring-boot:2.3.0 , it throws an exception:

org.springframework.dao.InvalidDataAccessApiUsageException: ConnectionCallback; isValid; nested exception is java.sql.SQLFeatureNotSupportedException: isValid

Is this a bug or compatibility problem ?

WEIZIBIN:
Fixes #5882.

- AbstractConnectionAdapter implement Connection#isValid(int timeout) method
- CircuitBreakerConnection implement Connection#isValid(int timeout) method
- ShardingSphereConnectionTest

### DIFF
--- a/shardingsphere-jdbc/shardingsphere-jdbc-core/src/main/java/org/apache/shardingsphere/driver/jdbc/adapter/AbstractConnectionAdapter.java
+++ b/shardingsphere-jdbc/shardingsphere-jdbc-core/src/main/java/org/apache/shardingsphere/driver/jdbc/adapter/AbstractConnectionAdapter.java
@@ -239,6 +239,16 @@ public abstract class AbstractConnectionAdapter extends AbstractUnsupportedOpera
         recordMethodInvocation(Connection.class, "setTransactionIsolation", new Class[]{int.class}, new Object[]{level});
         forceExecuteTemplate.execute(cachedConnections.values(), connection -> connection.setTransactionIsolation(level));
     }
+
+    @Override
+    public final boolean isValid(final int timeout) throws SQLException {
+        for (Connection connection : cachedConnections.values()) {
+            if (!connection.isValid(timeout)) {
+                return false;
+            }
+        }
+        return true;
+    }
     
     // ------- Consist with MySQL driver implementation -------
     
--- a/shardingsphere-jdbc/shardingsphere-jdbc-orchestration/src/main/java/org/apache/shardingsphere/driver/orchestration/internal/circuit/connection/CircuitBreakerConnection.java
+++ b/shardingsphere-jdbc/shardingsphere-jdbc-orchestration/src/main/java/org/apache/shardingsphere/driver/orchestration/internal/circuit/connection/CircuitBreakerConnection.java
@@ -25,6 +25,7 @@ import org.apache.shardingsphere.driver.orchestration.internal.circuit.statement
 import java.sql.Connection;
 import java.sql.DatabaseMetaData;
 import java.sql.PreparedStatement;
+import java.sql.SQLException;
 import java.sql.SQLWarning;
 import java.sql.Statement;
 
@@ -127,7 +128,12 @@ public final class CircuitBreakerConnection extends AbstractUnsupportedOperation
     public PreparedStatement prepareStatement(final String sql, final String[] columnNames) {
         return new CircuitBreakerPreparedStatement();
     }
-    
+
+    @Override
+    public boolean isValid(final int timeout) throws SQLException {
+        return true;
+    }
+
     @Override
     public Statement createStatement() {
         return new CircuitBreakerStatement();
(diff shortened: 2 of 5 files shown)


shardingsphere-jdbc/shardingsphere-jdbc-orchestration/src/main/java/org/apache/shardingsphere/driver/orchestration/internal/circuit/connection/CircuitBreakerConnection.java
```diff
@@ -127,7 +128,12 @@ public PreparedStatement prepareStatement(final String sql, final int[] columnIn
     public PreparedStatement prepareStatement(final String sql, final String[] columnNames) {
         return new CircuitBreakerPreparedStatement();
     }
-    
+
+    @Override
+    public boolean isValid(final int timeout) throws SQLException {
+        return true;
```
kimmking, line 134: why true?

shardingsphere-jdbc/shardingsphere-jdbc-core/src/main/java/org/apache/shardingsphere/driver/jdbc/adapter/AbstractConnectionAdapter.java
```diff
@@ -239,6 +239,16 @@ public final void setTransactionIsolation(final int level) throws SQLException {
         recordMethodInvocation(Connection.class, "setTransactionIsolation", new Class[]{int.class}, new Object[]{level});
         forceExecuteTemplate.execute(cachedConnections.values(), connection -> connection.setTransactionIsolation(level));
     }
+
+    @Override
+    public final boolean isValid(final int timeout) throws SQLException {
+        for (Connection connection : cachedConnections.values()) {
+            if (!connection.isValid(timeout)) {
+                return false;
```
kimmking, line 247: is there some cases need return true, such as m-s?

shardingsphere-jdbc/shardingsphere-jdbc-orchestration/src/main/java/org/apache/shardingsphere/driver/orchestration/internal/circuit/connection/CircuitBreakerConnection.java
WEIZIBIN: As far as the health check is concerned, the service should still be up when the CircuitBreak  is open, so I think isValid method returns true.

shardingsphere-jdbc/shardingsphere-jdbc-core/src/main/java/org/apache/shardingsphere/driver/jdbc/adapter/AbstractConnectionAdapter.java
WEIZIBIN: Do you mean when master is up and slave goes down, it should return true?
This seems better, i will try to identify which connections are from the master data source.

shardingsphere-jdbc/shardingsphere-jdbc-orchestration/src/main/java/org/apache/shardingsphere/driver/orchestration/internal/circuit/connection/CircuitBreakerConnection.java
TheOldDriver: After upgrading spring-boot to version 2.3.1, I also encountered "new SQLFeatureNotSupportedException("isValid")".

 spring-boot-actuator will perform database heartbeat detection according to the configuration file.

 In the "org.springframework.boot.actuate.jdbc.DataSourceHealthIndicator#doDataSourceHealthCheck" method, different heartbeats will be detected based on whether the'query' variable in the DataSourceHealthIndicator class has a value.

 If'query' is not assigned, it will directly call java .sql.Connection#isValid This specific implementation method.
Because shardingJDBC does not implement the java.sql.Connection#isValid method, it will directly report the error "new SQLFeatureNotSupportedException("isValid")".

Currently shardingJDBC does not support the java.sql.Connection#isValid method. 

In order to solve the above problem, you can inherit the "DataSourceHealthContributorAutoConfiguration" method, override the createIndicator method, and set the "query" variable in the DataSourceHealthIndicator class, for example, set it to "select 1".

After setting the variable "query",
 the doDataSourceHealthCheck method will call the "org.springframework.jdbc.core.JdbcTemplate#query(java.lang.String, org.springframework.jdbc.core.RowMapper<T>)" method to complete the database Heartbeat detection.

bert82503:
The complete exception call stack as follow.

[SQLFeatureNotSupportedException-isValid.log](https://github.com/apache/shardingsphere/files/12675182/SQLFeatureNotSupportedException-isValid.log)

The solution is to temporarily disable the monitoring of DB by the HealthEndpoint of Actuator.

```properties
management.health.db.enabled=false
```

PR #6030, opened 2020-06-14
fix issue_6029, add listener to cache before start cache, and avoid a…

### LINKED ISSUE #6029: Add listener to zookeeper TreeCache before start the cache, and avoid adding listener of same path to cache multiple times.
CuratorZookeeperCenterRepository.java
```java
    @Override
    public void watch(final String key, final DataChangedEventListener dataChangedEventListener) {
        final String path = key + "/";
        if (!caches.containsKey(path)) {
            addCacheData(key);
        }
        TreeCache cache = caches.get(path);
        cache.getListenable().addListener((client, event) -> {
            ChildData data = event.getData();
            if (null == data || null == data.getPath()) {
                return;
            }
            DataChangedEvent.ChangedType changedType = getChangedType(event);
            if (DataChangedEvent.ChangedType.IGNORED != changedType) {
                dataChangedEventListener.onChange(new DataChangedEvent(data.getPath(), null == data.getData() ? null : new String(data.getData(), Charsets.UTF_8), changedType));
            }
        });
    }

    private void addCacheData(final String cachePath) {
        TreeCache cache = new TreeCache(client, cachePath);
        try {
            cache.start();
            // CHECKSTYLE:OFF
        } catch (final Exception ex) {
            // CHECKSTYLE:ON
            CuratorZookeeperExceptionHandler.handleException(ex);
        }
        caches.put(cachePath + "/", cache);
    }
```
1.When we start to watch the data change stored in zookeeper, we should add listener first, then start the TreeCache.
2.We should avoid adding listener of same path to cache multiple times.

tswstarplanet:
…dd listener of same path to cache multiple times

Fixes #6029.

-
-
-


tswstarplanet:
Maybe we should use sync ? If one path has multiple listeners, Is there chances to cause multiple consumes of an data change event ?

PR #6149, opened 2020-06-21
add sharing conditions in update or delete sql

Lucas-307:
add sharing conditions in update or delete sql


Lucas-307:
resubmit later

PR #6247, opened 2020-07-02
Optimize startup time, ignore load unConfigured table's column and in…

### LINKED ISSUE #6212: Optimize startup time， control which tables can load when startup
I find the startup time is too slow due to the database full tables metadata scan,  
for our case. the database has 200 tables, we are using micro service, one app loads about 10-20 tables,  it's not acceptable to load all tables on startup

tuohai666:

We have received several similar reports about the slow startup recently, mostly due to loading unconfigured tables for meta data purpose of ShardingSphere.

https://github.com/apache/shardingsphere/issues/6179 : 7000+ tables cost much time.
Similar issues: https://github.com/apache/shardingsphere/issues/6161 , https://github.com/apache/shardingsphere/issues/5755 , https://github.com/apache/shardingsphere/issues/5999 .

Loading table (column and index information) is expensive, there're 7 SQLs interactive for each table.
image

After investigate, I found that for tables unconfigured will not use column or index information any more after loading. So we can omit this process. 

I also checked the relevant issue https://github.com/apache/shardingsphere/issues/4460 , which will load collation for all columns in every table. If a table is unconfigured, the route result must be single route, a single route will not need merge any more. If the resultset does not need to merge, it is irrelevant with cache of collation. So I think we can safely remove the loading of unconfigured tables now.

Note that all unconfigured table names still need to be loaded for routing logic, this can simply obtain from one SQL for each database.

We can accomplish the goal by several steps.

- [ ] Remove the load procedure for unconfigured tables.
- [ ] Remove the refresh procedure for unconfigured tables
- [ ] Remove the multi-thread logic for unconfigured tables. (Configured tables still use multi-thread)
- [ ] Load table names for unconfigured tables.
- [ ] Sync up the new load logic with orchestration module.
- [ ] Update the unit test cases.

xbkaishui:
Fixes #6212.

- Remove the load procedure for unconfigured tables. load only table names for unconfigured tables.
- Remove the refresh procedure for unconfigured tables
- Remove the multi-thread logic for unconfigured tables. (Configured tables still use multi-thread)

### DIFF
--- a/shardingsphere-infra/shardingsphere-infra-common/src/main/java/org/apache/shardingsphere/infra/metadata/schema/RuleSchemaMetaDataLoader.java
+++ b/shardingsphere-infra/shardingsphere-infra-common/src/main/java/org/apache/shardingsphere/infra/metadata/schema/RuleSchemaMetaDataLoader.java
@@ -68,7 +68,6 @@ public final class RuleSchemaMetaDataLoader {
     @SuppressWarnings("unchecked")
     public RuleSchemaMetaData load(final DatabaseType databaseType, final Map<String, DataSource> dataSourceMap, final ConfigurationProperties props,
                                    final ListeningExecutorService executorService) throws SQLException {
-        //store rule actual table
         Collection<String> excludedTableNames = new TreeSet<>(String.CASE_INSENSITIVE_ORDER);
         SchemaMetaData configuredSchemaMetaData = new SchemaMetaData(new HashMap<>());
         for (Entry<ShardingSphereRule, RuleMetaDataLoader> entry : OrderedSPIRegistry.getRegisteredServices(rules, RuleMetaDataLoader.class).entrySet()) {
@@ -80,7 +79,7 @@ public final class RuleSchemaMetaDataLoader {
             configuredSchemaMetaData.merge(schemaMetaData);
         }
         configuredSchemaMetaData = decorate(configuredSchemaMetaData);
-        Map<String, SchemaMetaData> unConfiguredSchemaMetaDataMap = syncLoad(databaseType, dataSourceMap, excludedTableNames);
+        Map<String, SchemaMetaData> unConfiguredSchemaMetaDataMap = load(databaseType, dataSourceMap, excludedTableNames);
         return new RuleSchemaMetaData(configuredSchemaMetaData, unConfiguredSchemaMetaDataMap);
     }
     
@@ -140,7 +139,7 @@ public final class RuleSchemaMetaDataLoader {
         return load(databaseType, dataSourceMap, tableName, props);
     }
     
-    private Map<String, SchemaMetaData> syncLoad(final DatabaseType databaseType, final Map<String, DataSource> dataSourceMap, final Collection<String> excludedTableNames) throws SQLException {
+    private Map<String, SchemaMetaData> load(final DatabaseType databaseType, final Map<String, DataSource> dataSourceMap, final Collection<String> excludedTableNames) throws SQLException {
         Map<String, SchemaMetaData> result = new HashMap<>(dataSourceMap.size(), 1);
         for (Entry<String, DataSource> entry : dataSourceMap.entrySet()) {
             SchemaMetaData schemaMetaData = SchemaMetaDataLoader.load(entry.getValue(), databaseType.getName(), excludedTableNames);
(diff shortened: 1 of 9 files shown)


shardingsphere-sql-parser/shardingsphere-sql-parser-binder/src/main/java/org/apache/shardingsphere/sql/parser/binder/metadata/schema/SchemaMetaDataLoader.java
```diff
@@ -49,35 +41,33 @@
 @NoArgsConstructor(access = AccessLevel.PRIVATE)
 @Slf4j(topic = "ShardingSphere-metadata")
 public final class SchemaMetaDataLoader {
-    
+
```
tuohai666, line 44: Don't change original indentation.

tuohai666: image

xbkaishui: done

xbkaishui:
Sync up the new load logic with orchestration module. for this task. I think there is no code to change.

yu199195:
@xbkaishui  why remove the multi-thread logic for unconfigured tables？

kimmking:
another question is: when refresh metadata, unconfigured need to be refreshed too.

tuohai666:
Hi @yu199195 , @kimmking , please go through https://github.com/apache/shardingsphere/issues/6212 . unconfigured tables do not need to load the column and index information, we can remove them to improve startup speed.

xbkaishui:
> @xbkaishui why remove the multi-thread logic for unconfigured tables？

for unConfigured tables, only need to load table name,  no need table index and column metadata,  so only one sql will load all tables. no need multi thread logic

xbkaishui:
> another question is: when refresh metadata, unconfigured need to be refreshed too.

yes, for the unConfigured table. now is only load table name

shardingsphere-sql-parser/shardingsphere-sql-parser-binder/src/main/java/org/apache/shardingsphere/sql/parser/binder/metadata/schema/SchemaMetaDataLoader.java
```diff
@@ -57,27 +49,25 @@
     /**
      * Load schema meta data.
      *
-     * @param dataSource data source
-     * @param maxConnectionCount count of max connections permitted to use for this query
+     * @param dataSource   data source
```
tuohai666, line 52: redundant spaces

shardingsphere-sql-parser/shardingsphere-sql-parser-binder/src/main/java/org/apache/shardingsphere/sql/parser/binder/metadata/schema/SchemaMetaDataLoader.java
```diff
@@ -57,27 +49,25 @@
...
+        return load(dataSource, databaseType, Collections.emptyList());
     }
     
     /**
      * Load schema meta data.
      *
-     * @param dataSource data source
-     * @param maxConnectionCount count of max connections permitted to use for this query
-     * @param databaseType database type
+     * @param dataSource         data source
```
tuohai666, line 64: redundant spaces

shardingsphere-sql-parser/shardingsphere-sql-parser-binder/src/main/java/org/apache/shardingsphere/sql/parser/binder/metadata/schema/SchemaMetaDataLoader.java
```diff
@@ -57,27 +49,25 @@
...
+     * @param databaseType       database type
```
tuohai666, line 65: redundant spaces

shardingsphere-infra/shardingsphere-infra-common/src/main/java/org/apache/shardingsphere/infra/metadata/schema/RuleSchemaMetaDataLoader.java
```diff
@@ -72,6 +68,7 @@
     @SuppressWarnings("unchecked")
     public RuleSchemaMetaData load(final DatabaseType databaseType, final Map<String, DataSource> dataSourceMap, final ConfigurationProperties props,
                                    final ListeningExecutorService executorService) throws SQLException {
```
tuohai666, line 70: remove RuleSchemaMetaData if not use anymore

Issue #6240, opened 2020-07-02
ShardingProxy starts without rule configurations


tristaZero:
- [x] Review ShardingOrchestration
- [x] Review ShardingProxy
- [x] Concentrate orchestration in OrchestrationSchemaContexts
- [x] Create Rule configuration online

PR #6268, opened 2020-07-05
Persist  part of configurations to registry center

tristaZero:
Ref #6240


shardingsphere-control-panel/shardingsphere-orchestration/shardingsphere-orchestration-core/shardingsphere-orchestration-core-configcenter/src/main/java/org/apache/shardingsphere/orchestration/core/configcenter/ConfigCenter.java
```diff
@@ -98,29 +98,19 @@ public void persistGlobalConfiguration(final Authentication authentication, fina
     }
     
     private void persistDataSourceConfiguration(final String shardingSchemaName, final Map<String, DataSourceConfiguration> dataSourceConfigurations, final boolean isOverwrite) {
-        if (isOverwrite || !hasDataSourceConfiguration(shardingSchemaName)) {
+        if (isOverwrite) {
```
tristaZero, line 101: @wgy8283335 
Hi, could you give it a check? Why did you prefer persisting the configuration for the first time even though `isOverwrite == false`?
Do you think the change here could not cover all the scenarios?

PR #6318, opened 2020-07-10
fix document errors for Sharding chapter.

dongzl:


tristaZero:
Hi @dongzl 
Very appreciated your careful checking. Moreover, I am glad to see you are active so far!

PR #6338, opened 2020-07-13
Refactor registry center instance

menghaoranss:


shardingsphere-control-panel/shardingsphere-orchestration/shardingsphere-orchestration-core/shardingsphere-orchestration-core-registrycenter/src/main/java/org/apache/shardingsphere/orchestration/core/registrycenter/instance/OrchestrationInstance.java
```diff
@@ -32,7 +35,9 @@
     private String instanceId;
 
     private OrchestrationInstance() {
-
+        String tag = Strings.isNullOrEmpty(System.getProperty("port"))
```
kimmking, line 38: use a special key and make it constants

Issue #6289, opened 2020-07-07
insert into select is unsupport


ChuckPerry:
### versions:
shardingproxy version: 4.1.1
mysql: 5.7.21

### replay example:
`mysql>  create table test_t1(id int primary key auto_increment,name varchar(20));`
`mysql>  insert into test_t1 values(1,1);`
case 1:
`mysql> mysql> insert into test_t1(`name`) select t.`name` from test_t1 t;`
`ERROR 10002 (C1000): 2Unknown exception: [null]`
case 2:
`mysql> insert into test_t1(`name`) select '1' as name from test_t1;`
`ERROR 10002 (C1000): 2Unknown exception: [null]`
case 3:
`mysql> insert into test_t1 select 1,'1' ;`
`ERROR 10002 (C1000): 2Unknown exception: [null]`

strongduanmu:
@ChuckPerry The insert select statement and replace select statement have not been implemented yet, and I will try to support it.

ChuckPerry:
OK，thx for your reply

PR #6350, opened 2020-07-14
support mysql insert select statement parse and add some test cases

strongduanmu:
Ref #6289 .

- support mysql insert select statement parse
- add some test cases


shardingsphere-sql-parser/shardingsphere-sql-parser-statement/src/main/java/org/apache/shardingsphere/sql/parser/sql/segment/dml/assignment/InsertSelectSegment.java
```diff
@@ -0,0 +1,36 @@
...
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+import org.apache.shardingsphere.sql.parser.sql.segment.SQLSegment;
+
+/**
+ * Insert select segment.
+ */
+@RequiredArgsConstructor
+@Getter
+public class InsertSelectSegment implements SQLSegment {
```
tristaZero, line 29: Hi @strongduanmu 

This PR looks clean and complete with many unit tests. Good job!
Besides, here are some of my questions, waiting for your feedback.
1. `InsertSelectSegment` and `SelectSegment` looks similar, do you think it is possible to remain either of them?
2. The existing class of `SubquerySegment` and new `SelectSegment` are alike somehow. Could you give them a look again?
3. Are there any consequent PRs needed to complete this feature later?

strongduanmu: @tristaZero 
1. `InsertSelectSegment` and `SelectSegment` should be able to be merged into one class. The two classes were originally designed to be consistent with the syntax rules in antlr.
2. SelectSegment is created with reference to SubquerySegment. Is it misleading to use SubquerySegment because Select is not a subquery in the Insert Select statement?
3. The pr related to route and rewrite will be submitted later. In addition, subsequent pr will also contain replace select statement.

PR #6380, opened 2020-07-19
parse innerExpression and remove sql parameter in SelectStatementContext

strongduanmu:
Ref #6289.

- parse `innerExpression` in parse phrase to replace `innerExpressionStartIndex`;
- remove sql parameter in SelectStatementContext;

### DIFF
--- a/shardingsphere-sql-parser/shardingsphere-sql-parser-dialect/shardingsphere-sql-parser-mysql/src/main/java/org/apache/shardingsphere/sql/parser/mysql/visitor/MySQLVisitor.java
+++ b/shardingsphere-sql-parser/shardingsphere-sql-parser-dialect/shardingsphere-sql-parser-mysql/src/main/java/org/apache/shardingsphere/sql/parser/mysql/visitor/MySQLVisitor.java
@@ -20,6 +20,7 @@ package org.apache.shardingsphere.sql.parser.mysql.visitor;
 import lombok.AccessLevel;
 import lombok.Getter;
 import org.antlr.v4.runtime.ParserRuleContext;
+import org.antlr.v4.runtime.misc.Interval;
 import org.antlr.v4.runtime.tree.TerminalNode;
 import org.apache.shardingsphere.sql.parser.api.ASTNode;
 import org.apache.shardingsphere.sql.parser.autogen.MySQLStatementBaseVisitor;
@@ -301,7 +302,7 @@ public abstract class MySQLVisitor extends MySQLStatementBaseVisitor<ASTNode> {
         if (rightValue instanceof ColumnSegment) {
             return rightValue;
         }
-        return rightValue instanceof SubquerySegment ? new PredicateCompareRightValue(ctx.comparisonOperator().getText(), new SubqueryExpressionSegment((SubquerySegment) rightValue)) 
+        return rightValue instanceof SubquerySegment ? new PredicateCompareRightValue(ctx.comparisonOperator().getText(), new SubqueryExpressionSegment((SubquerySegment) rightValue))
                 : new PredicateCompareRightValue(ctx.comparisonOperator().getText(), (ExpressionSegment) rightValue);
     }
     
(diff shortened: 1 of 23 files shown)


shardingsphere-sql-parser/shardingsphere-sql-parser-dialect/shardingsphere-sql-parser-sqlserver/src/main/java/org/apache/shardingsphere/sql/parser/sqlserver/visitor/SQLServerVisitor.java
```diff
@@ -430,11 +431,11 @@ public final ASTNode visitAggregationFunction(final AggregationFunctionContext c
     
     private ASTNode createAggregationSegment(final AggregationFunctionContext ctx, final String aggregationType) {
         AggregationType type = AggregationType.valueOf(aggregationType.toUpperCase());
-        int innerExpressionStartIndex = ((TerminalNode) ctx.getChild(1)).getSymbol().getStartIndex();
+        String innerExpression = ctx.start.getInputStream().getText(new Interval(((TerminalNode) ctx.getChild(1)).getSymbol().getStartIndex(), ctx.stop.getStopIndex()));
```
tristaZero, line 434: The result of `ctx.start.getInputStream().getText()` and `ctx.getText()` are the same?

tristaZero: Do you think `ctx.getChild(1)).getSymbol().getStartIndex()` == `ctx.LP_().getSymbol().getStartIndex()`?

From the definition of `aggregationFunction `,
```xml
aggregationFunction
    : aggregationFunctionName LP_ distinct? (expr (COMMA_ expr)* | ASTERISK_)? RP_
```
What we want is from LP_ to RP_, right?

shardingsphere-sql-parser/shardingsphere-sql-parser-dialect/shardingsphere-sql-parser-mysql/src/main/java/org/apache/shardingsphere/sql/parser/mysql/visitor/MySQLVisitor.java
```diff
@@ -339,11 +340,11 @@ private PredicateSegment createInSegment(final PredicateContext ctx) {
     }
     
     private PredicateBracketValue createBracketValue(final PredicateContext ctx) {
-        PredicateLeftBracketValue predicateLeftBracketValue = null != ctx.subquery() 
-                ? new PredicateLeftBracketValue(ctx.subquery().LP_().getSymbol().getStartIndex(), ctx.subquery().LP_().getSymbol().getStopIndex()) 
+        PredicateLeftBracketValue predicateLeftBracketValue = null != ctx.subquery()
+                ? new PredicateLeftBracketValue(ctx.subquery().LP_().getSymbol().getStartIndex(), ctx.subquery().LP_().getSymbol().getStopIndex())
                 : new PredicateLeftBracketValue(ctx.LP_().getSymbol().getStartIndex(), ctx.LP_().getSymbol().getStopIndex());
-        PredicateRightBracketValue predicateRightBracketValue = null != ctx.subquery()
```
tristaZero, line 345: Here the check style is out of rule, so you correct them?

shardingsphere-sql-parser/shardingsphere-sql-parser-dialect/shardingsphere-sql-parser-sqlserver/src/main/java/org/apache/shardingsphere/sql/parser/sqlserver/visitor/SQLServerVisitor.java
strongduanmu: > The result of `ctx.start.getInputStream().getText()` and `ctx.getText()` are the same?

@tristaZero According to the documentation of the `ctx.getText()` method, we can find that the method does not return whitespace.
```
/** 
  *  Return the combined text of all leaf nodes. Does not get any
  *  off-channel tokens (if any) so won't return whitespace and
  *  comments if they are sent to parser on hidden channel.
  */
String getText();
```
In order to get the original sql that contains whitespace, we need to use the `ctx.start.getInputStream().getText()` method to specify the start and end interval through Interval.

strongduanmu: > Do you think ctx.getChild(1)).getSymbol().getStartIndex() == ctx.LP_().getSymbol().getStartIndex()?

Yes, maybe using `ctx.LP_().getSymbol().getStartIndex()` is more easier to read. I will optimise it.

> From the definition of aggregationFunction ,
> 
> aggregationFunction
>     : aggregationFunctionName LP_ distinct? (expr (COMMA_ expr)* | ASTERISK_)? RP_
> What we want is from LP_ to RP_, right?

Yes, the innerExpression include sql string from LP_ to RP_. The previous innerExpressionStartIndex is also the same logic.

shardingsphere-sql-parser/shardingsphere-sql-parser-dialect/shardingsphere-sql-parser-mysql/src/main/java/org/apache/shardingsphere/sql/parser/mysql/visitor/MySQLVisitor.java
strongduanmu: I will fix it.

PR #6423, opened 2020-07-23
fix 6354

### LINKED ISSUE #6354: How to get the result in MySQL JSON field
When I execute the following statement, I get two fields, but the result returns one field
`select content_json->>'$.nation' as nation,content_json->>'$.title' as title from  tb_content_json b where b.content_id=1`
image

This table is not divided into tables. I want to ask if the tables that are not partitioned can directly put my SQL into MySQL for execution without parsing SQL through shardingsphere, or how to support the SQL above me

jingshanglu:
Fixes #6354 .


tristaZero:
@jingshanglu Thanks for fixing.

PR #6438, opened 2020-07-25
support mysql insert select statement route and rewrite

### LINKED ISSUE #6289: insert into select is unsupport

strongduanmu:
Ref #6289 .

- support mysql insert select statement route and rewrite
- add some rewrite test cases

### DIFF
--- a/shardingsphere-features/shardingsphere-sharding/shardingsphere-sharding-common/src/main/java/org/apache/shardingsphere/sharding/rule/ShardingRule.java
+++ b/shardingsphere-features/shardingsphere-sharding/shardingsphere-sharding-common/src/main/java/org/apache/shardingsphere/sharding/rule/ShardingRule.java
@@ -312,6 +312,22 @@ public final class ShardingRule implements DataNodeRoutedRule {
     }
     
     /**
+     * Judge is generate key column or not.
+     *
+     * @param columnName column name
+     * @param tableName table name
+     * @return is generate key column or not
+     */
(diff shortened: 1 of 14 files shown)


tristaZero:
Another improvement we can consider is moving checking logics of `route()` in `ShardingStandardRoutingEngine` to `ShardingStatementValidator`. Besides, do you think it will become better to pass `SQLStatementContext` instead of `SQLStatement` to it? If so, `tableContexts` is possible to remove from the parameter list of `validate()`.

Since this interface was created by @terrymanu  , I will ask him for advice ASAP.

tristaZero:
@strongduanmu @terrymanu has no problem with it. Please think about it. :)

strongduanmu:
> @strongduanmu @terrymanu has no problem with it. Please think about it. :)

Thank you very much for confirming, I will optimize it. 😀

PR #6509, opened 2020-07-29
Fix test

jingshanglu:
- fix sql parser test case
- simplify sql parser test frame
- check CreateTableStatement
- fix ConstraintDefinition assert


jingshanglu:
@tristaZero CreateTable should not have index defined at the end, index define should in ConstraintDefinition, i add it and add assert for it.

Issue #5841, opened 2020-05-29
make example cover all posibility


taojintianxia:
### Is your feature request related to a problem?
no

### Describe the feature you would like.
Refactor the example module to cover all usage combinations

Followings are the combinations

<details>
  <summary>sharding-jdbc</summary>

- [ ] sharding
    - [ ] java
    - [ ] yaml
    - [ ] properties
    - [ ] springboot
    - [ ] spring-namespace
- [ ] master-slave
    - [ ] java
    - [ ] yaml
    - [ ] properties
    - [ ] springboot
    - [ ] spring-namespace
- [ ] encryption
    - [ ] java
    - [ ] yaml
    - [ ] properties
    - [ ] springboot
    - [ ] spring-namespace
- [ ] shadow
    - [ ] java
    - [ ] yaml
    - [ ] properties
    - [ ] springboot
    - [ ] spring-namespace
- [ ] replica
    - [ ] java
    - [ ] yaml
    - [ ] properties
    - [ ] springboot
    - [ ] spring-namespace
</details>
<details>
  <summary>proxy</summary>

- [ ] sharding
    - [ ] java
    - [ ] yaml
    - [ ] properties
    - [ ] springboot
    - [ ] spring-namespace
- [ ] master-slave
    - [ ] java
    - [ ] yaml
    - [ ] properties
    - [ ] springboot
    - [ ] spring-namespace
- [ ] encryption
    - [ ] java
    - [ ] yaml
    - [ ] properties
    - [ ] springboot
    - [ ] spring-namespace
- [ ] shadow
    - [ ] java
    - [ ] yaml
    - [ ] properties
    - [ ] springboot
    - [ ] spring-namespace
- [ ] replica
    - [ ] java
    - [ ] yaml
    - [ ] properties
    - [ ] springboot
    - [ ] spring-namespace
</details>

taojintianxia:
I‘d like to contribute the corresponding code

terrymanu:
expired

PR #6602, opened 2020-08-04
refactor the example for XA JDBC example

taojintianxia:
for #5841.

- refactor the test case for xa raw jdbc


examples/pom.xml
```diff
@@ -59,7 +60,7 @@
         <btm.version>2.1.3</btm.version>
 
         <seata.version>1.0.0</seata.version>
-        
+
```
terrymanu, line 63: Please keep original indent

Issue #6466, opened 2020-07-27
Support ShardingSphere-defined RDL SQL 


tristaZero:
Hi community,

We plan to provide **RDL(Rule Definition Language)** for `ShardingProxy` since the 5.x version. This feature simplifies the tedious workload for the Sharding configuration. 
The user can start ShardingProxy without `sharding rule` (However, the registry center is required for the rule durability), then the user can create a sharding rule or a sharding table through RDL SQL.

Presently, the execution part for `RDL` is basically done, and this issue is to parse RDL SQL out and provide the corresponding SQLParser.

strongduanmu:
This feature sounds great!

PR #6618, opened 2020-08-04
Add gitignore and optimize proxy backend interface.

tristaZero:
Ref #6466


.gitignore
```diff
@@ -1,5 +1,6 @@
 # maven ignore
 target/
+gen/
```
terrymanu, line 3: What is useful of this folder?

tristaZero: ? You suggested to add it to gitignore, didn't you? BTW, it is for code generating from antlr.

PR #6621, opened 2020-08-04
fix DatabaseCommunicationEngineFactoryTest 

kimmking:
fixTest


kimmking:
org.mockito.Mockito.RETURNS_DEEP_STUBS is to avoid the temp object mock for a chain methods call.
Such as we need mock `a.b().c().d()`.

The old way is:
```
mock d as D,
mock c as C,
mock c.d() return d;
mock b as B,
mock b.c() return c;
mock a as A,
mock a.b() return b;
```

If we mock with 'RETURNS_DEEP_STUBS':
```
mock d as D,
mock a as A with RETURNS_DEEP_STUBS,
mock a.b().c().d() return d;
```

PR #6689, opened 2020-08-07
Add forceShardingColumn to TableRule

### LINKED ISSUE #6688: Add forceShardingColumn to TableRule
Add forceShardingColumn configuration to TableRule
the default value of the field is False when it is True, DML SQL must have the sharding column otherwise throw exception

jimgreat:
Fixes #6688

-
-
-


terrymanu:
It is unnecessary to change the API.
forceShardingColumn is too complicated and it is not the general requirement.
I suggest do not change it.

PR #6752, opened 2020-08-10
refactor the XAOrderServiceTest (#6751)

### LINKED ISSUE #6751: refactor the XAOrderServiceTest
currently the test case depends on a mysql server , should change to H2

taojintianxia:
Fixes #6751 .

- add H2 configuration for test case


taojintianxia:
> Excuse me, why use h2 for the example?
> MySQL is more generic for example.

because these codes are for `Junit Test` . use in memory Database is better than mysql . otherwise we will get failed test if we execute `mvn clean install` in example module

terrymanu:
Example is faced on users, there is no test cases needed.

PR #6772, opened 2020-08-10
modify ci configuration

kimmking:
modify ci configuration


kimmking:
> What is the useful of this issue?

1. keep the same configuration with elastic job project
2. add a new autogen ignore rule for coverage report

PR #6905, opened 2020-08-18
fix null schema error

owen200008:
fix null schema error

- null point error, when schema is null


terrymanu:
@owen200008， Hi, I will close soon if no feedback

PR #6906, opened 2020-08-18
fixes SQLServer SQL “columnNames” parser, instead of “columnNameWithSort".

dongzl:

### DIFF
--- a/shardingsphere-sql-parser/shardingsphere-sql-parser-dialect/shardingsphere-sql-parser-sqlserver/src/main/antlr4/imports/sqlserver/BaseRule.g4
+++ b/shardingsphere-sql-parser/shardingsphere-sql-parser-dialect/shardingsphere-sql-parser-sqlserver/src/main/antlr4/imports/sqlserver/BaseRule.g4
@@ -119,7 +119,7 @@ name
     ;
 
 columnNames
-    : LP_ columnNameWithSort (COMMA_ columnNameWithSort)* RP_
+    : LP_ columnName (COMMA_ columnName)* RP_
     ;
 
 tableNames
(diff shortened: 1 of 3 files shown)


shardingsphere-sql-parser/shardingsphere-sql-parser-dialect/shardingsphere-sql-parser-sqlserver/src/main/antlr4/imports/sqlserver/BaseRule.g4
```diff
@@ -119,7 +119,7 @@ name
     ;
 
 columnNames
-    : LP_ columnNameWithSort (COMMA_ columnNameWithSort)* RP_
+    : LP_ columnName (COMMA_ columnName)* RP_
```
jingshanglu, line 122: @dongzl Maybe it's not appropriate,because `columnNames` is used by `createIndex` in DDLStatement.g4, and actually `columnNameWithSort` is right. url:https://docs.microsoft.com/zh-cn/sql/t-sql/statements/create-index-transact-sql?view=sql-server-ver15

dongzl: Hi @jingshanglu , `columnNames` also use in the `insertValuesClause`, if use `columnNameWithSort`, I test the `insert into t_order (order_id asc, user_id desc) values (1, 1)` is also a correct sql.

Do we need divide the different rule for `createIndex `  and `insert`?

jingshanglu: @dongzl It is better to divide rule for `createIndex` and `insert`, because it can ensure the accuracy of grammar, although there is some redundancy.

dongzl:
I communicate with jingshanglu, I will divide different rules for `createIndex` and `insert`.

Issue #6655, opened 2020-08-06
 SqlServer : no viable alternative at input 'WITH'


rongflag:
### Which version of ShardingSphere did you use?
ShardingSphere-JDBC 4.1.1 
### Which project did you use? ShardingSphere-JDBC or ShardingSphere-Proxy?
ShardingSphere-JDBC
### Example codes for reproduce this issue (such as a github link).

SQL: 
WITH selectTemp AS (SELECT TOP 100 PERCENT ROW_NUMBER() OVER (ORDER BY CURRENT_TIMESTAMP) as __row_number__, role_id, role_name, role_code, role_desc, ds_type, ds_scope, create_time, update_time, del_flag FROM sys_role WHERE del_flag = '0' AND tenant_id = 1) SELECT * FROM selectTemp WHERE __row_number__ BETWEEN 1 AND 20 ORDER BY __row_number__ 

LOG::
line 1:0 no viable alternative at input 'WITH'
line 1:0 mismatched input 'WITH' expecting {SELECT, INSERT, UPDATE, DELETE, CREATE, ALTER, DROP, TRUNCATE, GRANT, REVOKE, SET, IF, BEGIN, COMMIT, ROLLBACK, CALL, DENY, SAVE}

ERRORE:
Caused by: org.apache.shardingsphere.sql.parser.exception.SQLParsingException: Unsupported SQL of `WITH selectTemp AS (SELECT TOP 100 PERCENT  ROW_NUMBER() OVER (ORDER BY CURRENT_TIMESTAMP) as __row_number__,  role_id, role_name, role_code, role_desc, ds_type, ds_scope, create_time, update_time, del_flag FROM sys_role WHERE del_flag = '0' AND tenant_id = 1) SELECT * FROM selectTemp WHERE __row_number__ BETWEEN 1 AND 20 ORDER BY __row_number__`
	at org.apache.shardingsphere.sql.parser.core.parser.SQLParserExecutor.execute(SQLParserExecutor.java:49)
	at org.apache.shardingsphere.sql.parser.SQLParserEngine.parse0(SQLParserEngine.java:79)
	at org.apache.shardingsphere.sql.parser.SQLParserEngine.parse(SQLParserEngine.java:61)
	at org.apache.shardingsphere.underlying.route.DataNodeRouter.createRouteContext(DataNodeRouter.java:97)
	at org.apache.shardingsphere.underlying.route.DataNodeRouter.executeRoute(DataNodeRouter.java:89)
[...]

Issue #6478, opened 2020-07-28
[SQL Definition Collation][SQLServer]Collate the g4 files of SQL definitions


jingshanglu:
Hi community,

This issue is to collate the SQL definitions of SQLServer. **Therefore, we can support parsing more SQLs**. 
We need your interaction! If you 

> 1. Be familiar with `g4 file` of Antlr (The grammars are RegularExpression-like)
> 2. Understand the `SQLServer grammars`

then welcome!

### Background

**How is a SQL parsed?**
`SQL definition file` 
->(Parse to) 
`AST`
->(Visit it)  
`SQLStatment`  

### Process

1. Compare `SQL definitions` in [Oficial SQL Doc](https://docs.microsoft.com/zh-cn/sql/t-sql/queries/select-transact-sql?view=sql-server-ver15) and [ShardingSphere SQL Doc](https://github.com/apache/shardingsphere/tree/master/shardingsphere-sql-parser/shardingsphere-sql-parser-dialect/shardingsphere-sql-parser-sqlserver/src/main/antlr4/imports/sqlserver)
2. If there is any difference in ShardingSphere SQL Doc, please correct them referring to Official SQL Doc
3. Run `mvn install the current_file_module` 
4. Check whether there are any exceptions. If indeed, please fix them. (Especially `xxxVisitor.class`)
5. Add new corresponding SQL case in [SQL Cases](https://github.com/apache/shardingsphere/tree/master/shardingsphere-sql-parser/shardingsphere-sql-parser-test/src/test/resources/sql/supported) and expected parsed result in [Expected Statment XML](https://github.com/apache/shardingsphere/tree/master/shardingsphere-sql-parser/shardingsphere-sql-parser-test/src/test/resources/case)
6. Run [SQLParserParameterizedTest](https://github.com/apache/shardingsphere/blob/master/shardingsphere-sql-parser/shardingsphere-sql-parser-test/src/test/java/org/apache/shardingsphere/sql/parser/integrate/engine/SQLParserParameterizedTest.java) to make sure no exceptions.

### Notice
1.`Left recursion` can not work well in Antlr4, 
```
a
  : b
  | a b
  ;
b
  : a c
  ;
c
  : ...
  ;
```
you can consider the following one,
```
a
  : b*
  ;
b
  : a c
  ;
c
  : ...
  ;
```

2. [Here](https://shimo.im/docs/WrgrCt9JR3JDjq8H/read) is a Chinese version for reference.

### subtask
1、DML
[...]

PR #7111, opened 2020-08-28
Add with segment parser for SQLServer insert statement.

dongzl:
For #6655 .
For #6478 .


jingshanglu:
@dongzl @tristaZero withClause is optional, MySQL has not withClause in insertStatement, maybe it is better to create subclass of InsertStatement, what do you think?

tristaZero:
Hi @dongzl @jingshanglu Got it.
I suppose it is required to review all of the current SQL statements and consider splitting them into various subclasses in terms of SQL dialects. But it needs some time to reflect its process and architecture. Besides, If we deal with two of these things together, that will frustrate us. So I suggest merging this one and later for that splitting work.

dongzl:
Hi @tristaZero @jingshanglu , I understand for all the database SQL dialects, they have the same parts and there also will be some differences, we should take the same parts to the parent class, and take the differences to subclass. But split them into different subclasses will need more time, I suggest we can finish the SQL collation firstly and then consider splitting them.

Thanks!

jingshanglu (approved):

PR #7120, opened 2020-08-28
#7011, Investigate whether missing call close

### LINKED ISSUE #7011: Investigate whether missing call close method in UserExecutorGroup
Is there need to close UserExecutorGroup on FrontendChannelInboundHandler.channelInactive()?

tuohai666:
Fixes #7011.


terrymanu (changes requested):
If there is no need to call close, do we need to keep the close method?

tuohai666:
> If there is no need to call close, do we need to keep the close method?

Not really. The life cycle of UserExecutorGroup is identical with ShardingSphere-Proxy.

terrymanu (approved):

PR #7130, opened 2020-08-28
Fix problem for springboot to use shardingsphere.

291277058:
Fix problem for springboot to use shardingsphere.

`aes.key.value` can not read as map key in springboot application.properties.

If forbiden update sharding key, jpa save math can not execute success due to the exeception


291277058:
And table route function for update sql.

PR #7173, opened 2020-08-31
update opentracing version to 0.32.0

eightroad:
Fixes #ISSUSE_ID.

- update opentracing api from 0.30.0 to 0.32.0


kimmking:
We consider move open-tracing integration to externals and replace it with java agent plugin. 
So version updating is now not in our plan.

PR #7234, opened 2020-09-03
Fixes drop schema event

menghaoranss:


shardingsphere-infra/shardingsphere-infra-common/src/main/java/org/apache/shardingsphere/infra/eventbus/event/SchemaNameEvent.java
```diff
@@ -29,5 +29,5 @@
     
     private final String schemaName;
     
-    private final boolean overwrite; 
+    private final boolean isDrop;
```
tristaZero, line 32: Why did you change it?

tristaZero (changes requested):

PR #7236, opened 2020-09-03
Refactor: overload the scaling MetaDataManager

taojintianxia:
- overload the getTableMetaData because `TableMetaDataLoader.load` accept databaseType , then hard code an empty string is not a good way

### DIFF
--- a/shardingsphere-scaling/shardingsphere-scaling-core/src/main/java/org/apache/shardingsphere/scaling/core/metadata/MetaDataManager.java
+++ b/shardingsphere-scaling/shardingsphere-scaling-core/src/main/java/org/apache/shardingsphere/scaling/core/metadata/MetaDataManager.java
@@ -43,7 +43,7 @@ public final class MetaDataManager {
      * @return table meta data
      */
     public TableMetaData getTableMetaData(final String tableName) {
-        return getTableMetaData(tableName, null);
+        return getTableMetaData(tableName, "");
     }
     
     /**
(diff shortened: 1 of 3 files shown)


shardingsphere-scaling/shardingsphere-scaling-core/src/main/java/org/apache/shardingsphere/scaling/core/metadata/MetaDataManager.java
```diff
@@ -43,9 +43,20 @@
      * @return table meta data
      */
     public TableMetaData getTableMetaData(final String tableName) {
+        return getTableMetaData(tableName, "");
```
Lucas-307, line 46: Hi, if you add `getTableMetaData(final String tableName, final String databaseType)` function. This function do not have to exist. and you should update all quote.

How do you know which function can use in which case?

And `databaseType` only affect `Oracle` database, but we don't support `Oracle` in scaling.

So, I suggest keep current code.

Lucas-307 (changes requested):

shardingsphere-scaling/shardingsphere-scaling-core/src/main/java/org/apache/shardingsphere/scaling/core/metadata/MetaDataManager.java
taojintianxia: I understand

taojintianxia: my origin though is replace `getTableMetaData(final String tableName) ` with `getTableMetaData(final String tableName, final String databaseType) ` one by one , and drop `getTableMetaData(final String tableName) ` at last . since we don't support Oracle . enn , I see

PR #7246, opened 2020-09-04
Document: add attention for scaling

taojintianxia:
- add document for scaling attention


Lucas-307 (changes requested):
invalid

Lucas-307:
Explain: If a table don't have a `primary key`, it can't split by row and resume from break-point. But we can scaling it with whole table.

PR #7257, opened 2020-09-04
Minor refactoring

kimmking:
Minor refactoring for constant/logging, cutting redundant codes


shardingsphere-governance/shardingsphere-governance-core/shardingsphere-governance-core-facade/src/main/java/org/apache/shardingsphere/governance/core/facade/GovernanceFacade.java
```diff
@@ -100,6 +103,7 @@ public void onlineInstance() {
     @Override
     public void close() {
         repositoryFacade.close();
+        log.info("GovernanceRepositoryFacade has been closed.");
```
terrymanu, line 106: Any reason to add this log?
UNIX philosophy is "No news is good news", I prefer do not add unnecessary logs

terrymanu (changes requested):

terrymanu (approved):

Issue #7170, opened 2020-08-31
The refinement on categorizing all the SQL statements


tristaZero:
Hi community,

I know this issue will be another considerable workload for the community. However, we have to start and boost it to move forward.

The reason is that **the overmuch and detailed segments overload the corresponding SQL statements at presently**. 

For instance, there is a `SetAssignmentSegment` in `InsertStatement`. Yet, the dialect SQL `INSERT SET` is an exclusive SQL for MySQL and not available for other databases. Hence, an `MySQLInsertStatement` with `SetAssignmentSegment` is needed. Moveover, it is better to split the original `InsertStatement` into `OracleInsertStatement`, `PostgreSQLInsertStatment`, `SQLServerInsertStatement`.

The case above is just a toy example, and there are many similar cases here. We have to polish all the SQL Statements. **Welcome you, who are interested in this issue, to join the community to push this issue to move ahead!**

Cheers,
Trista

dongzl:
very great work!

strongduanmu:
@tristaZero Hi, I want to participate in this task, can you assign some tasks to me? 😀

PR #7292, opened 2020-09-07
Refactor DALStatement to MySQL, PostgreSQL DALStatement

strongduanmu:
Ref #7170 .

- Refactor DALStatement to MySQL, PostgreSQL DALStatement


shardingsphere-sql-parser/shardingsphere-sql-parser-statement/src/main/java/org/apache/shardingsphere/sql/parser/sql/common/statement/dal/ChecksumTableStatement.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+package org.apache.shardingsphere.sql.parser.sql.common.statement.dal;
+
+import org.apache.shardingsphere.sql.parser.sql.common.statement.AbstractSQLStatement;
+
+/**
+ * Checksum table statement.
+ */
+public abstract class ChecksumTableStatement extends AbstractSQLStatement {
```
tristaZero, line 25: Hi @strongduanmu Here `ChecksumTableStatement` is an abstract class. Does that mean we will have `MySQL ChecksumTableStatement`, or `OracleChecksumTableStatement`? But I only saw the definition of `MySQL ChecksumTableStatement`. Therefore, do you think the abstract class `ChecksumTableStatement` is required? Or maybe it is more suitable just to have a class named `MySQL ChecksumTableStatement`?

Other abstract classes have such an issue as well, so I would like to listen to what you think.

tristaZero (changes requested):

shardingsphere-sql-parser/shardingsphere-sql-parser-statement/src/main/java/org/apache/shardingsphere/sql/parser/sql/common/statement/dal/ChecksumTableStatement.java
strongduanmu: @tristaZero This way is to ensure the consistency of the processing logic. Whether there is one subclass or multiple subclasses, it can be processed towards the parent class. Only when there is a special SQL dialect, the subclass will be used. 😂
Of course, when the SQL statement has only one dialect, this may cause users to misunderstand that it is a general statement.

shardingsphere-sql-parser/shardingsphere-sql-parser-statement/src/main/java/org/apache/shardingsphere/sql/parser/sql/dialect/statement/mysql/dal/MySQLAnalyzeTableStatement.java
```diff
@@ -17,12 +17,12 @@
...
+import org.apache.shardingsphere.sql.parser.sql.common.statement.dal.AnalyzeTableStatement;
 import org.apache.shardingsphere.sql.parser.sql.common.statement.dal.DALStatement;
 import org.apache.shardingsphere.sql.parser.sql.dialect.statement.mysql.MySQLStatement;
 
 /**
- * Analyze table statement.
+ * MySQL analyze table statement.
  */
-public final class MySQLAnalyzeTableStatement extends AbstractSQLStatement implements DALStatement, MySQLStatement {
+public final class MySQLAnalyzeTableStatement extends AnalyzeTableStatement implements DALStatement, MySQLStatement {
```
tristaZero, line 27: Hi @strongduanmu 
Where is `MySQLAnalyzeTableStatement` used?

tristaZero (changes requested):

shardingsphere-sql-parser/shardingsphere-sql-parser-statement/src/main/java/org/apache/shardingsphere/sql/parser/sql/dialect/statement/mysql/dal/MySQLAnalyzeTableStatement.java
strongduanmu: > Hi @strongduanmu
> Where is `MySQLAnalyzeTableStatement` used?

@tristaZero Now, `MySQLAnalyzeTableStatement` is only used in the `MySQLDALVisitor` class for parsing. The execution of `anlyze table statement` is not yet supported.

tristaZero: Thanks for your explanation. Merged.

tristaZero (approved):

Issue #6872, opened 2020-08-16
Failure test cases of BaseIT for PostgreSQL


sandynz:
TODO:
- [x] Fix BaseIT.createTables:131 exception, support keyword table name for different databases (#6912)
- [x] Fix BaseIT.createDatabases:121 exception, databases not cleaned by previous BaseIT (#6912)
- [x] Current PostgreSQL login user's password is changed by DCL test (#6943)
- [x] Remove duplicated test case (https://github.com/apache/shardingsphere/pull/6962)
- [x] DQL dataset select_count_tilde_concat.xml not found cause GeneralDQLIT.assertExecuteQuery failed (#7007)
- [x] PostgreSQL dedicated column missed in t_order cause GeneralDQLIT.assertExecuteQuery failed (#7010)
- [x] Print exception context for engine (https://github.com/apache/shardingsphere/pull/7005)
- [x] Integration test of PostgreSQL failed to drop database (#7174)
- [x] LockSegment absent of PostgreSQL parser (#7200)
- [x] Unexpected SELECT DISTINCT result (#7201)
- [x] remove unexpected error log (https://github.com/apache/shardingsphere/pull/7205)
- [x] assertRows failed in PostgreSQL DML integration test (#7204)
- [x] INSERT...RETURING statement failed in PostgreSQL DML integration test (#7207)
- [x] DROP ROLE/USER throws Exception for all database types (https://github.com/apache/shardingsphere/pull/7215)
- [x] Failure test cases of DCL for PostgreSQL (https://github.com/apache/shardingsphere/pull/7224)
- [x] Failure test cases of DCL for PostgreSQL 1 (https://github.com/apache/shardingsphere/pull/7230)
- [x] [Improve Performance] remove unnecessary clear procedure (https://github.com/apache/shardingsphere/pull/7254)
- [x] [Improve Performance] concurrent truncate (https://github.com/apache/shardingsphere/pull/7300 , https://github.com/apache/shardingsphere/pull/7305)
- [x] [Improve Performance] do not close thread pool for each test (https://github.com/apache/shardingsphere/pull/7327)
- [x] [Improve Performance] use delete instead of truncate to improve performance (https://github.com/apache/shardingsphere/pull/7333)
- [x] [Improve Performance] remove connection test `SELECT 1` to improve performance (https://github.com/apache/shardingsphere/pull/7334)
- [x] [Improve Performance] concurrent insert (https://github.com/apache/shardingsphere/pull/7356)
[...]

PR #7306, opened 2020-09-07
#6872, add maven-surefire-plugin vm

tuohai666:
For #6872.


tuohai666:
Set argLine as a property in the pom.xml file will work. Do not need to add to configuration.

PR #7363, opened 2020-09-09
Delete the abstract super class of a single DCL SQL dialect

strongduanmu:
Ref #7170.

- delete the abstract super class of a single DCL SQL dialect


tristaZero (approved):
Spotless!

PR #7486, opened 2020-09-16
Fix shadow value in literal expression

### LINKED ISSUE #7280: Judge whether shadow logic should be upgrade
The current shadow judge engine logic split to `simple` and `prepared` which chosen by SQL parameters are empty.

The value will miss if there are parameters in prepared statement, but the shadow value is not in parameter.

yanickxia:
Fixes #7280.

- fix shadow value in literal expression


yanyzy:
Could you add some unit tests in ShadowPreparedStatementTest?

yanyzy (approved):

PR #7506, opened 2020-09-18
refactor InsertStatement to different dialect Statement

strongduanmu:
Ref #7170.

- refactor InsertStatement to different dialect Statement

### DIFF
--- a/shardingsphere-sql-parser/shardingsphere-sql-parser-statement/src/main/java/org/apache/shardingsphere/sql/parser/sql/common/statement/dml/InsertStatement.java
+++ b/shardingsphere-sql-parser/shardingsphere-sql-parser-statement/src/main/java/org/apache/shardingsphere/sql/parser/sql/common/statement/dml/InsertStatement.java
@@ -19,15 +19,11 @@ package org.apache.shardingsphere.sql.parser.sql.common.statement.dml;
 
 import lombok.Getter;
 import lombok.Setter;
-import org.apache.shardingsphere.sql.parser.sql.common.segment.dml.assignment.AssignmentSegment;
 import org.apache.shardingsphere.sql.parser.sql.common.segment.dml.assignment.InsertValuesSegment;
-import org.apache.shardingsphere.sql.parser.sql.common.segment.dml.assignment.SetAssignmentSegment;
 import org.apache.shardingsphere.sql.parser.sql.common.segment.dml.column.ColumnSegment;
 import org.apache.shardingsphere.sql.parser.sql.common.segment.dml.column.InsertColumnsSegment;
-import org.apache.shardingsphere.sql.parser.sql.common.segment.dml.column.OnDuplicateKeyColumnsSegment;
 import org.apache.shardingsphere.sql.parser.sql.common.segment.dml.expr.ExpressionSegment;
 import org.apache.shardingsphere.sql.parser.sql.common.segment.dml.expr.subquery.SubquerySegment;
-import org.apache.shardingsphere.sql.parser.sql.common.segment.generic.WithSegment;
 import org.apache.shardingsphere.sql.parser.sql.common.segment.generic.table.SimpleTableSegment;
 import org.apache.shardingsphere.sql.parser.sql.common.statement.AbstractSQLStatement;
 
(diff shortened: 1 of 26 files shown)


shardingsphere-sql-parser/shardingsphere-sql-parser-statement/src/main/java/org/apache/shardingsphere/sql/parser/sql/common/statement/dml/InsertStatement.java
```diff
@@ -95,107 +76,63 @@
...
-     */
-    public Optional<WithSegment> getWithSegment() {
-        return Optional.ofNullable(withSegment);
-    }
-    
     /**
      * Judge is use default columns or not.
      * 
      * @return is use default columns or not
      */
```
tristaZero, line 83: Hi, I am glad to see this PR raized since it is the`real` refactoring work touching some of the database dialect issues.
Separately reviewing this PR, I suppose it seems great except for a renaming issue. However, we can not avoid handling database dialects later. Therefore I want to give two points after my deliberation. Could you give them some consideration?

1. A interface named `SQLstatementUtil<T entends SQLStatement>` (Like ·SQLStatementContext<T extends SQLStatement>·) is needed to deal with the judgment or handling for different dialect SQLStatments in the process of rewriting, routing, etc. 

2. We should keep all the `SQLStatment` clear and simple enough, which means `SQLStatement` just provides the basic public functions. 

**What's the difference between `SQLStatement` and `SQLStatmentContext`. ?**

My answer is that `SQLStatment` is the parsed `OBJECT` result for SQL input. Hence, `SQLStatment` is supposed to publish elementary functions as possible. On the other hand, `SQLStatmentContext` will carry all the logic handlings for ShardingSphere using `SQLStatment`.

Based on that, some functions like `getValueCountForPerGroup` in `InsertStatment` maybe belongs to `SQLStatmentContext` instead of `SQLStatement`. We can consider improving it later. Plus, not only this one, other `SQLStatment's are also in our consideration. Actually speaking, this is skilled work. Thus  I don't feel comfortable letting others take charge of this task. But if you had an interest in it,  then welcome. :-)

tristaZero (changes requested):

tristaZero (approved):

shardingsphere-sql-parser/shardingsphere-sql-parser-statement/src/main/java/org/apache/shardingsphere/sql/parser/sql/common/util/SQLStatementUtils.java
```diff
@@ -0,0 +1,82 @@
...
+import org.apache.shardingsphere.sql.parser.sql.dialect.statement.sqlserver.SQLServerStatement;
+import org.apache.shardingsphere.sql.parser.sql.dialect.statement.sqlserver.dml.SQLServerInsertStatement;
+
+import java.util.Optional;
+
+/**
+ * SQL statement utility class.
+ */
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public final class SQLStatementUtils {
```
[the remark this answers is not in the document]
strongduanmu: > Is this class just for `InserSQLstatment`? If so, do you think `InsertSQLStatmentUtils` is better?

@tristaZero Yes, I will optimize it. 👍

shardingsphere-sql-parser/shardingsphere-sql-parser-statement/src/main/java/org/apache/shardingsphere/sql/parser/sql/common/statement/dml/InsertStatement.java
strongduanmu: @tristaZero Thank you very much, I think your suggestions are great. 👏 `InsertStatment` really should not be overly complicated, and methods such as `getValueCountForPerGroup` are also usually used in `SQLStatmentContext`, so these methods should be migrated to `SQLStatmentContext`. I will try to refactor `SQLStatement` code to make it more elegant. 😄

PR #7544, opened 2020-09-21
refactor SelectStatement, UpdateStatement to different dialect Statement

strongduanmu:
Ref #7170.

- refactor SelectStatement to different dialect Statement
- refactor UpdateStatement to different dialect Statement


tristaZero (approved):
Hi @strongduanmu 

This PR looks terrific. Not only the reasonable abstract and child class structure but the unit tests for `Helper` are created!

77 files showed your full-of-effort for this PR, a considerable change, excellent job!

Lastly, the name of `xxxHelper` seems not ... rather proper IMO. Considering other class names with similar functions like `extractor`, `handler`, is there any particular reason we import another new class name (or maybe a new concept) for them? Could we pick up one existing name in the project to rename `helper`, like `handler`? 

Since this PR is pretty spotless close and to avoid potential conflicts, I merged it. Next, we can consider a more sutible name for them in the next PR. 

Best,
Trista

strongduanmu:
@tristaZero I will submit another PR later to optimize this naming problem and the method migration in InsertStatement.

PR #7546, opened 2020-09-21
fix problem：Statement setFetchDirection not supported #6661

### LINKED ISSUE #6661: Statement setFetchDirection not supported
## Feature Request
setFetchDirection is not supported and would throw exception
org.apache.shardingsphere.shardingjdbc.jdbc.unsupported.AbstractUnsupportedOperationStatement#setFetchDirection

We are using apache mahout which uses this method in db initialization, but would throw exception.
image

It's better to support this feature, or just make it not final which could be overridden in our codes.
**
image
**

wenweibin:
Fixes #6661 

- **AbstractUnsupportedOperationStatement:** delete `getFetchDirection` and `setFetchDirection` method
- **AbstractStatementAdapter:**  add field `fecthDirection`; add the implementation of `getFetchDirection` and `setFetchDirection` method.
 The setting of the `setFetchDirection` are invalid for many DB drivers,such as MySQL,H2，`getFetchDirection` always return 1000. Instead of returning directly the `getFetchDirection` result of the first underlying Statement, I think adding a field  named `fecthDirection` is a better choice , which ensures that the getter and setter of `fetchDirection`are consistent in  ShardingSphere.
- **CircuitBreakerStatement:** add the implementation of `getFetchDirection` and `setFetchDirection` method.
- **UnsupportedOperationStatementTest:** delete `assertGetFetchDirection` and `assertSetFetchDirection`
- **StatementAdapterTest:** add the `assertSetFetchDirection` and `assertFetchDirection` method
This PR is to solve the problem in issue 6661.

### DIFF
--- a/shardingsphere-jdbc/shardingsphere-jdbc-core/src/main/java/org/apache/shardingsphere/driver/jdbc/adapter/AbstractStatementAdapter.java
+++ b/shardingsphere-jdbc/shardingsphere-jdbc-core/src/main/java/org/apache/shardingsphere/driver/jdbc/adapter/AbstractStatementAdapter.java
@@ -39,6 +39,8 @@ public abstract class AbstractStatementAdapter extends AbstractUnsupportedOperat
     private boolean poolable;
     
     private int fetchSize;
+
+    private int fecthDirection;
     
     private final ForceExecuteTemplate<Statement> forceExecuteTemplate = new ForceExecuteTemplate<>();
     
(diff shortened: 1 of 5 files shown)


terrymanu (changes requested):
Can you fix the CI?

wenweibin:
> Can you fix the CI?

Fixed.

terrymanu (approved):

PR #7583, opened 2020-09-24
ref support of subquery

jingshanglu:
ref #6497 .


tristaZero (changes requested):
Thanks for fixing, waiting for your polish.

tristaZero (approved):

PR #7594, opened 2020-09-25
fix lexer for string

### LINKED ISSUE #7562: Single quote mark is not supported in LIKE clause
5.0.0-RC1-SNAPSHOT, git commit hash: f601d558de

ShardingSphere-Proxy

### Expected behavior
mysql> select id from admin where fullname like "%a";
Empty set (0.00 sec)
mysql> select id from admin where fullname like "%a";
Empty set (0.00 sec)
mysql> select id from admin where fullname like '%a';
Empty set (0.00 sec)
mysql> select id from admin where fullname like 'a%';
Empty set (0.00 sec)

### Actual behavior
mysql> select id from admin where fullname like "%a";
Empty set (0.00 sec)
mysql> select id from admin where fullname like "%a";
Empty set (0.00 sec)
mysql> select id from admin where fullname like '%a';
ERROR 10002 (C1000): 2Unknown exception: [Format specifier '%a']
mysql> select id from admin where fullname like 'a%';
ERROR 10002 (C1000): 2Unknown exception: [Conversion = ''']

It seems like " (double quote mark) is supported in LIKE clause, but ' (single quote mark) is not supported.

jingshanglu:
Fixes #7562.

- fix lexer for string
- change `literals` ref out of `expr` to `expr`
- simple visitor of `expr`

### DIFF
--- a/shardingsphere-sql-parser/shardingsphere-sql-parser-dialect/shardingsphere-sql-parser-mysql/src/main/antlr4/imports/mysql/DDLStatement.g4
+++ b/shardingsphere-sql-parser/shardingsphere-sql-parser-dialect/shardingsphere-sql-parser-mysql/src/main/antlr4/imports/mysql/DDLStatement.g4
@@ -322,7 +322,7 @@ columnDefinition
 storageOption
     : dataTypeGenericOption
     | AUTO_INCREMENT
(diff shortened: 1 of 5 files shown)


tristaZero (approved):
Nice!

PR #7619, opened 2020-09-27
Parse SQLServer insert output clause.

dongzl:
For #6478 .

- Parse SQLServer insert output clause.
- add unit test.


shardingsphere-sql-parser/shardingsphere-sql-parser-test/src/main/java/org/apache/shardingsphere/test/sql/parser/parameterized/asserts/statement/dml/impl/InsertStatementAssert.java
```diff
@@ -59,6 +61,7 @@ public static void assertIs(final SQLCaseAssertContext assertContext, final Inse
         assertInsertSelectClause(assertContext, actual, expected);
         assertOnDuplicateKeyColumns(assertContext, actual, expected);
         assertWithClause(assertContext, actual, expected);
+        assertOutputClause(assertContext, actual, expected);
```
jingshanglu, line 64: Mybe it is better to make assertions based on specific DB.How do you think?

dongzl: > Mybe it is better to make assertions based on specific DB.How do you think?

@jingshanglu , yes, it's a very good idea, but I suggest we can do it in another PR, do you think it?

jingshanglu: @dongzl @tristaZero @strongduanmu    We should discuss aboult parser test framework,one:first determin db type and then assert; two:assert all segment include in all db in `InsertStatementAssert.java`,and then decide whether to assert or not according to DB type.What do you think?

strongduanmu: @jingshanglu @tristaZero @dongzl Now, the refactor task of `Statement` just splits different dialect statements, `StatementContext` and `StatementAssert` are not splitted. When `Statement` is used by `StatemenContext` and `StatementAssert`, different dialect statements are processed uniformly through the `StatementHandler` now.

I was wondering if we need to split `StatemenContext` and `StatementAssert`, since `StatementHandler` seems to handle logic well already.

jingshanglu: > @jingshanglu @tristaZero @dongzl Now, the refactor task of `Statement` just splits different dialect statements, `StatementContext` and `StatementAssert` are not splitted. When `Statement` is used by `StatemenContext` and `StatementAssert`, different dialect statements are processed uniformly through the `StatementHandler` now.
> 
> I was wondering if we need to split `StatemenContext` and `StatementAssert`, since `StatementHandler` seems to handle logic well already.

you mean the two way? decide whether to assert or not according to DB type in  specific function, like some segment  only contained by SQLServer.

dongzl: Hi @jingshanglu , @strongduanmu  I suggest we open a new issue to discuss this problem, This is a big code refactor job, we need discuss it with more people. :)

jingshanglu (approved):

strongduanmu (approved):

PR #7725, opened 2020-10-09
Fix mysql binlog type decode

avalon5666:


Lucas-307 (approved):

shardingsphere-db-protocol/shardingsphere-db-protocol-mysql/src/main/java/org/apache/shardingsphere/db/protocol/mysql/packet/binlog/row/column/value/string/MySQLStringBinlogProtocolValue.java
```diff
@@ -31,18 +31,32 @@
...
+        }
+        switch (MySQLColumnType.valueOf(type)) {
             case MYSQL_TYPE_ENUM:
-                return readEnumValue(columnDef.getColumnMeta() & 0xff, payload);
+                return readEnumValue(length, payload);
             case MYSQL_TYPE_SET:
                 return payload.getByteBuf().readByte();
             case MYSQL_TYPE_STRING:
-                return payload.readStringFix(payload.getByteBuf().readUnsignedByte());
+                return payload.readStringFix(readActualLength(payload, length));
```
Lucas-307, line 47: please make sure args is in the same order.

Lucas-307 (changes requested):

Lucas-307 (approved):

terrymanu (approved):

PR #7776, opened 2020-10-13
fix oracle table name must to upper

yu199195:

### DIFF
--- a/shardingsphere-infra/shardingsphere-infra-common/src/main/java/org/apache/shardingsphere/infra/metadata/model/physical/jdbc/handler/TableNamePatternHandler.java
+++ b/shardingsphere-infra/shardingsphere-infra-common/src/main/java/org/apache/shardingsphere/infra/metadata/model/physical/jdbc/handler/TableNamePatternHandler.java
@@ -25,10 +25,10 @@ import org.apache.shardingsphere.infra.spi.typed.TypedSPI;
 public interface TableNamePatternHandler extends TypedSPI {
     
     /**
-     * decorate table name pattern.
+     * Decorate table name pattern.
      * 
      * @param tableNamePattern table name pattern
-     * @return decorate of table name pattern
+     * @return decorated table name pattern
      */
     String decorate(String tableNamePattern);
 }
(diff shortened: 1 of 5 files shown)


shardingsphere-infra/shardingsphere-infra-common/src/main/java/org/apache/shardingsphere/infra/metadata/model/physical/jdbc/handler/TableNamePatternHandler.java
```diff
@@ -0,0 +1,34 @@
...
+
+import org.apache.shardingsphere.infra.spi.typed.TypedSPI;
+
+/**
+ * Table name pattern handler.
+ */
+public interface TableNamePatternHandler extends TypedSPI {
+    
+    /**
+     * decorate table name pattern.
```
terrymanu, line 28: Please upper case for first letter of java doc

shardingsphere-infra/shardingsphere-infra-common/src/main/java/org/apache/shardingsphere/infra/metadata/model/physical/jdbc/handler/TableNamePatternHandler.java
```diff
@@ -0,0 +1,34 @@
...
+     * 
+     * @param tableNamePattern table name pattern
+     * @return decorate of table name pattern
```
terrymanu, line 31: The java doc `decorate of table name pattern` is not good sentence,
It is better to revise it as `decorated table name pattern`

terrymanu (changes requested):

shardingsphere-infra/shardingsphere-infra-common/src/main/java/org/apache/shardingsphere/infra/metadata/model/physical/jdbc/handler/TableNamePatternHandler.java
yu199195: modify already

shardingsphere-infra/shardingsphere-infra-common/src/main/java/org/apache/shardingsphere/infra/metadata/model/physical/jdbc/handler/TableNamePatternHandler.java
yu199195: modify already

terrymanu (approved):

PR #7793, opened 2020-10-14
refactor database dialect get delimiter

yu199195:


terrymanu (changes requested):
The delimiter class has already existed in parser module, can we consider about reuse it?

Issue #7731, opened 2020-10-10
How to configure no sharding schema in 5.0.0-RC1


lwtdev:
## How to configure no sharding schema in 5.0.0-RC1 ?

In 4.1.1, my configuration is like this:

```yaml
schemaName: spsqltest_nosharding
#
dataSources:
  ds_0:
    url: jdbc:mysql://127.0.0.1:3306/spsqltest_nosharding?serverTimezone=UTC&useSSL=false
    username: root
    password: root135
    connectionTimeoutMilliseconds: 30000
    idleTimeoutMilliseconds: 60000
    maxLifetimeMilliseconds: 1800000
    maxPoolSize: 50
    
shardingRule:
  defaultDataSourceName: ds_0
  defaultDatabaseStrategy:
    none:
  defaultTableStrategy:
    none:
```

In 5.0.0-RC1 :

```yaml
schemaName: spsqltest_nosharding
#
dataSourceCommon:
  username: root
  password: root135
  connectionTimeoutMilliseconds: 30000
  idleTimeoutMilliseconds: 60000
  maxLifetimeMilliseconds: 1800000
  maxPoolSize: 50
  minPoolSize: 1
  maintenanceIntervalMilliseconds: 30000
#
dataSources:
  ds_00:
    url: jdbc:mysql://127.0.0.1:3306/spsqltest_nosharding?serverTimezone=UTC&useSSL=false

#
rules:
- !SHARDING
  defaultDatabaseStrategy:
    none:
  defaultTableStrategy:
    none:
```

But the effect is different for some SQL, such as the following:
```sql
ShardingProxy(4.1.1)No Sharding
dropTable[hasShardingKey:none];	 Support:true;	SQL: drop table customer;
ds_0|drop table customer|


ShardingProxy(5.0.0.RC1)No Sharding
dropTable[hasShardingKey:none];	 Support:false;	SQL: drop table customer;
com.mysql.jdbc.exceptions.jdbc4.MySQLSyntaxErrorException: This version of ShardingProxy doesn't yet support this SQL. 'Cannot find table rule with logic table: 'customer''
```

tristaZero:
@lwtdev Every time I see your issue, I guess there may be a bug likely. :-) Believe me, you are definitely a good QA engineer. 😃 
Back to this issue, `no sharding schema` you mentioned means `defaultDataSourceName: ds_0`, doesn't it?
[...]

PR #7796, opened 2020-10-15
fix multi-table validation error for DML modify statements

strongduanmu:
Ref #7731.

- optimize the validator check prompt message
- fix multi-table validation error for DML modify statements


shardingsphere-features/shardingsphere-sharding/shardingsphere-sharding-route/src/main/java/org/apache/shardingsphere/sharding/route/engine/validator/dml/ShardingDMLStatementValidator.java
```diff
@@ -31,11 +34,14 @@
...
      * Validate multiple table.
      *
+     * @param shardingRule sharding rule
      * @param sqlStatementContext sqlStatementContext
+     * @param sql sql
      */
-    protected void validateMultipleTable(final SQLStatementContext<T> sqlStatementContext) {
-        if (1 != ((TableAvailable) sqlStatementContext).getAllTables().size()) {
-            throw new ShardingSphereException("Cannot support Multiple-Table for '%s'.", sqlStatementContext.getSqlStatement());
+    protected void validateMultipleTable(final ShardingRule shardingRule, final SQLStatementContext<T> sqlStatementContext, final String sql) {
```
tristaZero, line 41: 1. From my perspective, It is unnecessary to have SQL as an input parameter. Precisely speaking, It is a little heavy to pass SQL just for an exception. Maybe `Cannot support such a Multiple-Table query for '$tableNames'` is enough. What do you think?
2. The constrictions for multi-tables looks like incomplete? How about broadcast tables or the single table in the same instance with sharding tables (Maybe a postValidate()?).

tristaZero (changes requested):

shardingsphere-features/shardingsphere-sharding/shardingsphere-sharding-route/src/main/java/org/apache/shardingsphere/sharding/route/engine/validator/dml/ShardingDMLStatementValidator.java
strongduanmu: > From my perspective, It is unnecessary to have SQL as an input parameter. Precisely speaking, It is a little heavy to pass SQL just for an exception. Maybe Cannot support such a Multiple-Table query for '$tableNames' is enough. What do you think?

@tristaZero I agree with you that the inclusion of SQL in the exception is indeed a bit verbose, especially when the SQL is very complex.

strongduanmu: > The constrictions for multi-tables looks like incomplete? How about broadcast tables or the single table in the same instance with sharding tables (Maybe a postValidate()?).

@tristaZero I'm very sorry, I will support the check of broadcast tables and single tables. 
In addition, I think these check should be placed in preValidate, because they should be executed before routing. 🤔

shardingsphere-features/shardingsphere-sharding/shardingsphere-sharding-route/src/main/java/org/apache/shardingsphere/sharding/route/engine/validator/dml/ShardingDMLStatementValidator.java
```diff
@@ -18,24 +18,38 @@
...
      * @param sqlStatementContext sqlStatementContext
      */
-    protected void validateMultipleTable(final SQLStatementContext<T> sqlStatementContext) {
-        if (1 != ((TableAvailable) sqlStatementContext).getAllTables().size()) {
-            throw new ShardingSphereException("Cannot support Multiple-Table for '%s'.", sqlStatementContext.getSqlStatement());
+    protected void validateShardingMultipleTable(final ShardingRule shardingRule, final SQLStatementContext<T> sqlStatementContext) {
+        Collection<String> tableNames = sqlStatementContext.getTablesContext().getTableNames();
+        Collection<String> shardingTableNames = shardingRule.getShardingLogicTableNames(tableNames);
+        if ((1 == shardingTableNames.size() || shardingRule.isAllBindingTables(shardingTableNames)) && !isAllValidTables(shardingRule, tableNames)) {
+            throw new ShardingSphereException("Cannot support Multiple-Table for '%s'.", tableNames);
```
tristaZero, line 44: What's the string output of `tableNames`? Do you think that will be `Cannot support Multiple-Table for 'tb1tb2tb3'`?

shardingsphere-features/shardingsphere-sharding/shardingsphere-sharding-route/src/main/java/org/apache/shardingsphere/sharding/route/engine/validator/dml/ShardingDMLStatementValidator.java
```diff
@@ -18,24 +18,38 @@
...
+     * @param shardingRule sharding rule
      * @param sqlStatementContext sqlStatementContext
      */
-    protected void validateMultipleTable(final SQLStatementContext<T> sqlStatementContext) {
-        if (1 != ((TableAvailable) sqlStatementContext).getAllTables().size()) {
-            throw new ShardingSphereException("Cannot support Multiple-Table for '%s'.", sqlStatementContext.getSqlStatement());
+    protected void validateShardingMultipleTable(final ShardingRule shardingRule, final SQLStatementContext<T> sqlStatementContext) {
+        Collection<String> tableNames = sqlStatementContext.getTablesContext().getTableNames();
+        Collection<String> shardingTableNames = shardingRule.getShardingLogicTableNames(tableNames);
+        if ((1 == shardingTableNames.size() || shardingRule.isAllBindingTables(shardingTableNames)) && !isAllValidTables(shardingRule, tableNames)) {
```
tristaZero, line 43: Do you think `if ( !isAllValidTables(shardingRule, tableNames))` is enough?

tristaZero (changes requested):
Thanks @strongduanmu , Please look at the comments when you are available. :-)

shardingsphere-features/shardingsphere-sharding/shardingsphere-sharding-route/src/main/java/org/apache/shardingsphere/sharding/route/engine/validator/dml/ShardingDMLStatementValidator.java
strongduanmu: > What's the string output of `tableNames`? Do you think that will be `Cannot support Multiple-Table for 'tb1tb2tb3'`?

@tristaZero The error message may like Cannot support Multiple-Table for '[user, order, order_item]'.

shardingsphere-features/shardingsphere-sharding/shardingsphere-sharding-route/src/main/java/org/apache/shardingsphere/sharding/route/engine/validator/dml/ShardingDMLStatementValidator.java
strongduanmu: > Do you think `if ( !isAllValidTables(shardingRule, tableNames))` is enough?

@tristaZero In order to prevent it from affecting all the `RoutineEngine`, temporarily use the condition `(1 == shardingTableNames.size() || shardingRule.isAllBindingTables(shardingTableNames))` to ensure that only the ShardingStandardRoutingEngine is checked. 
After I investigate the logic of all `RouteEngine`, I will optimize this logic in the future.

tristaZero (approved):

PR #7926, opened 2020-10-27
Update committer.en.md

yang-7777:
add committer routine


tristaZero (changes requested):
More details are needed.

tristaZero (approved):

PR #7957, opened 2020-10-29
minor refactoring for parser api

kimmking:
minor refactoring for parser api

### DIFF
--- a/shardingsphere-sql-parser/shardingsphere-sql-parser-engine/src/main/java/org/apache/shardingsphere/sql/parser/core/visitor/SQLVisitorFactory.java
+++ b/shardingsphere-sql-parser/shardingsphere-sql-parser-engine/src/main/java/org/apache/shardingsphere/sql/parser/core/visitor/SQLVisitorFactory.java
@@ -62,7 +62,7 @@ public final class SQLVisitorFactory {
             case RL:
                 return (ParseTreeVisitor) visitorFacade.getRLVisitorClass().getConstructor().newInstance();
             default:
-                throw new SQLParsingException("Can not support SQL statement type: `%s`", type);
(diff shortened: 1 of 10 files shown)


shardingsphere-sql-parser/shardingsphere-sql-parser-engine/src/main/java/org/apache/shardingsphere/sql/parser/core/visitor/SQLVisitorFactory.java
```diff
@@ -62,7 +62,7 @@
             case RL:
                 return (ParseTreeVisitor) visitorFacade.getRLVisitorClass().getConstructor().newInstance();
             default:
-                throw new SQLParsingException("Can not support SQL statement type: `%s`", type);
+                throw new SQLParsingException(String.format("Can not support SQL statement type: `%s`", type));
```
terrymanu, line 65: The SQLParsingException has already use String.format inside, it is unnecessary format string outside constructor

shardingsphere-sql-parser/shardingsphere-sql-parser-engine/src/main/java/org/apache/shardingsphere/sql/parser/exception/SQLParsingException.java
```diff
@@ -24,7 +24,7 @@
     
     private static final long serialVersionUID = -6408790652103666096L;
     
-    public SQLParsingException(final String message, final Object... args) {
-        super(String.format(message, args));
+    public SQLParsingException(final String message) {
```
terrymanu, line 27: Please revert this changes

terrymanu (changes requested):

terrymanu (approved):

PR #8022, opened 2020-11-03
update license files for proxy

kimmking:
add three license files for proxy


terrymanu (changes requested):
Please consider about dependencies' dependencies

terrymanu (changes requested):
It is better to leave apollo and nacos as provided scope because there are too many dependencies.

terrymanu (approved):

PR #8024, opened 2020-11-04
remove dbcp with mock in CenterConfigTest

kimmking:
remove dbcp with mock in CenterConfigTest


terrymanu (changes requested):
There is existed class `MockedDataSource`, could we consider about reuse it first?
If cannot, we should use same class name and content for same requirement.

terrymanu (approved):

PR #8052, opened 2020-11-06
Fix mysql grammar

jingshanglu:
### Background
SQL support is not perfect, such as `join table`,`alter table`, because there are some differences between the official document and the actual grammar.

### What is the purpose of the change
According to MySQL source code to improve the syntax of `Join table` and `alterTableStatement` and `setStatement`

### Changes in this pull request:
- improve support of `Join table`
- improve support of `alterTableStatement`
- improve support of `setStatement`
- Modify the wrong SQL test case


shardingsphere-sql-parser/shardingsphere-sql-parser-dialect/shardingsphere-sql-parser-mysql/src/main/java/org/apache/shardingsphere/sql/parser/mysql/visitor/format/impl/MySQLFormatSQLVisitor.java
```diff
@@ -19,11 +19,14 @@
 
 import lombok.AccessLevel;
 import lombok.Getter;
+import lombok.Setter;
```
tristaZero, line 22: This new function needs another PR, IMO.

tristaZero (changes requested):

shardingsphere-sql-parser/shardingsphere-sql-parser-dialect/shardingsphere-sql-parser-mysql/src/main/java/org/apache/shardingsphere/sql/parser/mysql/visitor/format/impl/MySQLFormatSQLVisitor.java
jingshanglu: Yes. @tristaZero

tristaZero (approved):

PR #8105, opened 2020-11-09
support subquery route without sharding column

strongduanmu:
- support subquery route without sharding column (check sharding condition only when the subquery and outer query all contain sharding columns)
- support range sharding condition value

### DIFF
--- a/shardingsphere-features/shardingsphere-sharding/shardingsphere-sharding-route/src/test/java/org/apache/shardingsphere/sharding/route/engine/type/standard/SubqueryRouteTest.java
+++ b/shardingsphere-features/shardingsphere-sharding/shardingsphere-sharding-route/src/test/java/org/apache/shardingsphere/sharding/route/engine/type/standard/SubqueryRouteTest.java
@@ -149,16 +149,6 @@ public final class SubqueryRouteTest extends AbstractSQLRouteTest {
         assertRoute(sql, parameters);
     }
     
-    @Test(expected = IllegalStateException.class)
-    public void assertSubqueryWithoutHint() {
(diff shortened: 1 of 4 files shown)


shardingsphere-features/shardingsphere-sharding/shardingsphere-sharding-route/src/test/java/org/apache/shardingsphere/sharding/route/engine/type/standard/SubqueryRouteTest.java
```diff
@@ -149,16 +149,6 @@ public void assertSubqueryForBinding() {
...
-    @Test(expected = IllegalStateException.class)
-    public void assertSubqueryWithoutHint() {
-        List<Object> parameters = new LinkedList<>();
-        parameters.add(1);
-        parameters.add(2);
-        parameters.add(5);
-        String sql = "select count(*) from t_hint_test where user_id = (select t_hint_test from t_hint_test where user_id in (?,?,?)) ";
-        assertRoute(sql, parameters);
-    }
-
```
jingshanglu, line 161: @strongduanmu Why delete this UT?

strongduanmu: > @strongduanmu Why delete this UT?

@jingshanglu After we support sub query route without sharding key, this exception case will not exist.

jingshanglu (approved):

PR #8116, opened 2020-11-10
Add sql formart visitor

jingshanglu:
- add sql formart visitor to support sql formart output

### DIFF
--- a/shardingsphere-sql-parser/shardingsphere-sql-parser-dialect/shardingsphere-sql-parser-mysql/src/main/java/org/apache/shardingsphere/sql/parser/mysql/visitor/format/impl/MySQLFormatSQLVisitor.java
+++ b/shardingsphere-sql-parser/shardingsphere-sql-parser-dialect/shardingsphere-sql-parser-mysql/src/main/java/org/apache/shardingsphere/sql/parser/mysql/visitor/format/impl/MySQLFormatSQLVisitor.java
@@ -30,7 +30,6 @@ import org.apache.shardingsphere.sql.parser.autogen.MySQLStatementParser;
  */
 @Getter(AccessLevel.PROTECTED)
 public abstract class MySQLFormatSQLVisitor extends MySQLStatementBaseVisitor<String> {
-    @Getter
     private StringBuilder result = new StringBuilder();
 
     private final boolean uperCase = true;
@@ -38,8 +37,6 @@ public abstract class MySQLFormatSQLVisitor extends MySQLStatementBaseVisitor<St
     private int indentCount;
 
     private int lines;
-
-    private int rowConstructorNumberOfLine = 5;
 
     private int projectionsCountOfLine = 3;
 
@@ -226,12 +223,12 @@ public abstract class MySQLFormatSQLVisitor extends MySQLStatementBaseVisitor<St
     public String visitExpr(final MySQLStatementParser.ExprContext ctx) {
         if (null != ctx.logicalOperator()) {
             MySQLStatementParser.ExprContext left = ctx.expr(0);
-            MySQLStatementParser.ExprContext right = ctx.expr(1);
             visit(left);
             println();
+            MySQLStatementParser.ExprContext right = ctx.expr(1);
             print(ctx.logicalOperator().getText());
             visit(right);
-        } else if (null != ctx.notOperator()){
+        } else if (null != ctx.notOperator()) {
             print(ctx.notOperator().getText());
             visit(ctx.expr(0));
         } else {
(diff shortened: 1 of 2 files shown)


shardingsphere-sql-parser/shardingsphere-sql-parser-dialect/shardingsphere-sql-parser-mysql/src/test/java/org/apache/shardingsphere/sql/parser/mysql/MySQLFormartTest.java
```diff
@@ -0,0 +1,91 @@
...
+                + "\tandname = ?\n"
+                + "\tanda = ?\n"
+                + "\tandb = ?\n"
+                + "\tandc = ?\n"
+                + "\tandd = ?;"});
+        testUnits.add(new String[]{"", "ALTER TABLE t_log ADD name varchar(10)", "ALTER TABLE t_log ADD name VARCHAR ( 10 )"});
+    }
+
+    @Test
+    public final void assertTest() {
```
tristaZero, line 80: This name seems to be vague for this test assertion

shardingsphere-sql-parser/shardingsphere-sql-parser-dialect/shardingsphere-sql-parser-mysql/src/test/java/org/apache/shardingsphere/sql/parser/mysql/MySQLFormartTest.java
```diff
@@ -0,0 +1,91 @@
...
+        for (String[] each : testUnits) {
+            CodePointBuffer buffer = CodePointBuffer.withChars(CharBuffer.wrap(each[1].toCharArray()));
+            MySQLLexer lexer = new MySQLLexer(CodePointCharStream.fromBuffer(buffer));
+            MySQLParser parser = new MySQLParser(new CommonTokenStream(lexer));
+            ParseTree tree = ((ParseASTNode) parser.parse()).getRootNode();
+            MySQLFormatSQLVisitor visitor = new MySQLDMLFormatSQLVisitor();
+            String result = visitor.visit(tree);
+            assertTrue("SQL format error", each[2].equals(result));
```
tristaZero, line 88: So this test...just intends to get an error assertion result? How about getting a standard and correct assertion?

tristaZero (changes requested):

shardingsphere-sql-parser/shardingsphere-sql-parser-dialect/shardingsphere-sql-parser-mysql/src/main/java/org/apache/shardingsphere/sql/parser/mysql/visitor/format/impl/MySQLFormatSQLVisitor.java
```diff
@@ -19,11 +19,423 @@
...
+import org.antlr.v4.runtime.tree.TerminalNode;
 import org.apache.shardingsphere.sql.parser.autogen.MySQLStatementBaseVisitor;
+import org.apache.shardingsphere.sql.parser.autogen.MySQLStatementParser;
 
 /**
  * MySQL Format SQL visitor for MySQL.
  */
 @Getter(AccessLevel.PROTECTED)
 public abstract class MySQLFormatSQLVisitor extends MySQLStatementBaseVisitor<String> {
+    private StringBuilder result = new StringBuilder();
```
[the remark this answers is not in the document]
jingshanglu: First, `FormartVisitor` is different from `StatementVisitors`, `StatementVisitors` need create different type `Segment`,  `FormartVisitor` just need `StringBuilder`, so using shared variables can reduce the cost of creating objects.
Second, there is no need to processe the return value of child elements when using shared variables
 ,just call visit(`childrule`), 
When a visitor function is added, we do not need to modify  its parent function.ItWill reduce the amount of code, otherwise, you need to create a visitor function for each rule. @tristaZero

shardingsphere-sql-parser/shardingsphere-sql-parser-dialect/shardingsphere-sql-parser-mysql/src/main/java/org/apache/shardingsphere/sql/parser/mysql/visitor/format/impl/MySQLFormatSQLVisitor.java
```diff
@@ -19,11 +19,423 @@
...
+
+            ParseTree c = node.getChild(i);
+            String childResult = c.accept(this);
+            result = aggregateResult(result, childResult);
+        }
+        return result;
+    }
+
+    private void print(final char value) {
+        if (this.result == null) {
```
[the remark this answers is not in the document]
jingshanglu: OK

shardingsphere-sql-parser/shardingsphere-sql-parser-dialect/shardingsphere-sql-parser-mysql/src/test/java/org/apache/shardingsphere/sql/parser/mysql/MySQLFormartTest.java
jingshanglu: Is UT to verify whether the actual output is consistent with the expected output? @tristaZero

tristaZero (approved):
Great progress. :)

PR #8325, opened 2020-11-24
Update vote.en.md and icia.en.md

HoldDie:
Update vote.en.md and icia.en.md.


tristaZero (approved):
Hi @HoldDie , Thanks for fixing these typo. :-)

PR #8422, opened 2020-11-30
Overwrite InventoryDataTaskSplitter.splitByPrimaryKeyRange

Lucas-307:


shardingsphere-scaling/shardingsphere-scaling-core/src/main/java/org/apache/shardingsphere/scaling/core/execute/executor/sqlbuilder/SQLBuilder.java
```diff
@@ -0,0 +1,72 @@
...
+
+import org.apache.shardingsphere.scaling.core.execute.executor.record.Column;
+import org.apache.shardingsphere.scaling.core.execute.executor.record.DataRecord;
+
+import java.util.Collection;
+
+/**
+ * SQL builder.
+ */
+public interface SQLBuilder {
```
terrymanu, line 28: Please rename this class because of  SQLBuilder is too generic and may conflict with SQL rewrite module.

terrymanu (changes requested):

terrymanu (approved):

PR #8526, opened 2020-12-08
Enhancement plugin class loader

dmsolr:


shardingsphere-agent/shardingsphere-agent-core/src/main/java/org/apache/shardingsphere/agent/core/plugin/AgentPluginLoader.java
```diff
@@ -191,7 +219,7 @@ public PluginAdviceDefinition loadPluginAdviceDefine(final TypeDescription typeD
      * To get or create instance of the advice class. Create new one and caching when it is not exist.
      *
      * @param classNameOfAdvice class name of advice
-     * @param <T> advice type
+     * @param <T>               advice type
```
yu199195, line 222: remove this

yu199195 (approved):

PR #8546, opened 2020-12-09
make parameterized output of SQL is configurable

jingshanglu:

### DIFF
--- a/shardingsphere-sql-parser/shardingsphere-sql-parser-dialect/shardingsphere-sql-parser-mysql/src/main/java/org/apache/shardingsphere/sql/parser/mysql/visitor/format/impl/MySQLFormatSQLVisitor.java
+++ b/shardingsphere-sql-parser/shardingsphere-sql-parser-dialect/shardingsphere-sql-parser-mysql/src/main/java/org/apache/shardingsphere/sql/parser/mysql/visitor/format/impl/MySQLFormatSQLVisitor.java
@@ -17,8 +17,8 @@
 
 package org.apache.shardingsphere.sql.parser.mysql.visitor.format.impl;
 
-import lombok.AccessLevel;
 import lombok.Getter;
+import lombok.Setter;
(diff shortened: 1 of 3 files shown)


shardingsphere-sql-parser/shardingsphere-sql-parser-dialect/shardingsphere-sql-parser-mysql/src/main/java/org/apache/shardingsphere/sql/parser/mysql/visitor/format/impl/MySQLFormatSQLVisitor.java
```diff
@@ -52,22 +52,27 @@
...
  */
-@Getter(AccessLevel.PROTECTED)
+@Getter
+@Setter
 public abstract class MySQLFormatSQLVisitor extends MySQLStatementBaseVisitor<String> {
 
     private StringBuilder result = new StringBuilder();
 
-    private final boolean uperCase = true;
+    private boolean uperCase = true;
```
tristaZero, line 73: Why do you remove `final` here?

tristaZero (changes requested):

shardingsphere-sql-parser/shardingsphere-sql-parser-dialect/shardingsphere-sql-parser-mysql/src/main/java/org/apache/shardingsphere/sql/parser/mysql/visitor/format/impl/MySQLFormatSQLVisitor.java
jingshanglu: @tristaZero It needs to be configured from the outside.

tristaZero (approved):

PR #8587, opened 2020-12-11
Add ShardingAlgorithmType enum to replace the sharing algorithm type with magic string. 

likly:
Add ShardingAlgorithmType enum to replace the sharing algorithm type with magic string.

### DIFF
--- a/shardingsphere-features/shardingsphere-sharding/shardingsphere-sharding-common/src/main/java/org/apache/shardingsphere/sharding/algorithm/sharding/ShardingAlgorithmType.java
+++ b/shardingsphere-features/shardingsphere-sharding/shardingsphere-sharding-common/src/main/java/org/apache/shardingsphere/sharding/algorithm/sharding/ShardingAlgorithmType.java
@@ -1,0 +1,86 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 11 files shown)


terrymanu (changes requested):
Sharding algorithms are dynamic and loaded via SPI.
The enum is static config and broken OCP.
So it is better to do not use enum to define sharding algorithm types.

likly:
This enum only be used in build-in sharding algorithms, just for easy manage the SPI type of build-in sharding algorithms.
The user also can define custom sharidng algotithm with magic string or custom enum.

likly:
> Sharding algorithms are dynamic and loaded via SPI.
> The enum is static config and broken OCP.
> So it is better to do not use enum to define sharding algorithm types.

This Enum only used for build-in sharding algorithms to manage the SPI types easily.
And this enum is not broken OCP,  Sharding algorithms still are dynamic and loaded via SPI.

terrymanu:
> > Sharding algorithms are dynamic and loaded via SPI.
> > The enum is static config and broken OCP.
> > So it is better to do not use enum to define sharding algorithm types.
> 
> This Enum only used for build-in sharding algorithms to manage the SPI types easily.
> And this enum is not broken OCP, Sharding algorithms still are dynamic and loaded via SPI.

If user custom a new sharding algorithm, do they need to update the enum? 
If yes, OCP is broken; 
If not the custom sharding algorithm is different with built-in one, we'd like keep the customization sharding algorithm consist with built-in one.

likly:
> > > Sharding algorithms are dynamic and loaded via SPI.
> > > The enum is static config and broken OCP.
> > > So it is better to do not use enum to define sharding algorithm types.
> > 
> > 
> > This Enum only used for build-in sharding algorithms to manage the SPI types easily.
> > And this enum is not broken OCP, Sharding algorithms still are dynamic and loaded via SPI.
> 
> If user custom a new sharding algorithm, do they need to update the enum?
> If yes, OCP is broken;
> If not the custom sharding algorithm is different with built-in one, we'd like keep the customization sharding algorithm consist with built-in one.

Broken the OCI maybe appecd when the  `getType()` method  return type change to a enum type.
But this enum just for replace the magic string with `enum.name()`.

The getType method just changed like this:

```java
    public String getType() {
        // old: return "INLINE";
        // now: return the name of `INLINE` enum constant.
        return ShardingAlgorithmType.INLINE.name();
    }
```

The method definition is not changed.

likly:
The user also can find how many built-in algorithms quickly throught the enum of ShardingAlgorithmType.

terrymanu:
we'd like keep the customization sharding algorithm consist with built-in one.

terrymanu:
May I close this pr?

PR #8657, opened 2020-12-16
Issue#7942

### LINKED ISSUE #7942: Configuration error in test-engine/performance-test page
There is still use old configuration in this page, we need to update them to new version

huanghao495430759:
Fixes #7942 .
-fix configuration error in test-engine/performance-test page.
-fix configuration error in test-engine/rewrite-test-engine page.


tristaZero (approved):
Thanks for fixing. Doc is first step for users!

PR #8693, opened 2020-12-20
Fix parseDateTime error when sharding value is Date or LocalDateTime

### LINKED ISSUE #8470: DateTimeParseException
## Bug Report

**Exception**
Caused by: java.time.format.DateTimeParseException: Text '2020-03-15 09:51:17.681' could not be parsed, unparsed text found at index 19
	at java.time.format.DateTimeFormatter.parseResolved0(DateTimeFormatter.java:1952)
	at java.time.format.DateTimeFormatter.parse(DateTimeFormatter.java:1851)
	at java.time.LocalDateTime.parse(LocalDateTime.java:492)

**Location**
org.apache.shardingsphere.sharding.algorithm.sharding.datetime.AutoIntervalShardingAlgorithm#parseDate

**Reason**
The parameter type of this method may be java.sql.Timestamp, directly calling the toString method will get milliseconds (like 2020-12-01 01:01:55.601), LocalDateTime uses yyyy-MM-dd HH:mm:ss, this Exceptions will be thrown.

klniu:
Fix #8470

- There is not need to import  com.google.common.base.Function, and it can't compile after importing.
- The second commit is fixing the parsing of dateTime. when the sharding value is LocalDateTime or Date, toString() will give the string with the format as yyyy-MM-dd'T'HH:mm:ss. and it can not be parsed when user already set the different date format in configuation.

### DIFF
--- a/shardingsphere-features/shardingsphere-sharding/shardingsphere-sharding-common/src/main/java/org/apache/shardingsphere/sharding/algorithm/sharding/datetime/AutoIntervalShardingAlgorithm.java
+++ b/shardingsphere-features/shardingsphere-sharding/shardingsphere-sharding-common/src/main/java/org/apache/shardingsphere/sharding/algorithm/sharding/datetime/AutoIntervalShardingAlgorithm.java
@@ -33,7 +33,11 @@ import java.time.LocalDateTime;
 import java.time.ZoneId;
 import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeParseException;
-import java.util.*;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Date;
+import java.util.LinkedHashSet;
+import java.util.Properties;
 
 /**
  * Auto interval sharding algorithm.
--- a/shardingsphere-features/shardingsphere-sharding/shardingsphere-sharding-common/src/main/java/org/apache/shardingsphere/sharding/algorithm/sharding/datetime/AutoIntervalShardingAlgorithm.java
+++ b/shardingsphere-features/shardingsphere-sharding/shardingsphere-sharding-common/src/main/java/org/apache/shardingsphere/sharding/algorithm/sharding/datetime/AutoIntervalShardingAlgorithm.java
@@ -30,12 +30,10 @@ import org.apache.shardingsphere.sharding.api.sharding.standard.StandardSharding
 import java.text.DecimalFormat;
 import java.time.Duration;
 import java.time.LocalDateTime;
+import java.time.ZoneId;
 import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeParseException;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.LinkedHashSet;
-import java.util.Properties;
+import java.util.*;
 
 /**
  * Auto interval sharding algorithm.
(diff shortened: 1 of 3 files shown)


tristaZero:
Hi @klniu ,

Thanks for your PR, could you firstly look at the CI test and fix them?

shardingsphere-features/shardingsphere-sharding/shardingsphere-sharding-common/src/main/java/org/apache/shardingsphere/sharding/algorithm/sharding/datetime/AutoIntervalShardingAlgorithm.java
```diff
@@ -124,7 +126,16 @@ private int getLastPartition(final Range<Comparable<?>> valueRange) {
     }
     
     private long parseDate(final Comparable<?> shardingValue) {
-        LocalDateTime dateValue = LocalDateTime.parse(shardingValue.toString(), DATE_TIME_FORMAT);
+        LocalDateTime dateValue;
```
yu199195, line 129: Thank you very much for your contribution. 

Now I would like to know when shardingValue is Date or LocalDateTime?

We can't judge the treatment so simply,  because we define DateTimeFormatter like this 

```
DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss")
```

if you data or config not follow this, An exception is thrown, so you need to follow this DateTimeFormatter .

klniu: It happened when the type of the sharding value is LocalDateTime. Here is the configuration:

```
          dt_partition_detection:
            actual-data-nodes: dsd$->{0..9}.dt_partition_detection_$->{2013..2025}0$->{1..4}
            database-strategy:
              standard:
                sharding-column: sys_org_code
                sharding-algorithmName: detection-db-sharding
            tableStrategy:
              standard:
                sharding-column: time
                sharding-algorithm-name: detection-table-sharding
            key-generate-strategy:
              type: SNOWFLAKE
              column: id

...
            detection-table-sharding:
              type: INTERVAL
              props:
                datetime-pattern: yyyy-MM-dd HH:mm:ss
                datetime-lower: '2013-01-01 00:00:00'
                sharding-suffix-pattern: yyyyQQ
                datetime-interval-amount: '3'
                datetime-interval-unit: MONTHS
```

Using INTERVAL sharding rule，when insert a entity, whose field **time** is LocalDateTime. 

After PreparedStatement.execute(); it will call doSharding using the arguments passed to PreparedStatement.

**time** is not called with the toString() method, but directly pass the raw value. Like the image showing:

image

yu199195: can you config  datetime-pattern: yyyy-MM-ddTHH:mm:ss ?

klniu: Thanks.

After I configured the config as you said, it is normal, but the value passed to doSharding is still LocalDateTime, The configuration just make the formatter yyyy-MM-ddTHH:mm:ss the same as LocalDateTime.toString() using. So the different configuaration is not respected.

image

yu199195: so this pr ,we can not merged, Do you have any other fix,  if not can you close this pr?

Issue #8060, opened 2020-11-06
Design global lock for registry in multiple proxies case


kimmking:
Design global lock for registry in multiple proxies case

Task list:

**1. Implement global asynchronous lock**
  - [x] 1.1. Initialize the distributed lock module in the governance project(#8320)
  - [x] 1.2. Realize global lock for zookeeper(#8342)
  - [x] 1.3. Design and implement proxy state queue(#8358)
  - [x] 1.4. Implement proxy to monitor global lock and  lock itsself(#8358)
  - [x] 1.5. Achieve unlock after the proxy completes metadata synchronization(#8472 )
  - [x] 1.6.Try to acquire the global lock before ddl execution, asynchronously notify the cluster to lock, and release the lock after the execution is complete(#8444 #8447 #8472  )
  - [x] 1.7. Improve unit tests(#8500)

**2.Synchronous processing & Blocking requests & Improve** 
  - [x] 2.1. Realize the lock confirmation process before executing ddl(#8509)
  - [x] 2.2. Implement blocking requests in proxy lock state(#8495 )
  - [x] 2.3. Realize the proxy blocking state after ddl execution is completed(#8472)
  - [x] 2.4. Optimize the process of ddl lock failure(#8535 )
  - [x] 2.5. Optimize the online and offline situation of the proxy during the lock process
  - [x] 2.6. Improve unit tests(#8578)
  - [x] 2.7. Add globallock for jdbc

menghaoranss:
We will deal with this issue in the following two parts：

part 1:

![image](https://user-images.githubusercontent.com/60651792/100062360-66245400-2e6a-11eb-987c-3a12cced923d.png)


part 2:

![image](https://user-images.githubusercontent.com/60651792/100062674-efd42180-2e6a-11eb-8ae2-3bd83f05b448.png)

menghaoranss:
Task list:

- [ ] 1. Implement global asynchronous lock
  - [ ] 1.1. Initialize the distributed lock module in the governance project
  - [ ] 1.2. Realize global lock
  - [ ] 1.3. Design and implement proxy state queue
  - [ ] 1.4. Implement proxy to monitor global lock and  lock itsself
  - [ ] 1.5. Achieve unlock after the proxy completes metadata synchronization
  - [ ] 1.6.Try to acquire the global lock before ddl execution, asynchronously notify the cluster to lock, and release the lock after the execution is complete
  - [ ] 1.7. Improve unit tests


[...]

PR #8706, opened 2020-12-21
Refactor global lock 

menghaoranss:
For #8060 .


terrymanu (changes requested):
RegistryCenter should not singleton, we need to consider about pass them in the method arg

terrymanu (approved):

PR #8836, opened 2020-12-30
Upgrade the version of Spring framework

JiekerTime:
Just upgrade the version of Spring framework. I will make more detailed changes later.


terrymanu (changes requested):
Can not upgrade shardingsphere version of example.
The un-release version can not compile.

Issue #8579, opened 2020-12-11
Design and implement the scaling workflow


menghaoranss:
Hi，everyone

We consider starting to integrate proxy and scaling to realize automatic data migration and data source switching  after  
configuration changes.

Here are the preliminary tasks：

- [x] 1. Integrating scaling API

- [x] 2. Design and implement configuration version management

- [x] 3. Generate new configuration version after RDL execution

- [x] 4. Design automatic scaling job generation strategy  

- [x] 5. Automatically generate scaling job according to the latest configuration version

- [x] 6. Check the data consistency after the scaling job is done

- [x] 7. Realize automatic data source switching

PR #8846, opened 2020-12-31
Add cache manager for configuration changes

menghaoranss:
For #8579

### DIFF
--- a/shardingsphere-governance/shardingsphere-governance-core/src/main/java/org/apache/shardingsphere/governance/core/config/ConfigCacheManager.java
+++ b/shardingsphere-governance/shardingsphere-governance-core/src/main/java/org/apache/shardingsphere/governance/core/config/ConfigCacheManager.java
@@ -1,0 +1,56 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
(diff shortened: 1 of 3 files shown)


shardingsphere-governance/shardingsphere-governance-core/src/main/java/org/apache/shardingsphere/governance/core/config/ConfigCacheManager.java
```diff
@@ -0,0 +1,56 @@
...
+     * @return cache id
+     */
+    public String cache(final String key, final String configuration) {
+        String cacheId = getCacheId();
+        repository.persist(Joiner.on(PATH_SEPARATOR).join(key, CACHE_KEY, cacheId), configuration);
+        return cacheId;
+    }
+    
+    private String getCacheId() {
+        return UUID.randomUUID().toString().replace("-", "");
```
kimmking, line 54: what's the meaning for cacheID and how to use it?

kimmking (changes requested):

shardingsphere-governance/shardingsphere-governance-core/src/main/java/org/apache/shardingsphere/governance/core/config/ConfigCacheManager.java
menghaoranss: After the rule is modified by RDL, the scaling job needs to be triggered automatically.  The rule will be persisted to ZK  only after the scaling job is completed, so it needs to be cached in advance. After the scaling job is completed, the rule configurations will be persisted to ZK and synchronized to memory according to the cashId.(May be it's a version number)

kimmking (approved):

PR #8852, opened 2020-12-31
examples-quick-start

JiekerTime:
The base README of examples


terrymanu (changes requested):
It is not good for create a new menu for example quick start.
The document is for ShardingSphere usage, example should be easy to unstander.
Can you consider move the document into `examples/README.md` and `examples/README_ZH.md`

I notice your new document is similar with the existed document, can you update content on the existed one?

Issue #8736, opened 2020-12-23
Support mysql jdbc's loadbalance


xunyard:
## Feature Request

During  use loadbalnce in jdbc url configuration, it will cause problem (related to bug #5531 [fail to switch schema] ).

## Conditions
- Spring boot application (v2.1.6)
- @Transicational annotaion(optional)
- sharding proxy(v4.1.1)
- mysql-connector-java (v8.0.19)

## Other stuff
jdbc url config example like:

`url: jdbc:mysql:loadbalance://127.0.0.1:3307,127.0.0.1:3407/db?allowMultiQueries=true&useUnicode=true&characterEncoding=UTF8&serverTimezone=Asia/Shanghai&tcpKeepAlive=true&autoReconnect=true`

key code in jdbc will cause problem:

`com.mysql.cj.jdbc.ha.MultiHostConnectionProxy#syncSessionState(com.mysql.cj.jdbc.JdbcConnection, com.mysql.cj.jdbc.JdbcConnection)`

`com.mysql.cj.jdbc.ha.LoadBalancedConnectionProxy#pickNewConnection`

yu199195:
now ， our not support it。

tristaZero:
@xunyard ,
Let me guess why you try to use this property. 
1. You intend to make your program visit ShardingProxy with load-balance?
2. You want ShardingProxy to visit the databases by load balance?

Which one?
Supposing the second one, You can add replica-query configuration in `/conf`. Otherwise, currently we do not support this function.

xunyard:
> @xunyard ,
> Let me guess why you try to use this property.
> 
> 1. You intend to make your program visit ShardingProxy with load-balance?
> 2. You want ShardingProxy to visit the databases by load balance?
> 
> Which one?
> Supposing the second one, You can add replica-query configuration in `/conf`. Otherwise, currently we do not support this function.

The first one. We want to use load-balance to actually make the result of HA for sharding-proxy.

xunyard:
> @xunyard ,
> Let me guess why you try to use this property.
> 
> 1. You intend to make your program visit ShardingProxy with load-balance?
> 2. You want ShardingProxy to visit the databases by load balance?
> 
> Which one?
> Supposing the second one, You can add replica-query configuration in `/conf`. Otherwise, currently we do not support this function.

The first. We want to use java jdbc's loadbance to eventually make the result of HA for sharding-proxy. It is a simple choice for  HA, so how about considering support it in future.

[...]

PR #8906, opened 2021-01-06
Support multiple databases for PostgreSQL.

kfwang90:
Fixes about #8736.

### DIFF
--- a/shardingsphere-features/shardingsphere-ha/shardingsphere-ha-provider/shardingsphere-ha-mgr/pom.xml
+++ b/shardingsphere-features/shardingsphere-ha/shardingsphere-ha-provider/shardingsphere-ha-mgr/pom.xml
@@ -22,7 +22,7 @@
     <modelVersion>4.0.0</modelVersion>
     <parent>
         <groupId>org.apache.shardingsphere</groupId>
-        <artifactId>shardingsphere-ha</artifactId>
+        <artifactId>shardingsphere-ha-provider</artifactId>
         <version>5.0.0-RC1-SNAPSHOT</version>
     </parent>
     <artifactId>shardingsphere-ha-mgr</artifactId>
--- a/shardingsphere-infra/shardingsphere-infra-common/src/main/java/org/apache/shardingsphere/infra/database/metadata/dialect/PostgreSQLDataSourceMetaData.java
+++ b/shardingsphere-infra/shardingsphere-infra-common/src/main/java/org/apache/shardingsphere/infra/database/metadata/dialect/PostgreSQLDataSourceMetaData.java
@@ -41,7 +41,7 @@ public final class PostgreSQLDataSourceMetaData implements DataSourceMetaData {
     
     private final String schema;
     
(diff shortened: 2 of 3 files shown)


shardingsphere-features/shardingsphere-ha/shardingsphere-ha-provider/shardingsphere-ha-mgr/pom.xml
```diff
@@ -22,7 +22,7 @@
     <modelVersion>4.0.0</modelVersion>
     <parent>
         <groupId>org.apache.shardingsphere</groupId>
-        <artifactId>shardingsphere-ha</artifactId>
+        <artifactId>shardingsphere-ha-provider</artifactId>
```
terrymanu, line 25: What is the proposal here?

terrymanu (changes requested):

shardingsphere-features/shardingsphere-ha/shardingsphere-ha-provider/shardingsphere-ha-mgr/pom.xml
kfwang90: > What is the proposal here?

When I use maven clean intall all projects, it has a error at here. And this project is the child of shardingsphere-ha-provider. So I modify it and install again, thers is no more errors.

shardingsphere-infra/shardingsphere-infra-common/src/main/java/org/apache/shardingsphere/infra/database/metadata/dialect/PostgreSQLDataSourceMetaData.java
```diff
@@ -41,7 +41,7 @@
     
     private final String schema;
     
-    private final Pattern pattern = Pattern.compile("jdbc:postgresql://([\\w\\-\\.]+):?([0-9]*)/([\\w\\-]+)", Pattern.CASE_INSENSITIVE);
+    private final Pattern pattern = Pattern.compile("jdbc:postgresql://([\\w\\-\\.]+):?([0-9]*),?.*?/([\\w\\-]+)?\\S*", Pattern.CASE_INSENSITIVE);
```
terrymanu, line 44: What is the different for use multiple hosts and replica-query rule?

We need to discuss the production design here. It is better to use same way for same feature. I prefer use replica-query rule to deal with all master-slave scenarios.

kfwang90: Like this secene: There is PG clusters, contains one master node and another async slave node. 
The replica-query file can config as this:
**master : jdbc:postgresql://masterIp:5432,slaveIp:5432/db?targerServerType=master**
**slave : jdbc:postgresql://masterIp:5432,slaveIp:5432/db?targerServerType=preferSlave**
When PG master node goes down, I can promote async node change to new master fast. And the replica-query config file don't have to change. And the application don't have to restart. JDBC will recovery over a moment time.

terrymanu (approved):

PR #8937, opened 2021-01-07
format the result of 'show resources'

### LINKED ISSUE #8932: Format result of "show resources"
## Feature Request
format result of `show resources;`

expect format of `show resources;` like this:
+------------+-------+-----------+------+-----------------------+-----------------+
| name      | type  | ip          | port | db                        |attribute         |
+------------+-------+-----------+------+-----------------------+-----------------+
| xxxx        | xxxxx | xxx       | xxx  | xxx                     |xxxx               |
+------------+-------+-----------+------+-----------------------+-----------------+

jingshanglu:
Fixes #8932 .


shardingsphere-proxy/shardingsphere-proxy-backend/src/main/java/org/apache/shardingsphere/proxy/backend/text/distsql/rql/impl/DataSourcesQueryBackendHandler.java
```diff
@@ -42,6 +46,8 @@
     
     private Map<String, DataSourceParameter> dataSourceParameterMap;
     
+    private String schemaName;
```
terrymanu, line 49: It is not good a design for use global field which assigned out of the constructor.
Can you consider about improve it?

terrymanu (changes requested):

shardingsphere-proxy/shardingsphere-proxy-backend/src/main/java/org/apache/shardingsphere/proxy/backend/text/distsql/rql/impl/DataSourcesQueryBackendHandler.java
jingshanglu: ok.

tristaZero (changes requested):
Is there any unit test?

terrymanu (approved):

jingshanglu:
> Is there any unit test?

@tristaZero  Maybe Integration testing is required here，Unit testing cannot be effective，i have tested it manually.

tristaZero (approved):
This UT seems does not cover these key points and just gives a simple test no `ResponseHender`.  I have to say better than nothing.

PR #8957, opened 2021-01-08
Remove hook interface 

yu199195:
remove hook interface because we maybe use shardingsphere-agent-tracing plugin....


terrymanu (changes requested):
I notice this PR delete SQLExecutionHook, but no change of this impl class TransactionalSQLExecutionHook.
I guess it will cause compile error.

terrymanu (approved):

PR #9234, opened 2021-01-31
Update inline-expression.cn.md

boomyamm:
fixed db${0..1}.t_order_0${0..9}, db${0..1}.t_order_${10..20} to db0.t_order_0${0..9}, db1.t_order_${10..20}


boomyamm:
I was wrong

Issue #9152, opened 2021-01-25
[SQL federation] Add unit test when encrypt rule, replicaQuery rule and sharding rule all exist


tristaZero:
Hi, community,

This is a little complex issue to add unit test for `SQL federation` when encrypt rule, replicaQuery rule and sharding rule all exist.

### Target files
CalciteStatementTest and CalcitePrepareStatementTest

### How
- Understand `encrypt rule`, `replicaQuery rule` and `sharding rule` (Rule examples in `shardingsphere-jdbc/shardingsphere-jdbc-core/src/test/resources/config` or `examples/shardingsphere-jdbc-example/sharding-example/sharding-raw-jdbc-example/src/main/resources/META-INF`)
- Provide a mix rule configuration with `encrypt rule`, `replicaQuery rule` and `sharding rule` (A single table with an encrypted column, a sharding table with an encrypted column and a simple single table)
- Add a test `join` SQL for a single table with an encrypted column and a simple single table in target files
- Add a test `join` SQL for a sharding table with encrypted column and a simple single table in target files

lmhmhl:
I wanna have a try. Thx.

lmhmhl:
Hi @tristaZero.
I have configured my `yml` file, but I encountered some troubles, I check my `yml` many times, and rewrite it, but it still doesn't work. I would be very glad if you can give me some advice. 

```
rules:
- !SHARDING
  tables:
    t_order_item_calcite_sharding:
      actualDataNodes: calcite_ds.t_order_item_calcite_sharding_${0..1}
      tableStrategy:
        standard:
          shardingColumn: item_id
          shardingAlgorithmName: table_inline
    t_user_encrypt_calcite:
      actualDataNodes: calcite_ds.t_user_encrypt_calcite{0..1}
      columns:
        pwd:
          plainColumn: plain_pwd
          cipherColumn: cipher_pwd
          assistedQueryColumn: assist_pwd
          encryptorName: test
    t_user_encrypt_calcite_sharding:
      actualDataNodes: calcite_ds.t_user_encrypt_calcite_sharding${0..1}
      columns:
        pwd:
          plainColumn: plain_pwd
          cipherColumn: cipher_pwd
          assistedQueryColumn: assist_pwd
          encryptorName: test
      tableStrategy:
        shardingStrategy:
          standard:
            shardingColumn: id
            shardingAlgorithmName: mod
  shardingAlgorithms:
    table_inline:
      type: INLINE
      props:
[...]

PR #9249, opened 2021-02-01
Test encrypt rule, replicaQuery rule and sharding rule all exist when do query in calcite.

lmhmhl:
Ref #9152


shardingsphere-jdbc/shardingsphere-jdbc-core/src/test/resources/config/config-calcite.yaml
```diff
@@ -23,10 +23,54 @@ rules:
         tableStrategy:
           standard:
             shardingColumn: item_id
-            shardingAlgorithmName: table_inline
+            shardingAlgorithmName: table_inline_item_id
+      t_user_encrypt_calcite_sharding:
+        actualDataNodes: calcite_jdbc_1.t_user_encrypt_calcite_sharding_${0..1}
```
tristaZero, line 28: > calcite_jdbc_1.t_user_encrypt_calcite_sharding_${0..1}

It is supposed to be `calcite_ds.t_user_encrypt_calcite_sharding_${0..1}` to leverage `replicaQuery` power.

shardingsphere-jdbc/shardingsphere-jdbc-core/src/test/resources/config/config-calcite.yaml
```diff
@@ -23,10 +23,54 @@ rules:
...
+          - calcite_jdbc_1
+        replicaDataSourceNames:
+          - calcite_jdbc_2
+        loadBalancerName: roundRobin
+    loadBalancers:
+      roundRobin:
+        type: ROUND_ROBIN
+props:
+  sql-show: true
+  query-with-cipher-column: false
```
tristaZero, line 75: Could we set `query-with-cipher-column` = `true` to indeed examine whether this SQL federation feature can work well with the encryption rule?

shardingsphere-jdbc/shardingsphere-jdbc-core/src/test/resources/config/config-calcite.yaml
```diff
@@ -23,10 +23,54 @@ rules:
...
+          pwd:
+            plainColumn: plain_pwd
+            cipherColumn: cipher_pwd
+            encryptorName: encryptor_aes
+  - !REPLICA_QUERY
+    dataSources:
+      calcite_ds:
+        name: calcite_ds
+        primaryDataSourceName:
+          - calcite_jdbc_1
```
tristaZero, line 66: If there is replication relationship between `calcite_jdbc_1` and `calcite_jdbc_2`, maybe `calcite_data_2.sql` is still needed.

tristaZero (changes requested):
Hi @lmhmhl ,

Your PR is beyond my expectation, really. I guess this PR took you a lot of time and effort. Great!

Your configuration has a correct structure of the mixed configuring way, basically. Let me give more explanation about it.

The keypoint locates on `actualDataNodes: calcite_ds.t_user_encrypt_calcite_sharding_${0..1}`.

First, we can know this **logic table** `t_user_encrypt_calcite_sharding` has two `shardings`(Sharding feature) and also `pwd` logic encrypted column mapping with other actual columns (Encryption feature).

Second, where is this `t_user_encrypt_calcite_sharding` logic table? The answer is obvious enough, i.e., `calcite_ds`, a logic datasource (ReplicaQuery feature).

Consequently, dataSource, table, column are all `logic` ones. To meet users' `hybrid` needs, just provide mixed configuration. : )

lmhmhl:
@tristaZero 
I have to say to there do exist some problems in my test, the datasource is created by `YamlShardingSphereDataSourceFactory` using yml file, in order to create two connections to different database instances to build different table structures, I separately get connections from `dataSourceMap` using database name. But the yml configuration are not work on these two sub connection, so I can only construct the sharding data all by myself, and the encrypt or replica query are not easy to test. Basically, I have to figure out why yml configuration are not work on sub connection fetched by `dataSourceMap`. Should I create two yml file or two datasources, and I have tried a lot but all failed, so this PR is  just a compromise, I am in confused now, I would be very glad if u can give me some advices. 
```java
    @Before
    public void initTable() {
        try {
            ShardingSphereConnection conn = dataSource.getConnection();
            Map<String, DataSource> dataSourceMap = conn.getDataSourceMap();
            Connection database0 = dataSourceMap.get("calcite_jdbc_0").getConnection();
            Connection database1 = dataSourceMap.get("calcite_jdbc_1").getConnection();
            RunScript.execute(database0, new InputStreamReader(Objects.requireNonNull(AbstractSQLTest.class.getClassLoader().getResourceAsStream("sql/calcite_data_0.sql"))));
            RunScript.execute(database1, new InputStreamReader(Objects.requireNonNull(AbstractSQLTest.class.getClassLoader().getResourceAsStream("sql/calcite_data_1.sql"))));
            conn.close();
            throw new RuntimeException(ex);
        }
    }
```

tristaZero:
Hi @lmhmhl , 

>  I have tried a lot but all failed, so this PR is just a compromise, I am in confused now, I would be very glad if u can give me some advices.

I know this task is a little tricky, but that's why I want to have you taking charge of it. : ) Basically, I believe you can do better than other newbies.
Secondly, I suppose we can get more things while doing community tasks. So please take it easy. If any confusion, we can discuss it and figure it out.

I am unsure whether I got your point wholly. However, the primary blocker you met looks like the function of `ShardingSphereDataSource` (Virtual DB) and the `Map<String, DataSource>` (Actual DBs) from `conn.getDataSourceMap()`.

### Map<String, DataSource> for setting up test env
We use all the actual data sources to prepare initial data for databases. There is unnecessary to consider `ShardingSphereDataSource`, therefore I think `initTable()` has no critical issues. Just use them to execute the corresponding initialization scrip.

### ShardingSphereDataSource for test
After having the expected test databases, we will use `ShardingSphereDataSource` to test `logic SQLs` in UT.
This phase's key point is to write the configuration file, which means configure logic DB, table, columns with these actual ones. Definitely, this part of the work mainly needs your deep understanding of different configurations.  If you have any questions about this one, please contact me.
Could you figure out which one is your concern?

lmhmhl:
> Hi @lmhmhl ,
> 
> > I have tried a lot but all failed, so this PR is just a compromise, I am in confused now, I would be very glad if u can give me some advices.
> 
> I know this task is a little tricky, but that's why I want to have you taking charge of it. : ) Basically, I believe you can do better than other newbies.
> Secondly, I suppose we can get more things while doing community tasks. So please take it easy. If any confusion, we can discuss it and figure it out.
> 
> I am unsure whether I got your point wholly. However, the primary blocker you met looks like the function of `ShardingSphereDataSource` (Virtual DB) and the `Map<String, DataSource>` (Actual DBs) from `conn.getDataSourceMap()`.
> 
> ### Map<String, DataSource> for setting up test env
> We use all the actual data sources to prepare initial data for databases. There is unnecessary to consider `ShardingSphereDataSource`, therefore I think `initTable()` has no critical issues. Just use them to execute the corresponding initialization scrip.
> 
> ### ShardingSphereDataSource for test
> After having the expected test databases, we will use `ShardingSphereDataSource` to test `logic SQLs` in UT.
> This phase's key point is to write the configuration file, which means configure logic DB, table, columns with these actual ones. Definitely, this part of the work mainly needs your deep understanding of different configurations. If you have any questions about this one, please contact me.
> Could you figure out which one is your concern?

@tristaZero 
Thanks for you comment, which help me make it clear and boost my confidence, I raise a new PR, please review it, I have to say you help me a lot. Thx.

tristaZero (approved):
Thanks for this so-great PR, Top notch!

PR #9280, opened 2021-02-02
Update mix.cn.md

lmhmhl:


terrymanu (approved):

docs/document/content/user-manual/shardingsphere-jdbc/configuration/yaml/mix.cn.md
```diff
@@ -3,4 +3,66 @@ title = "混合规则"
...
+        type: INLINE
+        props:
+          algorithm-expression: t_order_item_calcite_sharding_${item_id % 2}
+      table_inline_user_id:
+        type: INLINE
+        props:
+          algorithm-expression: t_user_encrypt_calcite_sharding_${user_id % 2}
+  - !ENCRYPT
+    encryptors:
+      encryptor_aes:
```
AI1186780944, line 36: In the process of using it, I found that "_" can't be used here, "-" should be used instead. Otherwise, this value will not be recognized, resulting in an error. I configured it in properties format. I don't know if yaml format will have this problem.

PR #9307, opened 2021-02-04
add exception for table rewrite error

jingshanglu:
Add exception for table rewrite error when routed table is not exist.


tristaZero (changes requested):
A unit test is needed.

shardingsphere-features/shardingsphere-sharding/shardingsphere-sharding-route/src/main/java/org/apache/shardingsphere/sharding/route/strategy/type/standard/StandardShardingStrategy.java
```diff
@@ -67,6 +68,8 @@ public StandardShardingStrategy(final String shardingColumn, final StandardShard
             target = shardingAlgorithm.doSharding(availableTargetNames, new PreciseShardingValue(shardingValue.getTableName(), shardingValue.getColumnName(), each));
             if (null != target && availableTargetNames.contains(target)) {
                 result.add(target);
+            } else if (null != target && !availableTargetNames.contains(target)) {
+                throw new ShardingSphereException(String.format("Route table %s is not exist, available actual table:%s", target, availableTargetNames));
```
tristaZero, line 72: 1. Route table %s does not exist, available actual table: %s
2. `availableTargetNames` is a collection, so what's the string value of this variable? Does it not need `Strings.format()`?

tristaZero (changes requested):

tristaZero (approved):

Issue #9104, opened 2021-01-20
More test cases for integration test engine.


tristaZero:
More test cases are needed to ensure  integration test engine covers all the scenarios as many as it can.

Here is a todo list to add cases,

1 For tables (Q1)
- [x] Broadcast table
- [ ] Single table
- [ ] DAL

2 For configuration  (Q1)
- [x] Encryption
- [x] Sharding + Encryption
- [x] Encryption + ReplicaQuery
- [x] Sharding + Encryption + ReplicaQuery

3 For governance
- [ ] Zookeeper
  - [ ] DDL+DML   (Q1)
    - [ ] CREATE TABLE
    - [ ] ALTER TABLE
    - [ ] DROP TABLE
    - [ ] CREATE INDEX
    - [ ] DROP INDEX
    - [ ] CREATE VIEW
    - [ ] DROP VIEW
    - [ ] INSERT
    - [ ] DELETE
    - [ ] UPDATE
    - [ ] CREATE USER
    - [ ] DROP USER
    - [ ] GRANT
    - [ ] REVOKE
  - [ ] RDL(#9964 )
- [ ] ETCD (Later)

4 For databases
- [ ] PostgreSQL JDBC
- [ ] PostgreSQL Proxy
- [ ] Oracle JDBC
- [ ] SQLServer JDBC

5 For Shadow
- [ ] Shadow
- [ ] Sharding + Encryption + ReplicaQuery + Shadow (Later)

6 Data agent
- [ ] Frontend: MySQL  Backend: PostgreSQL
- [ ] Frontend: MySQL  Backend: Oracle/Server
- [ ] Frontend: PostgreSQL  Backend: MySQL
- [ ] Frontend: PostgreSQL  Backend: Oracle/Server

### Tips
- Here is [DOC](https://shardingsphere.apache.org/document/current/cn/features/test-engine/integration-test/) presenting more details
- [Concepts](https://shardingsphere.apache.org/document/current/cn/features/)

lmhmhl:
I wanna have a try

JiekerTime:
I will do it!

tristaZero:
Hi @lmhmhl ,

Currently, @JiekerTime is taking this one. Do you want to try #9117? 
Plus, if you like, I can give more intro about this issue to get you in.

lmhmhl:
> Hi @lmhmhl ,
> 
> Currently, @JiekerTime is taking this one. Do you want to try #9117?
[...]

PR #9313, opened 2021-02-04
add broadcast cases of ddl

JiekerTime:
#9104


JiekerTime:
Update PR due to too many submission records.

PR #9371, opened 2021-02-07
add SQLServer with clause for update statement.

dongzl:
For  #6478 .

- add SQLServer with clause for update statement.
- add SQLServer test case for update with clause.


tristaZero (approved):
Great. : )

PR #9487, opened 2021-02-24
add docker env of dbtbl_with_replica_query_and_encrypt

JiekerTime:


terrymanu (changes requested):
Please do not add new table t_user
The new t_user will break origin scenario

JiekerTime:
change mind ! add encrypt fouction into t_order.So close it.

Issue #9129, opened 2021-01-22
Proofread the SQL definition `MERGE` of Oracle database


tristaZero:
### Introduction
[MERGE](https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/MERGE.html#GUID-5692CCB7-24D9-4C0E-81A7-A22436DC968F) SQL is made up of many [basic elements](https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/Expressions.html#GUID-E7A5363C-AEE9-4809-99C1-1A9C6E3AE017) or [keywords](https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/Oracle-SQL-Reserved-Words-and-Keywords.html#GUID-6A07BB21-AD82-4B47-80FA-9B1141CC23C2).
Hence it is a good thread to help us fix the vague definitions of these basic elements which will compose other complex SQL definition.

### Task
- Work out the vague definition of basic elements
- Proofread `MERGE` definition
- Add test for `MERGE`

### Target files
shardingsphere-sql-parser/shardingsphere-sql-parser-dialect/shardingsphere-sql-parser-oracle/src/main/antlr4/imports/oracle

ThanoshanMV:
Hi @tristaZero, I'll try this one.

ThanoshanMV:
Hi @tristaZero, I'm quite unsure about where (in which target files) to add SQL case and expected parsed result for `MERGE`. I couldn't able to find similar test cases for `MERGE` inside `shardingsphere-sql-parser/shardingsphere-sql-parser-test/src/main/resources`

tristaZero:
Hi @ThanoshanMV ,

Sorry for missing your msg.
Since `MERGE` is a new SQL type, I suggest we create a `merge.xml` in `shardingsphere-sql-parser/shardingsphere-sql-parser-test/src/main/resources`

ThanoshanMV:
Hi @tristaZero,

The count of SQL cases and SQL parser cases are mismatched when I try to run `OracleParserParameterizedTest`. I tried to write a basic one to ensure all okay but it's not considering my new case which is written inside `merge.xml`. Are there any configurations we need to do here?

I've attached the error for your consideration. 
![merge-error](https://user-images.githubusercontent.com/48581379/105805755-b766d600-5fc8-11eb-8aec-e95ab1413f38.png)

ThanoshanMV:
Here are the basic codes inside sql and in case files: 

[...]

PR #9599, opened 2021-03-05
Add `MERGE` statement test case

ThanoshanMV:
#9129 

- added a test case for `MERGE` statement


ThanoshanMV:
Hi @tristaZero, I've added a test case for `MERGE` statement. Can you please check whether I'm doing it in the correct way? (the segments I chose and visiting actual values from AST in `OracleDMLStatementSQLVisitor`) I'll change according to your feedback.

tristaZero:
Hi @ThanoshanMV ,

The PR looks almost correct. Nice!
Actually, it is challenging to visit out the corresponding `SQLStatement`. Anyway, we can firstly fill some necessary `Segment`s into the target `SQLStatement`.

Here is another issue needing your attention.

Could you figure out whether there are similar `MERGE` SQL definitions in other databases, like `SQLServer`, `SQL92`?

The reason is that I found you created an abstract class, `MergeStatement`. Generally, an abstract class `MergeStatement` is designed to imply that there will be many final child classes from various databases, like the `OracleMergeStatement` class from you.

So if `MERGE` SQL is specific to `Oracle`, it is suggested to remove abstract `MergeStatement`.

ThanoshanMV:
Hi @tristaZero, 

I tried to search and figured out that `SQL Server` has a similar `MERGE` [statement](https://docs.microsoft.com/en-us/sql/t-sql/statements/merge-transact-sql?view=sql-server-ver15). 

In the `OracleDMLStatementSQLVisitor`, why we're visiting each subrule? For example, `visitWhereClause`, `visitGroupByClause`. If we consider asserting `MERGE` test cases, we actually get values from `visitMerge` right? If it is, then why we need to visit each subrule? I'm sorry this could be a silly question.

As you said, I'll fill some necessary segments into the target `SQLStatement` to work wholly and finish up remaining all other test cases.

jingshanglu:
@ThanoshanMV May be it is better to use 'into' and 'using' than use 'tables' in `mergeStatement`. we visit each subrule in order to construct a convenient data structure.

ThanoshanMV:
@jingshanglu, do you mean like this for `MergeStatement`:
MergeStatement
and `MergeStatementTestCase`:

MergeStatementTestCase

jingshanglu:
@ThanoshanMV  Yes, it may be better to determine the meaning of a property by its name.

ThanoshanMV:
In [`MERGE` statement](https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/MERGE.html#GUID-5692CCB7-24D9-4C0E-81A7-A22436DC968F), Into clause specifies the target table or view and using clause specifies source table or view. 

So, changing the property names and XML element names as `source` and `target` like below would be okay? : 
MergeStatement
and `MergeStatementTestCase`:
MergeStatementTestCase

jingshanglu:
@ThanoshanMV I think that's ok.

ThanoshanMV:
@jingshanglu I tried to visit subrules `intoClause` and `usingClause` in the `OracleDMLStatementSQLVisitor`. I also visited `viewName` in `OracleStatementSQLVisitor` similar to `MySQLStatementSQLVisitor`. I added single parameter asserts in `MergeStatementAssert`. Can you please check it

jingshanglu (approved):

tristaZero:
Hi @ThanoshanMV @jingshanglu 
It is the right time to merge? I guess @ThanoshanMV needs to change the pattern of this PR as a regular one rather than a draft?

ThanoshanMV:
Sorry @tristaZero, I changed the pattern of this PR

tristaZero:
Nice work, do you have any plan next? @ThanoshanMV  :)

ThanoshanMV:
@tristaZero in the merge statement, `merge update clause`, `merge insert clause` and `error logging clause` needs test cases. I'll add try to add them and I'll prepare my application.

tristaZero:
@ThanoshanMV Got it. go ahead, good luck.

Issue #8966, opened 2021-01-10
[Enhancement] Authentication feature of ShardingProxy


tristaZero:
## Introduction

This feature is the same as the authentication of MySQL or PostgreSQL. ShardingProxy has to do access control when users login in. 

## Three key points need to clarify
- This issue is an enhancement feature for the current `Authentication` module
- All users configured in `server.yaml` of `ShardingProxy` still work and are seen as system users.
- We will leverage the access-control power of different Databases

### Load
- [x] Get users from `DB` and `Configuration`
- [x] Load the privilege information from databases
- [x] Merge privileges by rules
- [x] Refactor API and user configuration
- [x] Different implement of loading privileges from Oracle/Pg/SQLServer

### Refresh
- [x] ShardingProxy supports `create user` SQL, which will be pushed down to databases
- [x] ShardingProxy supports `grant` SQL, which will be pushed down to database
- [x] ShardingProxy supports `revoke` SQL, which will be pushed down to databases
- [x] Refresh users and reload privileges for the SQLs above

#### Login
- [x] Get the `IP` or `hostname` once users login
- [x] Get the `user name`, `password` once users login
- [x] Check users' name and password
- [x] SQL check interface and framework
- [x] SQL checking for `use DB` and `show DB`
- [x] SQL checking for other SQLs

### Tests
- [x] PrivilegeBuilder
- [x] PrivilegeMerger
- [x] PrivilegeLoader
- [x]  MetadataRefreshEngine

PR #9676, opened 2021-03-15
Merge privileges loaded from MySQL

jingshanglu:
Ref #8966 .


tristaZero (changes requested):
I guess `hashCode()` is still needed if there is `equals()`, Like `DataNode`

tristaZero (approved):

PR #9700, opened 2021-03-17
fix proxy

JiekerTime:
fix it test if proxy


JiekerTime:
problem fixed, so close it.

PR #9720, opened 2021-03-18
add Single-table's it test

JiekerTime:


JiekerTime:
codes confilct, so closed it！

PR #9759, opened 2021-03-21
Issue#9758

huanghao495430759:
Fixes Issue#9758.
- doc error of read-write-splitting configurations.

PR #9762, opened 2021-03-22
Fixes ci NPE

menghaoranss:


terrymanu (changes requested):
Please make sure when the props will be null.
In my opinion, the props should never be null, so maybe we need to change the test code to avoid NPE

huanghao495430759:
A ha, I also meet this NPE,and I fix the UT error method at  [https://github.com/apache/shardingsphere/pull/9759](url)
image

menghaoranss:
> A ha, I also meet this NPE,and I fix the UT error method at [https://github.com/apache/shardingsphere/pull/9759](url)
> image

ok, i will close this pr.

PR #9801, opened 2021-03-24
docs: docker build error

totalo:
- There are some errors in the docker help document, and I tried to fix it

### DIFF
--- a/docs/document/content/user-manual/shardingsphere-proxy/docker.cn.md
+++ b/docs/document/content/user-manual/shardingsphere-proxy/docker.cn.md
@@ -28,16 +28,16 @@ mvn clean package -Prelease,docker
 ## 运行 Docker
 
 ```bash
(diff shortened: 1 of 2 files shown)


terrymanu (changes requested):
The correct project name is `shardingsphere-proxy`.
It is unnecessary to change here.

totalo:
> The correct project name is `shardingsphere-proxy`.
> It is unnecessary to change here.

But what the program reads is the configuration in sharding-proxy, mounted to the shardingsphere-proxy directory, the project cannot be started, please check

terrymanu:
> > The correct project name is `shardingsphere-proxy`.
> > It is unnecessary to change here.
> 
> But what the program reads is the configuration in sharding-proxy, mounted to the shardingsphere-proxy directory, the project cannot be started, please check

It seems we need to update the docker file, is it?

totalo:
> > > The correct project name is `shardingsphere-proxy`.
> > > It is unnecessary to change here.
> > 
> > 
> > But what the program reads is the configuration in sharding-proxy, mounted to the shardingsphere-proxy directory, the project cannot be started, please check
> 
> It seems we need to update the docker file, is it?

Yes, let me to fix it.

Issue #9528, opened 2021-02-28
[SQL federation] Add unit test for the quries joining sharding tables


tristaZero:
Hi, community,

### Background
 
I am glad to share that ShardingSphere supports most of the join queries from different shardings (or database instances) after #9518 merged.

Accordingly, this is an issue to add unit tests for `SQL federation` when executing join queries on sharding tables.

### Target files
CalciteStatementTest and CalcitePrepareStatementTest

### How
- Add a unit test concerning`join` SQL for two sharding tables which  existed in `shardingsphere-jdbc/shardingsphere-jdbc-core/src/test/resources/config/config-calcite.yaml`

### Tip
- #9019 and #9249 for your references

tristaZero:
Hi @lmhmhl ,

I am curious whether you are interested in this one? As you can see, it will take some time to dive, but easy for you.

lmhmhl:
@tristaZero Sure, I'd like to do this. so please assign it tome.

sandynz:
Closed by PR #9859

PR #9859, opened 2021-03-28
Test Calcite doing query between two sharding tables.

lmhmhl:
Ref #9528


tristaZero (approved):
Hi @lmhmhl Finally, this PR came out, nice. :)

PR #9888, opened 2021-03-31
Remove authorizedSchemas config for Proxy.

RaigorJiang:
For https://github.com/apache/shardingsphere/pull/9300:

1. remove authorizedSchemas config for proxy example.
2. update authentication documents.


tristaZero (approved):
Thanks for fixing.

PR #9985, opened 2021-04-07
Fix oracle metadata loader problem

### LINKED ISSUE #9617: Load columns for all tables into metadata
We load rules configured tables into metadata with columns and table name only for un-configured tables now. 
By product design, it is necessary to load all columns for every table.

Load all columns are too slow if using raw JDBC interface. We should create a new SPI for quickly load columns via raw database way, such as load metadata via `information_schema` for MySQL. 

The tasks are: 

- [x] Create SPI for load metadata with native database dialect
- [x] Load all columns for default JDBC loader
- [x] Add MySQL dialect
- [x] Finish index metadata loading for MySQL dialect
- [ ] Add PostgreSQL dialect
- [x] Add SQLServer dialect
- [ ] Add Oracle dialect
- [ ] Add H2 dialect

Some issues are related with imperfect metadata, there are: #9615

I have already finished kernel framework adjust and MySQL implementation.
I just leave other database dialects for volunteer wanted.

Volunteers can reference `MySQLTableMetaDataLoader` and `MySQLTableMetaDataLoaderTest` to implement other dialect of databases.

zhujunxxxxx:
Fixes #9617.

- fix oracle metadata loader

as I know oracle does't have `information_schema` schema, but last contributor query metadata from `information_schema`.
it's wrong, so I according to `https://docs.oracle.com/en/database/oracle/oracle-database` to fix.


zhujunxxxxx:
@tristaZero this PR need you review. thanks.

tristaZero (changes requested):
Hi @zhujunxxxxx Thanks for your fixing.
@Liangda-w Could you have a look? Since this PR is working on your previous one.

Liangda-w:
> Hi @zhujunxxxxx Thanks for your fixing.
> @Liangda-w Could you have a look? Since this PR is working on your previous one.

Yes, it's right. Oracle don't have `information_schema` and we should use `ALL_TAB_COLUMNS` instead. Thank you for pointing it out!

zhujunxxxxx:
@tristaZero  need you review.

> > Hi @zhujunxxxxx Thanks for your fixing.
> > @Liangda-w Could you have a look? Since this PR is working on your previous one.
> 
> Yes, it's right. Oracle don't have `information_schema` and we should use `ALL_TAB_COLUMNS` instead. Thank you for pointing it out!

tristaZero (approved):
@zhujunxxxxx Aha, nice.
@Liangda-w thanks for double-checking.

PR #10011, opened 2021-04-09
Fixes #9991

### LINKED ISSUE #9991: The output is incorrect when add resource and database connection failed
Execute RDL `ADD RESOURCE` ，but the client outputs `Can not drop resources` : 

```
mysql> add resource ds_0(host=127.0.0.1,port=3306,db=test1, user=root, password=123456);
ERROR 2006 (HY000): MySQL server has gone away
No connection. Trying to reconnect...
Connection id:    1
Current database: db_1

ERROR 1104 (C1104): Can not drop resources, resources [ds_0] do not exist.
```

menghaoranss:
Fixes #9991 .

### DIFF
--- a/shardingsphere-infra/shardingsphere-infra-common/src/main/java/org/apache/shardingsphere/infra/config/datasource/DataSourceValidator.java
+++ b/shardingsphere-infra/shardingsphere-infra-common/src/main/java/org/apache/shardingsphere/infra/config/datasource/DataSourceValidator.java
@@ -18,10 +18,6 @@
 package org.apache.shardingsphere.infra.config.datasource;
 
 import javax.sql.DataSource;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.LinkedList;
-import java.util.Map;
 
 /**
  * Data source validator.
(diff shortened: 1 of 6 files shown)


shardingsphere-infra/shardingsphere-infra-common/src/main/java/org/apache/shardingsphere/infra/config/datasource/DataSourceValidator.java
```diff
@@ -30,23 +31,23 @@
     /**
      * Validate.
      *
-     * @param dataSources data sources.
-     * @return is valid or not
+     * @param dataSourceConfigurations map of data source configuration
+     * @return collection of invalid data sources
      */
-    public boolean validate(final Map<String, DataSourceConfiguration> dataSources) {
```
tristaZero, line 36: Do you think It is better to rename it as `getInvalidatedDataSources()`? Or maybe we can pass one `DataSourceConfiguration` for validation (my preference).
How do you think?

tristaZero (changes requested):

tristaZero (approved):

PR #10021, opened 2021-04-10
Rename authority provider type from STORAGE to NATIVE

terrymanu:
The YAML to config authority changed to:

```yaml
rules:
  - !AUTHORITY
    provider:
      type: NATIVE
```

### DIFF
--- a/shardingsphere-features/shardingsphere-authority/shardingsphere-authority-common/src/main/java/org/apache/shardingsphere/authority/algorithm/natived/NativeAuthorityProviderAlgorithm.java
+++ b/shardingsphere-features/shardingsphere-authority/shardingsphere-authority-common/src/main/java/org/apache/shardingsphere/authority/algorithm/natived/NativeAuthorityProviderAlgorithm.java
@@ -15,9 +15,9 @@
  * limitations under the License.
  */
 
-package org.apache.shardingsphere.authority.algorithm.storage;
+package org.apache.shardingsphere.authority.algorithm.natived;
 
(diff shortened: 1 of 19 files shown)


shardingsphere-features/shardingsphere-authority/shardingsphere-authority-common/src/main/java/org/apache/shardingsphere/authority/algorithm/natived/NativeAuthorityProviderAlgorithm.java
```diff
@@ -48,6 +48,6 @@ public void init(final Map<String, ShardingSphereMetaData> mataDataMap, final Co
     
     @Override
     public String getType() {
-        return "STORAGE";
+        return "NATIVE";
```
taojintianxia, line 51: why not enum ?

taojintianxia (commented):
for code as following, why not using enum ?

menghaoranss (approved):

shardingsphere-features/shardingsphere-authority/shardingsphere-authority-common/src/main/java/org/apache/shardingsphere/authority/algorithm/natived/NativeAuthorityProviderAlgorithm.java
terrymanu: The feature is provided by SPI which is dynamic.
Enum is static variable which user can not add new type with out change original enum.

PR #10030, opened 2021-04-10
Revise GovernanceAuthorityContext to reuse AuthorityProvideAlgorithm

terrymanu:


shardingsphere-features/shardingsphere-authority/shardingsphere-authority-common/src/main/java/org/apache/shardingsphere/authority/rule/AuthorityRule.java
```diff
@@ -32,14 +33,17 @@
 /**
  * Authority rule.
  */
+@Getter
 public final class AuthorityRule implements GlobalRule {
-
+
```
taojintianxia, line 38: emtpy line with 4 white space or that's the front of static qualifier

taojintianxia (commented):
```java
    static {
        ShardingSphereServiceLoader.register(AuthorityProvideAlgorithm.class);
    }
```

is there are 4 white spaces in an empty line upon the static qulifier ?

tristaZero (approved):

Issue #9964, opened 2021-04-07
Refactor DistSQL for Resource/DB recovery/Sharding/Read write splitting rules


tristaZero:
Hi,

Although we have finished the basic process for these rules, there are more reasonable and expressive distributed-SQL. Here is the new draft for them.

## RESOURCE

1. Add
```sql
ADD RESOURCE resource0 (
HOST=127.0.0.1,
PORT=3306,
DB=db0,
USER=root,
(PASSWORD=root)?
)
```

2. Drop resource
```sql
DROP RESOURCE resource0, resource1
```

4. Show resources
```sql
SHOW RESOURCES (FROM schemaName)?
```

## DB_DISCOVERY

1. Create
```sql
CREATE DB_DISCOVERY RULE ha_group_0 (
RESOURCES(resource0,resource1),
TYPE(NAME=mgr,PROPERTIES(groupName='92504d5b-6dec',keepAliveCron=''))
)
```

2. Alter
```sql
ALTER DB_DISCOVERY RULE ha_group_0 (
RESOURCES(resource0,resource1),
TYPE(NAME=mgr,PROPERTIES(groupName='92504d5b-6dec' ,keepAliveCron=''))
)
```

3. Drop
```sql
DROP DB_DISCOVERY RULE ha_group_0, ha_group_1
```

4. Show
```sql
SHOW DB_DISCOVERY RULES (FROM schemaName)?
```

## READWRITE_SPLITTING

1. Create
```sql
// Static
CREATE READWRITE_SPLITTING RULE ms_group_0 (
WRITE_RESOURCE=primary_ds,
READ_RESOURCES(replica_ds_0,replica_ds_1),
TYPE(NAME=random)
)

// Dynamic, joining with DB_DISCOVERY RULE
CREATE READWRITE_SPLITTING RULE ms_group_1 (
AUTO_AWARE_RESOURCE=group_0,
TYPE(NAME=random,PROPERTIES(read_weight='2:1'))
)
```

2. Alter 
```sql
ALTER READWRITE_SPLITTING RULE ms_group_1 (
WRITE_RESOURCE=primary_ds,
READ_RESOURCES(replica_ds_0,replica_ds_1,replica_ds_2),
TYPE(NAME=random,PROPERTIES(read_weight='2:0'))
)
```

3. Drop  
```sql
DROP READWRITE_SPLITTING RULE ms_group_1
```

4. Show  
```sql
[...]

PR #10068, opened 2021-04-13
Check if resource name is duplicate for add resource

menghaoranss:
refer #9964.


shardingsphere-proxy/shardingsphere-proxy-backend/src/main/java/org/apache/shardingsphere/proxy/backend/text/distsql/rdl/impl/AddResourceBackendHandler.java
```diff
@@ -67,6 +72,21 @@ public ResponseHeader execute(final String schemaName, final AddResourceStatemen
         return new UpdateResponseHeader(sqlStatement);
     }
     
+    private void check(final String schemaName, final AddResourceStatement sqlStatement) {
+        List<String> dataSourceNames = new ArrayList<>(sqlStatement.getDataSources().size());
+        for (DataSourceSegment dataSourceSegment : sqlStatement.getDataSources()) {
+            if (dataSourceNames.contains(dataSourceSegment.getName())) {
+                throw new DuplicateResourceException(dataSourceSegment.getName());
+            }
+            dataSourceNames.add(dataSourceSegment.getName());
```
tristaZero, line 81: Hi this process is different with the one of `validate dataSources`, isn't it?
Here,  the programming will break once there is an exception. On the other hand, `validate dataSources` will return a whole list of invalid data sources. Could we keep the handling consistent?

tristaZero (changes requested):

tristaZero (approved):

tristaZero (changes requested):
conflicts

tristaZero (approved):

PR #10084, opened 2021-04-14
fixes PostgreSQL SQLXML data type packet exception.

dongzl:


terrymanu (changes requested):
Please do not add extra dependency, maybe throw RuntimeException is fine too.

shardingsphere-db-protocol/shardingsphere-db-protocol-postgresql/src/main/java/org/apache/shardingsphere/db/protocol/postgresql/packet/command/query/text/PostgreSQLDataRowPacket.java
```diff
@@ -64,7 +65,7 @@ private void writeSQLXMLData(final PostgreSQLPacketPayload payload, final Object
             payload.writeInt4(((SQLXML) data).getString().getBytes().length);
```
terrymanu, line 65: Please remove useless @slf4j too

terrymanu (changes requested):

terrymanu (approved):

PR #10116, opened 2021-04-16
refactor AbstractSQLBuilder

### LINKED ISSUE #9510: Can not rewrite owner table name of projection if using sharding and encrypt together
If the original SQL is:

```sql
SELECT t_account.* FROM t_account WHERE account_id=?
```

The current rewrited SQL which is incorrect:

```sql
SELECT `t_account`.col1, `t_account`.col2, `t_account`.col_n FROM t_account_0 WHERE account_id=?
```

It should be rewrite to: 

```sql
SELECT `t_account_0`.col1, `t_account_0`.col2, `t_account_0`.col_n FROM t_account_0 WHERE account_id=?
```

The owner of projection can not rewrite to sharded table name.

The solution for now is adding alias for table, for example:

```sql
SELECT a.* FROM t_account a WHERE account_id=?
```
It can rewrite correctly

```sql
SELECT a.* FROM t_account_0 a WHERE account_id=?
```

Please see #9512 to find commented test cases for this issue.

huanghao495430759:
Fixes #9510 .

- refactor AbstractSQLBuilder, extract sub SQLToken from ComposableSQLToken before executeSQL build.

### DIFF
--- a/shardingsphere-infra/shardingsphere-infra-rewrite/shardingsphere-infra-rewrite-engine/src/main/java/org/apache/shardingsphere/infra/rewrite/sql/impl/AbstractSQLBuilder.java
+++ b/shardingsphere-infra/shardingsphere-infra-rewrite/shardingsphere-infra-rewrite-engine/src/main/java/org/apache/shardingsphere/infra/rewrite/sql/impl/AbstractSQLBuilder.java
@@ -17,7 +17,6 @@
 
 package org.apache.shardingsphere.infra.rewrite.sql.impl;
 
-import lombok.RequiredArgsConstructor;
 import org.apache.shardingsphere.infra.rewrite.context.SQLRewriteContext;
 import org.apache.shardingsphere.infra.rewrite.sql.SQLBuilder;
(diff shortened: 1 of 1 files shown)


shardingsphere-infra/shardingsphere-infra-rewrite/shardingsphere-infra-rewrite-engine/src/main/java/org/apache/shardingsphere/infra/rewrite/sql/impl/AbstractSQLBuilder.java
```diff
@@ -17,49 +17,53 @@
...
 
 /**
  * Abstract SQL builder.
  */
-@RequiredArgsConstructor
 public abstract class AbstractSQLBuilder implements SQLBuilder {
     
     private final SQLRewriteContext context;
     
+    private final List<SQLToken> sqlTokens = new LinkedList<>();
```
terrymanu, line 37: What is diff of `context.getSqlTokens()` and `sqlTokens`?
Should we just keep only one way to access sqlTokens?

terrymanu (changes requested):

PR #10132, opened 2021-04-19
Get generated key only if column type is integer

### LINKED ISSUE #10131: Exception occurred when inserting values into a table with varchar primary key
master branch.
But I think this issue exists in previous versions.

Proxy

### Expected behavior

```yaml
rules:
- !SHARDING
  tables:
    t_config:
      actualDataNodes: hello_ss_0.t_config_0
  defaultTableStrategy:
    none:
```

```sql
create table t_config (
  config_name varchar(100) primary key,
  config_value varchar(100)
);
insert into t_config (config_name, config_value) values ('key1', 'value1');
```

### Actual behavior
```
Bad value for type long : key1
```

TeslaCN:
Fixes #10131 

- Check key column type before get value.

### DIFF
--- a/shardingsphere-proxy/shardingsphere-proxy-backend/src/main/java/org/apache/shardingsphere/proxy/backend/communication/jdbc/executor/callback/ProxyJDBCExecutorCallback.java
+++ b/shardingsphere-proxy/shardingsphere-proxy-backend/src/main/java/org/apache/shardingsphere/proxy/backend/communication/jdbc/executor/callback/ProxyJDBCExecutorCallback.java
@@ -35,6 +35,7 @@ import org.apache.shardingsphere.sql.parser.sql.common.statement.SQLStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
+import java.sql.Types;
 import java.util.Optional;
 
 /**
@@ -85,7 +86,18 @@ public abstract class ProxyJDBCExecutorCallback extends JDBCExecutorCallback<Exe
     
     private long getGeneratedKey(final Statement statement) throws SQLException {
         ResultSet resultSet = statement.getGeneratedKeys();
-        return resultSet.next() ? resultSet.getLong(1) : 0L;
+        return resultSet.next() ? getGeneratedKeyIfInteger(resultSet) : 0L;
+    }
+    
+    private long getGeneratedKeyIfInteger(final ResultSet resultSet) throws SQLException {
+        switch (resultSet.getMetaData().getColumnType(1)) {
+            case Types.SMALLINT:
+            case Types.INTEGER:
+            case Types.BIGINT:
+                return resultSet.getLong(1);
+            default:
+                return 0L;
+        }
     }
     
     @Override


shardingsphere-proxy/shardingsphere-proxy-backend/src/main/java/org/apache/shardingsphere/proxy/backend/communication/jdbc/executor/callback/ProxyJDBCExecutorCallback.java
```diff
@@ -85,7 +86,18 @@ private QueryResult createQueryResult(final ResultSet resultSet, final Connectio
...
     private long getGeneratedKey(final Statement statement) throws SQLException {
         ResultSet resultSet = statement.getGeneratedKeys();
-        return resultSet.next() ? resultSet.getLong(1) : 0L;
+        return resultSet.next() ? getGeneratedKeyIfInteger(resultSet) : 0L;
+    }
+    
+    private long getGeneratedKeyIfInteger(final ResultSet resultSet) throws SQLException {
+        switch (resultSet.getMetaData().getColumnType(1)) {
+            case Types.SMALLINT:
+            case Types.INTEGER:
```
tristaZero, line 95:
```
create table t_config (
  config_name varchar(100) primary key,
  config_value varchar(100)
);
insert into t_config (config_name, config_value) values ('key1', 'value1');
```
I am curious what result MySQL or PostgreSQL returns in this case?

tristaZero (changes requested):

shardingsphere-proxy/shardingsphere-proxy-backend/src/main/java/org/apache/shardingsphere/proxy/backend/communication/jdbc/executor/callback/ProxyJDBCExecutorCallback.java
TeslaCN: ## PostgreSQL

I have captured the packets and found the following SQL:
```
PostgreSQL
    Type: Parse
    Length: 96
    Statement: 
    Query: insert into t_config_0 (config_name, config_value) values ('key2', 'value2')\nRETURNING *
    Parameters: 0
```
It returned what we inserted.

TeslaCN: ## MySQL

```
MySQL Protocol
    Packet Length: 77
    Packet Number: 0
    Request Command Query
        Command: Query (3)
        Statement: insert into t_config_0 (config_name, config_value) values ('key2', 'value2')
```

```
MySQL Protocol
    Packet Length: 7
    Packet Number: 1
    Response Code: OK Packet (0x00)
    Affected Rows: 1
    Server Status: 0x0002
        .... .... .... ...0 = In transaction: Not set
        .... .... .... ..1. = AUTO_COMMIT: Set
        .... .... .... .0.. = Multi query / Unused: Not set
        .... .... .... 0... = More results: Not set
        .... .... ...0 .... = Bad index used: Not set
        .... .... ..0. .... = No index used: Not set
        .... .... .0.. .... = Cursor exists: Not set
        .... .... 0... .... = Last row sent: Not set
        .... ...0 .... .... = Database dropped: Not set
        .... ..0. .... .... = No backslash escapes: Not set
        .... .0.. .... .... = Metadata changed: Not set
        .... 0... .... .... = Query was slow: Not set
        ...0 .... .... .... = PS Out Params: Not set
        ..0. .... .... .... = In Trans Readonly: Not set
        .0.. .... .... .... = Session state changed: Not set
    Message: 
```

When using MySQL, it returned nothing about generated keys. So the `resultSet.next()` is `false`.

terrymanu: It looks like only MySQL need lastInsertId.
For this pr, it has already fix the problem about return pk with varchar data type.
In the future, it is better to redesign about class UpdateResult with diff database protocols.

terrymanu (approved):

Issue #10123, opened 2021-04-19
Refactor DistSQL for create Sharding rules


menghaoranss:
Refer #9964

- [x] CREATE SHARDING TABLE RULE
- [x] remove CREATE SHARDING RULE
- [x] CREATE SHARDING BINDING TABLE RULES
- [x] CREATE SHARDING BROADCAST TABLE RULES
- [x] Verify duplicate table names
- [x] Fix the rules being overwrote in the registry center

menghaoranss:
Completed!

PR #10139, opened 2021-04-20
Add CREATE SHARDING BROADCAST TABLE RULES

menghaoranss:
refer #10123 .

### DIFF
--- a/shardingsphere-db-protocol/shardingsphere-db-protocol-core/src/main/java/org/apache/shardingsphere/db/protocol/error/CommonErrorCode.java
+++ b/shardingsphere-db-protocol/shardingsphere-db-protocol-core/src/main/java/org/apache/shardingsphere/db/protocol/error/CommonErrorCode.java
@@ -55,7 +55,7 @@ public enum CommonErrorCode implements SQLErrorCode {
     
     DUPLICATE_TABLE(1113, "C1113", "Duplicate table names %s."),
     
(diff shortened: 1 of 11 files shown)


shardingsphere-db-protocol/shardingsphere-db-protocol-core/src/main/java/org/apache/shardingsphere/db/protocol/error/CommonErrorCode.java
```diff
@@ -55,6 +55,8 @@
     
     DUPLICATE_TABLE(1113, "C1113", "Duplicate table names %s."),
     
+    SHARDING_BROADCAST_EXIST(1114, "C1114", "Sharding broadcast table rules already exists in schema %s."),
```
tristaZero, line 58: `exists` -> `exist`

tristaZero (changes requested):

shardingsphere-db-protocol/shardingsphere-db-protocol-core/src/main/java/org/apache/shardingsphere/db/protocol/error/CommonErrorCode.java
menghaoranss: modified

tristaZero (approved):

PR #10158, opened 2021-04-23
fix drop columns exception.

huanghao495430759:
fix drop columns exception.


shardingsphere-features/shardingsphere-encrypt/shardingsphere-encrypt-rewrite/src/main/java/org/apache/shardingsphere/encrypt/rewrite/token/generator/impl/EncryptAlterTableTokenGenerator.java
```diff
@@ -130,7 +130,7 @@ protected boolean isGenerateSQLTokenForEncrypt(final SQLStatementContext sqlStat
     private Collection<SQLToken> getDropColumnTokens(final String tableName, final String columnName, 
                                                      final ColumnSegment columnSegment, final DropColumnDefinitionSegment dropColumnDefinitionSegment) {
         Collection<SQLToken> result = new LinkedList<>();
-        result.add(new RemoveToken(dropColumnDefinitionSegment.getStartIndex() - 1, columnSegment.getStopIndex() + 1));
+        result.add(new RemoveToken(columnSegment.getStartIndex() - "DROP COLUMN".length() - 2, columnSegment.getStopIndex()));
```
terrymanu, line 133: "DROP COLUMN" is the literal which is incorrect.
The `drop column` may keep more than one spaces or include comment inside, for example: `drop    /*xxxx*/ column`

terrymanu (changes requested):

PR #10189, opened 2021-04-25
Revise #10186

TeslaCN:
Add submitted task count after execute `CommandExecutorTask`.


shardingsphere-proxy/shardingsphere-proxy-frontend/shardingsphere-proxy-frontend-core/src/main/java/org/apache/shardingsphere/proxy/frontend/state/impl/OKProxyState.java
```diff
@@ -40,5 +40,6 @@ public void execute(final ChannelHandlerContext context, final Object message, f
         ExecutorService executorService = CommandExecutorSelector.getExecutorService(
                 isOccupyThreadForPerConnection, supportHint, backendConnection.getTransactionStatus().getTransactionType(), context.channel().id());
         executorService.execute(new CommandExecutorTask(databaseProtocolFrontendEngine, backendConnection, context, message));
+        backendConnection.getSubmittedTaskCount().incrementAndGet();
```
terrymanu, line 43: How about change seq of line 42 and 43? 
It seems better to increment count before task execute.

TeslaCN: I think it is OK that add the submitted count after the task submitted.

TeslaCN: Maybe we can add count before execute and catch `RejectedExecutinoException`.

strongduanmu (approved):

PR #10246, opened 2021-05-03
fix example pom miss shardingsphere-read-write-splitting-api dependency

weiyue888999:
Fixes #ISSUSE_ID.

解决example模块下，缺乏shardingsphere-read-write-splitting-api依赖的问题

### DIFF
--- a/examples/pom.xml
+++ b/examples/pom.xml
@@ -167,6 +167,11 @@
                 <artifactId>shardingsphere-infra-common</artifactId>
                 <version>${shardingsphere.version}</version>
             </dependency>
+            <dependency>
+                <groupId>org.apache.shardingsphere</groupId>
+                <artifactId>shardingsphere-read-write-splitting-api</artifactId>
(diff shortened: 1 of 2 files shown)


examples/pom.xml
```diff
@@ -167,6 +167,11 @@
                 <artifactId>shardingsphere-infra-common</artifactId>
                 <version>${shardingsphere.version}</version>
             </dependency>
+            <dependency>
+                <groupId>org.apache.shardingsphere</groupId>
+                <artifactId>shardingsphere-read-write-splitting-api</artifactId>
```
terrymanu, line 172: The `shardingsphere-read-write-splitting-api` has already change to `shardingsphere-readwrite-splitting-api`, please update it

terrymanu (changes requested):

examples/pom.xml
weiyue888999: oh , oh , my fault .

PR #10330, opened 2021-05-13
Add create readwrite splitting rule

menghaoranss:
For #9964.


shardingsphere-distsql-parser/shardingsphere-distsql-parser-engine/src/main/java/org/apache/shardingsphere/distsql/parser/core/DistSQLVisitor.java
```diff
@@ -187,6 +187,50 @@ public ASTNode visitAlterShardingBroadcastTableRules(final DistSQLStatementParse
         return result;
     }
 
+    @Override
+    public ASTNode visitCreateReadwriteSplittingRule(final DistSQLStatementParser.CreateReadwriteSplittingRuleContext ctx) {
+        Collection<ReadwriteSplittingRuleSegment> readwriteSplittingRuleSegments = new LinkedList<>();
+        for (DistSQLStatementParser.ReadwriteSplittingRuleDefinitionContext each : ctx.readwriteSplittingRuleDefinition()) {
```
taojintianxia, line 193: you could consider lambda syntax

shardingsphere-distsql-parser/shardingsphere-distsql-parser-engine/src/main/antlr4/imports/Keyword.g4
```diff
@@ -82,7 +82,23 @@ FROM
 SHARDING
     : S H A R D I N G
     ;
-   
+
+READ_WRITE_SPLITTING
```
tristaZero, line 86:
```yaml
rules:
- !READWRITE_SPLITTING
```
shows us `readwrite_spliting`, is it better to keep the same keyword here?

tristaZero (changes requested):

shardingsphere-distsql-parser/shardingsphere-distsql-parser-engine/src/main/antlr4/imports/Keyword.g4
menghaoranss: yes, I will update the issue latter.

tristaZero (approved):

PR #10355, opened 2021-05-16
Add more assertions of GovernanceFacadeTest.assertInit()

### LINKED ISSUE #10345: Add more assertions of GovernanceFacadeTest.assertInit()
use reflection to assert attributes of GovernanceFacade.

Just search `// TODO` from class `GovernanceFacadeTest` to find hint

JooKS-me:
Fixes #10345.

### DIFF
--- a/shardingsphere-governance/shardingsphere-governance-core/src/test/java/org/apache/shardingsphere/governance/core/facade/GovernanceFacadeTest.java
+++ b/shardingsphere-governance/shardingsphere-governance-core/src/test/java/org/apache/shardingsphere/governance/core/facade/GovernanceFacadeTest.java
@@ -60,7 +60,7 @@ public final class GovernanceFacadeTest {
         assertThat(getField(governanceFacade, "listenerManager"), instanceOf(GovernanceListenerManager.class));
         GovernanceListenerManager listenerManager = (GovernanceListenerManager) getField(governanceFacade, "listenerManager");
         assertThat(getField(listenerManager, "registryCenterRepository"), is(registryCenterRepository));
-        assertThat(getField(listenerManager,"schemaNames"), is(Arrays.asList("schema_0", "schema_1")));
+        assertThat(getField(listenerManager, "schemaNames"), is(Arrays.asList("schema_0", "schema_1")));
     }
     
     @Test
--- a/shardingsphere-governance/shardingsphere-governance-core/src/test/java/org/apache/shardingsphere/governance/core/facade/GovernanceFacadeTest.java
+++ b/shardingsphere-governance/shardingsphere-governance-core/src/test/java/org/apache/shardingsphere/governance/core/facade/GovernanceFacadeTest.java
@@ -34,8 +34,12 @@ import java.util.Collections;
 import java.util.Map;
 import java.util.Properties;
 
+import static org.hamcrest.CoreMatchers.instanceOf;
+import static org.hamcrest.CoreMatchers.is;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertThat;
+import static org.junit.Assert.assertEquals;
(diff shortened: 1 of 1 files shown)


shardingsphere-governance/shardingsphere-governance-core/src/test/java/org/apache/shardingsphere/governance/core/facade/GovernanceFacadeTest.java
```diff
@@ -47,7 +48,9 @@ public void assertInit() {
         GovernanceConfiguration config = new GovernanceConfiguration("test_name", new RegistryCenterConfiguration("TEST", "127.0.0.1", new Properties()), false);
         governanceFacade.init(config, Arrays.asList("schema_0", "schema_1"));
         assertNotNull(governanceFacade.getRegistryCenter());
-        // TODO use reflection to assert attributes of GovernanceFacade
+        assertFalse((Boolean) getField(governanceFacade, "isOverwrite"));
```
SteNicholas, line 51: IMO, this should add validation for the field type of `isOverwrite`, which verify whether `isOverwrite ` is boolean type and then convert to `Boolean` type object.

shardingsphere-governance/shardingsphere-governance-core/src/test/java/org/apache/shardingsphere/governance/core/facade/GovernanceFacadeTest.java
```diff
@@ -47,7 +48,9 @@ public void assertInit() {
...
+        assertNotNull(getField(governanceFacade, "registryCenterRepository"));
```
SteNicholas, line 52: This could validate the field value of `registryCenterRepository`, not only verify whether the field value is not null.

SteNicholas (changes requested):

shardingsphere-governance/shardingsphere-governance-core/src/test/java/org/apache/shardingsphere/governance/core/facade/GovernanceFacadeTest.java
JooKS-me: OK, I will add it.

JooKS-me: @SteNicholas Hi, do you mean to check the attribute values of `registryCenterRepository` one by one? I have some confusion because I think it has no attribute values.

In this test code, `governanceFacade.init(config, Arrays.asList("schema_0", "schema_1"));` will generate a `RegistryCenterRepositoryFixture` class instance through the SPI mechanism, but we know that `RegistryCenterRepositoryFixture` does not have any Attributes, except for the `PATH_SEPARATOR` defined in the interface `RegistryCenterRepository`.

However, I can check the attributes of `listenerManager` one by one~

Is it right ?

JooKS-me: @SteNicholas Hi, do you mean to check the attribute values of `registryCenterRepository` one by one? I have some confusion because I think it has no attribute values.

In this test code, `governanceFacade.init(config, Arrays.asList("schema_0", "schema_1"));` will generate a `RegistryCenterRepositoryFixture` class instance through the SPI mechanism, but we know that `RegistryCenterRepositoryFixture` does not have any Attributes, except for the `PATH_SEPARATOR` defined in the interface `RegistryCenterRepository`.

However, I can check the attributes of `listenerManager` one by one.

Is it right ?

SteNicholas (approved):

SteNicholas:
@terrymanu , please help to take a look for this pull request. I have approved the changes.

terrymanu (approved):

Issue #10368, opened 2021-05-17
Investigate and optimize the SQL compatibility of routing to single data node


strongduanmu:
# Background

The sharding function of SS 4.X version is 100% compatible with routing to single data node SQL(including single table SQL (implemented by defaultDataSourceName) or sharding table). Due to the refactoring of the kernel in SS 5.X, some SQL execute abnormally, which needs to be investigated  and supported.

# Task

- [ ] MySQL routing to single data node SQL support
  - [ ] DALStatement（pending）
  - [ ] DCLStatement
  - [x] DDLStatement（core SQL statement has been supported）
  - [x] DMLStatement（core SQL statement has been supported）
  - [ ] RLStatement
  - [ ] TCLStatement
- [ ] PostgreSQL routing to single data node SQL support
  - [ ] DALStatement（pending）
  - [ ] DCLStatement
  - [ ] DDLStatement (doing)
  - [ ] DMLStatement (doing)
  - [ ] StoreProcedure
  - [ ] TCLStatement
- [ ] Oracle routing to single data node SQL support (low priority)
- [ ] SQLServer routing to single data node SQL support (low priority)
- [ ] SQL92 routing to single data node SQL support (low priority)

# Investigate results

The investigate results are tested using the following sharding rules.

```yaml
rules:
- !SHARDING
 tables:
   t_order:
     actualDataNodes: ds_${0..1}.t_order_${0..1}
     tableStrategy:
       standard:
         shardingColumn: order_id
         shardingAlgorithmName: t_order_inline
     keyGenerateStrategy:
       column: order_id
       keyGeneratorName: snowflake
   t_order_item:
     actualDataNodes: ds_${0..1}.t_order_item_${0..1}
     tableStrategy:
       standard:
         shardingColumn: order_id
         shardingAlgorithmName: t_order_item_inline
     keyGenerateStrategy:
       column: order_item_id
       keyGeneratorName: snowflake
   t_user:
     actualDataNodes: ds_0.t_user_0
 bindingTables:
   - t_order,t_order_item
 broadcastTables:
   - t_config
 defaultDatabaseStrategy:
   standard:
     shardingColumn: user_id
     shardingAlgorithmName: database_inline
 defaultTableStrategy:
   none:

 shardingAlgorithms:
   database_inline:
     type: INLINE
     props:
       algorithm-expression: ds_${user_id % 2}
   t_order_inline:
     type: INLINE
     props:
       algorithm-expression: t_order_${order_id % 2}
[...]

PR #10372, opened 2021-05-18
support mysql create udf statement execute

strongduanmu:
Ref #10368.

- support mysql create udf statement execute


tristaZero (changes requested):
Hi,

From [MySQL DOC](https://dev.mysql.com/doc/refman/8.0/en/create-function-loadable.html), `Udf` points `user defined functions`, `MySQLCreateUdfStatement` is proper to include all users' function?

strongduanmu:
> Hi,
> 
> From [MySQL DOC](https://dev.mysql.com/doc/refman/8.0/en/create-function-loadable.html), `Udf` points `user defined functions`, `MySQLCreateUdfStatement` is proper to include all users' function?

@tristaZero `MySQLCreateUdfStatement` is not very suitable for user defined functions. It is specifically used for loadable functions. Maybe it is better to rename it to `MySQLCreateLoadableFunctionStatement`.

According to the document, the loadable function is defined as follows:

> A loadable function is a way to extend MySQL with a new function that works like a native (built-in) MySQL function such as ABS() or CONCAT()

Because of the loaded shared library file, loadable functions can be used globally, while user-defined functions can only be used in the corresponding db range.

Due to these different characteristics, the loadable function is divided into DALStatement, and the user-defined function is divided into DDLStatement, which also leads to the statement is not convenient to reuse, because we will use different routing engines according to the statement type.

tristaZero (approved):

PR #10448, opened 2021-05-25
fix oracle meta data load error

strongduanmu:
- fix oracle load table meta data error
- fix oracle load non-materialized view index error


taojintianxia (approved):

shardingsphere-infra/shardingsphere-infra-common/src/main/java/org/apache/shardingsphere/infra/metadata/schema/builder/loader/IndexMetaDataLoader.java
```diff
@@ -54,6 +56,10 @@
                     result.add(new IndexMetaData(indexName));
                 }
             }
+        } catch (SQLException ex) {
```
tristaZero, line 59: final

tristaZero (changes requested):

shardingsphere-infra/shardingsphere-infra-common/src/main/java/org/apache/shardingsphere/infra/metadata/schema/builder/loader/IndexMetaDataLoader.java
strongduanmu: > final

@tristaZero fixed

Issue #10454, opened 2021-05-25
Refactor SHOW SHARDING/READWRITE_SPLITTING/DB_DISCOVERY/ENCRYPT RULES


menghaoranss:
refer #9964
- [x] SHOW SHARDING TABLE RULES (FROM schemaName)？
- [x] SHOW SHARDING TABLE RULE t_order (FROM schemaName)？
- [x] SHOW SHARDING BINDING TABLE  RULES (FROM schemaName)?
- [x] SHOW SHARDING BROADCAST TABLE RULES (FROM schemaName)?
- [x] SHOW READWRITE_SPLITTING RULES (FROM schemaName)?
- [x] SHOW DB_DISCOVERY RULES (FROM schemaName)?
- [x] SHOW ENCRYPT RULES (FROM schemaName)?
- [x] SHOW ENCRYPT TABLE RULE t_order (FROM schemaName)

menghaoranss:
**SHOW SHARDING BINDING TABLE RULES (FROM schemaName)?**

```
mysql> show sharding binding table rules from sharding_db;
+----------------------+
| shardingBindingTable |
+----------------------+
| t_order,t_order_item |
| t1,t2                |
+----------------------+
2 rows in set (0.00 sec)
```

menghaoranss:
**SHOW SHARDING BROADCAST TABLE RULES (FROM schemaName)?**
```
mysql> show sharding broadcast table rules;
+------------------------+
| shardingBroadcastTable |
+------------------------+
| t_1                    |
| t_2                    |
+------------------------+
2 rows in set (0.00 sec)
```

menghaoranss:
**SHOW READWRITE_SPLITTING RULES (FROM schemaName)?**


_static_
```
mysql> show readwrite_splitting rules;
+------------+-------------------------+---------------------+--------------------------+------------------+------------------------+
| name       | autoAwareDataSourceName | writeDataSourceName | readDataSourceNames      | loadBalancerType | loadBalancerProps      |
+------------+-------------------------+---------------------+--------------------------+------------------+------------------------+
| ms_group_0 | NULL                    | ds_primary          | [ds_slave_0, ds_slave_1] | random           | {}                     |
+------------+-------------------------+---------------------+--------------------------+------------------+------------------------+
1 row in set (0.00 sec)
```

_dynamic_
```
mysql> show readwrite_splitting rules from readwrite_splitting_db;
[...]

PR #10479, opened 2021-05-26
Show readwrite splitting & db discovery rules

menghaoranss:
For #10454

### DIFF
--- a/shardingsphere-proxy/shardingsphere-proxy-backend/src/main/java/org/apache/shardingsphere/proxy/backend/text/distsql/rql/impl/DatabaseDiscoveryRulesQueryBackendHandler.java
+++ b/shardingsphere-proxy/shardingsphere-proxy-backend/src/main/java/org/apache/shardingsphere/proxy/backend/text/distsql/rql/impl/DatabaseDiscoveryRulesQueryBackendHandler.java
@@ -1,0 +1,89 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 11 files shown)


shardingsphere-proxy/shardingsphere-proxy-backend/src/main/java/org/apache/shardingsphere/proxy/backend/text/distsql/rql/impl/DatabaseDiscoveryRulesQueryBackendHandler.java
```diff
@@ -0,0 +1,89 @@
...
+    private void loadRuleConfiguration(final String schemaName) {
+        Optional<DatabaseDiscoveryRuleConfiguration> ruleConfig = ProxyContext.getInstance().getMetaData(schemaName).getRuleMetaData().getConfigurations()
+                .stream().filter(each -> each instanceof DatabaseDiscoveryRuleConfiguration).map(each -> (DatabaseDiscoveryRuleConfiguration) each).findAny();
+        data = ruleConfig.map(optional -> optional.getDataSources().iterator()).orElse(Collections.emptyIterator());
+        discoverTypes = ruleConfig.map(DatabaseDiscoveryRuleConfiguration::getDiscoveryTypes).orElse(Maps.newHashMap());
+    }
+    
+    private List<QueryHeader> getQueryHeader(final String schemaName) {
+        List<QueryHeader> result = new LinkedList<>();
+        result.add(new QueryHeader(schemaName, "", "name", "name", Types.CHAR, "CHAR", 255, 0, false, false, false, false));
```
taojintianxia, line 70: I'm not quite familiar with the logic here, but the QueryHeader looks like magic number style. is that better if we define some constant QueryHeader like `public final QueryHeader XXX_QUERY_HEADER=new QueryHeader(schemaName, "", "name", "name", Types.CHAR, "CHAR", 255, 0, false, false, false, false)` ? or after 3 month , it's hard to understand the logic here even by yourself.

menghaoranss: Good idea, I will refactor later

tristaZero (approved):

PR #10544, opened 2021-05-29
fix sqlserver table metadata load exception

strongduanmu:
- fix execute sql exception
- fix nullpoint exception


tristaZero (approved):

strongduanmu:
Hi @totalo, because you are the author of this code, can you help with double check?

totalo:
> Hi @totalo, because you are the author of this code, can you help with double check?

of course, i found that  `sysindexes` is a system table, `sys.indexes` is a view, and there are more results from using `sysindexes`, I am not sure which one we need? `sys.objects `is the same as him.

Issue #10545, opened 2021-05-29
[Examples] shadow-example test issue


RaigorJiang:
For #10455

The `shadow-example` module test issue

- shadow-raw-jdbc-example
  1. some schems  are not defined in `nanual_schema.sql`:

  &emsp;&emsp; shadow_demo_ds_0, shadow_demo_ds_1, demo_shadow_write_ds
  
  2. `ShardingType` in `DataSourceFactory` and `YamlDataSourceFactory` not match.
  
- shadow-spring-boot-example
  1. `readwrite-splitting-shadow-databases`  
  	 shadow rule configuration error
  	
  2. `sharding-shadow-databases` 
     shadow rule configuration error
  
  3. `encrypt-shadow-databases`
     shadow rule configuration error, and An exception occurs:
     
     ```
     Caused by: org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'pwd_encryptor': Error setting property values; nested exception is org.springframework.beans.NotWritablePropertyException: Invalid property 'props' of bean class [org.apache.shardingsphere.example.fixture.TestQueryAssistedShardingEncryptAlgorithm]: Bean property 'props' is not writable or has an invalid setter method. Does the parameter type of the setter match the return type of the getter?
     	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.applyPropertyValues(AbstractAutowireCapableBeanFactory.java:1571)
     ```
  
- `shadow-spring-namespace-example`
  
   1. test with `application-encrypt-shadow-databases.xml` got exception:
```
Exception in thread "main" org.mybatis.spring.MyBatisSystemException: nested exception is org.apache.ibatis.exceptions.PersistenceException: 
### Error querying database.  Cause: java.lang.IndexOutOfBoundsException: Index: 0, Size: 0
### The error may exist in file [E:\gitee\shardingsphere\examples\example-core\example-spring-mybatis\target\classes\META-INF\mappers\ShadowUserMapper.xml]
### The error may involve org.apache.shardingsphere.example.core.mybatis.repository.MybatisShadowUserRepository.selectAllByShadow
### The error occurred while handling results
### SQL: SELECT * FROM t_user WHERE shadow = ?;
### Cause: java.lang.IndexOutOfBoundsException: Index: 0, Size: 0
	at org.mybatis.spring.MyBatisExceptionTranslator.translateExceptionIfPossible(MyBatisExceptionTranslator.java:79)
[...]

PR #10550, opened 2021-05-30
fix shadow-raw-jdbc-example and shadow-spring-boot-example

RaigorJiang:
For #10545
fix shadow-raw-jdbc-example and shadow-spring-boot-example


tristaZero (approved):
Nice, @RaigorJiang

PR #10551, opened 2021-05-30
remove UserChangedListenerFactory from governance doc

huanghao495430759:
- remove UserChangedListenerFactory from governance doc


tristaZero (approved):
Hi @huanghao495430759 ,

Thanks for your successive PRs. Actually, I am a little worried about these conflicts, whereas glad to see you fixed them and still update doc.
Your PRs are expected to be valued as an important feature for the next release. That's why you got my nudges a couple of times.
Thanks for your efforts, and if you like, please stay tuned here since we sincerely need you, another potential Apache committer.

PR #10642, opened 2021-06-03
fix calcite parse error when sql include semicolon

strongduanmu:
- fix calcite parse error when sql include semicolon
- fix check error when result include decimal type


shardingsphere-infra/shardingsphere-infra-executor/src/main/java/org/apache/shardingsphere/infra/executor/sql/federate/execute/FederateJDBCExecutor.java
```diff
@@ -82,7 +82,7 @@ public ResultSet getResultSet() throws SQLException {
     }
     
     private ResultSet execute(final String sql, final List<Object> parameters) throws SQLException {
-        PreparedStatement statement = getConnection().prepareStatement(sql);
+        PreparedStatement statement = getConnection().prepareStatement(sql.replace(";", ""));
```
TeslaCN, line 85: What if the SQL contains a string contains `;`?

strongduanmu: > contains

@TeslaCN Yes, it may delete the wrong semicolon.

strongduanmu:
Covered by PR #10719.

PR #10672, opened 2021-06-05
Refactor PostgreSQL Proxy

### LINKED ISSUE #10656: PostgreSQL Proxy may return unexpected Row Description in prepared select statement
master

Proxy

### Actual behavior

Client sent:
```
PostgreSQL
    Type: Bind
    Length: 22
    Portal: 
    Statement: S_1
    Parameter formats: 1
    Parameter values: 1
    Result formats: 0
PostgreSQL
    Type: Execute
    Length: 9
    Portal: 
    Returns: all rows
PostgreSQL
    Type: Sync
    Length: 4
```

Server response:
```
PostgreSQL
    Type: Bind completion
    Length: 4
PostgreSQL
    Type: Row description
    Length: 143
    Field count: 5
PostgreSQL
    Type: Data row
    Length: 119
    Field count: 5
PostgreSQL
    Type: Command completion
    Length: 7
    Tag:  0
```

```java
try (PreparedStatement preparedStatement = connection.prepareStatement("select * from  t_order limit ?")) {
    for (int i = 0; i < 6; i++) {
        preparedStatement.setInt(1, 1);
        try (ResultSet resultSet = preparedStatement.executeQuery()) {
            while (resultSet.next()) {
                System.out.println(resultSet.getLong("order_id"));
            }
        }
    }
}
```

TeslaCN:
Related to #10656 .

- Add ThreadLocal variable `PostgreSQLConnectionContext` in PostgreSQLCommandExecuteEngine.
- Command Bind, Describe, Close executors will pending until executing `PostgreSQLComExecuteExecutor`.

### DIFF
--- a/shardingsphere-db-protocol/shardingsphere-db-protocol-postgresql/src/main/java/org/apache/shardingsphere/db/protocol/postgresql/packet/command/query/PostgreSQLNoDataPacket.java
+++ b/shardingsphere-db-protocol/shardingsphere-db-protocol-postgresql/src/main/java/org/apache/shardingsphere/db/protocol/postgresql/packet/command/query/PostgreSQLNoDataPacket.java
@@ -15,23 +15,24 @@
  * limitations under the License.
  */
 
-package org.apache.shardingsphere.proxy.frontend.postgresql.command.query.binary.execute;
+package org.apache.shardingsphere.db.protocol.postgresql.packet.command.query;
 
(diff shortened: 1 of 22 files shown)


terrymanu (changes requested):
ThreadLocal is difficult to understand in the complex system.
In all ShardingSphere codes (except Hint and XA for their own implementation), we do not use thread local.
So, it is better to avoid use it. We need to discuss and redesign later.

PR #10676, opened 2021-06-06
10374 - code refactoring for process report feature

### LINKED ISSUE #10374: Clean code for Process report feature
Hey,

Welcome! This is a very simple task for newcomers, please collaborate in this community from this one.

#### How
Look at my comment of #10351 and fix it.

SivaTharun:
Fixes #10374 .

- replaced the if condition with ternary operator 
- used collection utils and objects class for null check, instead of comparison


tristaZero (changes requested):
Thanks @SivaTharun 
It seems CI failed, could you give it a look?

shardingsphere-infra/shardingsphere-infra-executor/src/main/java/org/apache/shardingsphere/infra/executor/sql/execute/engine/raw/RawExecutor.java
```diff
@@ -62,10 +63,8 @@
             // TODO Load query header for first query
             List<ExecuteResult> results = execute(executionGroupContext, (RawSQLExecutorCallback) null, callback);
             ExecuteProcessEngine.finish(executionGroupContext.getExecutionID());
-            if (null == results || results.isEmpty() || null == results.get(0)) {
-                return Collections.singleton(new UpdateResult(0, 0L));
-            }
-            return results;
+            return CollectionUtils.isEmpty(results) || Objects.isNull(results.get(0)) ? Collections
+                .singleton(new UpdateResult(0, 0L)) : results;
```
tristaZero, line 67: Hi @SivaTharun ,
Do you think `return null == results || results.isEmpty() || null == results.get(0) ? Collections.singleton(new UpdateResult(0, 0L)) : results`

tristaZero (changes requested):

SivaTharun:
closing the PR, as mistakenly used git rebase instead of merge on master branch.

shardingsphere-infra/shardingsphere-infra-executor/src/main/java/org/apache/shardingsphere/infra/executor/sql/execute/engine/raw/RawExecutor.java
SivaTharun: Hi @tristaZero,   `CollectionUtils.isEmpty` method inside apache common,  would check for `results == null || results.isEmpty()`. That essentially would cover both the checks.  i guess you don't want to have any dependency on apache commons library.

Also, `Objects.isNull` is the java utility method of `Objects` class, which encompasses the change for `null == results.get(0)`. i thought of using it, resulting in cleaner looking code.

SivaTharun:
> Thanks @SivaTharun
> It seems CI failed, could you give it a look?

@tristaZero i missed adding the import for `Objects` class in the PR, i have added it back, the PR check is passing now.

shardingsphere-infra/shardingsphere-infra-executor/src/main/java/org/apache/shardingsphere/infra/executor/sql/execute/engine/raw/RawExecutor.java
tristaZero: Hi @SivaTharun Thanks for your clarification which is fine with me.

tristaZero (approved):

tristaZero:
Thanks, @SivaTharun.

PR #10681, opened 2021-06-06
Adjust datatype

### LINKED ISSUE #10680: The shadow table mapping example has a field type problem
## Bug Report

5.0.0-RC1-SNAPSHOT
ShardingSphere-JDBC

### Expected behavior
springboot-jdbc example project . The normal execution statement should be: SELECT
org.apache.shardingsphere.example.core.jpa.service.ShadowUserServiceImpl#printData
userRepository.selectAll()

### Actual behavior
Actual SQL: shadow_ds ::: select shadowuser0_.user_id as user_id1_0_, shadowuser0_.pwd as pwd2_0_, shadowuser0_.shadow as shadow3_0_, shadowuser0_.user_name as user_nam4_0_ from t_shadow_user shadowuser0_  
Exception in thread "main" org.springframework.orm.jpa.JpaSystemException: Null value was assigned to a property of primitive type setter of org.apache.shardingsphere.example.core.api.entity.ShadowUser.shadow; nested exception is org.hibernate.PropertyAccessException: Null value was assigned to a property of primitive type setter of org.apache.shardingsphere.example.core.api.entity.ShadowUser.shadow

Since the shadow database type is bit(1), the corresponding Java data type is Boolean,
Although the shadow is a logical field and does not store the actual data, it may be found to be null when querying.
A boolean is bound to give an error.It should be a Boolean.
So the entity ShadowUserEntit defines the field shadow to be of type Boolean, not the base type boolean.

cocodroid:
Fixes #10680.

-
-
-

### DIFF
--- a/examples/example-core/example-api/src/main/java/org/apache/shardingsphere/example/core/api/entity/ShadowUser.java
+++ b/examples/example-core/example-api/src/main/java/org/apache/shardingsphere/example/core/api/entity/ShadowUser.java
@@ -33,7 +33,7 @@ public class ShadowUser implements Serializable {
     
     private String assistedQueryPwd;
     
-    private boolean shadow;
+    private Boolean shadow;
     
     public int getUserId() {
         return userId;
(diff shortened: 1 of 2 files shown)


terrymanu:
The value should not be null, could we adjust the value of example?

cocodroid:
@terrymanu No, because ShardingSphere will remove the shadow fields, so the database table fields will be empty when saved

tristaZero:
Thanks, @cocodroid
@terrymanu Since you got the context of this one, please give it a look again?

terrymanu:
I don't think the shadow is designed as good as enough.
I prefer to redesign it again.
But for now, it is the best way to resolve the current problem. 
Let me merge the pr.

terrymanu (approved):

PR #10727, opened 2021-06-09
Add schema prefix before table name to some SQL script in test

mlkui:
- Explictly add schema prefix before table names to some SQL scripts in test. For tools use 'Default Schema' like MySQL Workbench, these scripts should specify target schema through 'USE target_schema', or statements will be targted to the current 'Default Schema'. However, as 'USE' is not supported by all databases, like Postgre, we should then explictly add schema prefix before table names.


tristaZero (changes requested):
Hi please look at failed CI tests.

tristaZero:
Hi @mlkui Any progress?

tristaZero:
Hi I have to close this one since there is no feedback anymore.

Issue #10111, opened 2021-04-16
[SQL Definition Collation][Oracle]Collate the SQL definition of alter session


tristaZero:
This issue is the subsequent one from [#9603](https://github.com/apache/shardingsphere/pull/9603#issuecomment-796600731).

Oracle Ref -> https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlqr/SQL-Statements.html#GUID-CE009C87-4AA6-45BF-9A0D-2B2C5FFB47B3

tristaZero:
Hi @ThanoshanMV , this one is yours. :)
Could you comment here so that I can assign it to you?

ThanoshanMV:
Hi @tristaZero, I'll try this one.

ThanoshanMV:
Hi @tristaZero, next, I'll try [ALTER SYSTEM](https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/ALTER-SYSTEM.html#GUID-2C638517-D73A-41CA-9D8E-A62D1A0B7ADB).

tristaZero:
@ThanoshanMV Go ahead.
BTW, your GSoC 2021 application is approved now. Let us wait its timeline moving.

ThanoshanMV:
Hi @tristaZero, I'll start work on [Oracle SELECT statement](https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/SELECT.html#GUID-CFA006CA-6FF1-4972-821E-6996142A51C6) as it's hard and other DML statements depend on `SELECT` statement's rules.

tristaZero:
@ThanoshanMV Yep, but you can proofread it by successive PRs.

ThanoshanMV:
Okay @tristaZero, I'll proofread it by successive PRs.

tristaZero:
Done.

PR #10759, opened 2021-06-11
Add selectList, fromClause definitions for Oracle Select

ThanoshanMV:
#10111 

- Format some existing `ANTLR` rules in DMLStatement.
- Added SQL definition for [SELECT](https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/SELECT.html#GUID-CFA006CA-6FF1-4972-821E-6996142A51C6) statement's `selectList` and `fromClause`.
- Added corresponding test cases.
- Changed some `sql-case` database types to specific as some of them aren't supported for Oracle.


ThanoshanMV:
Hi @tristaZero, @wgy8283335 

While I was writing `ANTLR` definitions for `SELECT` statement's `selectList` and `fromClause`, I encountered an issue. 

After writing the definitions specified in this PR, I ran `mvn clean install` in the `shardingsphere-sql-parser-oracle` directory to auto-generate contexts but though the build succeeded, I had an error `Cannot resolve symbol 'OracleStatementParser'` in `OracleDMLStatementSQLVisitor`. 

OracleDMLError
 
Before writing these definitions as mentioned in the Code of Conduct, I ran `mvn clean install`. There was no issue in the `OracleDMLStatementSQLVisitor`. 

Surprisingly, I removed `selectFromClause` and its corresponding definitions and kept only `selectList` definition and its associated definitions such as `selectProjection` , `selectProjectionExprClause`  and ran `mvn clean install` and there were no errors. 

OracleDMLNoError

I think the problem is with `selectFromClause` but I couldn't able to identify it. Even if we write definitions incorrectly, it should create the contexts, right? I should have created an issue for this but since it's coming because of my definition, I created a draft PR to show the code to you.

wgy8283335:
@ThanoshanMV Hi, may be you could try `mvn clean install` in shardingsphere directory, or try to `maven->reload project` in IDEA.

ThanoshanMV:
Still couldn't able to resolve it. @wgy8283335 can you please check this code in your local and please let me know whether it works or not.

wgy8283335:
> Still couldn't able to resolve it. @wgy8283335 can you please check this code in your local and please let me know whether it works or not.

I've tried to reproduce the problem. And I think you could add rules one by one to find which rule causes the problem.
Why do you put DDLStatement in import? This is not recommended.
image

ThanoshanMV:
`inlineExternalTable` has `columnDefinition` and `inlineExternalTableProperties` has `externalTableDataProps` in the DMLStatement. Both `columnDefinition` and `externalTableDataProps` are in DDLStatement.

Initially I was able to put `externalTableDataProps` in BaseRule but couldn't able to put `columnDefinition` as its `inlineConstraint` has `createIndex` rule:

createIndexClause

I also agree on not to put DDLStatement in import. I've added  `externalTableDataProps` in BaseRule. How can we specify `columnDefinition` in DMLStatement?

ThanoshanMV:
I've removed `rowPatternClause`, `pivotClause` and `unpivotClause` from the definition. After removing them, I think now that error isn't coming. Please check on how it's like for you.

shardingsphere-sql-parser/shardingsphere-sql-parser-dialect/shardingsphere-sql-parser-oracle/src/main/antlr4/imports/oracle/BaseRule.g4
```diff
@@ -731,3 +735,39 @@ unitName
...
+
+condition
+    : expr
+    ;
+
+variableName
+    : identifier | STRING_
+    ;
+
+externalTableDataProps
```
wgy8283335, line 771: It's better to move the rule to DDLStatement. And rename the rule and copy the rule in DMLStatement.

wgy8283335:
> I've removed `rowPatternClause`, `pivotClause` and `unpivotClause` from the definition. After removing them, I think now that error isn't coming. Please check on how it's like for you.

Why do you remove `rowPatternClause`, `pivotClause` and `unpivotClause` ?

wgy8283335:
> > Why do you remove `rowPatternClause`, `pivotClause` and `unpivotClause` ?
> 
> Like you mentioned I tried to add rules one by one. I found out when trying to add `rowPatternClause`, `pivotClause` and `unpivotClause`, it gave me the error in `OracleDMLStatementSQLVisitor`:
> 
> Error
> 
> Even if I try to add one rule among those three rules, the error stays the same.
> 
> If I remove those three rules then there are no more errors and I can able to use auto-generated contexts:
> 
> NoError
> 
> I hope the above-mentioned errors aren't specific to me. Can you please check whether you can able to reproduce the same error?
There may be some bugs in your rule.

ThanoshanMV:
I'll check them.

ThanoshanMV:
Hi @wgy8283335. I checked the rules, I couldn't able to find where the bug resides. 

I added `rowPatternClause`, `pivotClause` and `unpivotClause`. I tried to reformat the rules in DMLStatement similar to DDLStatement and ensured that all rules in DMLStatement can't over 200 chars. Still, that issue persists. Can you please help me to find why this issue comes and to resolve it

wgy8283335:
> Hi @wgy8283335. I checked the rules, I couldn't able to find where the bug resides.
> 
> I added `rowPatternClause`, `pivotClause` and `unpivotClause`. I tried to reformat the rules in DMLStatement similar to DDLStatement and ensured that all rules in DMLStatement can't over 200 chars. Still, that issue persists. Can you please help me to find why this issue comes and to resolve it.

OK, I'll have a look on it.

wgy8283335:
@ThanoshanMV, hi. 
Maybe you could try to use this tool to generate rule.
image
https://github.com/wgy8283335/shardingsphere/tree/10104-oracle-antlr-transfer-tool

ThanoshanMV:
Hi @wgy8283335, Okay I'll check that.

ThanoshanMV:
Hi @wgy8283335, I've resolved the above issue 😄

Since the generated file size is too big, AppCode couldn't able to provide code assistance. We have to change the file size limit from the default value as specified in the [documentation](https://www.jetbrains.com/help/objc/configuring-file-size-limit.html#file-size-limit).

I'll try to finish the remaining works in this PR and let you know.

wgy8283335:
I'd like to take care of this merge request.

tristaZero:
So... it is ready to be merged?

wgy8283335:
> So... it is ready to be merged? @wgy8283335

Not yet. This pr is big, and I am still working on it.

wgy8283335 (approved):

Issue #10513, opened 2021-05-27
[GSoC 2021] Proofread the Oracle DDL/TCL SQL Definitions


Liangda-w:
This issue is related to the Google Summer of Code (GSoC) 2021 project - [Proofread the Oracle DDL/TCL SQL Definitions](https://summerofcode.withgoogle.com/projects/#5167124928004096). 
Meanwhile, it is a sub-issue of #6480.

### Background

ShardingSphere parser engine helps users parse a SQL to get the AST (Abstract Syntax Tree) and visit this tree to get SQLStatement (Java Object). At present, this parser engine can handle SQLs for MySQL, PostgreSQL, SQLServer, and Oracle, which means we have to understand different database dialects SQLs. 

This project is focused on proofreading all the **Oracle TCL** (Transaction Control Language) and **DDL** (Data Definition Language) SQL definitions except for ALTER, DROP, CREATE and TRUNCATE.

#### TCL
- [x] SET TRANSACTION
- [x] SET CONSTRAINT(S)

#### DDL
- [x] ANALYZE
- [x] ASSOCIATE STATISTICS
- [x] DISASSOCIATE STATISTICS
- [x] AUDIT (Unified Auditing)
- [x] NOAUDIT (Unified Auditing)
- [x] COMMENT
- [x] FLASHBACK DATABASE
- [x] FLASHBACK TABLE
- [x] PURGE
- [x] RENAME

Reference: Oracle [documentation](https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/SQL-Statements-ADMINISTER-KEY-MANAGEMENT-to-ALTER-JAVA.html#GUID-04D8EBB3-583D-4969-8344-E221000E9CD9) of SQL definitions

Liangda-w:
Remaining improvements: see discussion in #11126 - parse the following segments and add corresponding tests.

- [ ] function
- [ ] package

Liangda-w:
GSoC project completed:) Huge thanks to my mentors @wgy8283335 and @tristaZero

wgy8283335:
> GSoC project completed:) Huge thanks to my mentors @wgy8283335 and @tristaZero

Congratulations! It's my pleasure. @Liangda-w

tristaZero:
Hey, it's an enjoyable time to be your mentor during GSoC here. @Liangda-w Good luck!

PR #10768, opened 2021-06-11
Add Oracle SQL - ANALYZE

Liangda-w:
Relates #10513 

-  Add Oracle SQL - [ANALYZE](https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/ANALYZE.html#GUID-535CE98E-2359-4147-839F-DCB3772C1B0E)
-  Add corresponding test cases

@wgy8283335

### DIFF
--- a/shardingsphere-sql-parser/shardingsphere-sql-parser-engine/src/main/java/org/apache/shardingsphere/sql/parser/core/visitor/SQLVisitorRule.java
+++ b/shardingsphere-sql-parser/shardingsphere-sql-parser-engine/src/main/java/org/apache/shardingsphere/sql/parser/core/visitor/SQLVisitorRule.java
@@ -108,6 +108,8 @@ public enum SQLVisitorRule {
     
     DROP_VIEW("DropView", SQLStatementType.DDL),
 
+    ANALYZE("Analyze", SQLStatementType.DDL),
+
     CREATE_SEQUENCE("CreateSequence", SQLStatementType.DDL),
 
     ALTER_SEQUENCE("AlterSequence", SQLStatementType.DDL),
@@ -230,7 +232,7 @@ public enum SQLVisitorRule {
     
     CREATE_LOADABLE_FUNCTION("CreateLoadableFunction", SQLStatementType.DAL),
     
-    ANALYZE("AnalyzeTable", SQLStatementType.DAL),
+    ANALYZE_TABLE("AnalyzeTable", SQLStatementType.DAL),
     
     LOAD("Load", SQLStatementType.DAL),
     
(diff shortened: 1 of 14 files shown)


shardingsphere-sql-parser/shardingsphere-sql-parser-engine/src/main/java/org/apache/shardingsphere/sql/parser/core/visitor/SQLVisitorRule.java
```diff
@@ -108,6 +108,8 @@
     
     DROP_VIEW("DropView", SQLStatementType.DDL),
 
+    ANALYZE("Analyze", SQLStatementType.DDL),
```
wgy8283335, line 111: Why set `Analyze` as DDL?

shardingsphere-sql-parser/shardingsphere-sql-parser-test/src/main/java/org/apache/shardingsphere/test/sql/parser/parameterized/asserts/statement/ddl/DDLStatementAssert.java
```diff
@@ -87,6 +90,8 @@ public static void assertIs(final SQLCaseAssertContext assertContext, final DDLS
             AlterSessionStatementAssert.assertIs(assertContext, (OracleAlterSessionStatement) actual, (AlterSessionStatementTestCase) expected);
         } else if (actual instanceof AlterSystemStatement) {
             AlterSystemStatementAssert.assertIs(assertContext, (AlterSystemStatement) actual, (AlterSystemStatementTestCase) expected);
+        } else if (actual instanceof OracleAnalyzeStatement) {
```
wgy8283335, line 93: Are you sure that `analyze` should be put in the DDL？

wgy8283335 (changes requested):
The `Analyze` should be classified as DDL or DAL?

Liangda-w:
> The `Analyze` should be classified as DDL or DAL?

Hi @wgy8283335 I thought it will be DDL because of the [Oracle Doc](https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/Types-of-SQL-Statements.html#GUID-FD9A8CB4-6B9A-44E5-B114-EFB8DA76FC88):

image

But you're right, in [MySQL](https://dev.mysql.com/doc/refman/8.0/en/analyze-table.html) `ANALYZE` is classified as DAL. So do you think change to DAL is better?

wgy8283335:
> > The `Analyze` should be classified as DDL or DAL?
> 
> Hi @wgy8283335 I thought it will be DDL because of the [Oracle Doc](https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/Types-of-SQL-Statements.html#GUID-FD9A8CB4-6B9A-44E5-B114-EFB8DA76FC88):
> 
> image
> 
> But you're right, in [MySQL](https://dev.mysql.com/doc/refman/8.0/en/analyze-table.html) `ANALYZE` is classified as DAL. So do you think change to DAL is better?

I think DDL is better.

wgy8283335 (approved):

Issue #10618, opened 2021-06-02
group by + having not support


dbzzzzz:
### Is your feature request related to a problem?

In shardingsphere 5 proxy，group by + having does not support, the error is as below:
mysql> select USER_ID,sum(ORDER_ID) from t_order group by USER_ID having sum(ORDER_ID)>1100;
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'having sum(ORDER_ID)>1100' at line 1

strongduanmu:
PR #10620 fixes the grammatical problem of the wrong order after adding the `order by` clause when the `select ... group by ... having ... statement` is rewritten, but `having` clause is not handled correctly.

![image](https://user-images.githubusercontent.com/10829171/120632545-aad51b80-c49b-11eb-8f92-378dfa3d8111.png)

We need to remove the `having` clause in the rewriting logic, and filter the having conditions after group merge.

PR #10848, opened 2021-06-17
optimize select group by having with calcite

strongduanmu:
Ref #10618.

- optimize select group by having with calcite
- delete `HavingDecoratorMergedResult` and corresponding logic
- fix queryResults loss when execute `moveNext` in `FederateRowEnumerator`


shardingsphere-infra/shardingsphere-infra-executor/src/main/java/org/apache/shardingsphere/infra/executor/sql/federate/schema/table/generator/FederateExecutionSQLGenerator.java
```diff
@@ -35,14 +38,17 @@
...
      * Generate sql.
      *
      * @param table table
      * @return sql
      */
     public String generate(final String table) {
-        // TODO
-        return String.format("SELECT * FROM %s", table);
+        // TODO generate sql with filters
+        String project = null == projects || 0 == projects.length ? "*" : Arrays.stream(projects).mapToObj(each -> fields.get(each).getName()).collect(Collectors.joining(", "));
```
tristaZero, line 51: Which condition will make`project = null == projects || 0 == projects.length`  true?

shardingsphere-infra/shardingsphere-infra-optimize/src/main/java/org/apache/shardingsphere/infra/optimize/core/metadata/FederateTableMetadata.java
```diff
@@ -88,4 +91,13 @@ private DataSource getActualDataSource(final Map<String, DataSource> dataSources
         }
         return dataSources.get(result);
     }
+    
+    /**
+     * Get rel data type field.
+     * 
+     * @return rel data type field collection
+     */
+    public List<RelDataTypeField> getRelDataTypeField() {
```
tristaZero, line 100: I have no idea how long a bunch of calls for `relProtoDataType.apply(TYPE_FACTORY)` will take up. But suppose we just wanna get the list of columns, do you think it will be better to cache these column names here from `TableMetaData` or this function?

tristaZero (changes requested):
It seems we will support more complex SQL cases after this PR merged. Nice.

shardingsphere-infra/shardingsphere-infra-executor/src/main/java/org/apache/shardingsphere/infra/executor/sql/federate/schema/table/generator/FederateExecutionSQLGenerator.java
strongduanmu: > projects

@tristaZero It can be seen from the source code that when the original projects and the optimized projects are the same, projectInts will return null. Maybe the `0 == projects.length` condition is redundant.

```java
  private static TableScanNode createProjectableFilterable(Compiler compiler,
      TableScan rel, ImmutableList<RexNode> filters, ImmutableIntList projects,
      ProjectableFilterableTable pfTable) {
    final DataContext root = compiler.getDataContext();
    final ImmutableIntList originalProjects = projects;
    for (;;) {
      final List<RexNode> mutableFilters = Lists.newArrayList(filters);
      final int[] projectInts;
      if (projects == null
          || projects.equals(TableScan.identity(rel.getTable()))) {
        projectInts = null;
      } else {
        projectInts = projects.toIntArray();
      }
      final Enumerable<Object[]> enumerable1 =
          pfTable.scan(root, mutableFilters, projectInts);
...
}
```

shardingsphere-infra/shardingsphere-infra-optimize/src/main/java/org/apache/shardingsphere/infra/optimize/core/metadata/FederateTableMetadata.java
strongduanmu: > I have no idea how long a bunch of calls for `relProtoDataType.apply(TYPE_FACTORY)` will take up. But suppose we just wanna get the list of columns, do you think it will be better to cache these column names here from `TableMetaData` or this function?

@tristaZero Yes, this does have performance problems, and I will consider caching column names.

shardingsphere-infra/shardingsphere-infra-optimize/src/main/java/org/apache/shardingsphere/infra/optimize/core/metadata/FederateTableMetadata.java
```diff
@@ -88,4 +96,13 @@ private DataSource getActualDataSource(final Map<String, DataSource> dataSources
         }
         return dataSources.get(result);
     }
+    
+    /**
+     * Get column names.
+     * 
+     * @return column names collection
+     */
+    public List<String> getColumnNames() {
```
tristaZero, line 105: Maybe a @Getter is another better option?

shardingsphere-infra/shardingsphere-infra-executor/src/main/java/org/apache/shardingsphere/infra/executor/sql/federate/schema/table/generator/FederateExecutionSQLGenerator.java
```diff
@@ -35,14 +37,17 @@
...
      * Generate sql.
      *
      * @param table table
      * @return sql
      */
     public String generate(final String table) {
-        // TODO
-        return String.format("SELECT * FROM %s", table);
+        // TODO generate sql with filters
+        String project = null == projects ? "*" : Arrays.stream(projects).mapToObj(columnNames::get).collect(Collectors.joining(", "));
```
tristaZero, line 50: Considering `projects` is for SQL optimization, at this moment, we can consider returning the result only with `Arrays.stream(projects).mapToObj(columnNames::get).collect(Collectors.joining(", ")`. How do you think?

tristaZero (changes requested):

shardingsphere-infra/shardingsphere-infra-optimize/src/main/java/org/apache/shardingsphere/infra/optimize/core/metadata/FederateTableMetadata.java
strongduanmu: > Maybe a @getter is another better option?

I will optimize it.

shardingsphere-infra/shardingsphere-infra-executor/src/main/java/org/apache/shardingsphere/infra/executor/sql/federate/schema/table/generator/FederateExecutionSQLGenerator.java
strongduanmu: > Considering `projects` is for SQL optimization, at this moment, we can consider returning the result only with `Arrays.stream(projects).mapToObj(columnNames::get).collect(Collectors.joining(", ")`. How do you think?

@tristaZero Good suggestion, I will replace all columns with `*`.

tristaZero (approved):

PR #11084, opened 2021-07-01
[bug:authority] fix after refreshing authority,The effect is not as  expected

lwclover:
Under ALL_PRIVILEGES_PERMITTED，after refreshing authority,The effect is not as  expected，that means partial ShardingSphereUsers are missing.
Under NATIVE, proxy when run grant cmd, will be throwing Unknow exception: [SQL check failed, Error message: .]that means partial ShardingSphereUsers are missing.


tristaZero (approved):
@lwclover Thanks for this refactoring work.

lwclover:
OK，please walk to pull request #11099

Issue #11075, opened 2021-06-30
Support SQL statement execution with schema


strongduanmu:
## Feature Request

Now, Proxy supports the configuration of multiple logic databases. The `use sharding_db` statement can be used to switch the logic databases, but it is not convenient. It is hoped to support the execution of SQL statements with schema and route to the corresponding logic database for execution according to the schema specified in the SQL statement.

For example, we have configured two logical databases, `sharding_db_01` and `sharding_db_02`, and executed the following SQL, hoping to be routed to the `sharding_db_01` database for execution.

```sql
USE sharding_db_02;
SELECT * FROM sharding_db_01.test;
```
When the current SQL does not specify a schema, routing is performed according to the currently used schema.

```sql
SELECT * FROM test;
```

strongduanmu:
After the merger of PR #11150, the master branch can support the following statements with all features.

## DAL Statement

* AnalyzeTableStatement
* ExplainStatement
* MySQLShowColumnsStatement
* MySQLShowCreateTableStatement
* MySQLShowIndexStatement

## DCL Statement

* GrantStatement
* RevokeStatement

## DDL Statement

* AlterTableStatement
* CreateIndexStatement
* CreateTableStatement
* DropIndexStatement
* DropTableStatement
* TruncateStatement

## DML Statement

* DeleteStatement
* InsertStatement
* SelectStatement
* UpdateStatement

strongduanmu:
When we execute sql after proxy start, this exception will occur:

![image](https://user-images.githubusercontent.com/10829171/126732201-53433c46-e377-44a0-b7dc-ba8d5ca92cfc.png)

When executing SQL without tables, like `select @@version_comment limit 1;`, the logic for obtaining the schema needs to be optimized.

strongduanmu:
Multiple different schemas containing the same table lead to incorrect routing results.

![image](https://user-images.githubusercontent.com/10829171/126953403-23226542-3e43-4aa0-ab99-bd56bfb67565.png)

strongduanmu:
All work has been completed.

PR #11091, opened 2021-07-01
extract SchemaAvailable interface & add validate for SelectStatement

strongduanmu:
Ref #11075.

- extract SchemaAvailable interface for Proxy to handle SQL with schema
- add default method `initSchemaBasedContext` in `SQLStatmentContext` to split context init logic
- add schema validate for SelectStatement


shardingsphere-jdbc/shardingsphere-jdbc-core/src/main/java/org/apache/shardingsphere/driver/jdbc/core/statement/ShardingSpherePreparedStatement.java
```diff
@@ -367,7 +367,8 @@ private ExecutionContext createExecutionContext() {
     private LogicSQL createLogicSQL() {
         List<Object> parameters = new ArrayList<>(getParameters());
         ShardingSphereSchema schema = metaDataContexts.getDefaultMetaData().getSchema();
-        SQLStatementContext<?> sqlStatementContext = SQLStatementContextFactory.newInstance(schema, parameters, sqlStatement);
+        SQLStatementContext<?> sqlStatementContext = SQLStatementContextFactory.newInstance(parameters, sqlStatement);
+        sqlStatementContext.initSchemaBasedContext(schema);
```
tristaZero, line 371: Is it necessary to use `initSchemaBasedContext` instead of `SQLStatementContextFactory.newInstance`?

tristaZero (changes requested):

shardingsphere-jdbc/shardingsphere-jdbc-core/src/main/java/org/apache/shardingsphere/driver/jdbc/core/statement/ShardingSpherePreparedStatement.java
strongduanmu: > Is it necessary to use `initSchemaBasedContext` instead of `SQLStatementContextFactory.newInstance`?

@tristaZero Because it is currently dependent on the schema when initializing the context, but the schema is not clear when using the context to extract the schema, so the context initialization needs to be split into two steps. The first step is to initialize the context independent of the schema, and the second step is to initialize the context related to the schema.

tristaZero (approved):

Issue #11100, opened 2021-07-02
[DistSQL] Enhance the 'AddResource' syntax to allow custom connection properties


RaigorJiang:
Hi community,

When I use DistSQL to add resource, the proxy always gives an alert because the `useSSL` parameter is not specified when creating a  connection.
> WARN: Establishing SSL connection without server's identity verification is not recommended. According to MySQL 5.5.45+, 5.6.26+ and 5.7.6+ requirements SSL connection must be established by default if explicit option isn't set. For compliance with existing applications not using SSL the verifyServerCertificate property is set to 'false'. You need either to explicitly disable SSL by setting useSSL=false, or set useSSL=true and provide truststore for server certificate verification.

Similarly, if use the driver of version 8.0 or higher, `serverTimezone` is also an indispensable parameter.
Therefore, I consider enhancing the `add resource` function to allow users to configure the above parameters, as well as any other optional parameters.

#### Before:

```
ADD RESOURCE dataSource [, dataSource] ...

dataSource:
    dataSourceName(HOST=hostName,PORT=port,DB=dbName,USER=user [, PASSWORD=password])
```

#### after:

```
ADD RESOURCE dataSource [, dataSource] ...

dataSource:
    dataSourceName(HOST=hostName,PORT=port,DB=dbName,USER=user [, PASSWORD=password][,PROPERTIES("key"="value","key"="value")])
```

for example:

```
ADD RESOURCE resource_0 (
    HOST=127.0.0.1,
    PORT=3306,
    DB=db0,
    USER=root,
    PASSWORD=root,
    PROPERTIES("useSSL"="false","serverTimezone"="UTC")
);

```

If you have any suggestions, welcome to discuss.

reference:
- [DistSQL: DATA SOURCE](https://shardingsphere.apache.org/document/current/en/features/dist-sql/syntax/rdl/rdl-resource/)
- [MySQL: connector-j-connp-props-security](https://dev.mysql.com/doc/connector-j/8.0/en/connector-j-connp-props-security.html)
- [MySQL: changes with Connector/J 8.0](https://dev.mysql.com/doc/connector-j/8.0/en/connector-j-other-changes.html)
- [PostgreSQL: Initializing the Driver](https://jdbc.postgresql.org/documentation/head/connect.html)

terrymanu:
Good idea

RaigorJiang:
[...]

PR #11101, opened 2021-07-02
For #11100, add custom properties for `add resource`;

RaigorJiang:
For #11100, add custom properties for `add resource`;

### DIFF
--- a/shardingsphere-distsql-parser/shardingsphere-distsql-parser-engine/src/main/antlr4/imports/RDLStatement.g4
+++ b/shardingsphere-distsql-parser/shardingsphere-distsql-parser-engine/src/main/antlr4/imports/RDLStatement.g4
@@ -64,5 +64,5 @@ connectionProperties
     ;
 
 connectionProperty
-    : key=(IDENTIFIER | STRING) EQ value=(NUMBER | INT | STRING)
+    : key=IDENTIFIER EQ value=IDENTIFIER
     ;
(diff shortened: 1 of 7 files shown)


terrymanu (approved):

shardingsphere-distsql-parser/shardingsphere-distsql-parser-engine/src/main/antlr4/imports/RDLStatement.g4
```diff
@@ -58,3 +58,11 @@ user
 password
     : IDENTIFIER | INT | STRING
     ;
+
+connectionProperties
+    : connectionProperty (COMMA connectionProperty)*
+    ;
+
+connectionProperty
+    : key=IDENTIFIER EQ value=IDENTIFIER
```
menghaoranss, line 67: Other `g4` file  define  properties as follows:
```
key=(IDENTIFIER | STRING) EQ value=(NUMBER | INT | STRING)
```

please confirm whether need to consider consistency.

RaigorJiang: `key=(IDENTIFIER | STRING) EQ value=(NUMBER | INT | STRING)`
Can not match the following statement: 
```
"useSSL"="false"
```

RaigorJiang: We may update to `key=(IDENTIFIER | STRING) EQ value=(IDENTIFIER | STRING)`

PR #11140, opened 2021-07-04
For #11100 , add additional properties to create data source.

RaigorJiang:
For #11100 , add additional properties to create data source.


tristaZero (changes requested):
I am a little confused,

- Except for `SSL` and `timeZone`, will we have other props?
- How to use these two props to set dataSource
- Are they similar to `readOnly` and other props?

RaigorJiang:
> I am a little confused,
> 
> * Except for `SSL` and `timeZone`, will we have other props?
> * How to use these two props to set dataSource
> * Are they similar to `readOnly` and other props?

- Yes, users can define any other properties for JDBC connector.  Refer to [MySQL:configuration-properties](https://dev.mysql.com/doc/connector-j/8.0/en/connector-j-reference-configuration-properties.html)
- For example :  ```ADD RESOURCE resource_0 (
    HOST=127.0.0.1,
    PORT=3306,
    DB=db0,
    USER=root,
    PASSWORD=root,
    PROPERTIES("useSSL"="false","serverTimezone"="UTC", "tinyInt1isBit"="false"));```
- No, `readOnly` is a prop for `HikariDataSource`, as well as `maxPoolSize`, not for `JDBC Connection`.

RaigorJiang:
Design adjustment,  this PR will be closed. please refer to #11100.

PR #11200, opened 2021-07-07
Delete the common configuration, adjust the position of datasource.driver-class-name and type, because the code in #8355 has deleted the reading of the common configuration

yx9o:
Delete the common configuration, adjust the position of datasource.driver-class-name and type, because the code in #8355 has deleted the reading of the common configuration


tristaZero (approved):
Thanks for tweaking @yx9o

Issue #10861, opened 2021-06-18
Double quote is removed when using AVG aggregation function 


ysyysjbama:
## Bug Report
* Sharding-JDBC version:  4.1.1 
* database : PostgreSQL
  In my database , table column is  **uppercase**,  because PostgreSQL is case sensitive,I have to use **double quote** in the sql query column . It is OK until I use AVG aggregation function in my query statement.
The original SQL :
```
SELECT 
	AVG ( A."DIRECTION" )
FROM
	b_flow_vihicle_link AS A 
	LEFT JOIN b_flow_vihicle_travel b ON A."VIHICLE_TRAVEL_ID" = b."ID"
	LEFT JOIN t_vihicle_info C ON b."VIHICLE_ID" = C."ID" 
```

the rewrited SQL (Sharding-JDBC ):
```
SELECT 
	AVG ( A."DIRECTION" ),
	COUNT ( A.DIRECTION ) AS AVG_DERIVED_COUNT_0,
	SUM ( A.DIRECTION ) AS AVG_DERIVED_SUM_0 
FROM
	b_flow_vihicle_link_34 AS A 
	LEFT JOIN b_flow_vihicle_travel_34 b ON A."VIHICLE_TRAVEL_ID" = b."ID"
	LEFT JOIN t_vihicle_info C ON b."VIHICLE_ID" = C."ID" 
```
In the rewrited SQL , COUNT ( A.DIRECTION ) and SUM ( A.DIRECTION ) are generated by Sharding-JDBC , but the double quote of A.DIRECTION is removed , so I get the exection like this ： org.postgresql.util.PSQLException: ERROR: column A.direction does not exist.

## Reason analyze (If you can)
* I found the code in the class **org.apache.shardingsphere.sql.parser.sql.util.SQLUtil** where the double quote is removed ,but I don‘t know why it should be removed. 
thank you !

strongduanmu:
@ysyysjbama Thank you for your feedback, are you interested in submitting a PR to fix this problem?

iamkyu:
@ysyysjbama @strongduanmu 
any progress? I interested in this issue

strongduanmu:
> @ysyysjbama @strongduanmu
> any progress? I want to work for this issue if nothing has started

@iamkyu Welcome, i will assign this issue to you.

PR #11204, opened 2021-07-07
Fix AggregationProjection return expression without cleaning

iamkyu:
Fixes https://github.com/apache/shardingsphere/issues/10861

- Fix AggregationProjection return expression without cleaning

---
- I Checked with only MySQL (5.7.34 / 8.0.25) and Postgres (10.5)
- Two databases have different column identifiers spec
- Postgres needs to double-quotes for using upper case column name.
- However, MySQL needs to change SQL mode ([ANSI_QUOTES](https://dev.mysql.com/doc/refman/8.0/en/sql-mode.html#sqlmode_ansi_quotes)) for using double-quotes.
- Therefore, AggregationProjection returns expression as passed from Original SQL.

### DIFF
--- a/shardingsphere-infra/shardingsphere-infra-binder/src/main/java/org/apache/shardingsphere/infra/binder/segment/select/projection/impl/AggregationProjection.java
+++ b/shardingsphere-infra/shardingsphere-infra-binder/src/main/java/org/apache/shardingsphere/infra/binder/segment/select/projection/impl/AggregationProjection.java
@@ -24,7 +24,6 @@ import lombok.Setter;
 import lombok.ToString;
 import org.apache.shardingsphere.infra.binder.segment.select.projection.Projection;
 import org.apache.shardingsphere.sql.parser.sql.common.constant.AggregationType;
-import org.apache.shardingsphere.sql.parser.sql.common.util.SQLUtil;
 
 import java.util.ArrayList;
 import java.util.List;
@@ -52,7 +51,7 @@ public class AggregationProjection implements Projection {
     
     @Override
     public final String getExpression() {
-        return SQLUtil.getExactlyValue(type.name() + innerExpression);
+        return type.name() + innerExpression;
     }
     
     @Override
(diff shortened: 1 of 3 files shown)


tristaZero (changes requested):
@iamkyu Hi, thanks for your PR, could you add a unit test for it?

iamkyu:
Hi @tristaZero , 
I Added some test cases

tristaZero:
Thanks @iamkyu , @strongduanmu  could you have a double-check?

shardingsphere-infra/shardingsphere-infra-binder/src/main/java/org/apache/shardingsphere/infra/binder/segment/select/projection/impl/AggregationProjection.java
```diff
@@ -52,7 +51,7 @@
     
     @Override
     public final String getExpression() {
-        return SQLUtil.getExactlyValue(type.name() + innerExpression);
```
strongduanmu, line 55: @iamkyu I found that there are many scenarios where `getExpression` is used for logical judgment. Should these scenarios use `SQLUtil.getExactlyValue(each.getExpression())` to ensure that exact values are used for judgment?

iamkyu: @strongduanmu thank you for the review.
I'll check it out. 
Or do you have any better ideas than fix the `AggregationProjection#getExpression()`?

tristaZero:
Hi @iamkyu @strongduanmu How is it going on?

iamkyu:
Hi @tristaZero , Working in progress. I need more time

strongduanmu:
> Hi @tristaZero , Working in progress. I need more time

@iamkyu Great, please go ahead. If you need help, please ping me.

shardingsphere-infra/shardingsphere-infra-binder/src/main/java/org/apache/shardingsphere/infra/binder/segment/select/projection/impl/AggregationProjection.java
iamkyu: Hi @strongduanmu,
I checked where "`getExpression` is used, 
and corrected some scenarios where exact values are needed.

tristaZero:
Hi I just saw @iamkyu pushed some commits here. @strongduanmu Please have a look.

strongduanmu (approved):

strongduanmu:
@iamkyu Thank you for your efforts, we will merge this pr later.

tristaZero (approved):
@iamkyu Aha, nice.

PR #11210, opened 2021-07-08
Modify the sharding doc

zjcnb:
Modify the sharding doc


examples/shardingsphere-parser-example/src/main/java/org/apache/shardingsphere/example/parser/mysql/statement/MySQLParserStatementExample.java
```diff
@@ -0,0 +1,63 @@
...
+import org.antlr.v4.runtime.tree.ParseTree;
+import org.apache.shardingsphere.sql.parser.api.SQLParserEngine;
+import org.apache.shardingsphere.sql.parser.api.SQLVisitorEngine;
+import org.apache.shardingsphere.sql.parser.sql.dialect.statement.mysql.MySQLStatement;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.Properties;
+
+public class MySQLParserStatementExample {
```
tristaZero, line 29: Is it better to define it as a final class?

examples/shardingsphere-parser-example/src/main/java/org/apache/shardingsphere/example/parser/mysql/format/MySQLParserFormatExample.java
```diff
@@ -0,0 +1,65 @@
...
+
+import org.antlr.v4.runtime.tree.ParseTree;
+import org.apache.shardingsphere.sql.parser.api.SQLParserEngine;
+import org.apache.shardingsphere.sql.parser.api.SQLVisitorEngine;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.Properties;
+
+public class MySQLParserFormatExample {
```
tristaZero, line 28: Is it better to define it as a final class?

tristaZero (changes requested):

tristaZero (approved):

PR #11221, opened 2021-07-08
Opengauss adapt

### LINKED ISSUE #11218: Bug in parse sharding value from binding table rules
5.0.0-beta

ShardingSphere-JDBC

### Expected behavior
correctly analysis sharding value from binding table 

### Actual behavior
get wrong sharding value analized from binding table under certain conditions. 

in code bellow, it get `bindingTableRule` by `shardingRule.findBindingTableRule(logicTableName)`，which leads to the result of `bindingTableRule.isPresent() && bindingTableRule.get().hasLogicTable(logicTableName)` to always be `true` if we configed the binding relation on `logicTableName`. :
```
// source code from: org.apache.shardingsphere.sharding.route.engine.type.standard.ShardingStandardRoutingEngine
private List<ShardingConditionValue> getShardingValuesFromShardingConditions(final ShardingRule shardingRule, final Collection<String> shardingColumns, final ShardingCondition shardingCondition) {
        List<ShardingConditionValue> result = new ArrayList<>(shardingColumns.size());
        for (ShardingConditionValue each : shardingCondition.getValues()) {
            Optional<BindingTableRule> bindingTableRule = shardingRule.findBindingTableRule(logicTableName);
            if ((logicTableName.equals(each.getTableName()) || bindingTableRule.isPresent() && bindingTableRule.get().hasLogicTable(logicTableName)) 
                    && shardingColumns.contains(each.getColumnName())) {
                result.add(each);
            }
        }
        return result;
    }
```

Imagine the following scenario:
we have 4 tables: T1、T2、T3、T4，and they have same sharding columns named 'col'，if we config binding relation as follows:
```
- T1, T2
- T3, T4
```
the Logic SQL bellow will leads to only one Actual SQL even if the table `T1` and `T3` are not bounded.
```
SELECT a.col1, a.col FROM T1 a JOIN T3 b ON a.col = b.col WHERE a.col = '0' 
```

Imagine the following scenario:
we have 4 tables: T1、T2、T3、T4，and they have same sharding columns named 'col'，and they all have 2 sharding tables(physic table) as follows: 
- T1_0、T1_1
- T2_0、T2_1
- T3_0、T3_1
- T4_0、T4_1
if we config binding relations as follows:
```
- T1, T2
- T3, T4
```
the Logic SQL bellow will leads to only one Actual SQL even if the table `T1` and `T3` are not bounded.
the Logic SQL is:
```
SELECT a.col1, a.col FROM T1 a JOIN T3 b ON a.col = b.col WHERE a.col = '0' 
```
the possible Actual SQL is（assume the sharding value "0" is in sharding table `T1_0`）:
```
SELECT a.col1, a.col FROM T1_0 a JOIN T3_0 b ON a.col = b.col WHERE a.col = '0' 
```

setamv:


tristaZero:
Hi @setamv Thanks for your PR here. @TeslaCN could you have a review?

TeslaCN:
This PR may be opened by mistake.

PR #11254, opened 2021-07-09
fix bug: get wrong sharding value analized from binding table

### LINKED ISSUE #11218: Bug in parse sharding value from binding table rules

setamv:
…ertain conditions. reference issue[11218]: https://github.com/apache/shardingsphere/issues/11218

Fixes #11218 .

- pass correct logic table names while find binding table rule in method: org.apache.shardingsphere.sharding.route.engine.type.standard.ShardingStandardRoutingEngine#getShardingValuesFromShardingConditions
-
-

### DIFF
--- a/shardingsphere-features/shardingsphere-sharding/shardingsphere-sharding-core/src/main/java/org/apache/shardingsphere/sharding/route/engine/type/standard/ShardingStandardRoutingEngine.java
+++ b/shardingsphere-features/shardingsphere-sharding/shardingsphere-sharding-core/src/main/java/org/apache/shardingsphere/sharding/route/engine/type/standard/ShardingStandardRoutingEngine.java
@@ -176,7 +176,7 @@ public final class ShardingStandardRoutingEngine implements ShardingRouteEngine 
     private List<ShardingConditionValue> getShardingValuesFromShardingConditions(final ShardingRule shardingRule, final Collection<String> shardingColumns, final ShardingCondition shardingCondition) {
         List<ShardingConditionValue> result = new ArrayList<>(shardingColumns.size());
         for (ShardingConditionValue each : shardingCondition.getValues()) {
-            Optional<BindingTableRule> bindingTableRule = shardingRule.findBindingTableRule(logicTableName);
+            Optional<BindingTableRule> bindingTableRule = shardingRule.findBindingTableRule(each.getTableName().toLowerCase());
             if ((logicTableName.equals(each.getTableName()) || bindingTableRule.isPresent() && bindingTableRule.get().hasLogicTable(logicTableName)) 
                     && shardingColumns.contains(each.getColumnName())) {
                 result.add(each);


tristaZero:
@setamv Thanks for this PR, @strongduanmu please have a look here.

tristaZero:
@strongduanmu Hi?

shardingsphere-features/shardingsphere-sharding/shardingsphere-sharding-core/src/main/java/org/apache/shardingsphere/sharding/route/engine/type/standard/ShardingStandardRoutingEngine.java
```diff
@@ -176,7 +176,7 @@ private boolean isGettingShardingValuesFromHint(final ShardingStrategy shardingS
     private List<ShardingConditionValue> getShardingValuesFromShardingConditions(final ShardingRule shardingRule, final Collection<String> shardingColumns, final ShardingCondition shardingCondition) {
         List<ShardingConditionValue> result = new ArrayList<>(shardingColumns.size());
         for (ShardingConditionValue each : shardingCondition.getValues()) {
-            Optional<BindingTableRule> bindingTableRule = shardingRule.findBindingTableRule(logicTableName);
+            Optional<BindingTableRule> bindingTableRule = shardingRule.findBindingTableRule(each.getTableName().toLowerCase());
```
strongduanmu, line 179: @setamv Why does it need to be lowercase?

strongduanmu:
Hi @setamv, can you give a feedback?

shardingsphere-features/shardingsphere-sharding/shardingsphere-sharding-core/src/main/java/org/apache/shardingsphere/sharding/route/engine/type/standard/ShardingStandardRoutingEngine.java
setamv: because the value of `each.getTableName()` is case sensitive，but table name in sharding rule is lower case.

setamv: the table names are uppercase in out schema，and we config the binding rules with table name in uppercase like `- T_XX_MAIN_004,T_XX_DETAIL_004_EXT`, but the table names in `org.apache.shardingsphere.sharding.rule.ShardingRule#bindingTableRules` are converted to lowercase.
when we execute sql with table names in uppercase, the table name in ShardingConditionValue are kept in uppercase, so, i   must convert it's to lowercase the match the table name in sharding rule.

setamv:
> Hi @setamv, can you give a feedback?
done. @strongduanmu

strongduanmu:
> > Hi @setamv, can you give a feedback?
> > done. @strongduanmu

@setamv I will review this pr.

shardingsphere-features/shardingsphere-sharding/shardingsphere-sharding-core/src/main/java/org/apache/shardingsphere/sharding/route/engine/type/standard/ShardingStandardRoutingEngine.java
strongduanmu: > ShardingConditionValue

@setamv Thank you for reply.

strongduanmu (approved):

PR #11266, opened 2021-07-11
Modify the spi address quoted in the faq

yx9o:
Modify the spi addresses quoted in #15 and #16 in the faq, the current access report is 404, because the spi directory has been adjusted in #5861.


tristaZero (approved):
@yx9o Hi thanks for your tweak.

Issue #11282, opened 2021-07-12
Code optimization for shardingsphere-shadow project


soulasuna:
## Code optimization for shardingsphere-shadow project

Optimization goal to improve code robustness and readability

## Optimization direction

- Perform type judgment before forced type conversion to avoid type conversion failure.

- Reduce nested if judgments. The logic is clearer and easier to read.

- Optimize large methods with complex logic and increase the readable lines for multiple small methods.

soulasuna:
Refactor shadow. see #11661

PR #11285, opened 2021-07-12
Code optimization for shardingsphere-shadow project. (#11282)

soulasuna:
Code optimization for shardingsphere-shadow project. 
For #11282

-  Optimize the code under the rewrite package.
-  Add junit test method for ShadowConditionEngine.


shardingsphere-features/shardingsphere-shadow/shardingsphere-shadow-core/src/main/java/org/apache/shardingsphere/shadow/rewrite/parameter/ShadowParameterRewriter.java
```diff
@@ -41,4 +41,8 @@ public final boolean isNeedRewrite(final SQLStatementContext sqlStatementContext
     }
     
     protected abstract boolean isNeedRewriteForShadow(SQLStatementContext sqlStatementContext);
+    
+    protected String getShadowColumn() {
+        return shadowRule.getColumn();
```
tristaZero, line 46: Why do we need another function?

tristaZero (changes requested):

tristaZero (approved):

shardingsphere-features/shardingsphere-shadow/shardingsphere-shadow-core/src/main/java/org/apache/shardingsphere/shadow/rewrite/parameter/ShadowParameterRewriter.java
soulasuna: Each subclass is called multiple times, and protected methods are extracted from the parent class.

Issue #11281, opened 2021-07-12
Sharding-JDBC Poor performance, need to Improve.


nanzheng:
### Env
Version：shardingsphere 4.0.1
Test tool: jmeter 5.3
Test Case: shardingsphere benchmark 4.1.1
Database ：mysql
Threads: 10
Each thread loops: 1000000

### Test cases
Case A(JDBC): select * from a where a.aa= 'ciphertext' and a.bb='ciphertext';
Case B(Sharding-JDBC): select * from a where a.aa= 'plaintext' and a.bb='plaintext';

### Result diagram
JDBC base compare profiler output [Case A](https://github.com/nanzheng/ShardingSphere-Oracle-Test/blob/main/20210712-152815.svg)

sharding-jdbc base compare profiler output [Case B](https://github.com/nanzheng/ShardingSphere-Oracle-Test/blob/main/20210712-152322.svg)

### Summary
As you can see
 from these two diagram， there is about 35% performance loss where i use two encryption filed as query condition .

terrymanu:
All improvements are finished.

PR #11294, opened 2021-07-12
Replace unnecessary TreeMap with CaseInsensitiveMap

TeslaCN:
Related to  #11281 

I found the order is unnecessary. Please remind me if I missed something.

image

image


terrymanu (approved):

MingxingLAI:
Which tool is used for this performance graph, can you teach me ? @TeslaCN

MingxingLAI:
Why use `CaseInsensitiveMap`，is `HashMap` okay ? @TeslaCN

TeslaCN:
@MingxingLAI You may take a look at https://github.com/jvm-profiling-tools/async-profiler
The label is case-insensitive. HashMap cannot do that.

PR #11308, opened 2021-07-13
Completely support integration test under PG database.

JiekerTime:
about issue #9104


JiekerTime:
The test results were observed, so PR was temporarily turned off to avoid wasting resources.

JiekerTime:
Hi @tristaZero ! I have finished the micro-refactoring of the integration test framework, and after the current PR merge, our integration test framework will fully support PG database.


However, I have temporarily turned off the testing of PG due to the failure of some cases. You can turn on the PG test by the following：
image


I will fix the problems that occur with the cases in the future.

tristaZero:
Do these cases only not work in docker env? Or the native env is same?

tristaZero (changes requested):
Do these cases only not work in docker env? Or the native env is same?
Could we just skip these to-do cases?

tristaZero (approved):

tristaZero:
@TeslaCN

JiekerTime:
> Do these cases only not work in docker env? Or the native env is same?

It only works in docker environment, we can switch to H2's jdbc scenario to run in our local environment.

PR #11479, opened 2021-07-23
Fix the ExecutorService does not shutdown when loading table meta data failed.

### LINKED ISSUE #11468: ExecutorService does not shutdown when loading table meta data failed.
## Bug Report

5.0.0-RC1-SNAPSHOT

ShardingSphere-Proxy

### Expected behavior
ExecutorService is supposed to shutdown for process to exit when loading table meta data failed.

### Actual behavior
ExecutorService does not shutdown when loading table meta data failed, that causes the process to not exit correctly.

image
When a task calling loadTableByDataNode failed, the exception will be thrown resulting in ExecutorService cannot shutdown.

1. Set the property check-table-metadata-enabled to true
2. Create scenarios where calling loadTableByDataNode failed.

sccc803:
Fixes #11468.

- Shutdown the ExecutorService in catch code block.


shardingsphere-features/shardingsphere-sharding/shardingsphere-sharding-core/src/main/java/org/apache/shardingsphere/sharding/metadata/ShardingTableMetaDataBuilder.java
```diff
@@ -98,6 +98,7 @@
             try {
                 getTableMetaData(value).ifPresent(tableMetaData -> result.put(key, tableMetaData));
             } catch (final InterruptedException | ExecutionException | TimeoutException ex) {
+                executorService.shutdownNow();
```
RaigorJiang, line 101: Why not close the resource in the finally block？

tristaZero (approved):
@sccc803 Hey, thanks for your kind tweak.

PR #11515, opened 2021-07-26
Revise #11513

terrymanu:
Fixes #11513.


strongduanmu (approved):

yx9o:
@terrymanu hi, I looked at the source code of `HashMap` and tested that the `threshold` value calculated by `new HashMap<>(3)` and `new HashMap<>(3, 1)` is the same, and it will not expand when storing three elements, so I I chose `new HashMap<>(3)` this concise way.

terrymanu:
> @terrymanu hi, I looked at the source code of `HashMap` and tested that the `threshold` value calculated by `new HashMap<>(3)` and `new HashMap<>(3, 1)` is the same, and it will not expand when storing three elements, so I I chose `new HashMap<>(3)` this concise way.

The `DEFAULT_LOAD_FACTOR` is `0.75f`, set loadFactor to `1` to avoid re-calculate is better

yx9o:
`DEFAULT_LOAD_FACTOR=0.75f` is no problem, but I actually tested `new HashMap<>(3);` and `new HashMap<>(3, 1);`, they will not expand when storing 3 elements, thank you, I will use the syntax of `new HashMap<>(3,1)` in the future.

PR #11544, opened 2021-07-27
Check the initial capacity of HashMap

yx9o:
Check the initial capacity of HashMap in ss, Keep the code style consistent.

### DIFF
--- a/shardingsphere-infra/shardingsphere-infra-authority/shardingsphere-infra-authority-common/src/main/java/org/apache/shardingsphere/authority/provider/schema/builder/SchemaPrivilegeBuilder.java
+++ b/shardingsphere-infra/shardingsphere-infra-authority/shardingsphere-infra-authority-common/src/main/java/org/apache/shardingsphere/authority/provider/schema/builder/SchemaPrivilegeBuilder.java
@@ -67,7 +67,7 @@ public final class SchemaPrivilegeBuilder {
      */
     private static Map<ShardingSphereUser, String> convertSchemas(final String userSchemaMapping) {
         String[] userSchemaPair = userSchemaMapping.trim().split("=");
-        Map<ShardingSphereUser, String> result = new HashMap<>();
+        Map<ShardingSphereUser, String> result = new HashMap<>(1, 1);
 
         String yamlUser = userSchemaPair[0];
         String username = yamlUser.substring(0, yamlUser.indexOf("@"));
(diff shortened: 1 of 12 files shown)


shardingsphere-infra/shardingsphere-infra-authority/shardingsphere-infra-authority-common/src/main/java/org/apache/shardingsphere/authority/provider/schema/builder/SchemaPrivilegeBuilder.java
```diff
@@ -67,7 +67,7 @@
      */
     private static Map<ShardingSphereUser, String> convertSchemas(final String userSchemaMapping) {
         String[] userSchemaPair = userSchemaMapping.trim().split("=");
-        Map<ShardingSphereUser, String> result = new HashMap<>();
+        Map<ShardingSphereUser, String> result = new HashMap<>(1, 1);
```
tristaZero, line 70: Hi Is there any possibility that it's not 1?

tristaZero (changes requested):
Hi @yx9o Sorry for my late response. ;(
But, you really take good care of these particulars, nice.

shardingsphere-infra/shardingsphere-infra-authority/shardingsphere-infra-authority-common/src/main/java/org/apache/shardingsphere/authority/provider/schema/builder/SchemaPrivilegeBuilder.java
yx9o: Hi, I read the context-related code. This method only processes one userSchemaMapping at a time. Line 76 of the code will put only one element in the map and then execute return. In subsequent processes, only traversal of the map is involved and no element will be put anymore.

tristaZero (approved):
@yx9o Sorry I came back late.
I saw the context of `SchemaPrivilegeBuilder`, which looks it has some room to improve. But thanks for your tweak here. ;-)

yx9o:
Hi, I think if there is only one `userSchemaMapping` at a time, the `convertSchemas` and `merge` methods can be merged, and the logic will be clearer.

tristaZero:
@yx9o If you like, welcome the next PR to improve it. ;-)

yx9o:
ok

PR #11552, opened 2021-07-28
Add proxy state, uptime metrics and more sql or dist sql metrics

dachuan9e:

Add proxy state and uptime metrics


dachuan9e:
Will use another pull request because has some errors in this.

PR #11587, opened 2021-07-30
fix(transaction):fix transaction statement commit error while using shared thread pool and local transaction manager

### LINKED ISSUE #11583: transaction statement commit error while using shared thread pool and local transaction manager
## Bug Report

I found a concurrency issues when exeucte transaction statement sqls.
here is my sql statemtns(MySQL)
```
set autocommit=false
select  * from tablexx where sharding_key=?
update  tablexx set x=x where sharding_key=?
commit
```
the client create 200 connections to the proxy and concurrently execute this transaction statements to get the TPS data

As above，here is 4 statements
the first is a begin statement and the LocalTransactionManager will add connectionPostProcessors to invoke the real connection`s setAutoCommit(false) once a real connection is borrow from datasource in the future

the last statement commit will invoke the real connection`s commit method to submit the transaction and affect to the database

ps:
statement is execute in the thread pool: UserExecutorGroup.getInstance().getExecutorService(),it is a shard pool for all commandExecutorTask

5.0.0-RC1-SNAPSHOT
ShardingSphere-Proxy
### Expected behavior
Every transaction statement commit success and affect to the MySQL
### Actual behavior
It is hard to find the error,according to my tests,only 0.015% probability to get the error below:
Can't call commit when autocommit=true
As I analyzed , the reason has two parts
1.BackendConnection`s field connectionPostProcessors is not thread safe

image
although this is not the main reason,but it`s a potential issue
the 4 statement will execute serially along a connection.
let's imagine the 4 statement is execute in multiple threads 
```
set autocommit=false     -- sql1:thread1
select  * from tablexx where sharding_key=?  -- sql2:thread2
update  tablexx set x=x where sharding_key=?  -- sql3:thread3
commit -- sql4:thread4
```
set autocommit=false will let the BackendConnection`s localTransactionManager add a connectionPostProcessors  to invoke the real connection`s setAutoCommit(false) once a real connection is borrow from datasource
when sql2 execute in thread 2, a real connection will be get and apply the connectionPostProcessors,but may not see the latest 
connectionPostProcessors ,cause the real connection never execute setAutoCommit(false)

2.CommandExecutorTask`s method run has sequence problem

I think this sequence problem cause concurrency issues on a single connection,and it's dangerous

image
code1 will lead the client get response immediately，and client know this transaction is finished and could reuse the connection to run the next statement.
code2 clear the resources under the backendConncection like clean cachedConnections and connectionPostProcessors (etc.)when transaction is finished

so what happened ? because code1 flush the response in thread4,the client  send next statement "set autocommit=false" for another transaction,and thread6 execute it and add a connectionPostProcessors,then thread4 run the code2,clear the resources include thread6 just created,so the next transaction has no connectionPostProcessors,so the real connections will not execute "setAutoCommit(false)", and finally got error

```
set autocommit=false     -- sql1:thread1
select  * from tablexx where sharding_key=?  -- sql2:thread2
update  tablexx set x=x where sharding_key=?  -- sql3:thread3
commit -- sql4:thread4

#next sql statement 
set autocommit=false     -- sql1:thread6
select  * from tablexx where sharding_key=?  -- sql2:thread7
...
```
It is a sequence problem,but make the object backendConnection be modified concurrently.and cause the error

This is a problem that can be analyzed and concluded and hard to reproduce unless rewrite the code ,and a sleep between code1 and code2.
but in product environment,long running will make this error happen at lasat

jianliu:
Fixes #11583.

- move code backendConnection.getConnectionStatus().switchToReleased() in BackendConnection method:closeConnections to the finally code block of CommandExecutorTask`s method  "run" to avoid switchToRelease when begin statement is called.

### DIFF
--- a/shardingsphere-proxy/shardingsphere-proxy-backend/src/main/java/org/apache/shardingsphere/proxy/backend/communication/jdbc/connection/BackendConnection.java
+++ b/shardingsphere-proxy/shardingsphere-proxy-backend/src/main/java/org/apache/shardingsphere/proxy/backend/communication/jdbc/connection/BackendConnection.java
@@ -307,7 +307,6 @@ public final class BackendConnection implements ExecutorJDBCManager {
         }
         cachedConnections.clear();
         connectionPostProcessors.clear();
-        connectionStatus.switchToReleased();
         return result;
     }
     
--- a/shardingsphere-proxy/shardingsphere-proxy-backend/src/test/java/org/apache/shardingsphere/proxy/backend/communication/jdbc/connection/BackendConnectionTest.java
+++ b/shardingsphere-proxy/shardingsphere-proxy-backend/src/test/java/org/apache/shardingsphere/proxy/backend/communication/jdbc/connection/BackendConnectionTest.java
@@ -274,7 +274,6 @@ public final class BackendConnectionTest {
         verify(connection, times(1)).close();
         assertTrue(cachedConnections.isEmpty());
         verifyConnectionPostProcessorsEmpty();
-        verify(connectionStatus, times(1)).switchToReleased();
     }
     
     @Test
@@ -285,7 +284,6 @@ public final class BackendConnectionTest {
         Connection connection = prepareCachedConnections();
         backendConnection.closeConnections(true);
         verify(connection, never()).rollback();
-        verify(connectionStatus, times(1)).switchToReleased();
     }
     
     @Test
@@ -296,7 +294,6 @@ public final class BackendConnectionTest {
         Connection connection = prepareCachedConnections();
         backendConnection.closeConnections(true);
         verify(connection, times(1)).rollback();
-        verify(connectionStatus, times(1)).switchToReleased();
     }
     
     @Test
--- a/shardingsphere-proxy/shardingsphere-proxy-frontend/shardingsphere-proxy-frontend-core/src/main/java/org/apache/shardingsphere/proxy/frontend/command/CommandExecutorTask.java
+++ b/shardingsphere-proxy/shardingsphere-proxy-frontend/shardingsphere-proxy-frontend-core/src/main/java/org/apache/shardingsphere/proxy/frontend/command/CommandExecutorTask.java
@@ -84,6 +84,7 @@ public final class CommandExecutorTask implements Runnable {
             if (!backendConnection.getTransactionStatus().isInConnectionHeldTransaction()) {
                 exceptions.addAll(backendConnection.closeDatabaseCommunicationEngines(true));
                 exceptions.addAll(backendConnection.closeConnections(false));
+                backendConnection.getConnectionStatus().switchToReleased();
             }
             processClosedExceptions(exceptions);
         }
--- a/shardingsphere-proxy/shardingsphere-proxy-frontend/shardingsphere-proxy-frontend-core/src/test/java/org/apache/shardingsphere/proxy/frontend/command/CommandExecutorTaskTest.java
+++ b/shardingsphere-proxy/shardingsphere-proxy-frontend/shardingsphere-proxy-frontend-core/src/test/java/org/apache/shardingsphere/proxy/frontend/command/CommandExecutorTaskTest.java
@@ -44,6 +44,7 @@ import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.anyBoolean;
 import static org.mockito.ArgumentMatchers.anyInt;
 import static org.mockito.Mockito.atLeast;
+import static org.mockito.Mockito.doThrow;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
@@ -105,6 +106,7 @@ public final class CommandExecutorTaskTest {
         actual.run();
         verify(connectionStatus).waitUntilConnectionRelease();
         verify(connectionStatus).switchToUsing();
+        verify(connectionStatus).switchToReleased();
         verify(queryCommandExecutor).close();
         verify(backendConnection).closeDatabaseCommunicationEngines(true);
     }
@@ -124,6 +126,7 @@ public final class CommandExecutorTaskTest {
         actual.run();
         verify(connectionStatus).waitUntilConnectionRelease();
         verify(connectionStatus).switchToUsing();
+        verify(connectionStatus).switchToReleased();
         verify(handlerContext).write(databasePacket);
         verify(handlerContext).flush();
         verify(engine.getCommandExecuteEngine()).writeQueryData(handlerContext, backendConnection, queryCommandExecutor, 1);
@@ -147,6 +150,7 @@ public final class CommandExecutorTaskTest {
         actual.run();
         verify(connectionStatus).waitUntilConnectionRelease();
         verify(connectionStatus).switchToUsing();
+        verify(connectionStatus).switchToReleased();
         verify(handlerContext).write(databasePacket);
         verify(handlerContext).flush();
         verify(commandExecutor).close();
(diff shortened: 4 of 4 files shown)


tristaZero (changes requested):
Hi, @jianliu  Thanks for your nice polishment here. @TeslaCN Could you have a double check?

shardingsphere-proxy/shardingsphere-proxy-backend/src/main/java/org/apache/shardingsphere/proxy/backend/communication/jdbc/connection/BackendConnection.java
```diff
@@ -81,7 +82,7 @@
     
     private final Collection<DatabaseCommunicationEngine> inUseDatabaseCommunicationEngines = Collections.newSetFromMap(new ConcurrentHashMap<>(64));
     
-    private final Collection<ConnectionPostProcessor> connectionPostProcessors = new LinkedList<>();
+    private final Collection<ConnectionPostProcessor> connectionPostProcessors = new CopyOnWriteArrayList<>();
```
TeslaCN, line 85: Could we fix this issue by defer the flush only? CopyOnWriteArrayList may impact the performance.

TeslaCN (changes requested):

shardingsphere-proxy/shardingsphere-proxy-backend/src/main/java/org/apache/shardingsphere/proxy/backend/communication/jdbc/connection/BackendConnection.java
jianliu: connectionPostProcessors has memory visibility issues here,linkedList is not thread-safe,if we don`t use CopyOnWriteArrayList, we can use Collections.synchronizedList(new LinkedList<>()) instead.
Do you have any recommendations？

jianliu: I took a look at the api doc,Collections.synchronizedList(new LinkedList()) need use synchronized (list) when Iterate over it.
so if we want to keep use LinkedList,we need to synchronized (connectionPostProcessors /backendConnection) when add item or foreach it.

TeslaCN: My understanding is if we defer the flush, client won't send next packet until the resources are closed and the processors will be operated by only one thread.

jianliu: As you said, processors are indeed only operated in one thread at a time, but not promise in the same thread. The changes of thread1 may not be visible to thread2, which is why we use volatile.I`ll test it

jianliu:
@TeslaCN I can`t reproduce the problem of connectionPostProccesors, I have remove the change of connectionPostProccesors and only keep context.flush
Please have a  check,thanks

TeslaCN (approved):
LGTM

TeslaCN (changes requested):
Hi @jianliu 
I checked again and found this issue may be caused by incorrect connection status in BackendConnection. Maybe we need further discussion about this issue.

jianliu:
@TeslaCN Thanks all the same.
If you think this is a problem that needs to be dealt with in time, it is recommended to deal with it as soon as possible, because once it occurs, it will lead to business failure

tristaZero:
@jianliu Sure, could you give a look at @TeslaCN 's comment, i.e., `be caused by incorrect connection status in BackendConnection`
Do you think the root cause is that the incorrect trx status of `BackendConnection`? Since you can see,
```java
# CommandExecutorTask
if (!backendConnection.getTransactionStatus().isInConnectionHeldTransaction()) {
                connectionStatus.waitUntilConnectionRelease();
                connectionStatus.switchToUsing();
            }
```
This task will suspend until all the connections are released.

jianliu:
@tristaZero 
This task will suspend until all the connections are released.

Yes,the thread is suspend too,but the client will receive the response when context.flush is invoked,not the time of task or thread be finished.In other words

the client may send next transaction sql like setcommit=false to proxy in the same moment,the proxy will submit next task to execute "setcommit=false",cause one BacendConnection handle its resources in two thread
you can have a look at this issue`s reason analyz part 2.

image

TeslaCN:
> @tristaZero
> This task will suspend until all the connections are released.
> 
> Yes,the thread is suspend too,but the client will receive the response when context.flush is invoked,not the time of task or thread be finished.
> In other words,the client may send next transaction sql like setcommit=false to proxy at the same moment,the proxy will submit a new task to execute "setcommit=false",cause one BacendConnection handle its resources in two thread
> you can have a look at this issue`s reason analyz part 2.
> 
> image

Hi @jianliu 

My understanding is if the connection status is correct, even the next command was sent to Proxy and a new task was put into the ThreadPool, the new task will `waitUntilConnectionRelease`.

The current connection status may be switched to `release` incorrectly when transaction started. Maybe correct the connection status may fix this issue.

jianliu:
The code
``` 
    public void waitUntilConnectionRelease() {
        while (isUsing) {
            resourceLock.doAwait();
        }
    }
``` 
```
    public void doAwait() {
        lock.lock();
        try {
            condition.await(DEFAULT_TIMEOUT_MILLISECONDS, TimeUnit.MILLISECONDS);
        } finally {
            lock.unlock();
        }
    }
```
await has a timeout which default is 200ms,so this is not always ok

jianliu:
@tristaZero @TeslaCN  Sorry for the deviation of the previous reason analysis.
I have add a lot of logs to reproduce error. and I have added a comment to  the issue like below.

I have made a mistake when analyz the reason because of missing an important info,that between pre tps and next tps.
```
#first statements
set autocommit=false     -- sql1:thread1
select  * from tablexx where sharding_key=?  -- sql2:thread2
update  tablexx set x=x where sharding_key=?  -- sql3:thread3
commit -- sql4:thread4
context.flush 
close resources because there is no transaction yet after commit

#the second sql by framework (hikari)
set autocommit=true-- sql1:thread10 this sql is request by datasource framework when connection is close(back to pool)
context.flush
close resources because there is no transaction at all

#the thrid statement 
set autocommit=false     -- sql1:thread6
select  * from tablexx where sharding_key=?  -- sql2:thread7
```
Here is the logs:
image

The second sql (setAutoCommit(true) between the first and the thrid statement send by framework cause this erorr,because it removes the third sql(setAutoCommit:false) `s connectionPostProcessor,and lead this error finally.

The PR could resolve this error too. or you could make close resources and add resources synchronzied,it`s alose resolve this issue too.

TeslaCN:
Hi @jianliu 
I think the root cause of this issue is the incorrect connection status, merging this PR may solve this issue, but the connection status is still incorrect. How about moving the `connectionStatus.switchToReleased()` from the method `closeConnections` to `CommandExecuteTask`?

Moving the following line
https://github.com/apache/shardingsphere/blob/183f3b85e3ccea9fc410c91f7f4488ed8cc85b3b/shardingsphere-proxy/shardingsphere-proxy-backend/src/main/java/org/apache/shardingsphere/proxy/backend/communication/jdbc/connection/BackendConnection.java#L310

to here
https://github.com/apache/shardingsphere/blob/183f3b85e3ccea9fc410c91f7f4488ed8cc85b3b/shardingsphere-proxy/shardingsphere-proxy-frontend/shardingsphere-proxy-frontend-core/src/main/java/org/apache/shardingsphere/proxy/frontend/command/CommandExecutorTask.java#L84-L87

jianliu:
@TeslaCN I have find it too this morning.. this is the real reason. I had test it and no error occured

TeslaCN (approved):
LGTM

tristaZero (approved):
Thanks @jianliu and @TeslaCN to make it fixed. Please give it a test on your env, any questions, please speak up. Good job.

PR #11641, opened 2021-08-04
Modify the conditional expression

yx9o:
Adjust the conditional expression according to the ss coding standard.

### DIFF
--- a/shardingsphere-jdbc/shardingsphere-jdbc-core/src/main/java/org/apache/shardingsphere/driver/jdbc/core/resultset/ResultSetUtil.java
+++ b/shardingsphere-jdbc/shardingsphere-jdbc-core/src/main/java/org/apache/shardingsphere/driver/jdbc/core/resultset/ResultSetUtil.java
@@ -104,7 +104,7 @@ public final class ResultSetUtil {
         if (null == value) {
             return convertNullValue(BigDecimal.class);
         }
-        if (value.getClass() == BigDecimal.class) {
+        if (BigDecimal.class == value.getClass()) {
             return adjustBigDecimalResult((BigDecimal) value, needScale, scale);
         }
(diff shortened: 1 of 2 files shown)


shardingsphere-jdbc/shardingsphere-jdbc-core/src/main/java/org/apache/shardingsphere/driver/jdbc/core/resultset/ResultSetUtil.java
```diff
@@ -115,14 +115,14 @@ public static Object convertBigDecimalValue(final Object value, final boolean ne
...
-        }
-        return value;
+        if (!needScale){
+            return value;
+        }
+        try {
+            return value.setScale(scale);
+        } catch (final ArithmeticException ex) {
+            return value.setScale(scale, BigDecimal.ROUND_HALF_UP);
+        }
```
terrymanu, line 125: This is guard clauses, please keep the original codes

terrymanu (changes requested):

shardingsphere-jdbc/shardingsphere-jdbc-core/src/main/java/org/apache/shardingsphere/driver/jdbc/core/resultset/ResultSetUtil.java
yx9o: Hi, the code has been rolled back, please review. 
But I think the modified statement conforms to the guard clauses, and the code is more concise.

terrymanu: Thank you

terrymanu (approved):

PR #11653, opened 2021-08-05
fix letter case problems as follows: table names in sql executed are …

setamv:
fix letter case problems as follows: table names in sql executed are upper case, and in database schema are lower case. in such case, the judgement like Collection#contains will goes wrong as the letter case is not consistent with parameter and value in collection.

the debugger info as follows:
table names in parameter `logicTableNames` are upper case which consistent with letter case in sql executed, but table names in `broadcastTables` are converted to lower case:
ShardingRule-getShardingBroadcastTableNames
ShardingRule-getShardingBroadcastTableNames-2

ShardingRule-isAllTablesInSameDataSource

Fixes #ISSUSE_ID.

- 
-
-

### DIFF
--- a/shardingsphere-features/shardingsphere-sharding/shardingsphere-sharding-core/src/main/java/org/apache/shardingsphere/sharding/rule/ShardingRule.java
+++ b/shardingsphere-features/shardingsphere-sharding/shardingsphere-sharding-core/src/main/java/org/apache/shardingsphere/sharding/rule/ShardingRule.java
@@ -171,7 +171,7 @@ public final class ShardingRule implements FeatureRule, SchemaRule, DataNodeCont
     @Override
     public Collection<String> getAllTables() {
         Collection<String> result = new HashSet<>(getTables());
-        result.addAll(getAllActualTables());
+        result.addAll(getAllActualTables().stream().map(String::toLowerCase).collect(Collectors.toList()));
         result.addAll(broadcastTables);
         return result;
     }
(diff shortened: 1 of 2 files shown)


shardingsphere-features/shardingsphere-sharding/shardingsphere-sharding-core/src/main/java/org/apache/shardingsphere/sharding/rule/ShardingRule.java
```diff
@@ -171,7 +171,7 @@ private BindingTableRule createBindingTableRule(final String bindingTableGroup)
     @Override
     public Collection<String> getAllTables() {
         Collection<String> result = new HashSet<>(getTables());
-        result.addAll(getAllActualTables());
+        result.addAll(getAllActualTables().stream().map(String::toLowerCase).collect(Collectors.toList()));
```
strongduanmu, line 174: @setamv Is it better to convert lowercase in the `getAllActualTables` method?

setamv: I think so too, but getAllActualTables is referenced in other places, i don't test it. so..

setamv: please wait and i'll test it.

strongduanmu:
@setamv Can you fix code conflicting?

strongduanmu:
Close because long time no response.

PR #11666, opened 2021-08-05
Insert clause not support routing to multiple dataNodes when is not broadcastTable. #11623

### LINKED ISSUE #11623: Insert statements with expressions should not be routed to all datanodes
## Bug Report

5.0.0-RC1-SNAPSHOT
ShardingSphere-Proxy
### Expected behavior
When inserting statement with expressions of a non-broadcast table, an exception should be thrown to avoid routing to all datanodes, which will cause dirty data problem.
### Actual behavior
Insert clause when sharding key with expression will routing to all dataNodes, cause dirty data problem.
image
When performing SQL parsing, sharding key with expression will be parsed as BinaryOperationExpression, but when generating sharding conditions, only parsing the three types of SimpleExpressionSegment, NowExpression, and NullExpression, other types will not be parsed, so the final ShardingCondition list is Empty, when routing based on shardingCondition list, will full routing.
1.order_id is sharding key.
2.execute sql  : 
INSERT INTO t_order (order_id, user_id, status) VALUES (1+1, 2, 'insert');
3.it will route to all datanodes, every datanode will have a record.

chengh1:
Fixes #11623

-if it is singleRouting, then return.
-if table is a broadcast Table, then return.
-if one of values is routing to multiple datanodes, then throwing a exception to avoid generating dirty data in every datanode.

### DIFF
--- a/shardingsphere-features/shardingsphere-sharding/shardingsphere-sharding-core/src/main/java/org/apache/shardingsphere/sharding/route/engine/validator/dml/impl/ShardingInsertStatementValidator.java
+++ b/shardingsphere-features/shardingsphere-sharding/shardingsphere-sharding-core/src/main/java/org/apache/shardingsphere/sharding/route/engine/validator/dml/impl/ShardingInsertStatementValidator.java
@@ -97,5 +97,17 @@ public final class ShardingInsertStatementValidator extends ShardingDMLStatement
         if (needCheckDatabaseInstance) {
             Preconditions.checkState(routeContext.isSingleRouting(), "Sharding value must same with subquery.");
         }
+        if (routeContext.isSingleRouting()) {
+            return;
+        }
+        String tableName = sqlStatementContext.getSqlStatement().getTable().getTableName().getIdentifier().getValue();
+        if (shardingRule.isBroadcastTable(tableName)) {
+            return;
+        }
+        routeContext.getOriginalDataNodes().forEach(dataNodes -> {
+            if (dataNodes.size() > 1) {
+                throw new ShardingSphereException("Insert clause not support routing to multiple dataNodes when is not broadcastTable.");
+            }
+        });
     }
 }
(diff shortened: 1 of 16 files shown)


chengh1:
@strongduanmu hi, the 2 failing checkes are the same reason. They are caused by the ClassCast when creating the data source during running the integration test. However, I did not modify the code on this side, and I also run the integration test code locally, they are all passed, so I need your help, thanks.

strongduanmu:
@chengh1 Execute `mvn clean install -Prelease -T 1C` locally, and see what exception message appears.

shardingsphere-features/shardingsphere-sharding/shardingsphere-sharding-core/src/main/java/org/apache/shardingsphere/sharding/route/engine/validator/dml/impl/ShardingInsertStatementValidator.java
```diff
@@ -97,5 +97,17 @@ public void postValidate(final ShardingRule shardingRule, final SQLStatementCont
         if (needCheckDatabaseInstance) {
             Preconditions.checkState(routeContext.isSingleRouting(), "Sharding value must same with subquery.");
         }
+        if (routeContext.isSingleRouting()) {
```
strongduanmu, line 100: @chengh1 Please add test case for this change.

tristaZero:
@chengh1 Hi, first time to see you here, welcome.

shardingsphere-features/shardingsphere-sharding/shardingsphere-sharding-core/src/main/java/org/apache/shardingsphere/sharding/route/engine/validator/dml/impl/ShardingInsertStatementValidator.java
chengh1: The integration test module already contains my test scenarios, so I don’t need to add a new test case.

PR #11678, opened 2021-08-06
Refactor proxy init and start

dachuan9e:

- Refactor proxy frontend start

### DIFF
--- a/shardingsphere-proxy/shardingsphere-proxy-frontend/shardingsphere-proxy-frontend-core/src/main/java/org/apache/shardingsphere/proxy/frontend/ShardingSphereProxy.java
+++ b/shardingsphere-proxy/shardingsphere-proxy-frontend/shardingsphere-proxy-frontend-core/src/main/java/org/apache/shardingsphere/proxy/frontend/ShardingSphereProxy.java
@@ -67,8 +67,7 @@ public final class ShardingSphereProxy {
         createEventLoopGroup();
         ServerBootstrap bootstrap = new ServerBootstrap();
         initServerBootstrap(bootstrap);
-        ChannelFuture future = bootstrap.bind(port).sync();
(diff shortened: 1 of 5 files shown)


shardingsphere-proxy/shardingsphere-proxy-frontend/shardingsphere-proxy-frontend-core/src/main/java/org/apache/shardingsphere/proxy/frontend/ShardingSphereProxy.java
```diff
@@ -58,20 +54,28 @@
...
             BackendExecutorContext.getInstance().getExecutorEngine().close();
         }
     }
     
+    private ChannelFuture startInternal(final int port) throws InterruptedException {
+        createEventLoopGroup();
+        ServerBootstrap bootstrap = new ServerBootstrap();
+        initServerBootstrap(bootstrap);
+        ChannelFuture future = bootstrap.bind(port).sync();
+        return future;
```
terrymanu, line 71: It is better to inline return expr

terrymanu (changes requested):

shardingsphere-proxy/shardingsphere-proxy-frontend/shardingsphere-proxy-frontend-core/src/main/java/org/apache/shardingsphere/proxy/frontend/ShardingSphereProxy.java
dachuan9e: ok.

terrymanu (approved):

terrymanu (approved):

Issue #11621, opened 2021-08-03
ShardingSphere-jdbc configuration add schema name


zjcnb:
Hi,

   ShardingSphere-jdbc configuration add schema name

## shardingsphere-jdbc-core

- [x] yml config add schemaName:
```
schemaName: logic_db
```

- [x] properties config add schemaName:  
```
spring.shardingsphere.schema.name: logic_db
```

- [x] namespace xml add schemaName   

```
<shardingsphere:data-source id="dataSource" data-source-names="ds_0_write" rule-refs="shardingRule" schema-name="logic_db"></shardingsphere:data-source>
```

## shardingsphere-jdbc-governance

- [x] yml config add schemaName:  
```
schemaName: logic_db
```

- [x] properties config add schemaName:  
```
spring.shardingsphere.schema.name: logic_db
```

- [x] namespace xml add schemaName:  
```
<governance:data-source id="shardingDatabasesTablesDataSource" data-source-names="demo_ds_0, demo_ds_1" reg-center-ref="regCenter" rule-refs="shardingRule" schema-name="logic_db"/>
```

## Remove getDefaultSchemaMetadata()  interface

- [x] remove getDefaultMetaData() interface
```
    /**
     * Get default mata data.
     *
     * @return default mata data
     */
    ShardingSphereMetaData getDefaultMetaData();
```

##  Use real schema name replace logic_db

- [x] use real schema name replace logic_db
```
    /**
     * Schema name.
     */
    public static final String LOGIC_NAME = "logic_db";
```

zjcnb:
example for yml:

`schemaName: logic_db

rules:
- !SHARDING
  tables:
    t_order:
      actualDataNodes: jdbc_${0..1}.t_order_${0..1}
    t_order_item:
      actualDataNodes: jdbc_${0..1}.t_order_item_${0..1}
      keyGenerateStrategy:
        column: item_id
        keyGeneratorName: increment
    t_config:
      actualDataNodes: jdbc_${0..1}.t_config
    t_user:
      actualDataNodes: jdbc_0.t_user_${0..1}
      tableStrategy:
        standard:
          shardingColumn: id
          shardingAlgorithmName: table_inline
      keyGenerateStrategy:
        column: id
        keyGeneratorName: reset_increment
  autoTables:
    t_order_auto:
      actualDataSources: jdbc_0, jdbc_1
      shardingStrategy:
        standard:
          shardingColumn: order_id
          shardingAlgorithmName: mod
[...]

PR #11679, opened 2021-08-06
JDBC-core module delete getDefaultMetaData()

zjcnb:
For #11621 

jdbc-core module delete getDefaultMetaData()

### DIFF
--- a/shardingsphere-jdbc/shardingsphere-jdbc-core/src/main/java/org/apache/shardingsphere/driver/executor/DriverJDBCExecutor.java
+++ b/shardingsphere-jdbc/shardingsphere-jdbc-core/src/main/java/org/apache/shardingsphere/driver/executor/DriverJDBCExecutor.java
@@ -48,10 +48,13 @@ public final class DriverJDBCExecutor {
     
     private final JDBCLockEngine jdbcLockEngine;
     
-    public DriverJDBCExecutor(final MetaDataContexts metaDataContexts, final JDBCExecutor jdbcExecutor) {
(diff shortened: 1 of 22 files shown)


shardingsphere-jdbc/shardingsphere-jdbc-core/src/main/java/org/apache/shardingsphere/driver/executor/DriverJDBCExecutor.java
```diff
@@ -48,10 +48,13 @@
     
     private final JDBCLockEngine jdbcLockEngine;
     
-    public DriverJDBCExecutor(final MetaDataContexts metaDataContexts, final JDBCExecutor jdbcExecutor) {
+    private final String schemaName;
+    
+    public DriverJDBCExecutor(final MetaDataContexts metaDataContexts, final JDBCExecutor jdbcExecutor, final String schemaName) {
```
menghaoranss, line 53: I don't think it's a good idea to pass `schemaName` in Executor

menghaoranss (changes requested):

shardingsphere-jdbc/shardingsphere-jdbc-core/src/main/java/org/apache/shardingsphere/driver/executor/DriverJDBCExecutor.java
strongduanmu: It may be a better way to pass in the schemaName during execution.

Issue #11669, opened 2021-08-05
Make governance from independent config to plugged rules


terrymanu:
Governance module is the main feature of the product for now. 
I just consider about refactor it to plugged features by the plugged architecture.

The tasks are:

- [x] Proxy YAML configuration change

The current YAML configuration for proxy is:

```yaml
governance:
  registryCenter:
    type: ZooKeeper
    namespace: governance_ds
    serverLists: localhost:2181
  overwrite: false

rules:
  - !AUTHORITY
    users:
      - root@%:root
    provider:
      type: NATIVE
```

It should be changed to:

```yaml
rules:
  - !AUTHORITY
    users:
      - root@%:root
    provider:
      type: NATIVE
  
  - !GOVERNANCE
    registryCenter:
      type: ZooKeeper
      namespace: governance_ds
      serverLists: localhost:2181
    overwrite: false
```

- [ ] JDBC YAML configuration change
- [ ] JDBC java API change
- [ ] JDBC spring namespace change
- [ ] JDBC spring boot starter change
- [ ] Move governance module from top level to the submodule of feature
- [ ] Adjust DistMetadataPersistService as plugged rule
- [ ] Resolve conflict if independent DistMetadataPersistService rule and governance rule in the same config file

After finish all tasks, I prefer to discuss about change governance module name to cluster.

terrymanu:
We need to redesign here, just relate with #11709

terrymanu:
I just close this issue because the design is changed

PR #11700, opened 2021-08-08
Add back StandardTransactionContexts and GovernanceTransactionContexts

terrymanu:
For #11669.

- Add back StandardTransactionContexts
- Add back GovernanceTransactionContexts
- Use GovernanceTransactionContexts on GovernanceShardingSphereDataSource


tristaZero (approved):

shardingsphere-transaction/shardingsphere-transaction-core/src/main/java/org/apache/shardingsphere/transaction/context/impl/StandardTransactionContexts.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+    
+    @Override
+    public ShardingTransactionManagerEngine getDefaultEngine() {
+        return engines.get(DefaultSchema.LOGIC_NAME);
+    }
+    
+    @Override
+    public void close() throws Exception {
+        for (ShardingTransactionManagerEngine each : engines.values()) {
```
taojintianxia, line 49: why not lambda ?

taojintianxia (approved):

shardingsphere-transaction/shardingsphere-transaction-core/src/main/java/org/apache/shardingsphere/transaction/context/impl/StandardTransactionContexts.java
terrymanu: lambda cannot work if the inline code declare checked exception

PR #11702, opened 2021-08-08
BeanClassName may be empty

danianyu:
…ClassName may be empty in the case of factory bean reference, the beanclassname needs to be judged

Fixes #ISSUSE_ID.

-
-
-


terrymanu (commented):
When the bean name will null?
Can you give an example?

danianyu:
E

> When the bean name will null?
> Can you give an example?

Sorry, the example is in the company's intranet environment and it is not convenient to provide.

This method will not work in the case factory method is used. As Java Doc says:
_Also, this may just be the class that a factory method is called on, or **it may even be empty in case of a factory bean reference that a method is called on.** Hence, do not consider this to be the definitive bean type at runtime but rather only use it for parsing purposes at the individual bean definition level._

This is an example of others.
_https://stackoverflow.com/questions/33697722/spring-bean-definition-class-name-is-null-when-using-javaconfig_

terrymanu (approved):

terrymanu:
Thank you for the excellent explain, I just merge the pr.

PR #11724, opened 2021-08-09
reset jdbc config in it 

JiekerTime:


terrymanu (changes requested):
The config of yaml has already changed, what is the reason to revert it?

terrymanu (approved):

Issue #10626, opened 2021-06-02
PostgreSQL Proxy performance optimization


TeslaCN:
### Is your feature request related to a problem?

We've tested a Proxy with PostgreSQL backend according to the document
[Performance Test(sysbench)](https://shardingsphere.apache.org/document/current/en/features/test-engine/performance-test-sysbench/). Compared to sysbench PostgreSQL directly, I think the performance of the Proxy still has some room for improvement.

### Describe the feature you would like.
Optimizing performance of the PostgreSQL Proxy.

PR #11791, opened 2021-08-12
use GroovyShell CompileStatic to improve inline expression performance

strongduanmu:
Ref #10626, #11281

- use GroovyShell CompileStatic to improve inline expression performance


terrymanu (changes requested):
Could you provide the performance test data?

strongduanmu:
After testing, it was found that the performance improvement using static compilation was not obvious, so we decided to close this PR.

PR #11835, opened 2021-08-16
For #11783, rename rule package and 'getType' method name.

RaigorJiang:
For #11783, rename rule package and 'getType' method name.


RaigorJiang:
CI failed because of `PostgreSQLStatementMemoryStrictlyFetchSizeSetterTest`, which has nothing to do with the current PR.

tristaZero (approved):

Issue #11677, opened 2021-08-06
Refactoring of SCTL


RaigorJiang:
Hi community,

Currently, the [SCTL](https://shardingsphere.apache.org/document/current/en/features/dist-sql/syntax/sctl/sctl/) syntax is identified by the pre-keyword'sctl:', and special processing is required in the SQL parsing process, which also makes SCTL statements not like SQL statements.

Now we consider the refactoring of SCTL, which mainly includes two aspects:
1. Remove the pre-keyword and redesign the SCTL syntax, similar to [RDL](https://shardingsphere.apache.org/document/current/en/features/dist-sql/syntax/rdl/rdl-sharding-rule/) and [RQL](https://shardingsphere.apache.org/document/current/en/features/dist-sql/syntax/rql/rql-sharding-rule/). 
   Here are some examples:

| Before                            | After                             |
|:---------------------------------------|:---------------------------------------|
|sctl:set transaction_type=XX            |set variable transaction_type=XX |
|sctl:show transaction_type              |show variable transaction_type |
|sctl:show cached_connections            |show variable cached_connections |
|sctl:explain SQL                        |preview SQL   |
|sctl:hint set PRIMARY_ONLY=true         |set readwrite_splitting hint source = [auto / write] |
|sctl:hint set DatabaseShardingValue=yy  |set sharding hint database_value = yy; |
|sctl:hint addDatabaseShardingValue xx=yy|add sharding hint database_value xx= yy; |
|sctl:hint addTableShardingValue xx=yy   |add sharding hint table_value xx = yy |
|sctl:hint clear                         |clear [hint / sharding hint / readwrite_splitting hint] |
|sctl:hint show status                   |show [sharding / readwrite_splitting] hint status  |
|sctl:hint show table status             |- |

2. Process SCTL in accordance with the process of DistSQL without special judgment; <br/> <br/> 

The tasks are:
- [x] The syntax of `set variable`
- [x] The syntax of `show variable`
- [x] The syntax of `preview SQL`
- [x] The syntax of `set sharding hint`
- [x] The syntax of `add sharding hint`
- [x] The syntax of `set readwrite_splitting hint`
- [x] The syntax of `clear [hint / sharding hint / readwrite_splitting hint]`
- [x] The syntax of `show [sharding / readwrite_splitting] hint status`
- [x] Adjust the document
- [x]  Update Proxy hint example
- [x] Delete old SCTL code.

<br/> 
If you have any ideas and suggestions, welcome to discuss.

PR #11874, opened 2021-08-18
Remove SCTL

lanchengx:
For  #11677.

- Delete SCTL related classes;
- Modify the import of the class and the corresponding test class;


shardingsphere-sql-parser/shardingsphere-sql-parser-statement/src/main/java/org/apache/shardingsphere/sql/parser/sql/common/util/SQLUtil.java
```diff
@@ -256,4 +260,21 @@ public static ExpressionSegment createLiteralExpression(final ASTNode astNode, f
...
+     */
+    public static String trimComment(final String sql) {
+        String result = sql;
+        if (sql.startsWith(COMMENT_PREFIX)) {
+            result = result.substring(sql.indexOf(COMMENT_SUFFIX) + 2);
+        }
+        if (sql.endsWith(SQL_END)) {
+            result = result.substring(0, result.length() - 1);
+        }
+        return result.trim();
```
tristaZero, line 278: Why do we need to remove the comment from the SQL?

tristaZero (changes requested):

shardingsphere-sql-parser/shardingsphere-sql-parser-statement/src/main/java/org/apache/shardingsphere/sql/parser/sql/common/util/SQLUtil.java
lanchengx: In order to verify whether there are only comments.

tristaZero (approved):

PR #11879, opened 2021-08-18
Fix #11861

### LINKED ISSUE #11861: Suggestion: adjust the name of some example modules.
Hi community,

I have some doubts when reading the example project:
Modules like `sharding-spring-boot-jpa-example` and `sharding-spring-boot-mybatis-example` 
can know from the name that `mybatis` or `jpa` is used.
`government-spring-boot-example` integrates `mybatis` but is not named according to the above format, looks like the `ORM` is not used?

Similarly, from the name of `transaction-2pc-xa-bitronix-raw-jdbc-example`, it can be seen that `bitronix` is used, while `transaction-2pc-xa-raw-jdbc-example` uses `Atomikos` but cannot be seen from the name.

Therefore, I suggest to adjust the names of the modules to make them more intuitive and reduce user confusion.
The following are the modules I think need to be optimized:
1. governance-spring-boot-example (mybatis used)
2. governance-spring-namespace-example (mybatis used) 
3. transaction-2pc-xa-raw-jdbc-example (atomikos used)
4. encrypt-spring-boot-example (mybatis used)
5. encrypt-spring-namespace-example (mybatis used)
6. shadow-spring-boot-example (mybatis used)
7. shadow-spring-namespace-example (mybatis used)

Note that if the module name is modified,
[README.md](https://github.com/apache/shardingsphere/blob/master/examples/README.md)  and 
[README_ZH.md](https://github.com/apache/shardingsphere/blob/master/examples/README_ZH.md)also need to be modified accordingly.

yx9o:
Fixes #11861 .

Refactor module name, modify directory tree.


RaigorJiang (approved):
This PR is very important, thank you！

tristaZero (approved):

Issue #11876, opened 2021-08-18
Add skywalking-eyes for license header check.


wu-sheng:
## Feature Request

I am asking whether someone could add [skywalking-eyes](https://github.com/apache/skywalking-eyes) as a GHA in the SharingSphere repo(s). It provides a very easy way to check license header, and directly tag them on the PR.
Also, in the next 0.2.0 release, it would provide license resolution for dependency libs, we could make sure all dependencies' licenses are expected.

### Is your feature request related to a problem?

License check.

totalo:
hi, Can I do this?

wu-sheng:
Assigned to you, this should be not hard. And many repo(s) in SkyWalking, APISIX, Apollo, TiDB have been using this. You could choose which repo you want to take a reference.

totalo:
> Assigned to you, this should be not hard. And many repo(s) in SkyWalking, APISIX, Apollo, TiDB have been using this. You could choose which repo you want to take a reference.

ok.thx

tristaZero:
Hi what's the difference between Apache rat and this one? Is there any doc?

PR #11883, opened 2021-08-18
Add skywalking-eyes to check License.

totalo:
For #11876.


.mvn/wrapper/MavenWrapperDownloader.java
```diff
@@ -1,18 +1,20 @@
...
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
  *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
- *      http://www.apache.org/licenses/LICENSE-2.0
+ *     http://www.apache.org/licenses/LICENSE-2.0
```
terrymanu, line 9: I am not sure whether can change license of `MavenWrapperDownloader`, it is not source code of ShardingSphere.
Could you contact skywalking community to confirm it?

terrymanu (changes requested):

.mvn/wrapper/MavenWrapperDownloader.java
totalo: I found from the code of `skywalking` that this generation purchase is also licensed under this license. It seems that it cannot be changed. I will ignore the `.mvn` for now.

wu-sheng: Oh, this was on their original header, even it is wrong. But we don't know what should we change to. And all things not updated after that(2+ years ago)

wu-sheng: Welcome to fix this on SkyWalking site too. This was deliberate. @totalo

wu-sheng:
One tip, this should be able to replace rat plugin. 

FYI @kezhenxu94 @fgksgf we have progress here.

terrymanu (approved):

terrymanu:
Thank you for the wonderful job

.mvn/wrapper/MavenWrapperDownloader.java
totalo: Okay, I will fix this question.

PR #11924, opened 2021-08-20
Fixed execute select now() command result mistakes at proxy

### LINKED ISSUE #11617: java jdbc rs.getString(i) or rs.getTimestamp(i) fail
**apache-shardingsphere-5.0.0-RC1-SNAPSHOT**

image
 
using jdbc mysql-connector-java 8.0.13。
ResultSet rs;
canot using rs.getString(i) or rs.getTimestamp(i)

image

zjcnb:
Fixes #11617 

Fixed execute select now() command result mistakes at proxy

### DIFF
--- a/shardingsphere-db-protocol/shardingsphere-db-protocol-mysql/src/main/java/org/apache/shardingsphere/db/protocol/mysql/packet/command/query/text/MySQLTextResultSetRowPacket.java
+++ b/shardingsphere-db-protocol/shardingsphere-db-protocol-mysql/src/main/java/org/apache/shardingsphere/db/protocol/mysql/packet/command/query/text/MySQLTextResultSetRowPacket.java
@@ -24,6 +24,7 @@ import org.apache.shardingsphere.db.protocol.mysql.payload.MySQLPacketPayload;
 
 import java.math.BigDecimal;
 import java.sql.Timestamp;
+import java.time.LocalDateTime;
 import java.util.ArrayList;
 import java.util.Collection;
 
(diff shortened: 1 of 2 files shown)


shardingsphere-db-protocol/shardingsphere-db-protocol-mysql/src/main/java/org/apache/shardingsphere/db/protocol/mysql/packet/command/query/text/MySQLTextResultSetRowPacket.java
```diff
@@ -64,6 +65,8 @@ public void write(final MySQLPacketPayload payload) {
                     payload.writeStringLenenc(((BigDecimal) each).toPlainString());
                 } else if (each instanceof Boolean) {
                     payload.writeBytesLenenc((Boolean) each ? new byte[]{1} : new byte[]{0});
+                } else if (each instanceof LocalDateTime) {
+                    payload.writeStringLenenc(each.toString().replace("T", " "));
```
terrymanu, line 69: Is it better to consider about SimpleDataFormat?

zjcnb: I think, LocalDateTime is better than SimpleDateFormat. The each originally belong to LocalDateTime. So i  think it maybe needn't use SimpleDateFormat replaced.

terrymanu (changes requested):
Why getString for `DATETIME` type only?

terrymanu (approved):

PR #11956, opened 2021-08-23
Oracle load test

### LINKED ISSUE #11927: Oracle 12c dialect load `connection.getCatalog()` return null
master

shardingSphere-jdbc

### Expected behavior
Should return userName(schema name)

### Actual behavior
Return null
May be using `preparedStatement.setString(1, connection.getMetaData().getUserName());` to replace `preparedStatement.setString(1, connection.getCatalog());`
use sharding-jdbc with Oracle 12c

tuichenchuxin:
Fixes #11927.

- Oracle dialect load `collation` fix
- modify test case


tuichenchuxin:
@Liangda-w Hi, I saw you first commit about oracle load support. I tested on version 12.1. But `connection.getCatalog()` return null, I use `connection.getSchema()` to replace, then it works. I want to know why is `connection.getCatalog()` ,  `connection.getSchema()` can be ok?

Liangda-w:
> @Liangda-w Hi, I saw you first commit about oracle load support. I tested on version 12.1. But `connection.getCatalog()` return null, I use `connection.getSchema()` to replace, then it works. I want to know why is `connection.getCatalog()` , `connection.getSchema()` can be ok?

Hi, thanks for the correction! you can refer to [this](https://stackoverflow.com/questions/43109125/connection-getcatalog-returns-null-for-oracle)

strongduanmu (approved):

PR #11996, opened 2021-08-25
optimize and or condition parse and extract logic

strongduanmu:
Ref #10626, #11281

- optimize and or condition parse
- add test case for and or operator parse
- optimize `AndPredicate` extract logic


strongduanmu:
In different database dialects, the `and` operator has a higher priority than the `or` operator, so it should be parsed according to the priority when parsing.

However, the parsing engine does not distinguish the priority of `and` and `or`, which results in incorrect parsing results of the following sentences, and ultimately leads to extremely complicated logic for extracting `AndPredicate` when using the parsing results.

```sql
SELECT * FROM t_order WHERE a = 1 AND b = 2 OR c = 3 AND d = 4 OR e = 5 AND f = 6;
```

Current parse result:

image

Optimized parse result:

image

tristaZero (approved):

PR #12111, opened 2021-08-30
Remove useless code.

totalo:


JiekerTime (approved):
After observing, I think the deleted classes are belong to duplicate initialization classes and can be merged.

menghaoranss (approved):

PR #12162, opened 2021-09-01
Enhance EncryptAlgorithm and QueryAssistedEncryptAlgorithm

### LINKED ISSUE #11689: [enhancement]how about add context parameters on encrypt and decrypt method
## Feature Request

We need different key on each table/column to encrypt and decrypt with, how about add content parameters as below code, for example: logicTable, logicColumn etc? 

`/**
 * Encryptor.
 */
public interface Encryptor extends TypeBasedSPI {
    
    /**
     * Initialize.
     */
    void init();
    
    /**
     * Encode.
     * 
     * @param plaintext plaintext
     * @param encryptContextMap encrypt context map(logicTable and logicColumn)
     * @return ciphertext
     */
    String encrypt(Object plaintext, **Map<String, Object> encryptContextMap**);
    
    /**
     * Decode.
     * 
     * @param ciphertext ciphertext
     * @param decryptContextMap decrypt context map(logicTable and logicColumn)
     * @return plaintext
     */
    Object decrypt(String ciphertext, **Map<String, Object> decryptContextMap**);
}`

cheese8:
@tristaZero Please help to review, thanks.

Fixes #11689.

- modify encrypt method and add parameter Map<String, String> encryptContextMap
- modify queryAssistedEncrypt method and add parameter Map<String, String> encryptContextMap
- migrate RuleConfigurationLogger to log and debug the value set on EncryptConfiguration, will refactoring it later


shardingsphere-features/shardingsphere-encrypt/shardingsphere-encrypt-api/src/main/java/org/apache/shardingsphere/encrypt/spi/EncryptAlgorithm.java
```diff
@@ -29,9 +31,10 @@
      * Encode.
      *
      * @param plaintext plaintext
+     * @param encryptContextMap encrypt context map which include keys: database, table and column etc
      * @return ciphertext
      */
-    String encrypt(Object plaintext);
+    String encrypt(Object plaintext, Map<String, String> encryptContextMap);
```
taojintianxia, line 37: what does this and following several maps use for ? and why not final this time

strongduanmu:
Hi @cheese8 , can you fix ci error first?

PR #12185, opened 2021-09-03
bindingTable associated query should judge related column. #11793

### LINKED ISSUE #11793: When bindingTable for associated query and they are not associated by sharding key, they should use shardingFederatedRoutingEngine to Routing
## Bug Report

5.0.0-RC1-SNAPSHOT

ShardingSphere-Proxy

### Expected behavior
When bindingTable for associated query and they are not related by sharding key, they should use shardingFederatedRoutingEngine to routing.  If we want to treat them as binding tables, their association key should also be sharding key.

### Actual behavior
When bindingTable for associated query and they are not related by sharding key, they use ShardingStandardRoutingEngine to routing, this will cause some eligible data can't be querying out because of some table associations are missing.

image
When selecting the routing engine, ShardingStandardRoutingEngineis is selected only by judging that all tables are binding tables, but they should be judged whether associated by sharding key, if they are associated by sharding key , ShardingStandardRoutingEngine should be selected, otherwise ShardingFederatedRoutingEngine should be selected .

1. there are two binding tables t_order and t_order_item, their sharding key are order_id ,  they also have a column user_id .
2. execute this sql：
```
 select t1.* from t_order t1 join t_order_item t2 on t1.user_id = t2.user_id where t1.order_id in (10, 11);
```
if  value 10 routing to sharding 0， value 11 routing to sharding 1, they are binding tables ,so will use ShardingStandardRoutingEngineis to routing , and the rewrite sql will be:
```
 select t1.* from t_order_0 t1 join t_order_item_0 t2 on t1.user_id = t2.user_id where t1.order_id in (10, 11);
 select t1.* from t_order_1 t1 join t_order_item_1 t2 on t1.user_id = t2.user_id where t1.order_id in (10, 11);
```
but if t_order_0 and t_order_item_1 also have the same user_id value, the data can't be querying out. the number of correct rewrite sql should be 4, they are:
```
 select t1.* from t_order_0 t1 join t_order_item_0 t2 on t1.user_id = t2.user_id where t1.order_id in (10, 11);
 select t1.* from t_order_0 t1 join t_order_item_1 t2 on t1.user_id = t2.user_id where t1.order_id in (10, 11);
 select t1.* from t_order_1 t1 join t_order_item_0 t2 on t1.user_id = t2.user_id where t1.order_id in (10, 11);
 select t1.* from t_order_1 t1 join t_order_item_1 t2 on t1.user_id = t2.user_id where t1.order_id in (10, 11);
```
**conclusion: if we want to treat them as binding tables, their association key should also be sharding key.**

chengh1:
Fixes #11793.

- When choose routeEngine , excute isShardingStandardQuery method, if it is a associated query, add judging related conditions.
- add testcase in shardingRuleTest.


chengh1:
All checks that run fail are due to using calcite for association query. There may be some problems when using calcite for related queries.

chengh1:
> All checks that run fail are due to using calcite for association query. There may be some problems when using calcite for related queries.

For example, when I use unbinding tables to perform a related query, it will use calcite to excute this sql, if the related column is type of string, it will run fail and throw a exception with "java.lang.String cannot be cast to java.lang.Integer".

tristaZero:
@chengh1 Hi thanks for your thinking and PR. Let our committer give a check. @strongduanmu I presume @chengh1 needs some help, could you have a look？

strongduanmu:
Hi @chengh1, can you fix ci error first?

tristaZero:
Hi @chengh1 
I know you must be swamped. But when you have time off, could you have a look here?

chengh1:
> Hi @chengh1
> I know you must be swamped. But when you have time off, could you have a look here?

@tristaZero 
It’s been a long time since the pr was submitted, but there is a problem when running the integration test module. When using calcite to execute the join query, if the query field has a date type field, an exception which is  "java.lang.ClassCastException: java.sql.Date cannot be cast to java.lang.Number" will be thrown. I think there are some code problems in the integration test module when excuting with calcite, but I haven't found the reason yet, so it was delayed for a while. 
Because I modified the logic of selecting the routing engine, some sql need to be executed using calcite, so some sql of the original integration test module will have problems when executed with calcite. For example,  the following SQL which is "select t.*,o.item_id as item_id,(case when t.status = 'init' then '已启用' when t.status = 'failed' then '已停用' end) as stateName from t_order t left join t_order_item as o on o.order_id =t.order_id where t.order_id=1000 limit 1 " contains Chinese, it will throw an encoding exception when executed with calcite, I have used the method of adding a calcite configuration file to specify charset as utf8 to resolved it,But there is still a problem which I mentioned above, I haven't found a solution yet.
Problem link : https://github.com/apache/shardingsphere/pull/12185/checks?check_run_id=3583715210

strongduanmu:
Thank you for your effort @chengh1, calcite currently executes the SQL containing the time field abnormally, and it will take some time to fix it. 

BTW, this PR has a great impact on the concept of binding tables. We need to consider what conditions are met to be considered binding tables and update related documents.

chengh1:
> Thank you for your effort @chengh1, calcite currently executes the SQL containing the time field abnormally, and it will take some time to fix it.
> 
> BTW, this PR has a great impact on the concept of binding tables. We need to consider what conditions are met to be considered binding tables and update related documents.

 ok, I think when the join query is executed ,if we want to treat them as binding tables, their association key should also be sharding key.

strongduanmu:
Hi @chengh1, calcite cast exception has been fixed. Can you sync master branch to your pr?

chengh1:
@strongduanmu hi，some checks run fail in the ShardingRuleTest, the reason is that the method assertGetTables runs incorrectly. The expected value is "LOGIC_TABLE", "SUB_LOGIC_TABLE", but the actual result is "logic_table", "sub_logic_table". I found out that you changed this piece of code on September 2nd. Can you look at this problem? thank you.

strongduanmu:
> @strongduanmu hi，some checks run fail in the ShardingRuleTest, the reason is that the method assertGetTables runs incorrectly. The expected value is "LOGIC_TABLE", "SUB_LOGIC_TABLE", but the actual result is "logic_table", "sub_logic_table". I found out that you changed this piece of code on September 2nd. Can you look at this problem? thank you.

@chengh1 This change is to optimize shardingsphere's handling of upper and lower case, which is an expected change. You need to compare PR and master branches to ensure that your code is correct.

PR #12192, opened 2021-09-03
Fix create encrypt rule PROPERTIES doc

sandynz:
- Update create encrypt rule PROPERTIES doc

### DIFF
--- a/docs/document/content/features/dist-sql/syntax/rdl/rdl-encrypt-rule.cn.md
+++ b/docs/document/content/features/dist-sql/syntax/rdl/rdl-encrypt-rule.cn.md
@@ -36,18 +36,18 @@ algorithmProperty:
 ```sql
 CREATE ENCRYPT RULE t_encrypt (
 COLUMNS(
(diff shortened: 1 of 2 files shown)


sandynz:
Test result:
```
shard1=> CREATE ENCRYPT RULE t_encrypt (
COLUMNS(
(NAME=user_id,PLAIN=user_plain,CIPHER=user_cipher,TYPE(NAME=AES,PROPERTIES('aes-key-value'=123456abc))),
(NAME=order_id,CIPHER=order_cipher,TYPE(NAME=MD5))
));

ERROR:  You have an error in your SQL syntax
```

From `RDLStatement.g4`
```
algorithmProperty
    : key=(IDENTIFIER | STRING) EQ value=(NUMBER | INT | STRING)
    ;
```
`value=(NUMBER | INT | STRING)`, `123456abc` does not match. Update it to `'123456abc'` solve it.

menghaoranss (approved):

PR #12197, opened 2021-09-03
Remove the check method of sharding rule

### LINKED ISSUE #12195: When there is only one `sharding table rule` on ZK,  `drop sharding table` rule will fail
Hi community,
There may be a problem about `ShardingRuleConfigurationChecker`

image

image

This method will be used when persistent configuration to zk.
So when only `rules.shardingAlgorithms` is configured in my configuration file, proxy startup will fail.
In addition, when the `schema` has only one `sharding table rule`, using DistSQL to delete the `sharding table rule` will also fail

image

lanchengx:
Fixes #12195

- Remove the check method of sharding rule

When I use `drop sharding table rule` to delete the last rule, the `tables` and `autoTables` in the local `ShardingSphereMetaData` will be empty, which will cause persistence failure.

When starting and modifying `sharding table rule`, local data should be persisted to ZK


tristaZero (approved):

tristaZero (changes requested):
A draft, right?

lanchengx:
> A draft, right?

Right, after discussing with @menghaoranss , the plan needs to be readjusted. This pr can be closed.

PR #12324, opened 2021-09-09
Improve resource release logic to prevent memory leaks.

yx9o:
Improve resource release logic to prevent memory leaks.

### DIFF
--- a/shardingsphere-kernel/shardingsphere-authority/shardingsphere-authority-core/src/main/java/org/apache/shardingsphere/authority/provider/natived/builder/dialect/MySQLPrivilegeHandler.java
+++ b/shardingsphere-kernel/shardingsphere-authority/shardingsphere-authority-core/src/main/java/org/apache/shardingsphere/authority/provider/natived/builder/dialect/MySQLPrivilegeHandler.java
@@ -56,12 +56,13 @@ public final class MySQLPrivilegeHandler implements StoragePrivilegeHandler {
     @Override
     public Collection<ShardingSphereUser> diff(final Collection<ShardingSphereUser> users, final DataSource dataSource) throws SQLException {
         Collection<Grantee> grantees = new LinkedList<>();
-        try (Connection connection = dataSource.getConnection()) {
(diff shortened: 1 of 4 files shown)


shardingsphere-kernel/shardingsphere-authority/shardingsphere-authority-core/src/main/java/org/apache/shardingsphere/authority/provider/natived/builder/dialect/MySQLPrivilegeHandler.java
```diff
@@ -56,12 +56,13 @@
...
-                while (resultSet.next()) {
-                    grantees.add(new Grantee(resultSet.getString("user"), resultSet.getString("host")));
-                }
+        try (
+                Connection connection = dataSource.getConnection();
+                Statement statement = connection.createStatement();
+                ResultSet resultSet = statement.executeQuery(getGlobalPrivilegesSQL(users))
+        ) {
+            while (resultSet.next()) {
+                grantees.add(new Grantee(resultSet.getString("user"), resultSet.getString("host")));
```
terrymanu, line 65: Why the original code may resource leak?
The connection, statement and resultSet are all in try with resource

terrymanu (changes requested):

shardingsphere-kernel/shardingsphere-authority/shardingsphere-authority-core/src/main/java/org/apache/shardingsphere/authority/provider/natived/builder/dialect/MySQLPrivilegeHandler.java
yx9o: > Why the original code may resource leak?
> The connection, statement and resultSet are all in try with resource

`Statement statement = connection.createStatement();` not in try with resource.

yx9o:
@terrymanu Hi, please review, thank you.

PR #12350, opened 2021-09-10
Fix : refactor the document since the namespace for jdbc url has been…

taojintianxia:


terrymanu (changes requested):
The `url` or `jdbc-url` are dependent with the attribute of data source connection pool
for dbcp is `url`, for hikari is `jdbc-url`.

I notice the document is not mention about data source connection pool.
It is better to add description of data source connection pool, but not change the customized configuration directly.

Issue #12305, opened 2021-09-09
Add unit tests for DefaultDataSourceCreator


strongduanmu:
This issue is for adding unit tests for DefaultDataSourceCreator.
If you are interested, please leave a message.

CodingBingo:
I am interested to provide unit tests for DefaultDataSourceCreator;
I will try to finish this task soon

strongduanmu:
@CodingBingo Welcome.

CodingBingo:
> @CodingBingo Welcome.

Hi, this issue can be closed now.

strongduanmu:
> > @CodingBingo Welcome.
> 
> Hi, this issue can be closed now.

Hi @CodingBingo, next time you can use `Fixes #11589` format, after the PR is merged, the issue will be closed automatically.

strongduanmu:
Fixed by #12367.

PR #12367, opened 2021-09-11
add some unit test for DefaultDataSourceCreator

CodingBingo:
https://github.com/apache/shardingsphere/issues/12305
add some unit test for DefaultDataSourceCreator

### DIFF
--- a/shardingsphere-infra/shardingsphere-infra-common/src/test/java/org/apache/shardingsphere/infra/config/datasource/creator/DefaultDataSourceCreatorTest.java
+++ b/shardingsphere-infra/shardingsphere-infra-common/src/test/java/org/apache/shardingsphere/infra/config/datasource/creator/DefaultDataSourceCreatorTest.java
@@ -1,0 +1,70 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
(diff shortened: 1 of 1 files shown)


shardingsphere-infra/shardingsphere-infra-common/src/test/java/org/apache/shardingsphere/infra/config/datasource/creator/DefaultDataSourceCreatorTest.java
```diff
@@ -0,0 +1,71 @@
...
+import org.apache.shardingsphere.infra.config.datasource.creator.impl.DefaultDataSourceCreator;
+import org.junit.Test;
+
+import javax.sql.DataSource;
+
+import static org.hamcrest.CoreMatchers.instanceOf;
+import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assert.assertThat;
+
+public class DefaultDataSourceCreatorTest {
```
tristaZero, line 31: Hi @CodingBingo It looks nice. One suggestion here is that this class is better to be a final one. How do you think?

tristaZero (changes requested):

shardingsphere-infra/shardingsphere-infra-common/src/test/java/org/apache/shardingsphere/infra/config/datasource/creator/DefaultDataSourceCreatorTest.java
CodingBingo: Yeah, sorry for not careful enough. I will change it right now.

tristaZero (approved):
@CodingBingo Hey, nice job.

PR #12412, opened 2021-09-14
Refactor jdbc url doc

taojintianxia:
add comment for jdbc-url and url


terrymanu (changes requested):
It is not for jdbc-url only.
Every attribute name may different with different connection poll

PR #12413, opened 2021-09-14
[DistSQL] Add set and show syntax for agent plugins variable

RaigorJiang:
- support `set variable agent_plugins_enabled = xxx`,  xxx must be `true` or `false`, if not, `false` will be setted;
- support `show variable agent_plugins_enabled`;
- add test cases;

### DIFF
--- a/shardingsphere-proxy/shardingsphere-proxy-backend/src/main/java/org/apache/shardingsphere/proxy/backend/context/ProxyContext.java
+++ b/shardingsphere-proxy/shardingsphere-proxy-backend/src/main/java/org/apache/shardingsphere/proxy/backend/context/ProxyContext.java
@@ -131,4 +131,25 @@ public final class ProxyContext {
         result.addAll(contextManager.getMetaDataContexts().getGlobalRuleMetaData().getRules());
         return result;
     }
+    
+    /**
+     * Set system property.
+     *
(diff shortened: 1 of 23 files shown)


shardingsphere-proxy/shardingsphere-proxy-backend/src/main/java/org/apache/shardingsphere/proxy/backend/context/ProxyContext.java
```diff
@@ -131,4 +131,25 @@ public StateContext getStateContext() {
...
         return result;
     }
+    
+    /**
+     * Set system property.
+     *
+     * @param key property key
+     * @param value property value
+     */
+    public void setSystemProperty(final String key, final String value) {
```
taojintianxia, line 141: for me, I prefer to extract these two methods into a higher level of util or something like that.en .... you decide that, this is not mandatory .

RaigorJiang: Thank you @taojintianxia .
I also had this idea and discussed it yesterday, but we didn't find a suitable place in the end.

tristaZero (approved):

PR #12423, opened 2021-09-14
Use port instead of pid in ClusterInstance for proxy

### LINKED ISSUE #12396: Use port instead of pid in ClusterInstance for proxy

menghaoranss:
Fixes #12396 .

### DIFF
--- a/shardingsphere-jdbc/shardingsphere-jdbc-core/src/main/java/org/apache/shardingsphere/driver/jdbc/core/datasource/ShardingSphereDataSource.java
+++ b/shardingsphere-jdbc/shardingsphere-jdbc-core/src/main/java/org/apache/shardingsphere/driver/jdbc/core/datasource/ShardingSphereDataSource.java
@@ -74,7 +74,7 @@ public final class ShardingSphereDataSource extends AbstractUnsupportedOperation
         Map<String, Collection<RuleConfiguration>> schemaRuleConfigs = Collections.singletonMap(
                 schemaName, ruleConfigs.stream().filter(each -> each instanceof SchemaRuleConfiguration).collect(Collectors.toList()));
         Collection<RuleConfiguration> globalRuleConfigs = ruleConfigs.stream().filter(each -> each instanceof GlobalRuleConfiguration).collect(Collectors.toList());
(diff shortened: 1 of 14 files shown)


shardingsphere-jdbc/shardingsphere-jdbc-core/src/main/java/org/apache/shardingsphere/driver/jdbc/core/datasource/ShardingSphereDataSource.java
```diff
@@ -74,7 +74,7 @@ private ContextManager createContextManager(final String schemaName, final ModeC
         Map<String, Collection<RuleConfiguration>> schemaRuleConfigs = Collections.singletonMap(
                 schemaName, ruleConfigs.stream().filter(each -> each instanceof SchemaRuleConfiguration).collect(Collectors.toList()));
         Collection<RuleConfiguration> globalRuleConfigs = ruleConfigs.stream().filter(each -> each instanceof GlobalRuleConfiguration).collect(Collectors.toList());
-        return ContextManagerBuilderFactory.newInstance(modeConfig).build(modeConfig, dataSourcesMap, schemaRuleConfigs, globalRuleConfigs, props, isOverwrite);
+        return ContextManagerBuilderFactory.newInstance(modeConfig).build(modeConfig, dataSourcesMap, schemaRuleConfigs, globalRuleConfigs, props, isOverwrite, null);
```
taojintianxia, line 77: there are 7 parameters, looks like not that graceful . Haoran, this method is build, why not refactor it as builder ? so you could build the parameter like ContextManagerBuilderFactory.newInstance(modeConfig).builder().modeConfig(modeConfig).overwrite(isOverwrite).xxx(xxx).build();

taojintianxia: BTW, there is an anotation may help you ,  `@Builder` in lombok

taojintianxia: and, again, null is not a good choice.

terrymanu (approved):

PR #12451, opened 2021-09-15
Fix Issue#12449 : Scaling job does not stop properly if exception thrown before inventory task start

### LINKED ISSUE #12449: If a scaling job failed or stopped then it could not be started by DistSQL except restarting proxy
5.0.0-RC1-SNAPSHOT

ShardingSphere-Scaling, commit: 46aeeff1b2e837fe514df3097e4945057b911e7a

### Expected behavior
Job could be stopped when exception occur.

### Actual behavior
It need to stop job manually, else it couldn't be started again.

In `ScalingJobExecutor.java`
```
    private void execute(final JobConfigurationPOJO jobConfigPOJO) {
        if (EXECUTING_JOBS.add(jobConfigPOJO.getJobName())) {
            new OneOffJobBootstrap(ScalingAPIFactory.getRegistryCenter(), new ScalingJob(), jobConfigPOJO.toJobConfiguration()).execute();
        }
    }
```
job name is still cached in `EXECUTING_JOBS`, since job is not stopped automatically.

Submit wrong srouce/target parameter in HTTP api `/scaling/job/start`.

sandynz:
Fixes #12449.

- Stop job if exception thrown before inventory task start
- `dumper` might be null in inventory and incremental task `onFailure`

### DIFF
--- a/shardingsphere-scaling/shardingsphere-scaling-core/src/main/java/org/apache/shardingsphere/scaling/core/api/ScalingWorker.java
+++ b/shardingsphere-scaling/shardingsphere-scaling-core/src/main/java/org/apache/shardingsphere/scaling/core/api/ScalingWorker.java
@@ -18,6 +18,7 @@
 package org.apache.shardingsphere.scaling.core.api;
 
 import com.google.common.eventbus.Subscribe;
+import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.shardingsphere.infra.eventbus.ShardingSphereEventBus;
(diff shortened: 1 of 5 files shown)


sandynz:
Two questions to verify:
- If job failed in a proxy instance, whether need to stop job in other proxy instances (in proxy cluster) immediately
- If `ScalingJob.execute` (in `OneOffJobBootstrap`) just throw exception, looks quartz will retry 150 times quickly, maybe we need to wait completion event from ElasticJob, but not just stop job on exception caught

I'll continue to do it next week.

sandynz:
There're too much conflicts now.

It will be closed now, and will be started after scaling job lock done (only one proxy instance could run scaling job in proxy cluster).

PR #12530, opened 2021-09-17
fix wrong rewrite result without KeyGenerateStrategy

### LINKED ISSUE #12528: Wrong sql rewrite when user does not config KeyGenerateStrategy

master branch

proxy or jdbc

### Expected behavior

execute successfully

### Actual behavior

Throw exception because wrong sql rewrite and user does not config KeyGenerateStrategy.

image

Because DialectTableMetaDataLoader loaded the original information of the table, and did not process the generated attribute, a logical error occurred when judging whether to generate a distributed primary key.

```java
/**
 * Generated key insert column token generator.
 */
public final class GeneratedKeyInsertColumnTokenGenerator extends BaseGeneratedKeyTokenGenerator {
    
    @Override
    protected boolean isGenerateSQLToken(final InsertStatementContext insertStatementContext) {
        Optional<InsertColumnsSegment> sqlSegment = insertStatementContext.getSqlStatement().getInsertColumns();
        return sqlSegment.isPresent() && !sqlSegment.get().getColumns().isEmpty()
                && insertStatementContext.getGeneratedKeyContext().isPresent()
                && !insertStatementContext.getGeneratedKeyContext().get().getGeneratedValues().isEmpty();
    }
...
}
```

strongduanmu:
Fixes #12528.

- Modify `TableMetaDataLoader` to load the real generated column
- modify the decorate method of the `ShardingTableMetaDataBuilder` to meet the needs of kernel rewriting

### DIFF
--- a/shardingsphere-features/shardingsphere-sharding/shardingsphere-sharding-core/src/main/java/org/apache/shardingsphere/sharding/metadata/ShardingTableMetaDataBuilder.java
+++ b/shardingsphere-features/shardingsphere-sharding/shardingsphere-sharding-core/src/main/java/org/apache/shardingsphere/sharding/metadata/ShardingTableMetaDataBuilder.java
@@ -117,13 +117,9 @@ public final class ShardingTableMetaDataBuilder implements RuleBasedTableMetaDat
     }
     
     private Collection<ColumnMetaData> getColumnMetaDataList(final TableMetaData tableMetaData, final TableRule tableRule) {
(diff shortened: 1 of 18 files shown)


zjcnb:
id column is primary key auto_increment and  configuration SNOWFLAKE algorithm, i think id value is SNOWFLAKE value, but it's database increment when i insert one record.  When i batch insert, it's SNOWFLAKE value.

tristaZero (approved):

PR #12579, opened 2021-09-18
Support to build with JDK 11

### LINKED ISSUE #12578: Support JDK 11 build

I want to build shadingsphere with JDK 11

zhfeng:
Fixes #12578

- Add dependency of ```javax.xml.bind::jaxb-api``` and ```javax.annotation::javax.annotation-api```
- minor fix the generic type complitation error in ```SchemaYamlSwapper:swapToYamlConfiguration```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -107,6 +107,9 @@
         <calcite.version>1.27.0</calcite.version>
         <embedded-mysql.version>4.6.1</embedded-mysql.version>
         <embedded-postgresql.version>2.10</embedded-postgresql.version>
+
+        <jaxb.version>2.3.0</jaxb.version>
+        <annotation-api.version>1.3.2</annotation-api.version>
         
         <!-- Plugin versions -->
         <apache-rat-plugin.version>0.12</apache-rat-plugin.version>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -887,6 +890,27 @@
         </plugins>
     </build>
     
+    <profiles>
```
TeslaCN, line 893: Hi @zhfeng 
Could you fix the indent of the tags?

pom.xml
```diff
@@ -887,6 +890,27 @@
...
+        <profile>
+            <id>jdk11+</id>
+            <activation>
+                <jdk>[11,)</jdk>
```
TeslaCN, line 897: The javax annotation was removed since JDK 9. So maybe the low bound should be 1.9?

pom.xml
zhfeng: Done

pom.xml
zhfeng: Since JDK 9 is not a LTS verion, I think it will not be in the supported JDK list.

TeslaCN (approved):
LGTM

zhfeng:
Thanks @terrymanu and it looks like there are some failing tests when runnning with JDK 11. I'm investigating and will creat following PRs to fix them.

PR #12645, opened 2021-09-23
Refactor federation meta data related classes

terrymanu:


shardingsphere-infra/shardingsphere-infra-optimize/src/main/java/org/apache/shardingsphere/infra/optimize/core/metadata/FederationTableMetaData.java
```diff
@@ -29,34 +31,38 @@
...
-import java.util.stream.Collectors;
 
 /**
  * Federation table meta data.
  */
 @Getter
 public final class FederationTableMetaData {
     
-    private static final RelDataTypeFactory TYPE_FACTORY = new SqlTypeFactoryImpl(RelDataTypeSystem.DEFAULT);   
+    private static final RelDataTypeFactory REL_DATA_TYPE_FACTORY = new SqlTypeFactoryImpl(RelDataTypeSystem.DEFAULT);
```
taojintianxia, line 41: except the final constant, do we need to make the name upper case even it's a class ?

strongduanmu (approved):

PR #12657, opened 2021-09-23
Read data source disabled

lanchengx:
- Add setStatementExecutor
- Move the `setVariable` function to the executor
- Add `setStatusExecutor`
- Add tests


shardingsphere-proxy/shardingsphere-proxy-backend/src/main/java/org/apache/shardingsphere/proxy/backend/text/distsql/ral/common/set/excutor/SetStatusExecutor.java
```diff
@@ -0,0 +1,73 @@
...
+        if (null == schemaName) {
+            throw new NoDatabaseSelectedException();
+        }
+        if (!ProxyContext.getInstance().getAllSchemaNames().contains(schemaName)) {
+            throw new SchemaNotExistedException(schemaName);
+        }
+        String resourceName = sqlStatement.getResourceName();
+        Collection<String> notExistedResources = ProxyContext.getInstance().getMetaData(schemaName).getResource().getNotExistedResources(Collections.singleton(resourceName));
+        DistSQLException.predictionThrow(notExistedResources.isEmpty(), new RequiredResourceMissedException(schemaName, Collections.singleton(resourceName)));
+        if (sqlStatement.getFeatureName().equals(READWRITE_SPLITTING)) {
```
RaigorJiang, line 61: Do we need to ignore case?

lanchengx: > Do we need to ignore case?

There may be new cases, like `circuit breaker` and `shadow`

RaigorJiang (approved):

Issue #11661, opened 2021-08-05
Refactoring plan of the shardingsphere-shadow project


soulasuna:
Hi,community:

I am very happy to bring you the refactoring plan of theshardingsphere-shadow  project. @terrymanu @tristaZero @wu-sheng

Combining the characteristics of the shardingsphere-shadow project's own functions and the configuration style of other functions of the shardingsphere.Make changes to the user API configuration.

In the future, shardingsphere-shadow will become an important part of data routing in full-link stress test scenarios.

The process of project reconstruction will not have any impact on the original functions.

#### After the refactoring is completed, the user API configuration changes are as follows:
- Spring-namespace xml file configuration

```xml
<beans xmlns="http://www.springframework.org/schema/beans"
       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
       xmlns:shadow="http://shardingsphere.apache.org/schema/shardingsphere/shadow"
       xsi:schemaLocation="http://www.springframework.org/schema/beans 
                           http://www.springframework.org/schema/beans/spring-beans.xsd
                           http://shardingsphere.apache.org/schema/shardingsphere/shadow
                           http://shardingsphere.apache.org/schema/shardingsphere/shadow/shadow.xsd
                           ">
                           
    <shadow:shadow-algorithm id="user-id-match-algorithm" type="COLUMN-REGULAR-MATCH">
        <props>
            <prop key="operation">UPDATE</prop>
            <prop key="column">user_id</prop>
            <prop key="regex">[a]</prop>
        </props>
    </shadow:shadow-algorithm>
    <shadow:shadow-algorithm id="note-algorithm" type="SIMPLE-NOTE">
        <props>
            <prop key="shadow">true</prop>
        </props>
    </shadow:shadow-algorithm>
    
    <shadow:rule id="shadowRule">
        <shadow:data-source id="shadow-data-source" source-data-source-name="ds" shadow-data-source-name="ds-shadow"/>

        <shadow:shadow-table name="t_order">
            <shadow:shadow-algorithm ref= "user-id-match-algorithm" />
            <shadow:shadow-algorithm ref= "note-algorithm" />
        </shadow:shadow-table>
        <shadow:shadow-table name="t_user">
            <shadow:shadow-algorithm ref= "user-id-match-algorithm" />
        </shadow:shadow-table>
    </shadow:rule>
</beans>
```

- Spring-boot properties file configuration

```properties
[...]

PR #12660, opened 2021-09-23
Shadow non-DML route

soulasuna:
##  Add shadow non-DML statement routing engine.

For  #11661.

- Refactor note shadow algorithm.
- Add shadow non-DML statement routing engine.
- Fix ut.
- Add sql note case example.
- Refactor future shadow spring boot mybatis example.


examples/example-core/example-api/src/main/java/org/apache/shardingsphere/example/core/api/entity/FutureShadowUser.java
```diff
@@ -0,0 +1,91 @@
...
+    
+    private String userName;
+    
+    private String userNamePlain;
+    
+    private String pwd;
+    
+    private String assistedQueryPwd;
+    
+    public int getUserId() {
```
taojintianxia, line 38: is this generated by some framework ? why not use lombok

taojintianxia (approved):

tristaZero (approved):

Issue #12619, opened 2021-09-22
Update meta data context in memory for RDL


menghaoranss:
refer #11709 

Search the comment `// TODO update meta data context in memory` and fix them.

PR #12820, opened 2021-09-29
Refresh memory meta data for create and drop database

menghaoranss:
For #12619 .

### DIFF
--- a/shardingsphere-mode/shardingsphere-mode-core/src/main/java/org/apache/shardingsphere/mode/manager/ContextManager.java
+++ b/shardingsphere-mode/shardingsphere-mode-core/src/main/java/org/apache/shardingsphere/mode/manager/ContextManager.java
@@ -18,13 +18,29 @@
 package org.apache.shardingsphere.mode.manager;
 
 import lombok.Getter;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.shardingsphere.infra.config.RuleConfiguration;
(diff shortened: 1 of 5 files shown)


strongduanmu (approved):

shardingsphere-mode/shardingsphere-mode-core/src/main/java/org/apache/shardingsphere/mode/manager/ContextManager.java
```diff
@@ -63,6 +79,74 @@ public synchronized void renewTransactionContexts(final TransactionContexts tran
...
+            metaDataContexts.getOptimizerContext().getParserContexts().remove(schemaName);
+            metaDataContexts.getOptimizerContext().getPlannerContexts().remove(schemaName);
+            ShardingSphereMetaData removeMetaData = metaDataContexts.getMetaDataMap().remove(schemaName);
+            closeDataSources(removeMetaData);
+            closeTransactionEngine(schemaName);
+        }
+    }
+    
+    private MetaDataContexts buildNewMetaDataContext(final String schemaName) throws SQLException {
+        Map<String, Map<String, DataSource>> dataSourcesMap = Collections.singletonMap(schemaName, new HashMap<>());
```
taojintianxia, line 115: map inside a map ? I'm not sure if multimap is better here . not formaliar with that . 

multimap ... from guava.

taojintianxia (approved):

Issue #12964, opened 2021-10-09
PostgreSQL Proxy integration test failed


TeslaCN:
### Which version of ShardingSphere did you use?
master 6a70d74de389b16dd2b1a30c62c15f9f49954d52

### Expected behavior
Pass PostgreSQL Proxy integration test

### Actual behavior
Integration tests failed.

### Reason analyze (If you can)

Proxy returned duplicated ReadyForQuery when error occurred, which messed up the frontend connection.

The 168 ReadyForQuery is duplicated, which messed up the frontend connection.

![image](https://user-images.githubusercontent.com/20503072/136653586-9a688938-c956-4f69-b487-a6a0f9372d83.png)

[pg_proxy_it_failed.pcapng.zip](https://github.com/apache/shardingsphere/files/7315345/pg_proxy_it_failed.pcapng.zip)

### Steps to reproduce the behavior, such as: SQL to execute, sharding rule configuration, when exception occur etc.

```
./mvnw -B clean install -am -pl shardingsphere-test/shardingsphere-integration-test/shardingsphere-integration-test-suite -DskipTests -Dmaven.javadoc.skip=true -Dcheckstyle.skip=true -Drat.skip=true -Djacoco.skip=true -T1C -Ddockerfile.useProxy=false -f pom.xml -Dit.env.docker
./mvnw -B install -Dit.adapters=proxy -Dit.databases=PostgreSQL -Dit.scenarios=db -Dit.env.type=DOCKER -f shardingsphere-test/shardingsphere-integration-test/shardingsphere-integration-test-suite/pom.xml
```

PR #12987, opened 2021-10-11
Fix PostgreSQL Proxy duplicate ReadyForQuery

TeslaCN:
Related to #12964


TeslaCN:
Integration tests may failed because of #13015

PR #13073, opened 2021-10-17
Think of it a different way to solve BeanDefinition is the abstract c…

### LINKED ISSUE #10366: springxml define abstract bean result in NullPointerException
Bug Report
xml define abstract bean as the figure shows
image
NullPointerException bugs as shown in the figure
image

notzero1995:
Sorry to push  the code so late ， although it has been repaired ， but I provide a new implementation

Fixes #10366.


tristaZero:
Hi @notzero1995 anyway, glad to see you come back.
@strongduanmu  or @TeslaCN could you have a review here?

PR #13152, opened 2021-10-19
change columnName to columnLabel for information_schema

### LINKED ISSUE #13090: header of information_schema is always ColumnName instead of ColumnLabel
5.0.0-RC1

ShardingSphere-Proxy

### Expected behavior
header of information_schema should be column label

### Actual behavior
header of information_schema is column name, for example 
```
select `COLUMN_NAME` as `cn`  from information_schema.columns where table_schema = 'test' and table_name = 'tbl'
```
should return cn, where actually return COLUMN_NAME now.

```
select `COLUMN_NAME` as `cn`  from information_schema.columns where table_schema = 'test' and table_name = 'tbl'
```

roodkcab:
Fixes #13090


lanchengx:
Hi, @roodkcab please synchronize the  master branch to the current pr to solve the problem of CI failure

PR #13163, opened 2021-10-20
Update licenses for dependencies

TeslaCN:


taojintianxia:
@TeslaCN just curiosity, what did log4j update in this version plus.

PR #13169, opened 2021-10-20
Adding validation for invalid dataNode String

### LINKED ISSUE #9596: DataNode constructor 'DataNode(final String dataNode)' may be throw invalid Exception when the args 'dataNode'  without db name or table name
## Bug Report

As the title

shardingshpere-5.0.0-RC1-SNAPSHOT
ShardingSphere-JDBC

when I  create a new DataNode instance as:
`
new DataNode("db.");
new DataNode(".table");
`
It can work normally，but actually they are invalid instance, so I think the method isValidDataNode should be check this case.

puneetjakhar:
Fixes #9596.

Added validations for invalid dataNode String.
Handled empty data source name or table name in dataNode String.


puneetjakhar:
@strongduanmu thanks for pointing out, updated PR with that change.

Issue #12973, opened 2021-10-10
SqlNode is not the same as SQL


wgy8283335:
### sqlNode is not the same as SQL
SQL:
SELECT order_id, user_id, (SELECT MAX(user_id) FROM t_user_info) FROM t_order_federate

After convert to SQLNode:
SELECT `order_id`, `user_id`, (SELECT MAX(*) FROM `t_user_info`) FROM `t_order_federate` 

This will cause an error in validation.
```
org.apache.calcite.runtime.CalciteContextException: At line 0, column 0: Unknown identifier '*'

	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
	at org.apache.calcite.runtime.Resources$ExInstWithCause.ex(Resources.java:506)
	at org.apache.calcite.sql.SqlUtil.newContextException(SqlUtil.java:917)
	at org.apache.calcite.sql.SqlUtil.newContextException(SqlUtil.java:902)
	at org.apache.calcite.sql.validate.SqlValidatorImpl.newValidationError(SqlValidatorImpl.java:5271)
	at org.apache.calcite.sql.validate.SqlValidatorImpl$DeriveTypeVisitor.visit(SqlValidatorImpl.java:6324)
	at org.apache.calcite.sql.validate.SqlValidatorImpl$DeriveTypeVisitor.visit(SqlValidatorImpl.java:6244)
	at org.apache.calcite.sql.SqlIdentifier.accept(SqlIdentifier.java:325)
	at org.apache.calcite.sql.validate.SqlValidatorImpl.deriveTypeImpl(SqlValidatorImpl.java:1867)
	at org.apache.calcite.sql.validate.SqlValidatorImpl.deriveType(SqlValidatorImpl.java:1852)
	at org.apache.calcite.sql.SqlOperator.constructArgTypeList(SqlOperator.java:678)
	at org.apache.calcite.sql.SqlFunction.deriveType(SqlFunction.java:249)
	at org.apache.calcite.sql.SqlFunction.deriveType(SqlFunction.java:231)
[...]

PR #13304, opened 2021-10-27
Support mysql regular function visit 

tuichenchuxin:
Refer to #12973 #13314

- support regular function visit
- modify test case


strongduanmu (changes requested):
@tuichenchuxin Some integration tests throw exceptions, please check if there are any problems.

tuichenchuxin:
> @tuichenchuxin Some integration tests throw exceptions, please check if there are any problems.

it's ok now, it failed before, because of the poor internet.

strongduanmu (approved):

PR #13187, opened 2021-10-21
support XA transaction for proxy #12771

roodkcab:
Fixes #12771 .

Changes proposed in this pull request:
add MySQL XA transaction for proxy

PR #13362, opened 2021-10-29
Fix #12771 to support XA statements in proxy

### LINKED ISSUE #12771: how to use xa transaction in shardingsphere-proxy
## Question

I tried to use xa transaction on shardingsphere-proxy but got something wrong.

the first problem is mysql does not support xa (start|begin) 'xxxxx' (join|resume), I made a little change here
image

then I set autocommit = 0 before use xa start 'xxxxx', but according to code here, it seems shardingsphere is trying to execute xa start on every schema using a broadcastdatabasebackendhandler 
image
which lead to 
image

the test sql is quite simple

```
set autocommit = 0;
xa start '1298390128390218903182903801923890218390';
xa end '1298390128390218903182903801923890218390';
xa prepare '1298390128390218903182903801923890218390';
xa commit '1298390128390218903182903801923890218390';
```

it seems I have to configure a xa datasource, but I can't find any document or examples for such configuration, please help, THX!

zhfeng:
* minor issues in XA statement parse

* add XATransactionBackendHandler

* add inXA in TransactionStatus

Fixes #12771

Replace #13187

### DIFF
--- a/shardingsphere-proxy/shardingsphere-proxy-backend/src/main/java/org/apache/shardingsphere/proxy/backend/communication/jdbc/transaction/TransactionStatus.java
+++ b/shardingsphere-proxy/shardingsphere-proxy-backend/src/main/java/org/apache/shardingsphere/proxy/backend/communication/jdbc/transaction/TransactionStatus.java
@@ -34,6 +34,9 @@ public final class TransactionStatus {
     
     @Setter
     private volatile boolean inTransaction;
+
+    @Setter
+    private volatile boolean inXA;
     
     private volatile TransactionType transactionType;
     
--- a/shardingsphere-proxy/shardingsphere-proxy-backend/src/main/java/org/apache/shardingsphere/proxy/backend/text/transaction/TransactionBackendHandler.java
+++ b/shardingsphere-proxy/shardingsphere-proxy-backend/src/main/java/org/apache/shardingsphere/proxy/backend/text/transaction/TransactionBackendHandler.java
@@ -39,17 +39,24 @@ public final class TransactionBackendHandler implements TextProtocolBackendHandl
     private final TCLStatement tclStatement;
     
     private final TransactionOperationType operationType;
-    
+
+    private final BackendConnection backendConnection;
+
     private final BackendTransactionManager backendTransactionManager;
-    
+
     public TransactionBackendHandler(final TCLStatement tclStatement, final TransactionOperationType operationType, final BackendConnection backendConnection) {
         this.tclStatement = tclStatement;
         this.operationType = operationType;
+        this.backendConnection = backendConnection;
         backendTransactionManager = new BackendTransactionManager(backendConnection);
     }
     
     @Override
     public ResponseHeader execute() throws SQLException {
+        if (backendConnection.getTransactionStatus().isInXA()) {
+            throw new SQLException("There is an active XA transaction");
+        }
+
         switch (operationType) {
             case BEGIN:
                 backendTransactionManager.begin();
--- a/shardingsphere-proxy/shardingsphere-proxy-backend/src/main/java/org/apache/shardingsphere/proxy/backend/text/transaction/TransactionBackendHandlerFactory.java
+++ b/shardingsphere-proxy/shardingsphere-proxy-backend/src/main/java/org/apache/shardingsphere/proxy/backend/text/transaction/TransactionBackendHandlerFactory.java
@@ -24,6 +24,7 @@ import org.apache.shardingsphere.proxy.backend.communication.jdbc.connection.Bac
 import org.apache.shardingsphere.proxy.backend.text.TextProtocolBackendHandler;
 import org.apache.shardingsphere.proxy.backend.text.data.impl.BroadcastDatabaseBackendHandler;
 import org.apache.shardingsphere.proxy.backend.text.skip.SkipBackendHandler;
+import org.apache.shardingsphere.proxy.backend.text.transaction.xa.XATransactionBackendHandler;
 import org.apache.shardingsphere.sql.parser.sql.common.statement.tcl.BeginTransactionStatement;
 import org.apache.shardingsphere.sql.parser.sql.common.statement.tcl.CommitStatement;
 import org.apache.shardingsphere.sql.parser.sql.common.statement.tcl.ReleaseSavepointStatement;
@@ -32,6 +33,7 @@ import org.apache.shardingsphere.sql.parser.sql.common.statement.tcl.RollbackToS
 import org.apache.shardingsphere.sql.parser.sql.common.statement.tcl.SavepointStatement;
 import org.apache.shardingsphere.sql.parser.sql.common.statement.tcl.SetAutoCommitStatement;
 import org.apache.shardingsphere.sql.parser.sql.common.statement.tcl.TCLStatement;
+import org.apache.shardingsphere.sql.parser.sql.common.statement.tcl.XAStatement;
 import org.apache.shardingsphere.transaction.core.TransactionOperationType;
 
 /**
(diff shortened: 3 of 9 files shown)


zhfeng:
@roodkcab Please can you check with this fix to see if it works in your case ? Thanks !

shardingsphere-proxy/shardingsphere-proxy-backend/src/main/java/org/apache/shardingsphere/proxy/backend/communication/jdbc/transaction/TransactionStatus.java
```diff
@@ -34,6 +34,9 @@
     
     @Setter
     private volatile boolean inTransaction;
+
+    @Setter
+    private volatile boolean inXA;
```
TeslaCN, line 39: Could we use the existing`transactionType` and `inTransaction`?

TeslaCN (changes requested):

shardingsphere-proxy/shardingsphere-proxy-backend/src/main/java/org/apache/shardingsphere/proxy/backend/communication/jdbc/transaction/TransactionStatus.java
zhfeng: No, I don't think so. They are two different transaction types and I think ```inXA``` is used in cases which handle the XA related statments to mark that the current transaction is in XA mode.

TeslaCN: If not `inTransaction`, connections will be returned to DataSource after each `CommandExecuteTask` finished. And which thread pool is selected to execute the task is related to `transactionType`(Refer to the class `OKProxyState`). I think this issue is not so simple that just passing the SQL to actual databases.

zhfeng: Well, can you point out the codes related ? I can take a close look.

TeslaCN: You may check the following classes
- `org.apache.shardingsphere.proxy.frontend.command.CommandExecutorTask`
- `org.apache.shardingsphere.proxy.frontend.state.impl.OKProxyState`

zhfeng: Thanks - I will take look.

zhfeng: So I think it should change the following
https://github.com/apache/shardingsphere/blob/8fc35c9afee87627dd43381c086dfe81df062b47/shardingsphere-proxy/shardingsphere-proxy-backend/src/main/java/org/apache/shardingsphere/proxy/backend/communication/jdbc/transaction/TransactionStatus.java#L56-L64

with ```return (InTransaction && TransactionType.BASE != transactionType) || inXA;```

So we can hold the connections during the XA transaction. Does it make sense ?

TeslaCN: I see your point. The `TransactionType.XA` and `inXa` seem confusing. I think the `inXa` is something like `TransactionType.MANUAL`.
From perspective of performance, broadcasting the SQL may cause cause high latency, especially if there is a large quantity of actual data nodes.

zhfeng: yeah, exactly - It should be user's resposiblity to control the transactions. And I can understand the performance issues,  if the bussiness logic has a strong data consistent, that must be a tradeofff.

So, can we introduce the ```TransactionType.MANUAL```? and if the user select this type, the proxy will broadcast all of the transaction related statements to the backend.

Does it make sense ?

zhfeng: By the way, can you show the transaction document on shardingsphere website ? I think we must have some clear ```NOTE``` to mention customers that the different behaviors of the transaction types.

tristaZero:
@jingshanglu Hey, please give this PR a look.

shardingsphere-proxy/shardingsphere-proxy-backend/src/main/java/org/apache/shardingsphere/proxy/backend/communication/jdbc/transaction/TransactionStatus.java
jingshanglu: > So I think it should change the following
> 
> https://github.com/apache/shardingsphere/blob/8fc35c9afee87627dd43381c086dfe81df062b47/shardingsphere-proxy/shardingsphere-proxy-backend/src/main/java/org/apache/shardingsphere/proxy/backend/communication/jdbc/transaction/TransactionStatus.java#L56-L64
> 
> with `return (InTransaction && TransactionType.BASE != transactionType) || inXA;`
> 
> So we can hold the connections during the XA transaction. Does it make sense ?

@zhfeng @TeslaCN Maybe we can use
```
return inTransaction && (TransactionType.XA == transactionType || .....);
```

zhfeng: Well,  I think ```TransactionType.XA``` here is different semantics with ```inXA``.

* ```TransactionType.XA``` is used by shardingsphere to manage the distributed transaction **INTERNAL**
* ```inXA``` means that users will be responsible to manage the transaction by themself, espcially using XA. So I'd like to change to use ```TransactionType.MANUAL```.

Also I'd like to understand the desgin strategy of transactions in shardingsphere in details. @terrymanu is there any document can be shared ?

jingshanglu: > Well, I think `TransactionType.XA` here is different semantics with ```inXA``.
> 
> * `TransactionType.XA` is used by shardingsphere to manage the distributed transaction **INTERNAL**
> * `inXA` means that users will be responsible to manage the transaction by themself, espcially using XA. So I'd like to change to use `TransactionType.MANUAL`.
> 
> Also I'd like to understand the desgin strategy of transactions in shardingsphere in details. @terrymanu is there any document can be shared ?
how about `TransactionType.XA` and `inTransaction` equals to `inXA`? @zhfeng

zhfeng: No, I'd like to say them are totaly different here.

zhfeng:
Sine we have not a decision to support XA, I withdraw this PR and just commit a fix for mysql xa parser. see #14075

PR #13380, opened 2021-11-01
Add primary data source name for show db discovery document

### LINKED ISSUE #13287: The issue of the dependency order of `feature-core` modules
Hi community,

Currently, `ShardingSphere-JDBC` and `ShardingSphere-Proxy` introduce their capabilities by depending on the core modules of each feature.
The core module of each feature will depend on other modules of feature, such as feature-api and feature-distsql. . .
The dependency tree is as follows:

```
shardingsphere-proxy-backend
│
├── shardingsphere-sharding-core
│   ├── shardingsphere-sharding-api
│   ├── shardingsphere-sharding-distsql-handler
│   ├── shardingsphere-sharding-distsql-parser
│   └── ...
├── shardingsphere-readwrite-splitting-core
│   ├── shardingsphere-readwrite-splitting-api
│   ├── shardingsphere-readwrite-splitting-distsql-handler
│   ├── shardingsphere-readwrite-splitting-distsql-parser
│   └── ...
├── shardingsphere-db-discovery-core
│   ├── shardingsphere-db-discovery-api
│   ├── shardingsphere-db-discovery-distsql-handler
│   ├── shardingsphere-db-discovery-distsql-parser
│   └── ...
└── ...
```

However, I found some problems:

1. `ShardingSphere-JDBC` cannot use `DistSQL`, but still introduces the distsql modules through the dependency on feature-core modules;
2. In `ShardingSphere-Proxy`, the work of the feature-core module does not actually need to use distsql. The current structure is only to indirectly introduce distsql dependency for `shardingsphere-proxy-backend` module.
3. Feature-distsql may need to use the objects in feature-core, but because of the dependency order, it cannot be imported conveniently.

Therefore, I suggest adjusting the dependency between feature-core and feature-distsql:

1. Let feature-core(s) no longer depend on feature-distsql(s);
2. Let feature-distsql(s) depend on feature-core(s);
3. `ShardingSphere-JDBC` still depends on feature-core(s), which can reduce the volume of application deployment package;
4. `ShardingSphere-Proxy` depends on feature-distsql(s), and indirectly depends on feature-core(s).

zjcnb:
Fixes #13287


zjcnb:
image
image

menghaoranss (approved):

PR #13463, opened 2021-11-04
Improve error message for SQLException.

### LINKED ISSUE #13462: Connection timeout exception cannot be displayed correctly in Local transaction.

ShardingSphere-Proxy, master branch

### Expected behavior
show accurate error message

### Actual behavior
Some users have reported that if the database connection times out after the `LOCAL` transaction has begun for a certain period of time, an exception will occur when the commit statement is executed. (It is normal for an exception to occur, but the error message is inaccurate)

The actual exception is wrapped in nextException: 
image

Wait for a period of time (greater than wait_timeout) after insert, and then execute commit.

image

RaigorJiang:
Fixes #13462

- If the next exception is not null and `getMessage()` returns blank String, get the detail message from next exception.

## Before:
```sql
mysql> INSERT INTO t_order values(1,1,'ok');
Query OK, 1 row affected (0.19 sec)
 
mysql> INSERT INTO t_order values(2,2,'disabled');
Query OK, 1 row affected (0.01 sec)
 
mysql> commit;
1815 - Internal error: 
```

## After:
```sql
mysql> INSERT INTO t_order values(1,1,'ok');
Query OK, 1 row affected (0.18 sec)
 
mysql> INSERT INTO t_order values(2,2,'disabled');
Query OK, 1 row affected (0.01 sec)
 
mysql> commit;
1815 - Internal error: Communications link failure during commit(). Transaction resolution unknown.
```

image

### DIFF
--- a/shardingsphere-proxy/shardingsphere-proxy-frontend/shardingsphere-proxy-frontend-mysql/src/main/java/org/apache/shardingsphere/proxy/frontend/mysql/err/MySQLErrPacketFactory.java
+++ b/shardingsphere-proxy/shardingsphere-proxy-frontend/shardingsphere-proxy-frontend-mysql/src/main/java/org/apache/shardingsphere/proxy/frontend/mysql/err/MySQLErrPacketFactory.java
@@ -19,6 +19,7 @@ package org.apache.shardingsphere.proxy.frontend.mysql.err;
 
 import lombok.AccessLevel;
 import lombok.NoArgsConstructor;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.shardingsphere.db.protocol.error.CommonErrorCode;
 import org.apache.shardingsphere.db.protocol.mysql.constant.MySQLServerErrorCode;
 import org.apache.shardingsphere.db.protocol.mysql.packet.generic.MySQLErrPacket;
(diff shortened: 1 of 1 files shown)


shardingsphere-proxy/shardingsphere-proxy-frontend/shardingsphere-proxy-frontend-mysql/src/main/java/org/apache/shardingsphere/proxy/frontend/mysql/err/MySQLErrPacketFactory.java
```diff
@@ -125,4 +126,11 @@ public static MySQLErrPacket newInstance(final Exception cause) {
         }
         return new MySQLErrPacket(1, CommonErrorCode.UNKNOWN_EXCEPTION, cause.getMessage());
     }
+    
+    private static String getErrorMessage(final SQLException cause) {
+        if (null != cause.getNextException() && StringUtils.isBlank(cause.getMessage())) {
+            return cause.getNextException().getMessage();
```
TeslaCN, line 132: What if the next is not the tail of the link?

TeslaCN: Could we consider aggregating all messages in the exception link?

RaigorJiang: This does not need to be the tail of the link, just get a meaningful exception for user. 
According to all the code of `setNextException`, it makes sense to start from the first next.

RaigorJiang: The number of connections used by the transaction is unknown, if this messages are aggregated, the result may be difficult to read.

TeslaCN (approved):

RaigorJiang:
@TeslaCN  Thank you!

Issue #13460, opened 2021-11-04
Tests in shardingsphere-proxy-backend may fail


TeslaCN:
### Which version of ShardingSphere did you use?
master - ad529a8ee7e06fc76225608888729a23ec4473f4

### Expected behavior
Build `shardingsphere-proxy-backend` successfully.

### Actual behavior
```
[INFO] Scanning for projects...
[INFO] ------------------------------------------------------------------------
[INFO] Detecting the operating system and CPU architecture
[INFO] ------------------------------------------------------------------------
[INFO] os.detected.name: linux
[INFO] os.detected.arch: x86_64
[INFO] os.detected.version: 5.11
[INFO] os.detected.version.major: 5
[INFO] os.detected.version.minor: 11
[INFO] os.detected.release: ubuntu
[INFO] os.detected.release.version: 20.04
[INFO] os.detected.release.like.ubuntu: true
[INFO] os.detected.release.like.debian: true
[INFO] os.detected.classifier: linux-x86_64
[INFO] 
[INFO] -------< org.apache.shardingsphere:shardingsphere-proxy-backend >-------
[INFO] Building shardingsphere-proxy-backend 5.0.0-RC1-SNAPSHOT
[INFO] --------------------------------[ jar ]---------------------------------
[INFO] 
[INFO] --- maven-enforcer-plugin:3.0.0:enforce (enforce-maven-version) @ shardingsphere-proxy-backend ---
[INFO] 
[INFO] --- maven-enforcer-plugin:3.0.0:enforce (enforce-banned-dependencies) @ shardingsphere-proxy-backend ---
[INFO] 
[INFO] --- maven-checkstyle-plugin:3.1.0:check (validate) @ shardingsphere-proxy-backend ---
[INFO] 
[INFO] --- jacoco-maven-plugin:0.8.7:prepare-agent (default) @ shardingsphere-proxy-backend ---
[INFO] argLine set to -javaagent:/home/wuweijie/.m2/repository/org/jacoco/org.jacoco.agent/0.8.7/org.jacoco.agent-0.8.7-runtime.jar=destfile=/home/wuweijie/projects/shardingsphere/shardingsphere-proxy/shardingsphere-proxy-backend/target/jacoco.exec
[INFO] 
[INFO] --- templating-maven-plugin:1.0.0:filter-sources (filtering-proxy-version) @ shardingsphere-proxy-backend ---
[INFO] Coping files with filtering to temporary directory.
[...]

PR #13472, opened 2021-11-05
[DistSQL]Remove unnecessary assertions.

lanchengx:
Fixes https://github.com/apache/shardingsphere/issues/13460.

- After modifying the `ConfigurationKey`, the data will be updated to `zk`, and the local data will also be refreshed according to the data of `zk`, so the verification data here is the data before updating `zk`, not the final result


TeslaCN (commented):
I prefer using `@Ignore` and add TODO instead of removing assertions.

TeslaCN (changes requested):
I have some suggestions in the issue.

lanchengx:
I will close this pr and resubmit one

Issue #13488, opened 2021-11-06
Unify `@throws SQLException` comments.


yx9o:
Hi, community, I found that there are several styles of `@throws SQLException` comments in the current master.
for example:

![image](https://user-images.githubusercontent.com/12792261/140607379-2121dff4-734b-4b82-baa8-1b41d7d36ea7.png)
![image](https://user-images.githubusercontent.com/12792261/140607182-be07800e-caa3-47ee-89d2-400d420c6352.png)

I want to unify them into: `SQL exception`.

yx9o:
Fixed in #13500 .

PR #13500, opened 2021-11-08
For #13488, Unify @throws SQLException comments.

yx9o:
For #13488, Unify @throws SQLException comments.


tristaZero (approved):
@yx9o Nice.

Issue #13494, opened 2021-11-08
DistSQL error message is wrong.


yx9o:
### Which version of ShardingSphere did you use?

master

### Which project did you use? ShardingSphere-JDBC or ShardingSphere-Proxy?

ShardingSphere-Proxy

### Actual behavior

![image](https://user-images.githubusercontent.com/12792261/140701509-8e146875-10c3-40b1-ad0a-a21a8160ac44.png)

### Reason analyze (If you can)

![image](https://user-images.githubusercontent.com/12792261/140701046-06289d38-72af-480c-97e7-dc56e08b462f.png)

I want to fix it.

RaigorJiang:
@yx9o 
Thanks a lot.

yx9o:
Fixed in #13507 .

PR #13507, opened 2021-11-08
For #13494, Correct the DistSQL error message.

yx9o:
For #13494, Correct the DistSQL error message.


tristaZero (approved):
Hey, @yx9o thanks for your another PR.

Issue #13634, opened 2021-11-16
[Disscussion] Why we need an optional parse skiped phase in encrypt scenarioes.


cheese8:
## The purpose:

1、Enhance users' confidence in using SS.
2、Reduce test coverage and workload in using SS.

## Scenarios:
Encrypt only.

Why an optional parse skipped parse could enhance user confidence?

1、Encrypt and decrypt are generally compliance requirements, and when your system needs compliance, which means it is already large and the situation is complicated, for example, no unit test or unit test is not enough etc. Do you agree with this?

2、Then are there a lot of SQLs, and there are some SQLs containing encrypt tables/fields, others not at all.  It is the reality.

3、The first solution is developers manually modify the SQLs  which containing encrypt tables/fields one by one, it is the most simple, direct and easiest to accept solution, as the scope that needs to be developed and tested is fixed. Do you agree with this?

4、The SS's encrypt solution is almost perfect, but:
> all SQLs should be parsed and some SQLs are not parsed well.
> all SQLs need to be parsed, which cause the unacceptable by test.

If SS says that only the SQLs which containing encrypt tables/fields is parsed and rewritten, the other SQLs parse phase can be skipped. In this way, it can both improve the overall performance and be more acceptable to users.

Do you agree with this? Look forward to your opinions @strongduanmu @tristaZero @terrymanu

strongduanmu:
@cheese8 Parsing is necessary for now. Only after parsing, we can extract the table segment in the statement and then determine whether we need to rewrite it.

cheese8:
@strongduanmu ye, i know this by reading the source code, and I think the most important thing is the risk problems that all SQL will be parsed which the test can not cover it completely.

terrymanu:
Just as @strongduanmu said, SQL parser is the necessary step.
We can enhance the SQL parser if there are some bugs, but cannot accept to skip SQL parser which will break the basic concept of ShardingSphere.

PR #13672, opened 2021-11-17
skip parse phase

cheese8:
- add ParseSkippedStatement to skip parse phase
- add testcase

@strongduanmu


terrymanu (changes requested):
What is the proposal of this issue?

cheese8:
ref to: #13634 @terrymanu , and this is a PoC to test the feasibility.

strongduanmu:
@cheese8 Considering that parser is the basis of the entire project, skip parsing seems to be just a special solution, so we cannot accept this PR.

PR #13684, opened 2021-11-18
support query which table has column contains keyword like 'desc' in …

### LINKED ISSUE #13509: Can MySQL keyword query be supported？
this is a SQL，

> SELECT 'name', 'desc' FROM t1   (' is`)

it be rewirte to

> SELECT 'name' AS name, 'desc' AS desc FROM t1    (' is`)

but this SQL is error, because 'name', 'desc' is MySQL keword，
how to solve this error？

tuichenchuxin:
…encrypt.

fixes #13509.

- support query which table has column contains keyword like 'desc'
- modify test case.


strongduanmu:
@tuichenchuxin Can you take a look for ci error?

strongduanmu (approved):

PR #13628, opened 2021-11-15
Add SQLServer CREATE TABLE AS SELECT statement

ThanoshanMV:
For #6478

Hi @jingshanglu, I've added SQLServer [CREATE TABLE AS SELECT](https://docs.microsoft.com/en-us/sql/t-sql/statements/create-table-as-select-azure-sql-data-warehouse?view=aps-pdw-2016-au7#syntax) statement. Please check it. After successfully completing this PR, I'll add this statement into SQLServer `EXPLAIN` statement's `SQL_statement` (#13546).

Changes proposed in this pull request:
- Added  `CREATE TABLE AS SELECT` statement.
- Added corresponding test cases.
- In the  `CREATE TABLE AS SELECT` statement, seems like `OPTION <query_hint>` is not correct as it can't parse [`OPTION ( HASH JOIN )` ](https://docs.microsoft.com/en-us/sql/t-sql/statements/create-table-as-select-azure-sql-data-warehouse?view=aps-pdw-2016-au7#examples-for-query-hints). I've found the option clause for query hint [here](https://docs.microsoft.com/en-us/sql/t-sql/queries/option-clause-transact-sql?view=aps-pdw-2016-au7#syntax-for--and-) and utilized it.

PR #13709, opened 2021-11-19
Add SQLServer CREATE REMOTE TABLE AS SELECT statement

ThanoshanMV:
For #6478

Hi @jingshanglu, @strongduanmu, @tuichenchuxin  I've added  [CREATE REMOTE TABLE AS SELECT](https://docs.microsoft.com/en-us/sql/t-sql/statements/create-remote-table-as-select-parallel-data-warehouse?view=aps-pdw-2016-au7) statement. Please check it.

- Added  `CREATE REMOTE TABLE AS SELECT` statement.
- Added corresponding test case.

### DIFF
--- a/shardingsphere-sql-parser/shardingsphere-sql-parser-dialect/shardingsphere-sql-parser-sqlserver/src/main/antlr4/imports/sqlserver/DDLStatement.g4
+++ b/shardingsphere-sql-parser/shardingsphere-sql-parser-dialect/shardingsphere-sql-parser-sqlserver/src/main/antlr4/imports/sqlserver/DDLStatement.g4
@@ -20,7 +20,7 @@ grammar DDLStatement;
 import Symbol, Keyword, SQLServerKeyword, Literals, BaseRule, DMLStatement, DCLStatement;
 
 createTable
-    : CREATE TABLE tableName fileTableClause createDefinitionClause
+    : CREATE createTableSpecification TABLE tableName fileTableClause createDefinitionClause
     ;
 
 createIndex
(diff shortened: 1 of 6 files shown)


tuichenchuxin (approved):

strongduanmu (approved):

ThanoshanMV:
Hi @strongduanmu, @tuichenchuxin. I added SQLServer `CREATE REMOTE TABLE AS SELECT` and `CREATE TABLE AS SELECT` (#13628) with the existing `CREATE TABLE` statement since the `EXPLAIN` statement's `SQL_statement` can have both of them.

I defined [EXPLAIN](https://docs.microsoft.com/en-us/sql/t-sql/queries/explain-transact-sql?view=azure-sqldw-latest#syntax) statement's `SQL_statement` as:
```
explainableStatement
    : select | insert | update | delete
    ;
```

To include the `CREATE REMOTE TABLE AS SELECT` and `CREATE TABLE AS SELECT` in `EXPLAIN` statement's `SQL_statement`, can we have the definition as below:
```
explainableStatement
    : select | insert | update | delete | createTable
    ;
```
But it would parse standard `CREATE TABLE` statements as well. Does this approach acceptable or do you have any other suggestions? Please let me know.

tuichenchuxin:
@ThanoshanMV I didn't got your point. What will happen if you add `crateTable` in explainableStatement.
If there is a standard `create table`, it will not be parsed by `explain` grammar.

ThanoshanMV:
> What will happen if you add `crateTable` in explainableStatement.

@tuichenchuxin If we add `createTable` in `explainableStatement`, then standard `CREATE TABLE` statements other than `CREATE REMOTE TABLE AS SELECT` and `CREATE TABLE AS SELECT` would be parsed by `explain` grammar. For example, `EXPLAIN CREATE TABLE t_log(id int PRIMARY KEY, status varchar(10))` would be parsed but according to the documentation it should not be parsed.

tuichenchuxin:
@ThanoshanMV Oh, I got your point. I think maybe we can split `create table` and `create table as select` .

ThanoshanMV:
Yes @tuichenchuxin, it would be fine.

Issue #13456, opened 2021-11-04
Support more query statements to convert between SQLStatement and SqlNode


strongduanmu:
This issue is a subtask of #13091.

At present, we have carried out a parameterized unit test for SQLStatement convert in the `shardingsphere-optimize-test` module, but there are still many scenarios that do not support. We need to support more query statements and support more database dialects.

Considering the importance of this task, I would like to invite @LeeGuoPing to participate in this work.

In order to find out the query statements that are not currently supported, we can delete the `SUPPORTED_SQL_CASE_IDS` condition, let all scenarios execute the test, and then be able to find our target. Due to the relatively large workload, I suggest starting with MySQL support first.

LeeGuoPing:
please assign it to me

strongduanmu:
We can refer this convert logic——https://github.com/ApsaraDB/galaxysql/blob/main/polardbx-optimizer/src/main/java/com/alibaba/polardbx/optimizer/parse/visitor/FastSqlToCalciteNodeVisitor.java.

strongduanmu:
Ref #19580.

kanha-gupta:
@strongduanmu @LeeGuoPing 

Hey ! I would like to work on this. While adding support for new Sql queries, are there any Documents or Code available to refer to ? to understand the required logic & implementation purposes.
I am currently checking up previous PRs & Would start adding up more ASAP.

strongduanmu:
Hi @kanha-gupta, `SQLNodeConverterEngine` can be used as an entry point to understand this issue. At present, the federated query engine relies on SQLNodeConverterEngine to realize the conversion between ShardingSphere SQLStatement and Calcite SqlNode, so as to take advantage of some capabilities of Calcite.

kanha-gupta:
@strongduanmu Thanks ! So to add support for more SQL queries, is Antlr parsed Generated code is used ?

strongduanmu:
> 

Yes, the responsibility of SQLNodeConverterEngine is to convert the SQLStatement parsed and visited by ANTLR into Calcite SqlNode.

kanha-gupta:
Okay I'll look into it.
Thanks alot for guidance :)

strongduanmu:
This task is so large that no one can actually complete it, so I will close this issue and replace it with more smaller issues.

PR #13763, opened 2021-11-24
support distinct aggregation and Optimized converterContext use threadLocal holder

LeeGuoPing:
Ref #13456.

-
-
-

### DIFF
--- a/shardingsphere-infra/shardingsphere-infra-optimize/src/main/java/org/apache/shardingsphere/infra/optimize/converter/segment/expression/impl/FunctionConverter.java
+++ b/shardingsphere-infra/shardingsphere-infra-optimize/src/main/java/org/apache/shardingsphere/infra/optimize/converter/segment/expression/impl/FunctionConverter.java
@@ -36,7 +36,7 @@ import java.util.stream.Collectors;
 /**
  * Function converter.
  */
-public class FunctionConverter implements SQLSegmentConverter<FunctionSegment, SqlBasicCall> {
+public final class FunctionConverter implements SQLSegmentConverter<FunctionSegment, SqlBasicCall> {
     
     @Override
     public Optional<SqlBasicCall> convertToSQLNode(final FunctionSegment segment) {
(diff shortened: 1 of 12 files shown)


shardingsphere-infra/shardingsphere-infra-optimize/src/main/java/org/apache/shardingsphere/infra/optimize/converter/segment/expression/impl/FunctionConverter.java
```diff
@@ -0,0 +1,73 @@
...
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+/**
+ * Function converter.
+ */
+public class FunctionConverter implements SQLSegmentConverter<FunctionSegment, SqlBasicCall> {
```
strongduanmu, line 39: @LeeGuoPing Please add final for FunctionConverter.

shardingsphere-infra/shardingsphere-infra-optimize/src/main/java/org/apache/shardingsphere/infra/optimize/converter/segment/projection/impl/AggregationProjectionConverter.java
```diff
@@ -86,4 +137,15 @@ private SqlAggFunction convertOperator(final String operator) {
         Preconditions.checkState(REGISTRY.containsKey(operator), "Unsupported SQL operator: `%s`", operator);
         return REGISTRY.get(operator);
     }
+    
+    /**
+     * Is sqlBasicCall as operator aggregation type.
```
strongduanmu, line 142: @LeeGuoPing `Judge whether sqlBasicCall  is as operator aggregation type or not.` may be better.

strongduanmu (changes requested):

strongduanmu (approved):

PR #13780, opened 2021-11-24
Update _index.en.md

Zhoutzzz:
About shadow 'HINT' description

Fixes #ISSUSE_ID.

-
-
-

### DIFF
--- a/docs/document/content/reference/shadow/_index.en.md
+++ b/docs/document/content/reference/shadow/_index.en.md
@@ -254,23 +254,29 @@ When using HINT, ensure that the configuration items of `props` in the configura
 
 ```yaml
 simple-note-algorithm:
-      type: SIMPLE_HINT
-      props:
-        shadow: true
-        user_id: 2
+  type: SIMPLE_HINT
+  props:
+    shadow: true
+    user_id: 2
 ```
 For example, the 'props' item have `2` configure, the following syntax can be used in SQL:
 ```sql
-/*shadow:true,user_id:2*/、/*shadow:true,user_id:2,foo:bar,.....*/
+SELECT * FROM t_xxx_2 WHERE order_id = xxx /*shadow:true,user_id:2*/、
+```
+```sql
+SELECT * FROM t_xxx_2 WHERE order_id = xxx /*shadow:true,user_id:2,foo:bar,.....*/
 ```
 
 ```yaml
-simple-note-algorithm:
-      type: SIMPLE_HINT
-      props:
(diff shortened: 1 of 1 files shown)


docs/document/content/reference/shadow/_index.en.md
```diff
@@ -248,4 +248,29 @@ props:
...
+simple-note-algorithm:
+      type: SIMPLE_HINT
+      props:
+        shadow: true
+        user_id: 2
+```
+For example, the 'props' item have `2` configure, the following syntax can be used in SQL:
+```sql
+/*shadow:true,user_id:2*/、/*shadow:true,user_id:2,foo:bar,.....*/
+```
```
soulasuna, line 265: Preferably a SQL statement that completes the structure.

docs/document/content/reference/shadow/_index.en.md
```diff
@@ -248,4 +248,29 @@ props:
...
+```yaml
+simple-note-algorithm:
+      type: SIMPLE_HINT
+      props:
+        shadow: false
+```
+For example, the 'props' item have `1` configure, the following syntax can be used in SQL:
+```sql
+/*shadow:false*/、/*shadow:false,user_id:2,foo:bar,.....*/
+```
```
soulasuna, line 276: Preferably a SQL statement that completes the structure.

soulasuna (commented):
Thank you very much for your suggestions.
We need to describe the way under in-depth discussion.
Looking forward to your reply.

docs/document/content/reference/shadow/_index.en.md
```diff
@@ -248,4 +248,29 @@ props:
   sql-comment-parse-enabled: true
 ```
 
-**Note**: The default shadow algorithm only supports note shadow algorithm.
+**Note**:
+The default shadow algorithm only supports note shadow algorithm.
+When using HINT, ensure that the configuration items of `props` in the configuration file are less than or equal to those in the SQL comment, and that the configuration items in the configuration file are the same as those in the SQL comment. The fewer the configuration items in the configuration file, the looser the matching conditions are
```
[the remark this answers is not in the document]
Zhoutzzz: I agree with what you said, but the purpose of adding this document is to make others better understand the matching mode of rules, because the configuration of this part is less described in the current document. At the same time, I think my change does not induce others to use it, but to tell others how rules are matched

soulasuna (approved):

strongduanmu (approved):

PR #13823, opened 2021-11-26
Maven flatten plugins for examples

### LINKED ISSUE #13512: Introduce Maven Flatten Plugin for project examples
There are too many modules in examples and we have to update example versions in every example `pom.xml` during project release, which is tedious to maintain.

Using `maven-flatten-plugin` in examples and replace version with variable in `pom.xml`.

Swastyy:
Fixes #13512 

Some points:
- The variable I used for the version of the parent directory that is `/examples` is revision as the revised version of the project.
- In the sub directories of `examples` Should I remove the plugin management lines? @TeslaCN 
- Let me know if there is anything that I am missing or doing incorrect.


Swastyy:
@TeslaCN Is the pom files inside the sub directories of examples (meaning the third level of hierarchy) need to have changed to the version variable or they will remain as it is?

examples/pom.xml
```diff
@@ -22,7 +22,7 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>org.apache.shardingsphere.example</groupId>
     <artifactId>shardingsphere-example</artifactId>
-    <version>5.0.1-SNAPSHOT</version>
+    <version>${revision}</version>
```
TeslaCN, line 25: Hi @Swastyy 
Sorry for the delay. You may need to define a value for `revision` in properties.

examples/shardingsphere-proxy-example/pom.xml
```diff
@@ -33,5 +33,38 @@
         <module>shardingsphere-proxy-boot-mybatis-example</module>
         <module>shardingsphere-proxy-hint-example</module>
     </modules>
+
+    <build>
+		<pluginManagement>
```
TeslaCN, line 38: Please fix the indent here and other.

examples/pom.xml
Swastyy: Yea, I have added that 👍

examples/shardingsphere-proxy-example/pom.xml
Swastyy: In ide it looks correctly indented, I am not sure why it looks like here this way, I will correct the individual files indentation if it still happens.

Swastyy: @TeslaCN The CI test are failing with error as they can't find the version of parent although I have added it correctly, and there are no dependencies in those pom files, I cannot understand what is missing, please help.

examples/pom.xml
```diff
@@ -22,7 +22,7 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>org.apache.shardingsphere.example</groupId>
     <artifactId>shardingsphere-example</artifactId>
-    <version>5.0.1-SNAPSHOT</version>
+    <version>${revision}${sha1}${changelist}</version>
```
TeslaCN, line 25:
```suggestion
    <version>${revision}</version>
```

examples/pom.xml
```diff
@@ -35,6 +35,9 @@
     </modules>
     
     <properties>
+        <revision>5.0.1</revision>
+        <changelist>-SNAPSHOT</changelist>
+        <sha1/>
```
TeslaCN, line 40:
```suggestion
        <revision>5.0.1-SNAPSHOT</revision>
```

TeslaCN (changes requested):
You may refer to this PR https://github.com/apache/shardingsphere-elasticjob/pull/1672

TeslaCN:
@Swastyy Could you figure out why apache/shardingsphere-elasticjob#1672 worked?

examples/shardingsphere-proxy-example/pom.xml
TeslaCN: Your indent are `tab`, replace them with `space`.

Swastyy:
> @TeslaCN Is the pom files inside the sub directories of examples (meaning the third level of hierarchy) need to have changed to the version variable or they will remain as it is?

Hey @TeslaCN I could see this as the problem, we are having so many levels and upto which one the parent version should be changed to variable, now in https://github.com/apache/shardingsphere-elasticjob/pull/1672 there is only two levels so it was handled easily but for more than 3 levels should I change every pom in the depth to revision?

TeslaCN:
> Hey @TeslaCN I could see this as the problem, we are having so many levels and upto which one the parent version should be changed to variable, now in [apache/shardingsphere-elasticjob#1672](https://github.com/apache/shardingsphere-elasticjob/pull/1672) there is only two levels so it was handled easily but for more than 3 levels should I change every pom in the depth to revision?

All versions of examples' pom should be changed to placeholder. What we expect is changing one place can change all examples' version.

Swastyy:
> > Hey @TeslaCN I could see this as the problem, we are having so many levels and upto which one the parent version should be changed to variable, now in [apache/shardingsphere-elasticjob#1672](https://github.com/apache/shardingsphere-elasticjob/pull/1672) there is only two levels so it was handled easily but for more than 3 levels should I change every pom in the depth to revision?
> 
> All versions of examples' pom should be changed to placeholder. What we expect is changing one place can change all examples' version.

I have completed it. Please review it once.

examples/example-core/pom.xml
```diff
@@ -36,4 +36,37 @@
         <module>example-spring-mybatis</module>
         <module>example-spring-jpa</module>
     </modules>
+
+    <build>
+        <pluginManagement>
+		    <plugins>
```
TeslaCN, line 42: The indents here and below mixed `tab` and `space` and they are still incorrect. How about just removing the `lifecycle-mapping`?

Swastyy: Yea, there is no specific use of it(a comment also mentions), I'll remove the `lifecycleMapping`. Also @TeslaCN there are too many changed files in this PR can I do it another PR?

TeslaCN: I prefer doing it in this PR. The examples updated in https://github.com/apache/shardingsphere/pull/13931.
You may need to update your branch.

Swastyy:
Hi @TeslaCN I have corrected the indents and updated the poms after #13931

TeslaCN (approved):

TeslaCN:
@Swastyy Thank you!

PR #13845, opened 2021-11-29
Replace java.util.stream with for each in high frequency codes

TeslaCN:
Related to #13848


TeslaCN:
Green for decrease and red for increase.

image

strongduanmu (approved):

PR #13904, opened 2021-12-02
Fix: Duplicate data appears when querying information_schemata

lanchengx:


lanchengx:
@RaigorJiang There is also a pr that needs to be merged

RaigorJiang (approved):

PR #13967, opened 2021-12-07
Increase judgment for debug log

totalo:


menghaoranss:
@totalo There are conflicting files.

totalo:
> @totalo There are conflicting files.

fixed! thx~

shardingsphere-agent/shardingsphere-agent-core/src/main/java/org/apache/shardingsphere/agent/core/bytebuddy/listener/LoggingListener.java
```diff
@@ -35,7 +35,9 @@ public void onDiscovery(final String typeName, final ClassLoader classLoader, fi
     
     @Override
     public void onTransformation(final TypeDescription typeDescription, final ClassLoader classLoader, final JavaModule module, final boolean loaded, final DynamicType dynamicType) {
-        log.debug("On transformation class {}", typeDescription.getTypeName());
+        if (log.isDebugEnabled()) {
+            log.debug("On transformation class {}", typeDescription.getTypeName());
```
terrymanu, line 39: Is it necessary? the symbol `{}` has already avoid assembly the strings.

terrymanu (changes requested):

shardingsphere-agent/shardingsphere-agent-core/src/main/java/org/apache/shardingsphere/agent/core/bytebuddy/listener/LoggingListener.java
totalo: I think that if its log level is higher than debug, it is unnecessary for the parameter in the debug method to be a method call.

Issue #13875, opened 2021-12-01
Refactoring: scaling as part of Apache ShardingSphere 3 layers architecture but not an individual tool


sandynz:
Hi community,

Scaling is an individual tool before, we've integrated scaling with proxy and removed scaling bootstrap and distribution in #13276, but we didn't refactor scaling.

We are planning to make scaling as part of Apache ShardingSphere [3 layers architecture](https://shardingsphere.apache.org/document/current/en/concepts/pluggable/#implementation ).

After several times of discussion and review by @tristaZero and @terrymanu , we've got the final design. In general, it includes following parts:
- Modules design
- Configuration design

Package name and class name refactoring won't be included here.

We'll refactor it in several stages.

### Origianl related modules
shardingsphere
- shardingsphere-scaling (include all scaling features)

### Current related modules
shardingsphere
- shardingsphere-kernel (original module)
	- shardingsphere-schedule (new module)
		- shardingsphere-schedule-core : scheduling core, facade of ElasticJob
		- shardingsphere-migration-common : fundamental functions for migration scenarios: scaling, data encryption
		- shardingsphere-migration-distsql : common DistSQL for migration scenarios
		- shardingsphere-migration-sql-core : dialect SPI interfaces
		- shardingsphere-migration-sql-dialect : migration dialect implementation
			- shardingsphere-migration-mysql
			- shardingsphere-migration-postgresql
			- shardingsphere-migration-opengauss
		- shardingsphere-cdc : change data capture
			- shardingsphere-cdc-core : data producer and data channel, `Dumper` and `Channel` in scaling
			- shardingsphere-cdc-dialect
				- shardingsphere-cdc-mysql
				- shardingsphere-cdc-postgresql
				- shardingsphere-cdc-opengauss
- shardingsphere-features (original module)
	- shardingsphere-encrypt (original module)
		- shardingsphere-encrypt-api : encrypt rule configuration upgrade
		- shardingsphere-encrypt-core : add encryption job here
	- shardingsphere-sharding (original module)
		- shardingsphere-sharding-api : sharding rule configuration upgrade
		- shardingsphere-sharding-core : move scaling job here, no shard could be considered as one shard rule
- shardingsphere-scaling (original module, disappeared after refactoring)

In summary, `shardingsphere-scaling` will be extracted into:
[...]

PR #14009, opened 2021-12-09
Ignore case about ModeScheduleContext.

zjcnb:
For #13875


sandynz (approved):
Currently, SPI loader ignore type case.

Issue #13725, opened 2021-11-21
Consider about add parser rule in kernel layer


totalo:
refer #13722.

In the process of reading the code, I found that some parsing rule configurations are still hard-coded in the code, and there are enough parsing-related configurations so that we can extract them and assemble them into a parse rule.
such as `sql-comment-parse-enabled`, sqlStatementCache option and parse tree cache option, and more about parse.

after extract:
```
- !SQL_PARSER
  sqlCommentParseEnabled: true
  sqlCacheInitialCapacity: 2000
  ......
```
todo list:

- [x] Support the configuration of rules
- [x] DistSQL RDL
- [x] DistSQL RQL
- [ ] add docs

Do you think this is a good idea? please give me some advices.

tuichenchuxin:
+1

totalo:
好的，我会推进这项工作

tuichenchuxin:
@totalo Hi, I want to know how can I use sqlCommentParseEnabled in sharding-jdbc for springboot. :)

totalo:
> @totalo Hi, I want to know how can I use sqlCommentParseEnabled in sharding-jdbc for springboot. :)

hi @tuichenchuxin. Sorry, it may not be supported yet. I am considering how to add it. I will submit a pr support tonight.

tuichenchuxin:
@totalo :) Not in a hurry

PR #14018, opened 2021-12-09
Add spring support for sql parse rule

totalo:
For #13725.

- Add spring support for sql parsing
- add ut


terrymanu:
It seems the configuration of spring is different with other rules, we need to keep consist with all rules.

totalo:
> It seems the configuration of spring is different with other rules, we need to keep consist with all rules.

Yes, I will work to unify them.

terrymanu (changes requested):
Please add spring namespace test case

terrymanu (approved):

PR #14031, opened 2021-12-10
improve load single table performance

### LINKED ISSUE #10938: about init  metadata of tables
shardingsphere jdbc 5
springboot2.3

I have a project it uses only one logic table，but there are 10000 other tables  in the same db for other projects
When my project starting，shardingsphere jdbc will init all tables' metadata,then the process seem like "sleep"

Can I exclude other tables in my sharding project? 
Thanks

tuichenchuxin:
Fixes #10938.

- improve load single table performance


tuichenchuxin:
I load about 14000 singletables.
From 140s To 17s.

shardingsphere-kernel/shardingsphere-single-table/shardingsphere-single-table-core/src/main/java/org/apache/shardingsphere/singletable/rule/SingleTableRule.java
```diff
@@ -56,11 +57,18 @@
...
+    private final Map<String, Collection<DataNode>> dataNodeMap;
+    
+    private final Map<String, String> tableNameMap;
+    
     public SingleTableRule(final SingleTableRuleConfiguration config, final DatabaseType databaseType, 
                            final Map<String, DataSource> dataSourceMap, final Collection<ShardingSphereRule> builtRules, final ConfigurationProperties props) {
         Map<String, DataSource> aggregateDataSourceMap = getAggregateDataSourceMap(dataSourceMap, builtRules);
         dataSourceNames = aggregateDataSourceMap.keySet();
         singleTableDataNodes = SingleTableDataNodeLoader.load(databaseType, aggregateDataSourceMap, getExcludedTables(builtRules), props);
+        dataNodeMap = singleTableDataNodes.values().stream().map(each -> new DataNode(each.getDataSourceName(), each.getTableName()))
```
strongduanmu, line 69: @tuichenchuxin Can we remove singleTableDataNodes and replace with dataNodeMap？

shardingsphere-kernel/shardingsphere-single-table/shardingsphere-single-table-core/src/main/java/org/apache/shardingsphere/singletable/rule/SingleTableRule.java
```diff
@@ -56,11 +57,18 @@
     
     private final Map<String, SingleTableDataNode> singleTableDataNodes;
     
+    private final Map<String, Collection<DataNode>> dataNodeMap;
+    
+    private final Map<String, String> tableNameMap;
```
strongduanmu, line 62: @tuichenchuxin Maybe it is better to define it as a collection.

strongduanmu (changes requested):

shardingsphere-kernel/shardingsphere-single-table/shardingsphere-single-table-core/src/main/java/org/apache/shardingsphere/singletable/rule/SingleTableRule.java
tuichenchuxin: I need to use concurrent map, And because when we drop table,may be the name is not the same case as the actual table, So I use the low case as the key.

strongduanmu (approved):

PR #14068, opened 2021-12-13
try to fix the error of 'MySQL-Proxy with Agent Tracing OpenTelemetry'

wsm12138:
Fixes #ISSUSE_ID.

-
-
-


totalo:
hi @wsm12138 ,I think the core problem is that the response is empty, not the test code. You can try to find out why the http response is empty.

PR #14075, opened 2021-12-13
Fix #14074 Mysql parser with XA statements

zhfeng:
Fixes #14074 

Changes proposed in this pull request:
- Fix the grammer
- Add  ```op``` and ```xid``` attribute in ```XAStatement```
- Add a test

PR #14130, opened 2021-12-17
add xa handler

### LINKED ISSUE #12771: how to use xa transaction in shardingsphere-proxy

roodkcab:
Fixes #12771.

- add xa handler to handle `xa` related statements.  
- based on [14075](https://github.com/apache/shardingsphere/pull/14075) and [13362](https://github.com/apache/shardingsphere/pull/13362)

### DIFF
--- a/docs/document/content/features/transaction/use-norms/xa.en.md
+++ b/docs/document/content/features/transaction/use-norms/xa.en.md
@@ -13,3 +13,101 @@ weight = 2
 ## Unsupported
 
 * Recover committing and rolling back in other machines after the service is down.
+
+## XA Transaction managed by XA Statement
+
+* When using XA START to open a XA Transaction, ShardingSphere will pass it to backend database directly, you have to manage this transaction by yourself;
+* When recover from a crush, you have to call XA RECOVER to check unfinished transaction, and choose to commit or rollback using xid. Or you can use ONE PHASE commit without PREPARE.
+
+MySQL [(none)]> use test1                                                                               │MySQL [(none)]> use test2
+Reading table information for completion of table and column names                                      │Reading table information for completion of table and column names
+You can turn off this feature to get a quicker startup with -A                                          │You can turn off this feature to get a quicker startup with -A
+                                                                                                        │
+Database changed                                                                                        │Database changed
+MySQL [test1]> XA START '61c052438d3eb';                                                                │MySQL [test2]> XA START '61c0524390927';
+Query OK, 0 rows affected (0.030 sec)                                                                   │Query OK, 0 rows affected (0.009 sec)
+                                                                                                        │
+MySQL [test1]> update test set val = 'xatest1' where id = 1;                                            │MySQL [test2]> update test set val = 'xatest2' where id = 1;
+Query OK, 1 row affected (0.077 sec)                                                                    │Query OK, 1 row affected (0.010 sec)
+                                                                                                        │
+MySQL [test1]> XA END '61c052438d3eb';                                                                  │MySQL [test2]> XA END '61c0524390927';
+Query OK, 0 rows affected (0.006 sec)                                                                   │Query OK, 0 rows affected (0.008 sec)
+                                                                                                        │
+MySQL [test1]> XA PREPARE '61c052438d3eb';                                                              │MySQL [test2]> XA PREPARE '61c0524390927';
+Query OK, 0 rows affected (0.018 sec)                                                                   │Query OK, 0 rows affected (0.011 sec)
+                                                                                                        │
+MySQL [test1]> XA COMMIT '61c052438d3eb';                                                               │MySQL [test2]> XA COMMIT '61c0524390927';
+Query OK, 0 rows affected (0.011 sec)                                                                   │Query OK, 0 rows affected (0.018 sec)
+                                                                                                        │
+MySQL [test1]> select * from test where id = 1;                                                         │MySQL [test2]> select * from test where id = 1;
++----+---------+                                                                                        │+----+---------+
+| id | val     |                                                                                        │| id | val     |
++----+---------+                                                                                        │+----+---------+
+|  1 | xatest1 |                                                                                        │|  1 | xatest2 |
++----+---------+                                                                                        │+----+---------+
+1 row in set (0.016 sec)                                                                                │1 row in set (0.129 sec)
+
+MySQL [test1]> XA START '61c05243994c3';                                                                │MySQL [test2]> XA START '61c052439bd7b';
+Query OK, 0 rows affected (0.047 sec)                                                                   │Query OK, 0 rows affected (0.006 sec)
+                                                                                                        │
+MySQL [test1]> update test set val = 'xarollback' where id = 1;                                         │MySQL [test2]> update test set val = 'xarollback' where id = 1;
+Query OK, 1 row affected (0.175 sec)                                                                    │Query OK, 1 row affected (0.008 sec)
+                                                                                                        │
+MySQL [test1]> XA END '61c05243994c3';                                                                  │MySQL [test2]> XA END '61c052439bd7b';
+Query OK, 0 rows affected (0.007 sec)                                                                   │Query OK, 0 rows affected (0.014 sec)
+                                                                                                        │
+MySQL [test1]> XA PREPARE '61c05243994c3';                                                              │MySQL [test2]> XA PREPARE '61c052439bd7b';
+Query OK, 0 rows affected (0.013 sec)                                                                   │Query OK, 0 rows affected (0.019 sec)
+                                                                                                        │
+MySQL [test1]> XA ROLLBACK '61c05243994c3';                                                             │MySQL [test2]> XA ROLLBACK '61c052439bd7b';
+Query OK, 0 rows affected (0.010 sec)                                                                   │Query OK, 0 rows affected (0.010 sec)
+                                                                                                        │
+MySQL [test1]> select * from test where id = 1;                                                         │MySQL [test2]> select * from test where id = 1;
++----+---------+                                                                                        │+----+---------+
+| id | val     |                                                                                        │| id | val     |
++----+---------+                                                                                        │+----+---------+
+|  1 | xatest1 |                                                                                        │|  1 | xatest2 |
++----+---------+                                                                                        │+----+---------+
+1 row in set (0.009 sec)                                                                                │1 row in set (0.083 sec)
+
+MySQL [test1]>  XA START '61c052438d3eb';
+Query OK, 0 rows affected (0.030 sec)
+
+MySQL [test1]> update test set val = 'recover' where id = 1;
+Query OK, 1 row affected (0.072 sec)
+
+MySQL [test1]> select * from test where id = 1;
++----+---------+
+| id | val     |
++----+---------+
+|  1 | recover |
++----+---------+
+1 row in set (0.039 sec)
+
+MySQL [test1]>  XA END '61c052438d3eb';
+Query OK, 0 rows affected (0.005 sec)
+
+MySQL [test1]> XA PREPARE '61c052438d3eb';
+Query OK, 0 rows affected (0.020 sec)
+
+MySQL [test1]> XA RECOVER;
++----------+--------------+--------------+---------------+
+| formatID | gtrid_length | bqual_length | data          |
++----------+--------------+--------------+---------------+
+|        1 |           13 |            0 | 61c052438d3eb |
++----------+--------------+--------------+---------------+
+1 row in set (0.010 sec)
+
+MySQL [test1]> XA RECOVER CONVERT XID;
++----------+--------------+--------------+------------------------------+
+| formatID | gtrid_length | bqual_length | data                         |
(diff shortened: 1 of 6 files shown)


shardingsphere-proxy/shardingsphere-proxy-backend/src/main/java/org/apache/shardingsphere/proxy/backend/text/transaction/TransactionXAHandler.java
```diff
@@ -0,0 +1,105 @@
...
+    @Override
+    public ResponseHeader execute() throws SQLException {
+        switch (tclStatement.getOp()) {
+            case "START":
+            case "BEGIN":
+                /**
+                 * we have to let session occupy the thread when doing xa transaction.
+                 * according to https://dev.mysql.com/doc/refman/5.7/en/xa-states.html XA and local transactions are mutually exclusive
+                 */
+                if (connectionSession.getTransactionStatus().isInTransaction()) {
```
zhfeng, line 74: It should throw ```new SQLException("can not start in a Active transaction")``` immediately but not ```commit``` the current transaction.

shardingsphere-proxy/shardingsphere-proxy-backend/src/main/java/org/apache/shardingsphere/proxy/backend/text/transaction/TransactionXAHandler.java
```diff
@@ -0,0 +1,105 @@
...
+            case "END":
+            case "PREPARE":
+            case "RECOVER":
+                return backendHandler.execute();
+            case "COMMIT":
+            case "ROLLBACK":
+                try {
+                    return backendHandler.execute();
+                } finally {
+                    connectionSession.getTransactionStatus().setManualXA(false);
```
zhfeng, line 98: It also need to ```setInTransaction(false)```

shardingsphere-kernel/shardingsphere-transaction/shardingsphere-transaction-core/src/main/java/org/apache/shardingsphere/transaction/core/TransactionType.java
```diff
@@ -22,7 +22,7 @@
  */
 public enum TransactionType {
     
-    LOCAL, XA, BASE;
+    LOCAL, XA, BASE, MANUALXA;
```
zhfeng, line 25: I'd like to suggest to change ```XA``` to ```JTA``` but it should be fixed in another PR in the future.

shardingsphere-proxy/shardingsphere-proxy-backend/src/main/java/org/apache/shardingsphere/proxy/backend/text/transaction/TransactionXAHandler.java
```diff
@@ -0,0 +1,105 @@
...
+                    TransactionHolder.clear();
+                }
+            default:
+                throw new SQLException("unrecognized XA statement");
```
zhfeng, line 102: It could be better to output the ```tclStatement.getOp()``` for debugging the issue.

zhfeng:
@roodkcab Well done ! and I just add my reviews. This PR looks good to me.

shardingsphere-proxy/shardingsphere-proxy-backend/src/main/java/org/apache/shardingsphere/proxy/backend/text/transaction/TransactionXAHandler.java
```diff
@@ -0,0 +1,105 @@
...
+                        throw new SQLException("already in XA transaction");
+                    }
+                }
+                TransactionHolder.setInTransaction();
+                connectionSession.setAutoCommit(false);
+                connectionSession.getTransactionStatus().setManualXA(true);
+                JDBCBackendConnection connection = (JDBCBackendConnection) connectionSession.getBackendConnection();
+                connection.closeDatabaseCommunicationEngines(true);
+                connection.closeConnections(false);
+                return backendHandler.execute();
```
zhfeng, line 88: I wonder that it only set connectionSession to manualXA only after ```backendHandler.execute()``` success.

shardingsphere-kernel/shardingsphere-transaction/shardingsphere-transaction-core/src/main/java/org/apache/shardingsphere/transaction/core/TransactionType.java
roodkcab: yes, how about JTAXA?

shardingsphere-proxy/shardingsphere-proxy-backend/src/main/java/org/apache/shardingsphere/proxy/backend/text/transaction/TransactionXAHandler.java
roodkcab: no problem

shardingsphere-kernel/shardingsphere-transaction/shardingsphere-transaction-core/src/main/java/org/apache/shardingsphere/transaction/core/TransactionType.java
zhfeng: ```JTAXA``` is weird but I think we can discuss it later.

shardingsphere-proxy/shardingsphere-proxy-backend/src/main/java/org/apache/shardingsphere/proxy/backend/text/transaction/TransactionXAHandler.java
roodkcab: yes, you are right. this handler is different JTA, and shouldn't close connection with backend database

zhfeng:
@roodkcab Is it possible to add a test for XAHandler ?

shardingsphere-proxy/shardingsphere-proxy-backend/src/main/java/org/apache/shardingsphere/proxy/backend/text/transaction/TransactionXAHandler.java
roodkcab: and I keep TransactionHolder here, it seems this object is only used in readwritesplit router.

roodkcab:
> @roodkcab Is it possible to add a test for XAHandler ?

I'm trying to figure out how does ss test normal transaction and then write some test for this the same way, any ideas?

shardingsphere-proxy/shardingsphere-proxy-backend/src/main/java/org/apache/shardingsphere/proxy/backend/text/transaction/TransactionXAHandler.java
roodkcab: yes, override `isInTransaction()` in `TransactionStatus` to avoid set it.

zhfeng (approved):
LGTM

jingshanglu:
@roodkcab  Are there any manual test report?

roodkcab:
@jingshanglu https://github.com/SphereEx/auto-test can we use this to give this feature a test? I've written a simple use case in PHP [here](https://github.com/roodkcab/shardingsphere-test)

jingshanglu:
> @jingshanglu https://github.com/SphereEx/auto-test can we use this to give this feature a test? I've written a simple use case in PHP [here](https://github.com/roodkcab/shardingsphere-test), but it definitely can't generate any test report like JUnit does

@roodkcab Yeah, we can use `auto-test`,we will run auto-test daily. `auto-test` have a test report like `Junit`,maybe we can implement test case in `auto-test`.

jingshanglu:
@roodkcab Should we add some check for number of datasource,whether it is in encrypt mode?

roodkcab:
> @roodkcab Should we add some check for number of datasource,whether it is in encrypt mode?

shall we add a variable in init progress to indicate no multi datasource configured , and check this variable when `xa start` ?

jingshanglu:
> Shall we add a variable in init progress to indicate no multi datasource configured , and check this variable when `xa start` ? Can we change datasource on fly using `distSQL` ? Or can we document this feature limitation, and leave it to user? I don't think this check is not that reliable.

@roodkcab Ok, now, when `distSQL` change datasource , metadata may be inconsistent.Maybe it is better to add `TODO` and document, and then manual test the feature, report the test result.

zhfeng:
Please add these following issues on github and we can track all of them easily

- create a feature test on ```auto-test```
- change database dynamically by using ```distSQL```
- improve the documentation and make this ```XA``` support as ```experimental``` or ```Pre TechView```

roodkcab:
image
manual commit test

image
manual rollback test

jingshanglu (approved):

tristaZero:
Hello @roodkcab @zhfeng @jingshanglu Glad to see this PR merged and really appreciate it, happy holidays!

docs/document/content/features/transaction/use-norms/xa.en.md
```diff
@@ -13,3 +13,101 @@ weight = 2
 ## Unsupported
 
 * Recover committing and rolling back in other machines after the service is down.
+
+## XA Transaction managed by XA Statement
```
terrymanu, line 17: What is the different with original XA? If same, why it appear here?

shardingsphere-kernel/shardingsphere-transaction/shardingsphere-transaction-core/src/main/java/org/apache/shardingsphere/transaction/core/TransactionType.java
terrymanu, line 25: I cannot agree to add new transaction type here.
I don't think `MANUALXA` is a valid transaction type, it is not a same abstract level with `LOCAL`, `XA` and `BASE`

terrymanu:
I think this PR need to be reverted.

docs/document/content/features/transaction/use-norms/xa.en.md
zhfeng: I think the difference between `XAMANUAL` and `XA` is that users can contronl the XA transaction by themselves in `XAMANUAL` and `sharding-proxy` contronls XA transction in `XA`.

I think the motivation to add this transaction type is that users want to do the XA operations by themselves in such case like involving in two or more resources in a transacion, e.g. 
```
tx.begin();
jms.send("/queue/test", "messages"); // send to a message broker
jdbc.update("insert into db_test"); // insert into database through the proxy
tx.commit();
```

And yeah, in this scenario, users should be responsible to the transaction operations. `sharding-proxy` just bypass all of the XA related statments to the backend databases;

shardingsphere-kernel/shardingsphere-transaction/shardingsphere-transaction-core/src/main/java/org/apache/shardingsphere/transaction/core/TransactionType.java
zhfeng: `MANUALXA` is not a good name. It might be better to use `NONE` or `TRANSPARENT`. you know, naming is a hard work!

zhfeng:
@terrymanu I think we need more discussion before reverting. Thanks!

zhfeng:
@terrymanu have you ever read the discussions from https://github.com/apache/shardingsphere/issues/12771? I think it is a long story.

docs/document/content/features/transaction/use-norms/xa.en.md
roodkcab: > What is the different with original XA? If same, why it appear here?

both of them are distribution transaction. however

`XA` could do it automatically by using `begin`, `commit`/`rollback`
while `MANUALXA` is controlled by user themself, i.e. triggered by `XA` statements.

roodkcab:
> I think this PR need to be reverted.

cloud we keep this feature, and do it in a right way?

zhfeng:
@roodkcab please check the discussion on https://github.com/apache/shardingsphere/issues/18636 and I think we need to improve the `TransactionXAHandler`

zhfeng:
And yeah, it's shame to lose the XA Statements support in sharding-proxy

PR #14160, opened 2021-12-19
Fix #14035, `select current_user` query configured user.

### LINKED ISSUE #14035: In the Proxy, `select current_user` does not query the configured user
The query result of `select current_user()` is users in `server.yaml`.

image

But `select current_user` is routed to the database to query the user in the database;

We expect that the result of the select current_user query should also be the users configured by `server.yaml`

yx9o:
Fixes #14035.

`select current_user` query configured user.


shardingsphere-proxy/shardingsphere-proxy-backend/src/main/java/org/apache/shardingsphere/proxy/backend/util/CurrentUserEnum.java
```diff
@@ -0,0 +1,35 @@
...
+
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+
+/**
+ * Current user enum.
+ */
+@RequiredArgsConstructor
+@Getter
+public enum CurrentUserEnum {
```
terrymanu, line 28: Do we need to define enum for every function?
It is better to keep consist with original impl

terrymanu (changes requested):

terrymanu (approved):

PR #14172, opened 2021-12-20
Use union all to merge sql route units for simple select

### LINKED ISSUE #14122: Consider using union all to merge sql routes to the same database
## Feature Request

According to the description of the official document——https://shardingsphere.apache.org/document/current/en/reference/sharding/execute/, the current execute engine will judge based on `maxConnectionsSizePerQuery` and the number of sqlUnits on a single database, choose `MEMORY_STRICTLY Mode` or `CONNECTION_STRICTLY Mode`. When using `MEMORY_STRICTLY Mode`, it will consume a lot of connections, and when using `CONNECTION_STRICTLY Mode`, it will consume more memory.

In order to alleviate the problems caused by these two modes, we consider using `UNION ALL` to merge multiple `SQLUnits` on the same database to reduce connection consumption. At the same time, since each result set can be held through the connection, we can use streaming merge instead of memory merge, which will also reduce memory usage.

No.

Use `UNION ALL` to merge multiple SQLUnits on the same database.

strongduanmu:
Fixes #14122.

- Use union all to merge sql route units for simple select
- add rewrite test case for simple select
- add unit test case


terrymanu (approved):

lif123:
Is there  any chance  or configuration to turn off this feature? union all may cost too much query time . If over mysql max_execution_time, exception 'maximum statement execution time exceeded' will be triggered. Thanks!

strongduanmu:
Hi @lif123, can you submit an issue to describe your problem?

PR #14197, opened 2021-12-21
14101 ignore case when execute show tables statement

### LINKED ISSUE #14101: Show tables statement return wrong result when mysql is case sensitive

5.0.0

ShardingSphere-Proxy

### Expected behavior

Get correct show tables result.

### Actual behavior

When I created a table with uppercase TEST and executed `SHOW TABLES LIKE'TEST'`, I couldn't get the correct result.

```sql
mysql> desc TEST;
+---------+--------------+------+-----+---------+-------+
| Field   | Type         | Null | Key | Default | Extra |
+---------+--------------+------+-----+---------+-------+
| ID      | int(11)      | YES  |     | NULL    |       |
| CONTENT | varchar(100) | YES  |     | NULL    |       |
+---------+--------------+------+-----+---------+-------+
2 rows in set (0.36 sec)

mysql> desc test;
ERROR 1146 (42S02): Table 'demo_ds_1.test' doesn't exist
mysql> SHOW TABLES LIKE 'test';
+-----------------------+------------+
| Tables_in_sharding_db | Table_type |
+-----------------------+------------+
| test                  | BASE TABLE |
+-----------------------+------------+
1 row in set (1.23 sec)

mysql> SHOW TABLES LIKE 'TEST';
Empty set (0.00 sec)
```

MingxingLAI:
Fixes #14101 .

- ignore case when execute show tables statement
- Add more unit test for ShowTablesExecutor

### DIFF
--- a/shardingsphere-proxy/shardingsphere-proxy-backend/src/main/java/org/apache/shardingsphere/proxy/backend/text/admin/mysql/executor/ShowTablesExecutor.java
+++ b/shardingsphere-proxy/shardingsphere-proxy-backend/src/main/java/org/apache/shardingsphere/proxy/backend/text/admin/mysql/executor/ShowTablesExecutor.java
@@ -78,7 +78,7 @@ public final class ShowTablesExecutor implements DatabaseAdminQueryExecutor {
     private Collection<String> getAllTableNames(final String schemaName) {
         Collection<String> allTableNames = ProxyContext.getInstance().getMetaData(schemaName).getSchema().getAllTableNames();
         if (showTablesStatement.getFilter().isPresent()) {
-            Optional<String> pattern = showTablesStatement.getFilter().get().getLike().map(each -> SQLUtil.convertLikePatternToRegex(each.getPattern()));
+            Optional<String> pattern = showTablesStatement.getFilter().get().getLike().map(each -> SQLUtil.convertLikePatternToRegex(each.getPattern().toLowerCase()));
             return pattern.isPresent() ? allTableNames.stream().filter(each -> each.matches(pattern.get())).collect(Collectors.toList()) : allTableNames;
         }
         return allTableNames;
--- a/shardingsphere-proxy/shardingsphere-proxy-backend/src/test/java/org/apache/shardingsphere/proxy/backend/text/admin/mysql/executor/ShowTablesExecutorTest.java
+++ b/shardingsphere-proxy/shardingsphere-proxy-backend/src/test/java/org/apache/shardingsphere/proxy/backend/text/admin/mysql/executor/ShowTablesExecutorTest.java
@@ -51,11 +51,8 @@ public final class ShowTablesExecutorTest {
     
     private static final String SCHEMA_PATTERN = "schema_%s";
     
-    private ShowTablesExecutor showTablesExecutor;
-    
     @Before
     public void setUp() throws NoSuchFieldException, IllegalAccessException {
-        showTablesExecutor = new ShowTablesExecutor(new MySQLShowTablesStatement());
         Map<String, ShardingSphereMetaData> metaDataMap = getMetaDataMap();
         Field contextManagerField = ProxyContext.getInstance().getClass().getDeclaredField("contextManager");
         contextManagerField.setAccessible(true);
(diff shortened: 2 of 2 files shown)


shardingsphere-proxy/shardingsphere-proxy-backend/src/main/java/org/apache/shardingsphere/proxy/backend/text/admin/mysql/executor/ShowTablesExecutor.java
```diff
@@ -78,7 +78,7 @@ private QueryResult getQueryResult(final String schemaName) {
     private Collection<String> getAllTableNames(final String schemaName) {
         Collection<String> allTableNames = ProxyContext.getInstance().getMetaData(schemaName).getSchema().getAllTableNames();
         if (showTablesStatement.getFilter().isPresent()) {
-            Optional<String> pattern = showTablesStatement.getFilter().get().getLike().map(each -> SQLUtil.convertLikePatternToRegex(each.getPattern()));
+            Optional<String> pattern = showTablesStatement.getFilter().get().getLike().map(each -> SQLUtil.convertLikePatternToRegex(each.getPattern().toLowerCase()));
             return pattern.isPresent() ? allTableNames.stream().filter(each -> each.matches(pattern.get())).collect(Collectors.toList()) : allTableNames;
```
strongduanmu, line 82: Hi @MingxingLAI, `allTableNames` returns lowercase table names now, which is the cause of confusion for users.

strongduanmu (changes requested):

shardingsphere-proxy/shardingsphere-proxy-backend/src/main/java/org/apache/shardingsphere/proxy/backend/text/admin/mysql/executor/ShowTablesExecutor.java
MingxingLAI: I think there are three ways to slove this problem：

1. Treat all table names in lower case（like this pull request)
2. Distinguish case of table name
3. Ignore uppercase table names

Which solution do you think is the most proper ?

If we use the second solution, it may cause compatibility problems for many users who has use ShardingSphere and thread all table names in lower case. So, I this the first solution is proper.

MingxingLAI: In addition，In the implement of ShardingSphere, treat all table names in lower case.

```java

public final class ShardingSphereSchema {
    
    private final Map<String, TableMetaData> tables;
    
    @SuppressWarnings("CollectionWithoutInitialCapacity")
    public ShardingSphereSchema() {
        tables = new ConcurrentHashMap<>();
    }
    
    public ShardingSphereSchema(final Map<String, TableMetaData> tables) {
        this.tables = new ConcurrentHashMap<>(tables.size(), 1);
        tables.forEach((key, value) -> this.tables.put(key.toLowerCase(), value));
    }
```

strongduanmu: @MingxingLAI I think we should distinguish case of table name for this issue. And the final result displayed to the user should be consistent with the table in the database, rather than converted to lowercase.

MingxingLAI: I also think distinguish case of table name is more accurate, but I am still more worried about compatibility problem, or we invite more people to participate in this discussion?

strongduanmu: > In addition，In the implement of ShardingSphere, treat all table names in lower case.
> 
> ```java
> public final class ShardingSphereSchema {
>     
>     private final Map<String, TableMetaData> tables;
>     
>     @SuppressWarnings("CollectionWithoutInitialCapacity")
>     public ShardingSphereSchema() {
>         tables = new ConcurrentHashMap<>();
>     }
>     
>     public ShardingSphereSchema(final Map<String, TableMetaData> tables) {
>         this.tables = new ConcurrentHashMap<>(tables.size(), 1);
>         tables.forEach((key, value) -> this.tables.put(key.toLowerCase(), value));
>     }
> ```

@MingxingLAI Only the key is lowercase, and all the real information is stored in the values. So, I think we can use values to handle show tables logic.

MingxingLAI: good idea, I will change the implement.

strongduanmu: Hi @MingxingLAI, does this pr has any progress?

Issue #12366, opened 2021-09-11
Example improvement plan


terrymanu:
The current example is not good enough. It is difficult to understand, which users can not find their desired example codes quickly, the disadvantages are:

1. The categories of examples are not in same abstract level, they are not show the overview of ShardingSphere features and structure clearly;
2. The `example-core` module is too complicated which contains too many common and abstract codes, user do not need to understander them. It is better to use "flat" codes and make them independent each other;
3. Proxy and JDBC need to provide configurations for multiple dimensions, such as different modes, different kernel providers and different feature configs;
4. JDBC example can be split to different adaptors, such as `jdbc`, `spring boot starter & jdbc`, `spring boot starter & jpa`, `spring boot starter & mybatis`, `spring namespace & jdbc`, `spring namespace & jpa` and , `spring namespace & mybatis`;
5. Proxy do not need java connect example, config only is enough.

It is more clear if adjust example structure as:

```
shardingsphere-example
  ├- shardingsphere-jdbc-example
     ├- shardingsphere-jdbc-${mode}-example
        ├- shardingsphere-jdbc-${mode}-${transaction}-${feature}-example
           ├- shardingsphere-jdbc-${mode}-${transaction}-${feature}-${framework}-example
  ├- shardingsphere-proxy-example
     ├- shardingsphere-proxy-${mode}-example
        ├- shardingsphere-proxy-${mode}-${transaction}-${feature}-example
  ├- shardingsphere-parser-example
```

`${xxx}` are variables which can be replaced by:

- `${mode}`: memory, standalone-file, cluster-zookeeper, cluster-etcd
- `${transaction}`: local, xa-atomikos, xa-bitronix, xa-narayana, base-seata
- `${feature}`: sharding, readwrite-splitting, db-discovery, encrypt, shadow and all permutation of them
- `${framework}`: jdbc, springboot-starter-jdbc, springboot-starter-jpa, springboot-starter-mybatis, springnamespace-jdbc, springnamespace-jpa, spring-namespace-mybatis

The duplicated codes may consider about using code-generation engine.

![example-code-gen-engine drawio](https://user-images.githubusercontent.com/5516298/132954397-406a0dbe-f28f-4e5b-89b2-1d79595422b4.png)

totalo:
hi, I want to finish this job. plz assgin to me. thanks.

terrymanu:
Welcome, we can discuss more about it.

totalo:
[...]

PR #14330, opened 2021-12-26
Add db-discovery spring namespace jdbc example and template

totalo:
For #12366.

- add db-discovery spring namespace jdbc example
- add db-discovery spring namespace jdbc template


tristaZero (approved):
Nice job.

PR #14364, opened 2021-12-27
Adjust shadow example and template to fix ci error

totalo:
refer #14355.

- adjust shadow example and template
- fix ci error


soulasuna (commented):
Thank you very much for your contribution. 
The problem here can be solved not only by modifying these. 
I am fixing these

totalo:
> Thank you very much for your contribution. The problem here can be solved not only by modifying these. I am fixing these

ok，thanks.

PR #14428, opened 2021-12-29
Fix show readwrite_splitting read resource disable data duplication.

lanchengx:


lanchengx:
@RaigorJiang could you have a  review ?

RaigorJiang (approved):

PR #14443, opened 2021-12-30
Add if exists in `Drop database` syntax.

lanchengx:


RaigorJiang (changes requested):
When execute `DROP`, the field should be `containsExistClause`, like which in `MySQLDropTableStatement`.

shardingsphere-sql-parser/shardingsphere-sql-parser-statement/src/main/java/org/apache/shardingsphere/sql/parser/sql/common/statement/ddl/DropDatabaseStatement.java
```diff
@@ -30,5 +30,7 @@
 @ToString
 public abstract class DropDatabaseStatement extends AbstractSQLStatement implements DDLStatement {
     
+    private boolean containsNotExistClause;
```
RaigorJiang, line 33: `containsExistClause`

RaigorJiang (approved):

PR #14501, opened 2022-01-04
set writeonly hint based on connection username

### LINKED ISSUE #14472: route sql to write datasource based on username or listen address.
read write split datasource can be routed through hint. but it'll be more convenient to use username or connection host to set hint automatically.

example:
when user name is `db_writeonly`, session could set writeonly hint automatically;

or we can add a new property `writeOnlyHost` to read write split configuration, if user connect to database use `writeOnlyHost` then set writeonly hint automatically on that session. (it may difficult to get connection host, since host is always converted to IP address before connect to database. however, if user connect to database use lo network, we can distinguish different host ip, like 127.0.0.1, 127.0.0.2, and etc)

roodkcab:
Fixes #14472 .

- set write only hint based on session username


terrymanu (changes requested):
This feature is for readwrite split, could we consider enclosure it into readwrite-splitting module?

roodkcab:
> This feature is for readwrite split, could we consider enclosure it into readwrite-splitting module?

then I guess we have to encapsulate session information into sql context ?

terrymanu:
> > This feature is for readwrite split, could we consider enclosure it into readwrite-splitting module?
> 
> then I guess we have to encapsulate session information into sql context ?

Yes, it is the better way

terrymanu:
This pull request is long time ago, I just close it.

PR #14512, opened 2022-01-04
Remove unnecessary static.

yx9o:
Remove unnecessary static.


terrymanu (changes requested):
It is better to use static here, the variable is class based

yx9o:
> It is better to use static here, the variable is class based

Ok, I will close this pr, do getColumnNames in other QueryResultSet classes need to be adjusted to static?

Issue #14706, opened 2022-01-12
Plan to split the readwrite-splitting static and dynamic configuration to SPI interfaces.


zjcnb:
Hi community,

Now, readwrite-splitting static and dynamic configuration are together, we think this is not clear enough. So, we plan to split the readwrite-splitting static and dynamic configuration to SPI interfaces.

`Before adjustment configuration :`

```
schemaName: readwrite_splitting_db

dataSources:
  write_ds:
    url: jdbc:mysql://127.0.0.1:3306/demo_write_ds?serverTimezone=UTC&useSSL=false
    username: root
    password: 123456
    connectionTimeoutMilliseconds: 30000
    idleTimeoutMilliseconds: 60000
    maxLifetimeMilliseconds: 1800000
    maxPoolSize: 50
    minPoolSize: 1
  read_ds_0:
    url: jdbc:mysql://127.0.0.1:3306/demo_read_ds_0?serverTimezone=UTC&useSSL=false
    username: root
    password: 123456
    connectionTimeoutMilliseconds: 30000
    idleTimeoutMilliseconds: 60000
    maxLifetimeMilliseconds: 1800000
    maxPoolSize: 50
    minPoolSize: 1

rules:
- !READWRITE_SPLITTING
  dataSources:
    read_wirte_ds:
      writeDataSourceName: write_ds
      readDataSourceNames:
        - read_ds_0
      loadBalancerName: random
  loadBalancers:
    random:
      type: RANDOM
```

`After adjustment dynamic configuration :`

```
schemaName: readwrite_splitting_db

dataSources:
  write_ds:
    url: jdbc:mysql://127.0.0.1:3306/demo_write_ds?serverTimezone=UTC&useSSL=false
    username: root
    password: 123456
    connectionTimeoutMilliseconds: 30000
    idleTimeoutMilliseconds: 60000
    maxLifetimeMilliseconds: 1800000
    maxPoolSize: 50
    minPoolSize: 1
  read_ds_0:
    url: jdbc:mysql://127.0.0.1:3306/demo_read_ds_0?serverTimezone=UTC&useSSL=false
    username: root
    password: 123456
    connectionTimeoutMilliseconds: 30000
    idleTimeoutMilliseconds: 60000
    maxLifetimeMilliseconds: 1800000
    maxPoolSize: 50
    minPoolSize: 1

rules:
  - !READWRITE_SPLITTING
    dataSources:
      read_wirte_ds:
        load-balancer-name: random
[...]

PR #14745, opened 2022-01-13
Adjust readwrite-splitting api about yaml and boot starter.

zjcnb:
For #14706


menghaoranss:
I will merge the PR first, and the errors of Integration Test will be fixed by next PR

PR #14822, opened 2022-01-17
Use JDBCMemoryQueryResult to avoid federation execution exception

strongduanmu:
Ref #14780.

- Use JDBCMemoryQueryResult to avoid federation execution exception


TeslaCN (approved):
LGTM.

Issue #14254, opened 2021-12-23
Support automatic generation of instanceId for Memory/Standalone/Cluster mode


menghaoranss:
We must manually configure the `work-id` for distributed key generator algorithm just as follows:

```
keyGenerators:
    snowflake:
      type: SNOWFLAKE
      props:
        worker-id: 123
```

For the cluster mode, we cannot specify different `work-id` for all instances. Therefore, we consider using the automatically generated `instance-id` instead of the manually configured `work-id`, and provide different generation algorithms for the Memory/Standalone/Cluster mode. After that, we no longer need to configure the `work-id`:

```
keyGenerators:
    snowflake:
      type: SNOWFLAKE
```

Ahoo-Wang:
@menghaoranss 

In memory / standalone mode, we use JDBC(*cosid-jdbc*) as the back-end storage of machine id. and in cluster mode, we use zookeeper(*cosid-zookeeper*) as the back-end storage. What do you think?

menghaoranss:
> @menghaoranss
> 
> In memory / standalone mode, we use JDBC(_cosid-jdbc_) as the back-end storage of machine id. and in cluster mode, we use zookeeper(_cosid-zookeeper_) as the back-end storage. What do you think?

@Ahoo-Wang the back-end storage of machine id is based on mode of ShardingSphere, therefore cosid does not need to care about storage, I will pass the `WorkerIdGenerator` to the cosid algorithm, and you can get worker id (machine id) through the interface.

Ahoo-Wang:
>  the back-end storage of machine id is based on mode of ShardingSphere, therefore cosid does not need to care about storage, I will pass the `WorkerIdGenerator` to the cosid algorithm, and you can get worker id (machine id) through the interface.

@menghaoranss That's a good idea.

menghaoranss:
@Ahoo-Wang this issue has been completed, and I have passed the `InstanceContext` to `CosIdKeyGenerateAlgorithm`, you can refactor this algorithm now,  and look forward to using it in the next version.

Ahoo-Wang:
@Ahoo-Wang ok.

Ahoo-Wang:
#15067

PR #14837, opened 2022-01-17
refactor unit test for SnowflakeKeyGenerateAlgorithmTest

menghaoranss:
For #14254


sandynz (approved):
`assertSetWorkerIdSuccess` is not necessary now, since `worker-id` is not in props any more.

Issue #14015, opened 2021-12-09
Support unsupported sql in parser.


tuichenchuxin:
Hi community,
There are some unsupported sql in [this file](https://github.com/apache/shardingsphere/blob/master/shardingsphere-test/shardingsphere-parser-test/src/main/resources/sql/unsupported/unsupported.xml)
You can help to support if you
>Be familiar with g4 file of Antlr (The grammars are RegularExpression-like)
Be familiar with java 

## Background
How is a SQL parsed?
`SQL definition file`
->(Parse to)
`AST`
->(Visit it)
`SQLStatment`

## Process
- Pick an unsupported sql and try to get why it's not supported.(antlr4 grammer? or not implement visit method. You can use [antlr4 plugins](https://plugins.jetbrains.com/plugin/7358-antlr-v4) to help analysis.You may need to visit offical doc to check the grammer.)
- After you fixing it, remember to add new corresponding SQL case in [SQL Cases](https://github.com/apache/shardingsphere/tree/master/shardingsphere-test/shardingsphere-parser-test/src/main/resources/sql/supported) and expected parsed result in [Expected Statment XML](https://github.com/apache/shardingsphere/tree/master/shardingsphere-test/shardingsphere-parser-test/src/main/resources/case)
- Run [SQLParserParameterizedTest](https://github.com/apache/shardingsphere/blob/master/shardingsphere-test/shardingsphere-parser-test/src/main/java/org/apache/shardingsphere/test/sql/parser/parameterized/engine/SQLParserParameterizedTest.java) to make sure no exceptions.
## Notice
1.Left recursion can not work well in Antlr4,
```
a
  : b
  | a b
  ;
b
  : a c
  ;
c
  : ...
  ;
```
you can consider the following one,
```
a
  : b*
  ;
b
  : a c
  ;
c
  : ...
  ;
```
[Here](https://shimo.im/docs/WrgrCt9JR3JDjq8H/read) is a Chinese version for reference.

## Sub Task

- [ ] assert_select_into_SQL
- [ ] assert_select_with_collate_with_marker
- [ ] assert_select_with_json_extract_sign_with_parameter_marker
- [ ] assert_select_with_json_unquote_extract_sign_with_parameter_marker
- [ ] assert_insert_with_first
- [ ] assert_dist_SQL_show_rule_parse_conflict
- [ ] create_bit_xor_table
- [ ] select_with_comment1
- [ ] select_with_comment2
- [ ] select_with_comment5
[...]

PR #14850, opened 2022-01-18
support alter default privileges for pg/og

tuichenchuxin:
for #14015.

- support alter default privileges for pg/og
- add test case
- delete unsupported case


strongduanmu (approved):
@tuichenchuxin Good job.

PR #14890, opened 2022-01-19
[DistSQL] Drop statements add the judgment of if exists

lanchengx:
Related statements
- drop db_discovery rule 
- drop single table rule
- drop encrypt rule
- drop readwrite splitting rule
- drop shadow rule
- drop shadow algorithm
- drop sharding table rule
- drop sharding algorithm
- drop resource


RaigorJiang:
@lanchengx This design still needs to be optimized, I just set as a draft.

Issue #14110, opened 2021-12-16
[Proofread SQL Definition][MySQL]Proofread the g4 files of SQL definitions


tuichenchuxin:
Hi community,

This issue is to proofread the SQL definitions of MySQL. Therefore, we can support parsing more SQLs.
We need your interaction! If you

>Be familiar with g4 file of Antlr (The grammars are RegularExpression-like)
Understand the MySQL grammars 

then welcome!

#### How is a SQL parsed?
`SQL definition file`
->(Parse to)
`AST`
->(Visit it)
`SQLStatment`

### Process
Compare SQL definitions in [Oficial SQL Doc](https://dev.mysql.com/doc/refman/8.0/en/sql-statements.htmll) and [ShardingSphere SQL Doc](https://github.com/apache/shardingsphere/tree/master/shardingsphere-sql-parser/shardingsphere-sql-parser-dialect/shardingsphere-sql-parser-mysql/src/main/antlr4/imports/mysql)
If there is any difference in ShardingSphere SQL Doc, please correct them referring to Official SQL Doc
You can also refer to [MySQL source code](https://github.com/mysql/mysql-server/blob/8.0/sql/sql_yacc.yy)
Run mvn install the current_file_module
Check whether there are any exceptions. If indeed, please fix them. (Especially xxxVisitor.class)
Add new corresponding SQL case in [SQL Cases](https://github.com/apache/shardingsphere/tree/master/shardingsphere-test/shardingsphere-parser-test/src/main/resources/sql/supported) and expected parsed result in[ Expected Statment XML](https://github.com/apache/shardingsphere/tree/master/shardingsphere-test/shardingsphere-parser-test/src/main/resources/case)
Run [SQLParserParameterizedTest](https://github.com/apache/shardingsphere/blob/master/shardingsphere-test/shardingsphere-parser-test/src/main/java/org/apache/shardingsphere/test/sql/parser/parameterized/engine/SQLParserParameterizedTest.java) to make sure no exceptions.
### Notice
- Left recursion can not work well in Antlr4,
```
a
  : b
  | a b
  ;
b
  : a c
  ;
c
  : ...
  ;
```
- you can consider the following one,
```
a
  : b*
  ;
b
  : a c
  ;
c
  : ...
  ;
```
[Here](https://shimo.im/docs/WrgrCt9JR3JDjq8H/read) is a Chinese version for reference.
### subtask
- Data Definition Statements
[...]

PR #14920, opened 2022-01-19
Support MySQL to use `COLLATE` in whereClasue

totalo:
For #14110.

- support MySQL to use `COLLATE` in whereClasue
- change the index name of the placeholder from `value` to `parameter-index`
- add test case


strongduanmu (approved):
@totalo Thank you for your pr.

PR #15047, opened 2022-01-25
Add ProxyGlobalConfiguration

terrymanu:


shardingsphere-proxy/shardingsphere-proxy-bootstrap/src/main/java/org/apache/shardingsphere/proxy/config/ProxyGlobalConfiguration.java
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * Proxy global configuration.
+ */
+@RequiredArgsConstructor
+@Getter
+public final class ProxyGlobalConfiguration {
+    
+    private final Collection<RuleConfiguration> rules;
+    
+    private final Properties properties;
```
totalo, line 36: How about `props`, other places are this variable name.

terrymanu: We'd like use full name for public method.
We just use @Getter to generate the get method, so getProperties is better method name.

strongduanmu (approved):

PR #15114, opened 2022-01-27
For #14722: Add release lock logic for failure cases during migration

### LINKED ISSUE #14722: Only one scaling job could be executed at the same time for one schema
No

In case of there're several scaling jobs are created that belong to the same schema, they will run concurrently, it might cause issue. So it's better to add some limitation:
1. Make sure there's only one scaling job is running if there're seveval scaling jobs that belong to the same schema.
2. Validate metadata, if target metadata is old than current enabled one, then it should not be executed. (It will be done after metadata multi-version is completed, not in this feature request.)

ReyYang:
Fixes #14722.

- Added release lock event
- Add release lock logic for failure cases
-


ReyYang:
- In the case of multiple Proxy nodes, this method is invalid
- Since the distributed lock is implemented as Zookeeper, the `repository.type` only supports Zookeeper
- Use distSQL to stop the migration task, the distributed lock will not be released

sandynz (approved):

Issue #14987, opened 2022-01-21
support encrypt config column data type


tuichenchuxin:
## Background
Currently, if we create a encrypt table like this
```
mysql> PREVIEW CREATE TABLE `t_encrypt` (`id` int(11) DEFAULT NULL, `user_id` varchar(100) DEFAULT NULL, `order_id` varchar(100) DEFAULT NULL) ENGINE=InnoDB;
+------------------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| data_source_name | sql                                                                                                                                                                                     |
+------------------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| ds_0             | CREATE TABLE `t_encrypt` (`id` int(11) DEFAULT NULL, user_cipher varchar(100) DEFAULT NULL, user_plain varchar(100) DEFAULT NULL, order_cipher varchar(100) DEFAULT NULL) ENGINE=InnoDB |
+------------------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
```
The cipher column's data type and length is same with the logic column.
But encrypt column will have some difference with logic column.
So, we think we need to add config for this.
## Task
- [x] Mete Data load encrypt table with config logic column data type
- [x] EncryptCreateTableTokenGenerator、EncryptAlterTableTokenGenerator support create token by config
- [x] DistSQL support creating encrypt rule with config

lanchengx:
@tuichenchuxin DistSQL has been completed, can this issue be closed?

tuichenchuxin:
@lanchengx Yes, I'll close it.

PR #15119, opened 2022-01-27
Support config encrypt column data type

tuichenchuxin:
For #14987.

- move data types into encrypt rule
- refactor encrypt column


strongduanmu (approved):
This refactoring makes the logic of encrypt data type clearer.

PR #15148, opened 2022-01-28
Add kernel new blog

yy2so:


shardingsphere-distsql/pom 2.xml
```diff
@@ -0,0 +1,36 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
menghaoranss, line 1: this pom file doesn't seem to be submitted

menghaoranss (changes requested):

menghaoranss (approved):

PR #15136, opened 2022-01-28
For #14722: Refactor the issue that cannot have multiple jobs under the same Schema name

ReyYang:
Fixes #14722.

Changes proposed in this pull request:
- Added `ZookeeperDistributeLock` class
- Modify the lock position to `PipelineJobExecutor`

PR #15172, opened 2022-01-29
For #15136: Modify the isUncompletedJobOfSameSchemaInJobList method to not judge the completed Job

### LINKED ISSUE #14722: Only one scaling job could be executed at the same time for one schema

ReyYang:
Fixes #14722 .

- Modify the isUncompletedJobOfSameSchemaInJobList method to not judge the completed Job

### DIFF
--- a/shardingsphere-kernel/shardingsphere-data-pipeline/shardingsphere-data-pipeline-core/src/main/java/org/apache/shardingsphere/data/pipeline/scenario/rulealtered/RuleAlteredJobWorker.java
+++ b/shardingsphere-kernel/shardingsphere-data-pipeline/shardingsphere-data-pipeline-core/src/main/java/org/apache/shardingsphere/data/pipeline/scenario/rulealtered/RuleAlteredJobWorker.java
@@ -19,7 +19,6 @@ package org.apache.shardingsphere.data.pipeline.scenario.rulealtered;
 
 import com.google.common.eventbus.Subscribe;
 import lombok.extern.slf4j.Slf4j;
-import org.apache.commons.collections4.CollectionUtils;
 import org.apache.commons.lang3.tuple.Pair;
 import org.apache.shardingsphere.data.pipeline.api.PipelineJobAPIFactory;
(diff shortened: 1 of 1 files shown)


shardingsphere-kernel/shardingsphere-data-pipeline/shardingsphere-data-pipeline-core/src/main/java/org/apache/shardingsphere/data/pipeline/scenario/rulealtered/RuleAlteredJobWorker.java
```diff
@@ -295,6 +299,11 @@ public static TaskConfiguration buildTaskConfig(final PipelineConfiguration pipe
     private boolean isUncompletedJobOfSameSchemaInJobList(final String schema) {
         boolean isUncompletedJobOfSameSchema = false;
         for (JobInfo each : PipelineJobAPIFactory.getRuleAlteredJobAPI().list()) {
+            List<JobProgress> unFinishedJob = PipelineJobAPIFactory.getRuleAlteredJobAPI().getProgress(each.getJobId()).values().stream()
+                    .filter(progress -> !JobStatus.FINISHED.equals(progress.getStatus())).collect(Collectors.toList());
```
sandynz, line 303: Use `stream().allMatch` could be better, `List<JobProgress> unFinishedJob` is not necessary.

sandynz (changes requested):

sandynz (approved):

sandynz:
Fix #15136. When there's completed scaling job, new scaling job could not be started.

PR #15226, opened 2022-01-31
Remove hard coded SO_BACKLOG

TeslaCN:
`128` is the default value in Linux and macOS. Netty can detect this value from `/proc/sys/net/core/somaxconn` or `sysctl`.

https://github.com/netty/netty/blob/62ef8f96e3ff5c21f4ff122756d8a21e6e6d1c01/common/src/main/java/io/netty/util/NetUtil.java#L154-L159


totalo:
Happy New Year.

totalo (approved):

PR #15231, opened 2022-02-01
Add MySQL parse support for `LOCK` Statement

totalo:
For #14110.

- add MySQL parse support for `LOCK` Statement
- add test case.


strongduanmu (approved):
Thank you @totalo, and happy chinese new year.

PR #15207, opened 2022-01-30
Fixed Refresh SingleTableRule tables.

zjcnb:
Fixes #15191

PR #15234, opened 2022-02-02
Adjust get single table data source name mapping.

### LINKED ISSUE #15191: Table changed not refresh SingleTableRule.

`master`

`ShardingSphere-Proxy`

### Expected behavior

Refresh SingleTableRule when table meta data changed.

### Actual behavior

No refresh.

image

zjcnb:
Fixes #15191, #15207

### DIFF
--- a/shardingsphere-infra/shardingsphere-infra-common/src/main/java/org/apache/shardingsphere/infra/rule/identifier/type/MutableDataNodeRule.java
+++ b/shardingsphere-infra/shardingsphere-infra-common/src/main/java/org/apache/shardingsphere/infra/rule/identifier/type/MutableDataNodeRule.java
@@ -48,5 +48,5 @@ public interface MutableDataNodeRule extends ShardingSphereRule {
      * @param tableName table name
      * @return single table data node
      */
-    Optional<DataNode> findSingleTableDataNodeByTableName(String tableName);
+    Optional<DataNode> findSingleTableDataNode(String tableName);
 }
--- a/shardingsphere-infra/shardingsphere-infra-common/src/main/java/org/apache/shardingsphere/infra/rule/identifier/type/MutableDataNodeRule.java
+++ b/shardingsphere-infra/shardingsphere-infra-common/src/main/java/org/apache/shardingsphere/infra/rule/identifier/type/MutableDataNodeRule.java
@@ -17,9 +17,10 @@
 
 package org.apache.shardingsphere.infra.rule.identifier.type;
 
+import org.apache.shardingsphere.infra.datanode.DataNode;
 import org.apache.shardingsphere.infra.rule.ShardingSphereRule;
 
-import java.util.Collection;
+import java.util.Optional;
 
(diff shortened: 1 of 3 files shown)


shardingsphere-infra/shardingsphere-infra-common/src/main/java/org/apache/shardingsphere/infra/rule/identifier/type/MutableDataNodeRule.java
```diff
@@ -42,9 +43,10 @@
...
     /**
-     * Get data source names.
+     * Find single data node by table name.
      *
-     * @return Get data source names
+     * @param tableName table name
+     * @return single table data node
      */
-    Collection<String> getDataSourceNames();
+    Optional<DataNode> findSingleTableDataNodeByTableName(String tableName);
```
strongduanmu, line 51: @zjcnb Why not change this method to findSingleTableDataNode directly?

shardingsphere-mode/shardingsphere-mode-core/src/main/java/org/apache/shardingsphere/mode/manager/ContextManager.java
```diff
@@ -251,9 +251,7 @@ public void alterSchema(final String schemaName, final ShardingSphereSchema sche
      */
     public void alterSchema(final String schemaName, final TableMetaData changedTableMetaData, final String deletedTable) {
         ShardingSphereMetaData metaData = metaDataContexts.getMetaData(schemaName);
-        if (!containsInDataNodeContainedRule(changedTableMetaData.getName(), metaData)) {
-            metaData.getRuleMetaData().findRules(MutableDataNodeRule.class).forEach(each -> each.put(changedTableMetaData.getName(), each.getDataSourceNames().iterator().next()));
-        }
+        singleTableDataSourceMappingRelation(metaData, changedTableMetaData);
```
strongduanmu, line 254: Can we use another meaningful method name? Just like 'alterSingleTableDataNodes'?

strongduanmu (changes requested):

shardingsphere-infra/shardingsphere-infra-common/src/main/java/org/apache/shardingsphere/infra/rule/identifier/type/MutableDataNodeRule.java
zjcnb: Yes, I will adjust it.

shardingsphere-mode/shardingsphere-mode-core/src/main/java/org/apache/shardingsphere/mode/manager/ContextManager.java
zjcnb: Yes, it' s a good idea.

strongduanmu (approved):

PR #15296, opened 2022-02-08
[issue-15295] upgrade dependencies due to cves

### LINKED ISSUE #15295: update example dependencies due to cves
## Bug Report

https://github.com/apache/shardingsphere/blob/master/examples/pom.xml has a number of old dependencies

dependencies with cves:
* https://mvnrepository.com/artifact/mysql/mysql-connector-java
* https://github.com/advisories/GHSA-v7wg-cpwc-24m4 (postgresql)
* https://mvnrepository.com/artifact/junit/junit
* https://mvnrepository.com/artifact/org.mybatis/mybatis
* https://mvnrepository.com/artifact/com.h2database/h2
* https://mvnrepository.com/artifact/ch.qos.logback/logback-core

I did not update snakeyaml because it is being handled in a separate issue

pjfanning:
Fixes #15295

- pom changes

### DIFF
--- a/examples/pom.xml
+++ b/examples/pom.xml
@@ -46,7 +46,7 @@
         <spring-boot.version>2.3.12.RELEASE</spring-boot.version>
         <hikari-cp.version>3.4.2</hikari-cp.version>
         <mysql-connector-java.version>8.0.28</mysql-connector-java.version>
-        <postgresql.version>42.2.25</postgresql.version>
+        <postgresql.version>42.3.2</postgresql.version>
         <h2.version>1.4.196</h2.version>
         <slf4j.version>1.7.7</slf4j.version>
         <logback.version>1.2.10</logback.version>
--- a/examples/pom.xml
+++ b/examples/pom.xml
@@ -45,13 +45,13 @@
         <spring-framework.version>5.2.19.RELEASE</spring-framework.version>
         <spring-boot.version>2.0.9.RELEASE</spring-boot.version>
         <hikari-cp.version>3.4.2</hikari-cp.version>
-        <mysql-connector-java.version>5.1.47</mysql-connector-java.version>
-        <postgresql.version>42.2.5</postgresql.version>
+        <mysql-connector-java.version>8.0.28</mysql-connector-java.version>
+        <postgresql.version>42.2.25</postgresql.version>
         <h2.version>1.4.196</h2.version>
         <slf4j.version>1.7.7</slf4j.version>
-        <logback.version>1.2.0</logback.version>
+        <logback.version>1.2.10</logback.version>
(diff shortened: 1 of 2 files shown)


examples/pom.xml
```diff
@@ -45,14 +45,14 @@
         <spring-framework.version>5.2.19.RELEASE</spring-framework.version>
         <spring-boot.version>2.0.9.RELEASE</spring-boot.version>
         <hikari-cp.version>3.4.2</hikari-cp.version>
-        <mysql-connector-java.version>5.1.47</mysql-connector-java.version>
-        <postgresql.version>42.2.5</postgresql.version>
+        <mysql-connector-java.version>8.0.28</mysql-connector-java.version>
```
taojintianxia, line 48: did you test the compatibility for these dependencies ?

examples/pom.xml
```diff
@@ -63,7 +63,7 @@
         
         <seata.version>1.4.2</seata.version>
         
-        <junit.version>4.12</junit.version>
+        <junit.version>4.13.2</junit.version>
```
TeslaCN, line 66: `assertThat` is marked as deprecated in junit 4.13. This require further consideration.

examples/pom.xml
pjfanning: @taojintianxia to be honest, I don't use shardingsphere - I am just an Apache member concerned about ASF projects relying on old dependencies with publicly disclosed attack vectors

pjfanning: my experience is that all of these updates are low impact - that in most scenarios, they can be upgraded without breaking compatibility

examples/pom.xml
linghengqian: Maybe https://shardingsphere.apache.org/community/en/contribute/code-conduct/ should be updated?

pjfanning: I reverted the junit piece - there is a CVE but it could dealt with in a separate issue

examples/pom.xml
taojintianxia: I see, anyway thanks for your contribution

examples/pom.xml
linghengqian: Refer to https://github.com/junit-team/junit4/pull/1150 , if the version is raised, it means a PR with keyword changes. (Similar to `org.junit.Assert.assertThat` is replaced by `org.hamcrest.MatcherAssert.assertThat` )

linghengqian:
The logback and postgresql versions have been bumped up, no more changes needed.

pjfanning:
@linghengqian I removed the mysql jar change because that seems to cause issues with tests - CI build passes now

linghengqian (approved):

terrymanu (approved):

PR #15307, opened 2022-02-09
[DistSQL]Add a check on whether the database is modified when modifying resources

lanchengx:
- Add a check on whether the database is modified when modifying resources

image


RaigorJiang (changes requested):
@lanchengx 
`hostname` and `port` should also be part of the check.

RaigorJiang (approved):

PR #15354, opened 2022-02-11
Optimize for NarayanaConfigurationFileGenerator

jingshanglu:
- optimize narayana config in server.yaml.
- add UT for NarayanaConfigurationFileGenerator.
- use JAXB to process narayana config file


shardingsphere-kernel/shardingsphere-transaction/shardingsphere-transaction-type/shardingsphere-transaction-xa/shardingsphere-transaction-xa-provider/shardingsphere-transaction-xa-narayana/src/main/java/org/apache/shardingsphere/transaction/xa/narayana/config/Entry.java
```diff
@@ -0,0 +1,43 @@
...
+import java.util.Collection;
+import java.util.LinkedList;
+
+/**
+ * Narayana config entry.
+ */
+@Getter
+@Setter
+@XmlAccessorType(XmlAccessType.FIELD)
+public final class Entry {
```
terrymanu, line 36: Please do not name as `Entry`, which is same with `Map.Entry`

terrymanu (changes requested):

terrymanu (approved):

PR #15365, opened 2022-02-11
Move authentication from AdapterContainer to ShardingSphereProxyContainer

terrymanu:


shardingsphere-infra/shardingsphere-infra-common/src/main/java/org/apache/shardingsphere/infra/metadata/user/yaml/config/YamlUsersConfigurationConverter.java
```diff
@@ -32,47 +32,47 @@
...
-
+    
     private static final UserYamlSwapper USER_YAML_SWAPPER = new UserYamlSwapper();
-
+    
     /**
-     * Convert to users yaml content.
+     * Convert to users YAML content.
      *
      * @param users sharding sphere users
```
taojintianxia, line 41: sharding sphere should be one word

taojintianxia (dismissed):

tristaZero (approved):

PR #15416, opened 2022-02-15
Remove unused property

dachuan9e:
remove unused property


terrymanu (changes requested):
Please global search, there still lots of `proxy-opentracing-enabled` in the source codes

terrymanu (approved):

PR #15491, opened 2022-02-18
support declare for pg/og

tuichenchuxin:
for #14015.

- support declare for pg/og
- add test case


strongduanmu (approved):
@tuichenchuxin Good job.

PR #15565, opened 2022-02-22
For #15541: FinishCheckJob will pass data consistency check when target table is not empty.

### LINKED ISSUE #15541: FinishCheckJob cannot do a completion check if the scaling job is incorrect
yes
In the current case, FinishCheckJob is a separate program running, so it will still check the scaling state when the Scaling job fails, and we want to skip the completion check when the Scaling job fails

ReyYang:
Fixes #15541.

- FinishedCheckJob add conditional to pass data consistency check

### DIFF
--- a/shardingsphere-kernel/shardingsphere-data-pipeline/shardingsphere-data-pipeline-core/src/main/java/org/apache/shardingsphere/data/pipeline/core/job/FinishedCheckJob.java
+++ b/shardingsphere-kernel/shardingsphere-data-pipeline/shardingsphere-data-pipeline-core/src/main/java/org/apache/shardingsphere/data/pipeline/core/job/FinishedCheckJob.java
@@ -51,7 +51,7 @@ public final class FinishedCheckJob implements SimpleJob {
                 continue;
             }
             String jobId = jobInfo.getJobId();
-            if (checkAllowDataCheckJobStatus(jobId)) {
+            if (isAllowDataCheckJobStatus(jobId)) {
                 continue;
             }
             try {
@@ -98,11 +98,11 @@ public final class FinishedCheckJob implements SimpleJob {
         }
     }
     
-    private boolean checkAllowDataCheckJobStatus(final String jobId) {
+    private boolean isAllowDataCheckJobStatus(final String jobId) {
         Map<Integer, JobProgress> jobProgressMap = ruleAlteredJobAPI.getProgress(jobId);
         boolean flag = false;
         for (JobProgress each : jobProgressMap.values()) {
-            if (each == null || !JobStatus.EXECUTE_INCREMENTAL_TASK.equals(each.getStatus())) {
+            if (null == each || !JobStatus.EXECUTE_INCREMENTAL_TASK.equals(each.getStatus())) {
                 flag = true;
                 break;
             }
--- a/shardingsphere-kernel/shardingsphere-data-pipeline/shardingsphere-data-pipeline-core/src/main/java/org/apache/shardingsphere/data/pipeline/core/job/FinishedCheckJob.java
+++ b/shardingsphere-kernel/shardingsphere-data-pipeline/shardingsphere-data-pipeline-core/src/main/java/org/apache/shardingsphere/data/pipeline/core/job/FinishedCheckJob.java
@@ -51,7 +51,7 @@ public final class FinishedCheckJob implements SimpleJob {
                 continue;
             }
             String jobId = jobInfo.getJobId();
-            if (checkJobStatus(jobId)) {
+            if (checkAllowDataCheckJobStatus(jobId)) {
                 continue;
             }
             try {
@@ -98,14 +98,11 @@ public final class FinishedCheckJob implements SimpleJob {
         }
     }
     
-    private boolean checkJobStatus(final String jobId) {
+    private boolean checkAllowDataCheckJobStatus(final String jobId) {
         Map<Integer, JobProgress> jobProgressMap = ruleAlteredJobAPI.getProgress(jobId);
         boolean flag = false;
         for (JobProgress each : jobProgressMap.values()) {
-            if (each == null) {
-                continue;
-            }
-            if (!JobStatus.EXECUTE_INCREMENTAL_TASK.equals(each.getStatus())) {
+            if (each == null || !JobStatus.EXECUTE_INCREMENTAL_TASK.equals(each.getStatus())) {
                 flag = true;
                 break;
             }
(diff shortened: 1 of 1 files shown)


ReyYang:
Preparation phase failure task would have been skipped, may affect performance

shardingsphere-kernel/shardingsphere-data-pipeline/shardingsphere-data-pipeline-core/src/main/java/org/apache/shardingsphere/data/pipeline/core/job/FinishedCheckJob.java
```diff
@@ -93,6 +99,19 @@ public void execute(final ShardingContext shardingContext) {
         }
     }
     
+    private boolean checkJobStatus(final String jobId) {
+        Map<Integer, JobProgress> jobProgressMap = ruleAlteredJobAPI.getProgress(jobId);
+        final AtomicBoolean flag = new AtomicBoolean(false);
```
LeeGuoPing, line 104: I think this place can use `boolean flag = false`;

shardingsphere-kernel/shardingsphere-data-pipeline/shardingsphere-data-pipeline-core/src/main/java/org/apache/shardingsphere/data/pipeline/core/job/FinishedCheckJob.java
```diff
@@ -93,6 +99,19 @@ public void execute(final ShardingContext shardingContext) {
...
+        for (JobProgress each : jobProgressMap.values()) {
+            if (JobStatus.FINISHED.equals(each.getStatus()) || JobStatus.PREPARING_FAILURE.equals(each.getStatus())
+                    || JobStatus.EXECUTE_INVENTORY_TASK_FAILURE.equals(each.getStatus()) || JobStatus.EXECUTE_INCREMENTAL_TASK_FAILURE.equals(each.getStatus())) {
+                flag.set(true);
+                break;
+            }
+        }
```
sandynz, line 111: Looks it could be simplified. All sharding items' progress statuses are `EXECUTE_INCREMENTAL_TASK`, then it could do data consistency check, else not.

sandynz (changes requested):

shardingsphere-kernel/shardingsphere-data-pipeline/shardingsphere-data-pipeline-core/src/main/java/org/apache/shardingsphere/data/pipeline/core/job/FinishedCheckJob.java
sandynz: And `JobProgress each` might be null.

ReyYang: - The judgment here should include all jobStatus because it is possible to get an exception before incrementalTask.
- JobProgress has added judgment NULL

shardingsphere-kernel/shardingsphere-data-pipeline/shardingsphere-data-pipeline-core/src/main/java/org/apache/shardingsphere/data/pipeline/core/job/FinishedCheckJob.java
```diff
@@ -101,15 +100,19 @@ public void execute(final ShardingContext shardingContext) {
...
+                continue;
+            }
             if (JobStatus.FINISHED.equals(each.getStatus()) || JobStatus.PREPARING_FAILURE.equals(each.getStatus())
-                    || JobStatus.EXECUTE_INVENTORY_TASK_FAILURE.equals(each.getStatus()) || JobStatus.EXECUTE_INCREMENTAL_TASK_FAILURE.equals(each.getStatus())) {
-                flag.set(true);
+                    || JobStatus.EXECUTE_INVENTORY_TASK_FAILURE.equals(each.getStatus())
+                    || JobStatus.EXECUTE_INCREMENTAL_TASK_FAILURE.equals(each.getStatus())) {
+                flag = true;
                 break;
             }
```
sandynz, line 113: It's better to depend on less status as previous comment discussed. Not only simplify code, but also if there's new status added, then it's not necessary to change code here.

`jobProgressMap` for each could be used to reduce code lines.

sandynz (changes requested):

shardingsphere-kernel/shardingsphere-data-pipeline/shardingsphere-data-pipeline-core/src/main/java/org/apache/shardingsphere/data/pipeline/core/job/FinishedCheckJob.java
```diff
@@ -49,6 +51,9 @@ public void execute(final ShardingContext shardingContext) {
                 continue;
             }
             String jobId = jobInfo.getJobId();
+            if (checkJobStatus(jobId)) {
+                continue;
+            }
```
sandynz, line 56: The method name `checkJobStatus` could be more meaningful.

sandynz (changes requested):

shardingsphere-kernel/shardingsphere-data-pipeline/shardingsphere-data-pipeline-core/src/main/java/org/apache/shardingsphere/data/pipeline/core/job/FinishedCheckJob.java
```diff
@@ -51,7 +51,7 @@ public void execute(final ShardingContext shardingContext) {
                 continue;
             }
             String jobId = jobInfo.getJobId();
-            if (checkJobStatus(jobId)) {
+            if (checkAllowDataCheckJobStatus(jobId)) {
                 continue;
             }
```
sandynz, line 56: Looks `checkAllowDataCheckJobStatus` means not allow, since there's `continue;`.

Use `isXxx` style could be more certain.

shardingsphere-kernel/shardingsphere-data-pipeline/shardingsphere-data-pipeline-core/src/main/java/org/apache/shardingsphere/data/pipeline/core/job/FinishedCheckJob.java
```diff
@@ -98,14 +98,11 @@ public void execute(final ShardingContext shardingContext) {
...
         Map<Integer, JobProgress> jobProgressMap = ruleAlteredJobAPI.getProgress(jobId);
         boolean flag = false;
         for (JobProgress each : jobProgressMap.values()) {
-            if (each == null) {
-                continue;
-            }
-            if (!JobStatus.EXECUTE_INCREMENTAL_TASK.equals(each.getStatus())) {
+            if (each == null || !JobStatus.EXECUTE_INCREMENTAL_TASK.equals(each.getStatus())) {
                 flag = true;
                 break;
```
sandynz, line 107: `each == null` should be `null == each`.

sandynz (changes requested):

sandynz (approved):

PR #15577, opened 2022-02-23
Reduce allocations in SubqueryExtractUtil

TeslaCN:
Related to #10626.

image


strongduanmu (approved):
@TeslaCN Thank you, maybe there are other similar utility classes that need to be optimized.

TeslaCN:
> @TeslaCN Thank you, maybe there are other similar utility classes that need to be optimized.

I'm working on it.

Issue #13317, opened 2021-10-27
Charset in ShardingSphere Proxy is determined by environment


TeslaCN:
## Feature Request

Charset of ShardingSphere Proxy is determined by
1. locale of environment
2. The property `file.encoding`

### Is your feature request related to a problem?
If the charset was unset in environment, any non-ASCII characters will always be malformed.

### Describe the feature you would like.

* [x] Parse Charset provides by client in protocol and store in connection scope. #13332 
* [x] Intercept SQL like `set names utf8` or `set client_encoding to 'UTF8'` and set the charset to current channel attribute. Duplicated with #12245

soulasuna:
I will try second task

TeslaCN:
Setting charset is available.

PR #15603, opened 2022-02-24
Set charset simple query supported in proxy

soulasuna:
## Set charset simple query supported in proxy

Related to #13317.

- Set charset simple query supported for PostgreSQL by `SET CLIENT_ENCODING TO 'charsetName';`.
- Set charset simple query supported for MySQL by `SET NAMES 'charsetName';`.


shardingsphere-proxy/shardingsphere-proxy-frontend/shardingsphere-proxy-frontend-mysql/src/main/java/org/apache/shardingsphere/proxy/frontend/mysql/command/query/text/query/MySQLComQueryPacketExecutor.java
```diff
@@ -71,10 +80,19 @@ public MySQLComQueryPacketExecutor(final MySQLComQueryPacket packet, final Conne
...
     private Collection<DatabasePacket<?>> processUpdate(final UpdateResponseHeader updateResponseHeader) {
-        responseType = ResponseType.UPDATE;
         return ResponsePacketBuilder.buildUpdateResponsePackets(updateResponseHeader);
     }
     
+    private Collection<DatabasePacket<?>> processClientEncoding(final ClientEncodingResponseHeader clientEncodingResponseHeader) {
+        String value = clientEncodingResponseHeader.getValue();
+        Optional<Charset> charset = MySQLCharacterSet.findByValue(value);
+        if (charset.isPresent()) {
+            clientEncodingResponseHeader.getConnectionSession().getAttributeMap().attr(CommonConstants.CHARSET_ATTRIBUTE_KEY).set(charset.get());
```
TeslaCN, line 90: There is another value `MySQLConstants.MYSQL_CHARACTER_SET_ATTRIBUTE_KEY` which represented MySQL charset and collation is stored in channel attributes map. What if the charset changed to `GBK` while the collation are still `utf8mb4`? The wrong collation id may be returned to client.

TeslaCN (changes requested):

shardingsphere-proxy/shardingsphere-proxy-frontend/shardingsphere-proxy-frontend-mysql/src/main/java/org/apache/shardingsphere/proxy/frontend/mysql/command/query/text/query/MySQLComQueryPacketExecutor.java
soulasuna: When using MySQL set character encoding not just for interacting with the client-side protocol. 
The storage of data will also changed.
I did test. The performance of proxy is consistent with MySQL.

shardingsphere-proxy/shardingsphere-proxy-backend/src/main/java/org/apache/shardingsphere/proxy/backend/text/encoding/SetClientEncodingBackendHandler.java
```diff
@@ -0,0 +1,69 @@
...
+import org.apache.shardingsphere.sql.parser.sql.common.statement.dal.SetStatement;
+
+import java.sql.SQLException;
+import java.util.Iterator;
+
+/**
+ * Set client encoding backend handler.
+ */
+@RequiredArgsConstructor
+public class SetClientEncodingBackendHandler implements TextProtocolBackendHandler {
```
TeslaCN, line 35: Consider refactoring this by `DatabaseAdminExecutor` in future.

TeslaCN (approved):

shardingsphere-proxy/shardingsphere-proxy-backend/src/main/java/org/apache/shardingsphere/proxy/backend/text/encoding/SetClientEncodingBackendHandler.java
soulasuna: OK. Only the Simple Query Protocol is currently supported.
Will be refactored to support the JDBC extension protocol.

PR #15608, opened 2022-02-24
Remove useless worker-id props in snowflake

strongduanmu:
- Remove useless worker-id props in snowflake


totalo:
I think maybe we have to support workid setting, if we want to set workid how should we set it?

strongduanmu:
> 

@totalo The shown setting workerId will cause problems in Cluster mode, so the snowflake algorithm has built-in workerId generation logic now.

totalo:
> 

Get it,thx~

totalo:
> > 
> 
> @totalo The shown setting workerId will cause problems in Cluster mode, so the snowflake algorithm has built-in workerId generation logic now.

Get it,thx~

totalo (approved):

PR #15617, opened 2022-02-24
`drop resource` syntax adds `if exist` keyword.

lanchengx:
For https://github.com/apache/shardingsphere/pull/15617

- Add `if exist` keyword

image
image


shardingsphere-distsql/shardingsphere-distsql-parser/src/main/antlr4/imports/Keyword.g4
```diff
@@ -215,6 +215,14 @@ CONCURRENCY_LEVEL
     : C O N C U R R E N C Y UL_ L E V E L
     ;
 
+IF
```
RaigorJiang, line 218: duplicated IF & EXISTS in file.

RaigorJiang (approved):

Issue #15682, opened 2022-03-01
Some sql for encrypt is not supported


tuichenchuxin:
### Which version of ShardingSphere did you use?
master - 2e43bd09bdae7625aad4eb61f4dfaf0086db9473

### Which project did you use? ShardingSphere-JDBC or ShardingSphere-Proxy?
proxy
### Expected behavior
support
### Actual behavior
not support
### Steps to reproduce the behavior, such as: SQL to execute, sharding rule configuration, when exception occur etc.
- proxy add encrypt althgrim
```
public final class QueryAssistedEncryptAlgorithmFixture implements QueryAssistedEncryptAlgorithm<Object, String> {
    
    @Override
    public void init() {
    }
    
    @Override
    public String encrypt(final Object plainValue, final EncryptContext encryptContext) {
        if (null == plainValue) {
            return null;
        }
        return DigestUtils.md5Hex(String.valueOf(plainValue));
    }
    
    @Override
    public Object decrypt(final String cipherValue, final EncryptContext encryptContext) {
        return cipherValue;
    }
    
    @Override
    public String queryAssistedEncrypt(final Object plainValue, final EncryptContext encryptContext) {
        if (null == plainValue) {
            return null;
        }
        return DigestUtils.md5Hex(String.valueOf(plainValue));
    }
    
    @Override
    public String getType() {
        return "ASSISTED_QUERY_ENCRYPT";
    }
}
```
- config
```
schemaName: encrypt_db

dataSources:
  ds_0:
    url: jdbc:mysql://127.0.0.1:3306/demo_ds_0?serverTimezone=UTC&useSSL=false
    username: root
    password: 123456
    connectionTimeoutMilliseconds: 30000
    idleTimeoutMilliseconds: 60000
    maxLifetimeMilliseconds: 1800000
    maxPoolSize: 50
    minPoolSize: 1
#  ds_1:
#    url: jdbc:mysql://127.0.0.1:3306/demo_ds_1?serverTimezone=UTC&useSSL=false
#    username: root
#    password:
#    connectionTimeoutMilliseconds: 30000
#    idleTimeoutMilliseconds: 60000
#    maxLifetimeMilliseconds: 1800000
#    maxPoolSize: 50
#    minPoolSize: 1

rules:
- !ENCRYPT
  encryptors:
    aes_encryptor:
      type: AES
      props:
        aes-key-value: 123456abc
    md5_encryptor:
      type: MD5
    assisted_query:
[...]

PR #15691, opened 2022-03-01
support select where sub query encrypt

tuichenchuxin:
for #15682.

- support `Select amount from t_account_bak where (select amount from t_account_bak) in ('100');`


strongduanmu (changes requested):
@tuichenchuxin Can we add performance test for this logic change?

tuichenchuxin:
`Select amount from t_account_bak where (select amount from t_account_bak) in ('100');`
This sql seems to be written to this way
`Select amount from t_account_bak where (select amount from t_account_bak where amount in ('100')) ;`
So I close this.

PR #15753, opened 2022-03-02
Set charset extended protocol supported in proxy

soulasuna:
## Set charset extended protocol supported in proxy

Related to #13317.

- Set charset extended protocol supported for PostgreSQL.
- Refactor Set charset backend handler.


shardingsphere-proxy/shardingsphere-proxy-backend/src/main/java/org/apache/shardingsphere/proxy/backend/response/header/update/ClientEncodingResponseHeader.java
```diff
@@ -20,18 +20,26 @@
...
+    private final String inputValue;
     
-    private final ConnectionSession connectionSession;
+    /**
+     * Get current charset value.
+     *
+     * @return current charset value
+     */
+    public Optional<String> getCurrentCharsetValue() {
+        return null == currentCharsetValue ? Optional.empty() : Optional.of(currentCharsetValue);
```
TeslaCN, line 43:
```suggestion
        return Optional.ofNullable(currentCharsetValue);
```

TeslaCN (approved):

PR #15776, opened 2022-03-03
Fix NumberFormatException when sharding algorithm config number props

### LINKED ISSUE #15769: HashModShardingAlgorithm getShardingCount() throw a NumberFormatException
## Bug Report

5.1.0
ShardingSphere-JDBC
### Expected behavior
HashModShardingAlgorithm can correct parsing of the sharding-count property
### Actual behavior
throw a java.lang.NumberFormatException: null
JDK1.8 Properties.getProperty(String key) only return value of type string, otherwise return null 

HashModShardingAlgorithm
image
Properties
image

image

image

if i config like this, it can work correctly

image

strongduanmu:
Fixes #15769.

- fix NumberFormatException when sharding algorithm config number props
- update unit test


qiaoheng110 (commented):
you handled it really fast

RaigorJiang (approved):

Issue #10641, opened 2021-06-03
addShutdownHook in ShardingExecutorService


whojes:
## Question

Hi there, i have a question. 

I'm currently using shardingsphere for a project with spring boot framework.
When i create sharding datasource, it registers a shutdown hook while creating its stuffs; ([code link](https://github.com/apache/shardingsphere/blob/master/shardingsphere-infra/shardingsphere-infra-executor/src/main/java/org/apache/shardingsphere/infra/executor/kernel/thread/ExecutorServiceManager.java#L47))

i'm using my sharding datasource in middle of business code, and i'm hoping that `ShardingExecutorService` would be destroyed after related beans stop working and has been released, according to spring's shutdown process. But the library registers shutdown hook which destroys a thread pool regardless of spring context.

i think `ExecutorServiceManager` should not register the shutdown hook, implement Closeable, and add destroy process in `close()` method. Is there a  reason for register the shutdown hook to Runtime?

terrymanu:
Is anybody interest this issue?

terrymanu:
Thanks for @cheese8 to fix this issue.

PR #15868, opened 2022-03-08
Remove useless ShutdownHook

cheese8:
Ref https://github.com/apache/shardingsphere/issues/10641

-try to enhance it, please figure it out any problems

### DIFF
--- a/shardingsphere-proxy/shardingsphere-proxy-frontend/shardingsphere-proxy-frontend-core/src/main/java/org/apache/shardingsphere/proxy/frontend/executor/UserExecutorGroup.java
+++ b/shardingsphere-proxy/shardingsphere-proxy-frontend/shardingsphere-proxy-frontend-core/src/main/java/org/apache/shardingsphere/proxy/frontend/executor/UserExecutorGroup.java
@@ -21,8 +21,6 @@ import com.google.common.util.concurrent.ListeningExecutorService;
 import lombok.Getter;
 import org.apache.shardingsphere.infra.executor.kernel.thread.ExecutorServiceManager;
 
-import java.util.concurrent.TimeUnit;
-
 /**
  * User executor group.
  */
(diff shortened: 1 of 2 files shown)


terrymanu:
What is the usage to remove `SHUTDOWN_EXECUTOR`?

cheese8:
> What is the usage to remove `SHUTDOWN_EXECUTOR`?

The original usage is register a delayed shutdown hook, while closing `ExecutorEngine` which cause by `AutoCloseable` using another thread `SHUTDOWN_EXECUTOR` to shutdown the `executorService` asynchronously.

I think it should be break by while closing `ExecutorEngine` and `UserExecutorGroup` which cause by `AutoCloseable` shutdown the `executorService` directly.

shardingsphere-proxy/shardingsphere-proxy-frontend/shardingsphere-proxy-frontend-core/src/main/java/org/apache/shardingsphere/proxy/frontend/executor/UserExecutorGroup.java
```diff
@@ -46,4 +48,21 @@ private UserExecutorGroup() {
...
+        if (null != executorService) {
+            try {
+                executorService.shutdown();
+                while (!executorService.awaitTermination(5, TimeUnit.SECONDS)) {
+                    executorService.shutdownNow();
+                }
+            } catch (final InterruptedException ex) {
+                Thread.currentThread().interrupt();
+            }
+        }
```
terrymanu, line 66: UserExecutorGroup is for proxy only, spring boot maybe not need change

terrymanu (changes requested):

shardingsphere-proxy/shardingsphere-proxy-frontend/shardingsphere-proxy-frontend-core/src/main/java/org/apache/shardingsphere/proxy/frontend/executor/UserExecutorGroup.java
cheese8: ok, revert it.

terrymanu (approved):

terrymanu:
We are talking about in PR

Issue #15722, opened 2022-03-01
FixBug: testcases override in public final class SQLRewriteEngineTestParametersBuilder.loadAllRewriteAssertionsRootEntities


cheese8:
## Bug Report

Testcases in shardingsphere-rewrite-test would be overrided with the same filename.xml

PR #15894, opened 2022-03-08
add / to show correct xml path in testcases

cheese8:
Ref #15722

The original fix miss the `/` in xml path as below

-add / to show correct xml path in testcases


terrymanu:
Could you upload the pic again, it does not display

cheese8:
> Could you upload the pic again, it does not display

Yes, reupload it.

terrymanu (approved):

PR #15957, opened 2022-03-10
Restart TransactionContext when alter transaction rule in narayana xa mode

### LINKED ISSUE #15405: Enhancements to proxy integration narayana
## Feature Request

If no response anymore and we cannot make decision by current information, we will **close it**.

### Background
When using the proxy with narayana, we need to configure the `xaRecoveryNodes` attribute for the proxy (the instanceid of the current proxy is used by default), and this attribute can be modified through distsql. The `xaRecoveryNodes` attribute is instance-level and needs to be bound to the instance

- [x] Store `xaRecoveryNodes` at the instance level  @menghaoranss 
- [x] add Event when `xaRecoveryId` changed @menghaoranss 
- [x] use distsql to show all instanceid @RaigorJiang 
- [x] use distsql to modify `xaRecoveryId` @RaigorJiang 
- [x] Ensure that `TM` is properly reinitialized after `xaRecoveryId` is modified @jingshanglu 
- [x] suppprt default jdbc recovery store for cluster mod. @jingshanglu

jingshanglu:
ref #15405 .

- Change to `for` from `stream` in fun `findRuleConfiguration`
- Restart transaction context after change transaction config

### Test 

- [x] `Memory`,`Standalone`: start a proxy, then use `distsql` to change transaction config, proxy will restart `TransactionContext`
- [x] `Cluster`: no  transaction props, there are `resource`, start proxy, proxy use the first resource as the address for store transaction log, then use `distsql` to change transaction config, proxy will restart `TransactionContext`


shardingsphere-infra/shardingsphere-infra-common/src/main/java/org/apache/shardingsphere/infra/metadata/rule/ShardingSphereRuleMetaData.java
```diff
@@ -64,7 +63,25 @@
...
+    /**
+     * Find single rule configuration by class.
+     *
+     * @param clazz target class
+     * @param <T> type of rule configuration
+     * @return found rule configurations
+     */
+    public <T extends RuleConfiguration> T findSingleRuleConfiguration(final Class<T> clazz) {
+        Collection<T> foundRuleConfiguration = findRuleConfiguration(clazz);
+        return foundRuleConfiguration.isEmpty() ? null : foundRuleConfiguration.iterator().next();
```
terrymanu, line 84: Please do not return null, could you use optional if necessary?

shardingsphere-kernel/shardingsphere-transaction/shardingsphere-transaction-api/src/main/java/org/apache/shardingsphere/transaction/config/TransactionRuleConfiguration.java
```diff
@@ -36,4 +37,18 @@
...
+     *
+     * @param transactionRuleConfiguration transaction rule configuration
+     * @return return true if the two transactionRuleConfiguration are the same
+     */
+    public boolean compare(final TransactionRuleConfiguration transactionRuleConfiguration) {
+        if (null == transactionRuleConfiguration) {
+            return false;
+        }
+        return Objects.equals(defaultType, transactionRuleConfiguration.defaultType) && Objects.equals(providerType, transactionRuleConfiguration.providerType)
+                && Objects.equals(props, transactionRuleConfiguration.props);
```
terrymanu, line 52: the props are synchronized and may caused by performance issue.
Is the method `compare` call frequently?

shardingsphere-mode/shardingsphere-mode-type/shardingsphere-cluster-mode/shardingsphere-cluster-mode-core/src/main/java/org/apache/shardingsphere/mode/manager/cluster/ClusterContextManagerBuilder.java
```diff
@@ -83,8 +84,14 @@ private Properties getTransactionProperties(final MetaDataContexts metaDataConte
                 ? TransactionConfigurationFileGeneratorFactory.newInstance(transactionRule.get().getProviderType()) : Optional.empty();
         if (schemaName.isPresent() && fileGenerator.isPresent()) {
```
terrymanu, line 85: please revert the condition judge, and return `new Properties()` firstly.

terrymanu (changes requested):

shardingsphere-kernel/shardingsphere-transaction/shardingsphere-transaction-api/src/main/java/org/apache/shardingsphere/transaction/config/TransactionRuleConfiguration.java
jingshanglu: It is just called when alter global rule.

shardingsphere-mode/shardingsphere-mode-type/shardingsphere-cluster-mode/shardingsphere-cluster-mode-core/src/main/java/org/apache/shardingsphere/mode/manager/cluster/ClusterContextManagerBuilder.java
jingshanglu: I don't change the condition judge, just fill the props to `MetaDataContexts`.

terrymanu (approved):

PR #15973, opened 2022-03-10
Optimized read-write splitting syntax check

lanchengx:
- When dropping a rule, add a check on the rule in use
- when adding a rule, add a check for duplicate resource names


RaigorJiang (approved):

RaigorJiang:
Good job!  @lanchengx

Issue #14670, opened 2022-01-10
insert into broadcasttable with longblob cause longblob break up to serveral part(the number of shards )  


ziliangzeng:
## Bug Report
      hi.
      example: i have 4 database . the broadcasttable was in 4 database.
      when I try to insert into broadcastTable with longblob. it inserts success.
      but interesting thing is that It segmentation 4 parts. each database have 1/4 longblob.
      in my mind . each data should have complete longblob .
      that was confuse.
      when I change the broadcasttable to sharding table .  the longblob do not spilt and successful insert. but sharding table is not the table type I want.
       
 broadcasttable:      
![1771kb](https://user-images.githubusercontent.com/27884890/148759782-0f260e13-c125-4ff6-ab7b-51ab67234f66.png)
![800kb](https://user-images.githubusercontent.com/27884890/148759795-c78a04e9-0b88-40e0-9b59-2c3121be9fc9.png)
sharding table:
![2171kb](https://user-images.githubusercontent.com/27884890/148759807-68664398-7b18-4f47-abd0-fb014bb5547d.png)

### Which version of ShardingSphere did you use?
       5.0.0
### Which project did you use? ShardingSphere-JDBC or ShardingSphere-Proxy?
         sharding - jdbc
### Expected behavior
        long blob insert into each broadcasttable success
### Actual behavior
       longblob was break up and insert into table

strongduanmu:
@ziliangzeng Thank you for your feedback, I will track this issue.

strongduanmu:
Hi @ziliangzeng, can you provide sql-show log? When ShardingSphere handle broadcast tables, there is no logic to split strings.

ziliangzeng:
@strongduanmu  hi . is it what you want ? 
![image](https://user-images.githubusercontent.com/27884890/148865656-a80310c6-ebba-4c79-89f0-9252b24f3cdf.png)

strongduanmu:
Hi @ziliangzeng, can you upload this demo? I can't reproduce this bug in master branch.

ziliangzeng:
```
dataSources:
  ds_0:
    dataSourceClassName: com.zaxxer.hikari.HikariDataSource
    driverClassName: com.mysql.cj.jdbc.Driver
    jdbcUrl: jdbc:mysql://localhost:3306/ds0?serverTimezone=UTC&useSSL=false&useUnicode=true&characterEncoding=UTF-8&allowPublicKeyRetrieval=true
    username: root
    password: 123456
[...]

PR #16083, opened 2022-03-15
Convert stream to reference data type（#16080）

### LINKED ISSUE #16080: Modifying the database lob column will report errors and steam cannot be read repeatedly
3.1.0
ShardingSphere-JDBC
### Actual behavior
When using sharding to connect to Oracle database, if the database field is CLOB or blob, the following error will be reported:

Error updating database. Cause: java. sql. SQLException: Invalid column type
The error may involve defaultParameterMap
The error occurred while setting parameters
 SQL: UPDATE BC_ CARD_ PROFILE_ RULE SET OBJ_ PROLONG_ RULE = ? WHERE CARD_ PROFILE_ RULE_ ID = ?
Cause: java. sql. SQLException: Invalid column type
; uncategorized SQLException; SQL state [99999]; error code [17004]; Invalid column type; nested exception is java. sql. SQLException: Invalid column type
In addition, when the table is a sharding table or broadcast table, if the CLOB field is updated on multiple tables, only one table can read the updated value, because the CLOB will be set with stringreader type in the code. After reading the first time, the index reaches the end of the stream, and the value cannot be read repeatedly
### Solution
Convert stream to reference data type

iwhalecloud-platform:
Fixes #16080 #14670.

-
-
-

### DIFF
--- a/shardingsphere-jdbc/shardingsphere-jdbc-core/src/main/java/org/apache/shardingsphere/driver/jdbc/adapter/AbstractPreparedStatementAdapter.java
+++ b/shardingsphere-jdbc/shardingsphere-jdbc-core/src/main/java/org/apache/shardingsphere/driver/jdbc/adapter/AbstractPreparedStatementAdapter.java
@@ -19,6 +19,7 @@ package org.apache.shardingsphere.driver.jdbc.adapter;
 
 import com.google.common.io.CharStreams;
 import lombok.Getter;
+import org.apache.commons.io.IOUtils;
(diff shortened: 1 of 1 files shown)


TeslaCN:
Hi @iwhalecloud-platform 
The dependency commons-io has been removed from ShardingSphere.

tuichenchuxin:
@iwhalecloud-platform If we copy the stream, we will lose the advantage of using stream.

terrymanu:
This pull request is long time ago, I just close it.

PR #16227, opened 2022-03-20
Supplementary encryption function specification

acmenlt:
- Supplementary encryption function specification


docs/document/content/features/encrypt/use-norms.cn.md
```diff
@@ -11,5 +11,6 @@ weight = 2
 ## 不支持项
 
 * 需自行处理数据库中原始的存量数据；
+* 加密字段无法支持查询不区分大小写功能；
```
strongduanmu, line 14: @acmenlt I think this is a bug that needs to be fixed.

strongduanmu (changes requested):

docs/document/content/features/encrypt/use-norms.cn.md
acmenlt: Mysql database queries are case insensitive by default. Case sensitive after encryption. I don't think this is a problem. It's correct.

strongduanmu: > Mysql database queries are case insensitive by default. Case sensitive after encryption. I don't think this is a problem. It's correct.

I got it.

strongduanmu (approved):

Issue #16234, opened 2022-03-21
[New Feature] Support system database and schema query statements for MySQL and PostgreSQL


strongduanmu:
## Feature Request

According to the feedback from the community, ShardingSphere currently does not support MySQL and PostgreSQL system databases and schema query statements. When the user executes the following statement, an exception will occur.

```sql
select count(*) from information_schema.TABLES where table_name = #{tableName}
# org.apache.shardingsphere.infra.exception.SchemaNotExistedException: Schema 'information_schema' doesn't exist.
```

### Is your feature request related to a problem?

#15327 #11994 #14728 #14347 #13586 #11167

### Describe the feature you would like.

In order to provide support for the MySQL and PostgreSQL system database/schema, we need to adjust the structure of the metadata first, so that the metadata can support the three-tier structure of `database -> schema -> table`.

- [x] adjust the structure of metadata in zookeeper
- [x] adjust metadata initialization logic and kernel usage metadata logic
- [x] initialize calcite metadata with new metadata

New metadata structure:

```
|----metadata
    |----sharding_db
        |----active_version: 0
        |----versions
            |----0
                |----dataSources
                |----rules
        |----schemas
            |----information_schema
                |----tables
                    |----views
            |----pg_catalog
            |----public
                |----tables
                    |----t_order
```

New ShardingSphereMetaData java class:

```java
/**
 * ShardingSphere meta data.
 */
@RequiredArgsConstructor
@Getter
public final class ShardingSphereMetaData {
    
    private final String name;
    
    private final ShardingSphereResource resource;
    
    private final ShardingSphereRuleMetaData ruleMetaData;
    
    private final Map<String, ShardingSphereSchema> schemas;
}
```

After completing the metadata structure adjustment, we need to optimize metadata initialization logic and manually build system databases and schemas for different database dialects. The overall process of initialization is as follows:

![image](https://user-images.githubusercontent.com/10829171/159218342-3633222a-1cfb-42f0-acd7-d242328a451e.png)

Then adjust the execution process, when the query statement contains system databases/tables, we use federation execution engine to execute. 

![image](https://user-images.githubusercontent.com/10829171/159218388-74f08372-a311-4b5e-9fdb-7798b8cfd39f.png)

It is worth noting that we will first return an empty system database/table result set, and then consider using elastic-job to regularly maintain system database/table information to zookeeper after all these tasks are completed.

- [x] optimize metadata initialization logic
- [x] manually build system databases and schemas for different database dialects
[...]

PR #16340, opened 2022-03-24
Adjust Goverence center structure to support three layers of metadata structure

zjcnb:
For #16234


shardingsphere-mode/shardingsphere-mode-type/shardingsphere-cluster-mode/shardingsphere-cluster-mode-core/src/main/java/org/apache/shardingsphere/mode/manager/cluster/ClusterContextManagerBuilder.java
```diff
@@ -149,10 +147,10 @@ private SchemaConfiguration createSchemaConfiguration(final String schemaName, f
...
     }
     
-    private void persistMetaData(final MetaDataPersistService metaDataPersistService, final Map<String, ShardingSphereSchema> schemaMap) {
-        for (Entry<String, ShardingSphereSchema> entry : schemaMap.entrySet()) {
-            metaDataPersistService.getSchemaMetaDataService().persist(entry.getKey(), entry.getValue());
-        }
+    private void persistMetaData(final MetaDataPersistService metaDataPersistService, final Map<String, ShardingSphereDatabase> databaseMap) {
+        databaseMap.entrySet().forEach(entry ->
+                entry.getValue().getSchemas().entrySet().forEach(schema ->
+                        metaDataPersistService.getSchemaMetaDataService().persist(schema.getKey(), schema.getValue())));
```
taojintianxia, line 153: invoke an interface in a loop, why not design another signature which support Map from very beginning

zjcnb: I want to get value by ZooKeeper api, but it 's not supprot batch query.

terrymanu (approved):

PR #16414, opened 2022-03-27
Workflow on Windows and MacOS scheduled daily

sandynz:
Purpose:
- Currently, CI has several jobs, include Linux-JDK8/Linux-JDK11/Linux-JDK17, windows-JDK8, macos-JDK8. windows-JDK8 and macos-JDK8 are the slowest jobs. Since ShardingSphere run on Linux at the most time, we'll test several JDK versions on Linux, and there should be almost no failure on Windows and MacOS when it pass on Linux. If we run workflow on Windows and MacOS daily but not every PR or push, it could speedup CI.

- Workflow on Windows and MacOS scheduled daily. 1) update concurrency group, since push to master and schedule both will trigger CI on master branch, 2) add condition for `windows` and `macos` jobs, just trigger CI when `github.event_name` is `schedule`.

It might need some discussion about whether need to do it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -121,7 +121,7 @@ jobs:
   
   macos:
     if: (github.event_name == 'schedule')
-    name: JDK ${{ matrix.java.version }} - on ${{ matrix.os }}
+    name: macos
     runs-on: ${{ matrix.os }}
     timeout-minutes: 60
     strategy:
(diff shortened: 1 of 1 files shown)


totalo:
image, There may be another problem here.

sandynz:
@totalo thanks, I'll check it.

totalo (approved):

PR #16428, opened 2022-03-28
Add database discovery pictures.

zjcnb:


tuichenchuxin (approved):

LeeGuoPing (commented):
The chinses picture use chinses as far as possible. Do you think so?

image

PR #16484, opened 2022-03-30
Reuse built artifacts and IT image in it.yml

TeslaCN:
This may save about 32 * 5 minutes which were spent on building projects.

Before:
image

After:
image


tristaZero (approved):

taojintianxia:
👍

Issue #16269, opened 2022-03-22
Proxy support schema lock for stop write in cluster mode


soulasuna:
## Feature Request

Proxy support global schema lock in cluster mode

### Describe the feature you would like.

Proxy support schema level stop write lock in cluster mode

TeslaCN:
Hi @soulasuna 
Any update here? Could we finish this in 5.1.1?

soulasuna:
Hi @TeslaCN
There will also be an update on intercepting DistSQL.
Others need to be supplemented by @sandynz.

sandynz:
Hi @TeslaCN , @soulasuna , scaling integration with schema lock is done, this part will be released in 5.1.1

soulasuna:
The basic function of the non-reentrant global distributed lock is completed.
Implemented:
1. Schema-level distributed lock based on zookeeper.
2. Lock and unlock can be performed in different `Proxy` nodes for the same Schema.
3. In the locked state, newly joined cluster `Proxy` nodes can lock the schema synchronously.

PR #16523, opened 2022-03-31
Fix is locked schema NPE

soulasuna:
## Fix: locked schema NPE

Related to #16269.

- Fix: locked schema NPE.

### DIFF
--- a/shardingsphere-mode/shardingsphere-mode-type/shardingsphere-memory-mode/shardingsphere-memory-mode-core/src/main/java/org/apache/shardingsphere/mode/manager/memory/lock/MemoryLockContext.java
+++ b/shardingsphere-mode/shardingsphere-mode-type/shardingsphere-memory-mode/shardingsphere-memory-mode-core/src/main/java/org/apache/shardingsphere/mode/manager/memory/lock/MemoryLockContext.java
@@ -58,6 +58,10 @@ public final class MemoryLockContext implements LockContext {
     
     @Override
     public boolean isLockedSchema(final String schemaName) {
-        return null != schemaName && null != locks.get(schemaName) && locks.get(schemaName).isLocked(schemaName);
+        if (null == schemaName) {
+            return false;
+        }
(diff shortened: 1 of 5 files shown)


shardingsphere-mode/shardingsphere-mode-type/shardingsphere-memory-mode/shardingsphere-memory-mode-core/src/main/java/org/apache/shardingsphere/mode/manager/memory/lock/MemoryLockContext.java
```diff
@@ -56,7 +58,6 @@ public ShardingSphereLock getOrCreateSchemaLock(final String schemaName) {
     
     @Override
     public boolean isLockedSchema(final String schemaName) {
-        ShardingSphereLock shardingSphereLock = locks.get(schemaName);
-        return null != shardingSphereLock && shardingSphereLock.isLocked(schemaName);
+        return null != schemaName && null != locks.get(schemaName) && locks.get(schemaName).isLocked(schemaName);
     }
```
sandynz, line 62: `locks.get(schemaName)` run 2 times, the second time of `locks.get(schemaName)` might return different value as the first result.

sandynz (changes requested):

shardingsphere-mode/shardingsphere-mode-type/shardingsphere-memory-mode/shardingsphere-memory-mode-core/src/main/java/org/apache/shardingsphere/mode/manager/memory/lock/MemoryLockContext.java
soulasuna: I will fix.

sandynz (approved):

sandynz:
Local test is passed.

PR #16678, opened 2022-04-08
Optimize performance, log MDC information only when sql-show is enabled.

RaigorJiang:
Optimize performance, log MDC information only when sql-show is enabled.


shardingsphere-proxy/shardingsphere-proxy-frontend/shardingsphere-proxy-frontend-core/src/main/java/org/apache/shardingsphere/proxy/frontend/command/CommandExecutorTask.java
```diff
@@ -66,8 +69,9 @@
     @Override
     public void run() {
         boolean isNeedFlush = false;
+        boolean sqlShowEnabled = isSQLShowEnabled();
         try (PacketPayload payload = databaseProtocolFrontendEngine.getCodecEngine().createPacketPayload((ByteBuf) message, context.channel().attr(CommonConstants.CHARSET_ATTRIBUTE_KEY).get())) {
-            fillLogMDC();
+            fillLogMDC(sqlShowEnabled);
```
TeslaCN, line 74: Could we consider checking the value before entering the method?

RaigorJiang: Sure, thanks for your advice.  😀

shardingsphere-proxy/shardingsphere-proxy-frontend/shardingsphere-proxy-frontend-core/src/main/java/org/apache/shardingsphere/proxy/frontend/command/CommandExecutorTask.java
```diff
@@ -66,8 +69,11 @@
     @Override
     public void run() {
         boolean isNeedFlush = false;
+        boolean sqlShowEnabled = isSQLShowEnabled();
```
TeslaCN, line 72:
```suggestion
        boolean sqlShowEnabled = ProxyContext.getInstance().getContextManager().getMetaDataContexts().getProps().getValue(ConfigurationPropertyKey.SQL_SHOW);
```
Could we consider doing inline manually?

RaigorJiang: Thank you @TeslaCN , updated.

TeslaCN (approved):

PR #16700, opened 2022-04-09
Unified debug level log adding judgment

longtai-cn:
- Unified debug level log adding judgment.

> Referring to the writing method of other debug level logs in the project, standardize the printing method of debug logs.


terrymanu (changes requested):
Is it effect the performance? maybe we need a performance test report

longtai-cn:
> Is it effect the performance? maybe we need a performance test report

@terrymanu This PR is mainly to unify the printing method of debug level logs, which may not greatly improve the performance

terrymanu:
> > Is it effect the performance? maybe we need a performance test report
> 
> @terrymanu This PR is mainly to unify the printing method of debug level logs, which may not greatly improve the performance

@longtai-cn The original codes look like more clear, is it necessary to add more judgment here?

longtai-cn:
@terrymanu You're right. I should focus on performance. This PR is coming to an end.

terrymanu:
Thank you very much

PR #16766, opened 2022-04-12
Add drop access method for PostgreSQL

Trydamere:
Ref #14015.

- Add `drop access method` for PostgreSQL
- Add test case
- Run Tests


strongduanmu (approved):
@Trydamere Good job, merged.

PR #16847, opened 2022-04-15
Fix the exception that there is no response when the pg client executes DistSQL successfully

lanchengx:
- When executing DistSQL successfully, return SUCCESS

**Before**
image

**After**
image


TeslaCN (approved):
LGTM

RaigorJiang (approved):

PR #16868, opened 2022-04-15
Add sharding algorithm example.

yx9o:
Add Interval Sharding Algorithm example.

2b7a1609e4566fe2ba539792ca362f5
1d2fa6ebf0a04ad5c67f92634dfa20d


examples/example-core/example-api/src/main/java/org/apache/shardingsphere/example/core/api/entity/OrderStatisticsInfo.java
```diff
@@ -0,0 +1,71 @@
...
+    
+    private Long id;
+    
+    private Long userId;
+    
+    private Date orderDate;
+    
+    private int orderNum;
+    
+    public Long getId() {
```
RaigorJiang, line 35: Hi @yx9o , can we use lombok to generate getters and setters?

examples/example-core/example-api/src/main/java/org/apache/shardingsphere/example/core/api/entity/OrderStatisticsInfo.java
```diff
@@ -0,0 +1,71 @@
...
+
+public class OrderStatisticsInfo implements Serializable {
+    
+    private static final long serialVersionUID = -1770007969944794302L;
+    
+    private Long id;
+    
+    private Long userId;
+    
+    private Date orderDate;
```
RaigorJiang, line 31: Since ShardingSphere is already based on Java8, I recommend to use LocalDateTime instead of Date.

yx9o: ok

examples/example-core/example-api/src/main/java/org/apache/shardingsphere/example/core/api/entity/OrderStatisticsInfo.java
yx9o: done

RaigorJiang (approved):

PR #16911, opened 2022-04-19
Add `rebootRequired` to `ConfigurationPropertyKey`

lanchengx:
- Add `requireReboot`


shardingsphere-infra/shardingsphere-infra-common/src/main/java/org/apache/shardingsphere/infra/config/props/ConfigurationPropertyKey.java
```diff
@@ -35,106 +35,108 @@ public enum ConfigurationPropertyKey implements TypedPropertyKey {
...
-    PROXY_MYSQL_DEFAULT_VERSION("proxy-mysql-default-version", "5.7.22", String.class);
+    PROXY_MYSQL_DEFAULT_VERSION("proxy-mysql-default-version", "5.7.22", String.class, false);
     
     private final String key;
     
     private final String defaultValue;
     
     private final Class<?> type;
     
+    private final boolean requireReboot;
```
RaigorJiang, line 138: Hi @lanchengx , 
Do you think `rebootRequired` would be better than  `requireReboot`?

RaigorJiang (approved):

PR #16963, opened 2022-04-20
Replace unnecessary Collectors.toCollection(LinkedList::new) to Collectors.toList()

terrymanu:


taojintianxia (approved):

yx9o (approved):

RaigorJiang:
Hi @lanchengx ,
I think the original `LinkedList` is to preserve the order of elements, could you please take a check?

lanchengx:
> Hi @lanchengx , I think the original `LinkedList` is to preserve the order of elements, could you please take a check?

The order is not important when verifying, but do we need to display it in the order of addition? @terrymanu

PR #17041, opened 2022-04-24
Use lombok to generate getters and setters.

yx9o:
Use lombok to generate getters and setters.


terrymanu (changes requested):
The example should keep simple, it is unnecessary to introduce lombok for enduser.

yx9o:
ok.

PR #17160, opened 2022-04-27
fix the CosIdModShardingAlgorithm lost the mod configuration parameter.

### LINKED ISSUE #17159: The `CosIdModShardingAlgorithm` lost the `mod` configuration parameter.
## Bug Report

v5.1.0

ShardingSphere-JDBC

### Expected behavior

The `CosIdModShardingAlgorithm` do not lost the `mod` configuration parameter.

### Actual behavior

The `CosIdModShardingAlgorithm` lost the `mod` configuration parameter.

When the configuration parameter value is not of type `String`, using `Properties.getProperty` will return null even if `key` exists.

Following is the implementation of `Properties.getProperty` method：

```java
    public String getProperty(String key) {
        Object oval = map.get(key);
        String sval = (oval instanceof String) ? (String)oval : null;  // the problem is here
        Properties defaults;
        return ((sval == null) && ((defaults = this.defaults) != null)) ? defaults.getProperty(key) : sval;
    }
```

Ahoo-Wang:
Fixes #17159.

- fix the `CosIdModShardingAlgorithm` lost the mod configuration parameter.


terrymanu (commented):
It is necessary to set String value to properties.
Do we need to support set other type into properties?

Ahoo-Wang:
> It is necessary to set String value to properties. Do we need to support set other type into properties?

Yes,When configured with spring-boot yaml the value is parsed as an int.

``` yaml
        sharding-algorithms:
          algorithm-name:
            type: COSID_MOD
            props:
              mod: 2
              logic-name-prefix: t_table_
```

terrymanu (approved):

PR #17182, opened 2022-04-28
Use spotless to format POM globally XML file

longtai-cn:
- Use spotless to format POM globally XML file.


pom.xml
```diff
@@ -28,6 +29,25 @@
     <version>5.1.2-SNAPSHOT</version>
     <packaging>pom</packaging>
     <name>${project.artifactId}</name>
+    <description>Build criterion and ecosystem above multi-model databases</description>
+
```
terrymanu, line 33: Could it possible to not change this part which is for description only, but the module and dependencies are most important, it is better to keep them in the front of the pom file

terrymanu (changes requested):

pom.xml
longtai-cn: Sorry, it cannot be modified here.

pom.xml
```diff
@@ -1313,29 +1334,28 @@
...
+                <pluginManagement>
+                    <plugins>
+                        <plugin>
+                            <artifactId>maven-surefire-plugin</artifactId>
+                            <configuration>
+                                <trimStackTrace>false</trimStackTrace>
+                                <excludes>
+                                    <exclude />
+                                </excludes>
+                                <argLine>--add-opens java.base/java.lang=ALL-UNNAMED --add-opens java.base/java.lang.reflect=ALL-UNNAMED</argLine>
```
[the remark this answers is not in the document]
longtai-cn: This configuration involves `shardingsphere-pipeline-test`, MAC + jdk1 8 there is no problem with local compilation.

terrymanu (approved):

Issue #17203, opened 2022-04-29
MySQL create database succeed, but can't use it at JDBC


azexcy:
### Which version of ShardingSphere did you use?

master 3d08a8d176d494239c575b84535d30408cc6b456

### Which project did you use? ShardingSphere-JDBC or ShardingSphere-Proxy?

ShardingSphere-Proxy

### Expected behavior

If create database succeed, can be used right now.

### Actual behavior

I was using ShardingSphere-Proxy with MySQL, only use server.yaml, not use config-sharding.yaml

Firstly, I create database sharding_db;
```
        try (Connection connection = DriverManager.getConnection("jdbc:mysql://localhost:3307/?serverTimezone=UTC&useLocalSessionState=true&characterEncoding=utf-8&rewriteBatchedStatements=true&useSSL=false&useServerPrepStmts=true"), "root", "root")) {
            connection.createStatement().execute("CREATE DATABASE sharding_db");
        }
```

Then I want to create new Connection and use `sharding_db` just created just now, but the exception will thrown.
<img width="805" alt="image" src="https://user-images.githubusercontent.com/101622833/165897838-2e985c9c-af9b-4a8b-8329-123f431dcb66.png">

Only MySQL has the problem, PostgreSQL I tested, it's ok. And after i execute `ADD RESOURCE xxx`, the problem will not appear.

RaigorJiang:
Hi @azexcy 
Thanks for your feedback.
@lanchengx  could you help investigate?

lanchengx:
@azexcy @RaigorJiang Thanks, I will check

lanchengx:
Hi @azexcy, you can try to use the latest code from master, it works.
<img width="1717" alt="image" src="https://user-images.githubusercontent.com/52209337/167583507-816b0f3c-cd25-4511-a3aa-a11a323b6b1b.png">

azexcy:
Ok, I will try later, Thanks

PR #17611, opened 2022-05-12
Fix the exception when specifying schema when JDBC connects to Proxy

lanchengx:
Fixes https://github.com/apache/shardingsphere/issues/17203

- Remove the judgment that the schema is not used


RaigorJiang (approved):

shardingsphere-proxy/shardingsphere-proxy-backend/src/main/java/org/apache/shardingsphere/proxy/backend/text/admin/mysql/MySQLAdminExecutorCreator.java
```diff
@@ -159,14 +159,12 @@ private boolean isQuerySpecialSchema(final SelectStatement sqlStatement, final S
     
     private DatabaseAdminExecutor mockExecutor(final String schemaName, final SelectStatement sqlStatement, final String sql) {
         boolean isNotUseSchema = !Optional.ofNullable(schemaName).isPresent() && sqlStatement.getFrom() == null;
+        if (!hasDatabases() || !hasResources()) {
+            return new NoResourceShowExecutor(sqlStatement);
+        }
+        String driverType = ProxyContext.getInstance().getContextManager().getMetaDataContexts().getProps().getValue(ConfigurationPropertyKey.PROXY_BACKEND_DRIVER_TYPE);
```
TeslaCN, line 165: Getting driver type can be moved into `if (isNotUseSchema)` block to reduce unnecessary overhead since `isNotUseSchema` is not always `true`.

lanchengx: I will adjust it in the next pr

Issue #16777, opened 2022-04-12
Reject execution of non-query DistSQL in transaction


RaigorJiang:
Hi community,

In ShardingSphere-Proxy, if user begins a transaction and executes DistSQL during the transaction:
- DistSQL will not wait for commit, the execution will take effect immediately.
- Changes made by DistSQL will not be rolled back when the transaction is rolled back.

After discussion with @tristaZero , we have a plan:
When a connection is in a transaction, the connection will no longer accept the execution of non-query DistSQL, that is, the execution of RDLStatement and UpdatableRALStatement will be rejected.

RaigorJiang:
fixed by #17632.

PR #17632, opened 2022-05-13
Disallow distsql in a transaction

jingshanglu:
ref #16777.


RaigorJiang (changes requested):
Hi @jingshanglu ,
It's not correct to only allow RQL. Because there are also some query statements in RAL, such as `SHOW TRANSACNTION TYPE`.

RaigorJiang:
In addition, `QueryableRALStatement` is also allowed to execute, which suffices.

RaigorJiang (approved):

PR #17818, opened 2022-05-20
Update the priority of converting to String

ZZemptypoint:
The issue Ref https://github.com/apache/shardingsphere/issues/17821

### DIFF
--- a/shardingsphere-jdbc/shardingsphere-jdbc-core/src/main/java/org/apache/shardingsphere/driver/jdbc/core/resultset/ResultSetUtil.java
+++ b/shardingsphere-jdbc/shardingsphere-jdbc-core/src/main/java/org/apache/shardingsphere/driver/jdbc/core/resultset/ResultSetUtil.java
@@ -61,7 +61,6 @@ public final class ResultSetUtil {
         if (value.getClass() == convertType) {
             return value;
         }
-        // converting to String need a higher priority
         if (String.class.equals(convertType)) {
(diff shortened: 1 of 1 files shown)


shardingsphere-jdbc/shardingsphere-jdbc-core/src/main/java/org/apache/shardingsphere/driver/jdbc/core/resultset/ResultSetUtil.java
```diff
@@ -61,6 +61,10 @@ public static Object convertValue(final Object value, final Class<?> convertType
         if (value.getClass() == convertType) {
             return value;
         }
+        // converting to String need a higher priority
```
strongduanmu, line 64: Please remove this comment.

strongduanmu (changes requested):
Hi @ZZemptypoint, can you submit an issue to describe this problem first? And then add some unit test for this logic?

shardingsphere-jdbc/shardingsphere-jdbc-core/src/main/java/org/apache/shardingsphere/driver/jdbc/core/resultset/ResultSetUtil.java
ZZemptypoint: As u wish.

PR #17855, opened 2022-05-22
Add spotless XML formatting rule

longtai-cn:
- Add spotless XML formatting rule.


terrymanu (changes requested):
pom is for build only, mix business logic here is not a good practice.

longtai-cn:
> pom is for build only, mix business logic here is not a good practice.

Roger, request to close

PR #18067, opened 2022-05-30
Fix encrypt examples EncryptColumnRuleConfiguration constructor

zhaoguhong:
Fix encrypt examples EncryptColumnRuleConfiguration constructor

### DIFF
--- a/examples/shardingsphere-example-generator/src/main/resources/template/jdbc/java/TestQueryAssistedShardingEncryptAlgorithm.ftl
+++ b/examples/shardingsphere-example-generator/src/main/resources/template/jdbc/java/TestQueryAssistedShardingEncryptAlgorithm.ftl
@@ -50,4 +50,9 @@ public final class TestQueryAssistedShardingEncryptAlgorithm implements QueryAss
     public String getType() {
         return "assistedTest";
     }
+
+    @Override
+    public void init(Properties props) {
+
+    }
 }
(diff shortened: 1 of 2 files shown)


examples/shardingsphere-example-generator/src/main/resources/template/jdbc/java/TestQueryAssistedShardingEncryptAlgorithm.ftl
```diff
@@ -50,4 +50,9 @@ public final class TestQueryAssistedShardingEncryptAlgorithm implements QueryAss
     public String getType() {
         return "assistedTest";
     }
+
+    @Override
+    public void init(Properties props) {
```
totalo, line 55: Thanks for your contribution, but this init method I think needs to initialize props, you can refer to the rest of the TestQueryAssistedShardingEncryptAlgorithm file

totalo (changes requested):

examples/shardingsphere-example-generator/src/main/resources/template/jdbc/java/TestQueryAssistedShardingEncryptAlgorithm.ftl
zhaoguhong: ok，I see

totalo:
Good job!

totalo (approved):

PR #18097, opened 2022-05-31
fix when table alias case different throw NullPointException

### LINKED ISSUE #18095: An NPE occurs when a field is identified by a table alias
## Bug Report
The sql like this
```sql
select count(1) from (        select T1.id as testId from table1 t1 left join table2 t2 on t1.id = T2.id) as a
```
It is will use alias 'T1' to find table 'table1', then throw NullPointException, because the alias 'T1' and alias referencce 't1' has case different

Zhoutzzz:
Fixes #18095 .

### DIFF
--- a/shardingsphere-features/shardingsphere-sharding/shardingsphere-sharding-core/src/main/java/org/apache/shardingsphere/sharding/merge/dql/groupby/GroupByMemoryMergedResult.java
+++ b/shardingsphere-features/shardingsphere-sharding/shardingsphere-sharding-core/src/main/java/org/apache/shardingsphere/sharding/merge/dql/groupby/GroupByMemoryMergedResult.java
@@ -37,15 +37,8 @@ import org.apache.shardingsphere.sql.parser.sql.common.constant.AggregationType;
 import org.apache.shardingsphere.sql.parser.sql.common.segment.generic.table.SimpleTableSegment;
 
 import java.sql.SQLException;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 import java.util.Map.Entry;
-import java.util.Objects;
 
 /**
  * Memory merged result for group by.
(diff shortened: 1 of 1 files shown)


shardingsphere-features/shardingsphere-sharding/shardingsphere-sharding-core/src/main/java/org/apache/shardingsphere/sharding/merge/dql/groupby/GroupByMemoryMergedResult.java
```diff
@@ -129,7 +130,8 @@ private List<Boolean> getValueCaseSensitive(final QueryResult queryResult, final
     
     private boolean getValueCaseSensitiveFromTables(final QueryResult queryResult,
                                                     final SelectStatementContext selectStatementContext, final ShardingSphereSchema schema, final int columnIndex) throws SQLException {
-        for (SimpleTableSegment each : selectStatementContext.getAllTables()) {
+        Collection<SimpleTableSegment> allTables = selectStatementContext.getAllTables();
```
tuichenchuxin, line 133: There's no need to modify this file.

shardingsphere-sql-parser/shardingsphere-sql-parser-statement/src/main/java/org/apache/shardingsphere/sql/parser/sql/common/extractor/TableExtractor.java
```diff
@@ -260,7 +260,7 @@ public void extractTablesFromUpdate(final UpdateStatement updateStatement) {
      */
     public boolean needRewrite(final OwnerSegment owner) {
         for (TableSegment each : tableContext) {
-            if (owner.getIdentifier().getValue().equals(each.getAlias().orElse(null))) {
+            if (owner.getIdentifier().getValue().equalsIgnoreCase(each.getAlias().orElse(null))) {
```
tuichenchuxin, line 263: Is this sql can be execute in different database. like postgres or oracle?

tuichenchuxin (changes requested):

shardingsphere-sql-parser/shardingsphere-sql-parser-statement/src/main/java/org/apache/shardingsphere/sql/parser/sql/common/extractor/TableExtractor.java
Zhoutzzz: this sql follow sql92 protocol

tuichenchuxin (approved):

Issue #17803, opened 2022-05-19
Refactor EncryptColumnRuleConfiguration to decouple cipher encryptor and assist encryptor


cheese8:
## Feature Request

How about refactor `EncryptColumnRuleConfiguration` as below:  @strongduanmu 

- [x] add assistEncryptorName
- [x] make assistEncryptorName work
- [ ] refactor QueryAssistedEncryptAlgorithm and deprecate it (the only difference of  the algorithm one is **symmetric** and other is **asymmetric**)
- [ ] validate the availabilities on cipherColumn & cipherEncryptorName pair, assistColumn & assistEncryptorName pair

![image](https://user-images.githubusercontent.com/25882819/169700196-4efe8170-4da9-4260-b58f-55e0836c0ebb.png)

tuichenchuxin:
@cheese8 Why need to decouple cipher encryptor and assist encryptor?

cheese8:
> @cheese8 Why need to decouple cipher encryptor and assist encryptor?

One for cipherColumn and one for assistColumn, that's more simple.

PR #18166, opened 2022-06-03
add AssistEncryptorName

cheese8:
For #17803 this is the 1st pr.

- add AssistEncryptorName


shardingsphere-features/shardingsphere-encrypt/shardingsphere-encrypt-api/src/main/java/org/apache/shardingsphere/encrypt/api/config/rule/EncryptColumnRuleConfiguration.java
```diff
@@ -37,5 +37,7 @@ public final class EncryptColumnRuleConfiguration {
     
     private final String encryptorName;
     
+    private final String assistedQueryEncryptorName;
```
terrymanu, line 40: Could you keep the old constructor for legacy API compatible?

terrymanu (approved):

PR #18235, opened 2022-06-08
support inline constraint support

### LINKED ISSUE #18230: Support inline constraint in Oracle CREATE TABLE statement
## English Description
When I input such a sql statement to create a table:

```sql
create table test1
(
  id int not null
);
```

I support that "not null" will be in the SqlStatement object.

## Chinese Description

当我解析这样一段SQL的时候：

```sql
create table test1
(
  id int not null
);
```

我希望在解析后“not null”，可以出现在SqlStatement的实现类中。

zhiyoucai:
Fixes #18230 .

- modify org.apache.shardingsphere.sql.parser.sql.common.segment.ddl.column.ColumnDefinitionSegment，support notNull constraint.
- modify org.apache.shardingsphere.sql.parser.oracle.visitor.statement.impl.OracleDDLStatementSQLVisitor, support what I changed above.


terrymanu (changes requested):
Hi, could you translate it to English?

terrymanu (approved):

PR #18359, opened 2022-06-14
Fix begin for postgresql & opengauss

jingshanglu:


shardingsphere-proxy/shardingsphere-proxy-backend/src/main/java/org/apache/shardingsphere/proxy/backend/exception/InTransactionException.java
```diff
@@ -0,0 +1,33 @@
...
+
+package org.apache.shardingsphere.proxy.backend.exception;
+
+import lombok.Getter;
+
+/**
+ * In transaction exception.
+ */
+@Getter
+public final class InTransactionException extends BackendException {
```
terrymanu, line 26: Please use @RequiredArgsConstructor

shardingsphere-proxy/shardingsphere-proxy-frontend/shardingsphere-proxy-frontend-core/src/main/java/org/apache/shardingsphere/proxy/frontend/command/CommandExecutorTask.java
```diff
@@ -114,7 +115,10 @@ private boolean executeCommand(final ChannelHandlerContext context, final Packet
             }
             return true;
         } catch (final SQLException ex) {
-            databaseProtocolFrontendEngine.handleException(connectionSession);
+            databaseProtocolFrontendEngine.handleException(connectionSession, ex);
+            throw ex;
+        } catch (final BackendException ex) {
+            databaseProtocolFrontendEngine.handleException(connectionSession, ex);
```
terrymanu, line 121: It is better to use `|` for multiple exception process

terrymanu (changes requested):

tristaZero (approved):

terrymanu (approved):

PR #18364, opened 2022-06-14
Add assert of stop scaling source writing

soulasuna:
## Add assert of stop scaling source writing

- Add assert of stop scaling source writing in test-scaling


soulasuna:
Migrate to #18374.

PR #18398, opened 2022-06-16
Remove the check method in ShardingRule.

yx9o:
Remove the `check` method in `ShardingRule`.


shardingsphere-features/shardingsphere-sharding/shardingsphere-sharding-distsql/shardingsphere-sharding-distsql-handler/src/main/java/org/apache/shardingsphere/sharding/distsql/handler/checker/ShardingTableRuleStatementChecker.java
```diff
@@ -98,6 +114,27 @@ private static void check(final ShardingSphereDatabase database,
         Map<String, List<AbstractTableRuleSegment>> groupedTableRule = groupingByClassType(rules);
         checkAutoTableRule(groupedTableRule.getOrDefault(AutoTableRuleSegment.class.getSimpleName(), Collections.emptyList()));
         checkTableRule(databaseName, currentRuleConfig, groupedTableRule.getOrDefault(TableRuleSegment.class.getSimpleName(), Collections.emptyList()));
+        if (!isCreate) {
```
menghaoranss, line 117: Does the rule need to be checked when binding table rule is created?

menghaoranss (changes requested):

menghaoranss (approved):

PR #18420, opened 2022-06-18
support forced route by hint and not influcenced with sharding rule config

gavenpeng:
Increment feature:support forced route by hint not influenced with sharding config

support force routing by hint with in code mode,as long as the thread context has the force-routing flag,may be route
to config database or config table,not influenced by sharding rule config.

demo：
public void demoFunc() {
    
    HintManager hintManager = HintManager.getInstance();
    //force route to config database 
    hintManager.addDatabaseShardingValue("logic_database_name", "real_database_name");
    // force route to 
    hintManager.addTableShardingValue("logic_table_name", "real_table_name");
    try {     
        //your code
    } catch (Throwable e) {
        xxxxxxx   
    } finally {
        // this is must
        hintManager.close();   
    }
}


terrymanu (changes requested):
The hint strategy contains hint value and hint algorithm, and ShardingSphere only can support config single strategy for sharding (hint or standard), how to use them together?

gavenpeng:
> The hint strategy contains hint value and hint algorithm, and ShardingSphere only can support config single strategy for sharding (hint or standard), how to use them together?

this feature is support force route when sharding strategy is standard, user can hint config routing database or table

terrymanu (changes requested):
The CI is broken, please fix them

PR #18426, opened 2022-06-19
Add StorageConnectorReusableRule to decouple connection reuse logic and concrete rule

terrymanu:


tristaZero (approved):

chenyx1007:
this interface look like could cache `ShardingspherePreparedStatement` How is it used? Ant what the restrictions it have.

PR #18473, opened 2022-06-21
Add query route primary data source when replica all offline.

### LINKED ISSUE #18441: Consider primary datasource provides query when all the replica are offline in Readwrite-Splitting Dynamic type
Hi community,

Execute the query throws an exception when we use `readwrite-splitting ` rule and` db-discovery` rule and all the replica are offline.

image

So, we consider add configuration switch in Dynamic `readwrite-splitting`, choose the primary data source by yourself to whether provide query when all the replica are offline.

zjcnb:
Fixes #18441

### DIFF
--- a/shardingsphere-features/shardingsphere-readwrite-splitting/shardingsphere-readwrite-splitting-core/src/main/java/org/apache/shardingsphere/readwritesplitting/route/impl/ReadwriteSplittingDataSourceRouter.java
+++ b/shardingsphere-features/shardingsphere-readwrite-splitting/shardingsphere-readwrite-splitting-core/src/main/java/org/apache/shardingsphere/readwritesplitting/route/impl/ReadwriteSplittingDataSourceRouter.java
@@ -50,7 +50,7 @@ public final class ReadwriteSplittingDataSourceRouter {
     }
     
     private boolean isPrimaryRoute(final SQLStatementContext<?> sqlStatementContext) {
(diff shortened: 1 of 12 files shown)


tristaZero (changes requested):
How about `write-data-source-query-enable`?

zjcnb:
> write-data-source-query-enable

Yeah, it 's better.

shardingsphere-features/shardingsphere-readwrite-splitting/shardingsphere-readwrite-splitting-core/src/main/java/org/apache/shardingsphere/readwritesplitting/route/impl/ReadwriteSplittingDataSourceRouter.java
```diff
@@ -68,4 +69,9 @@ private boolean containsLastInsertIdProjection(final SQLStatementContext<?> sqlS
     private boolean isHintWriteRouteOnly(final SQLStatementContext<?> sqlStatementContext) {
         return HintManager.isWriteRouteOnly() || (sqlStatementContext instanceof CommonSQLStatementContext && ((CommonSQLStatementContext<?>) sqlStatementContext).isHintWriteRouteOnly());
     }
+    
+    private boolean isAllowWriteDatasourceQuery() {
```
menghaoranss, line 73: Datasource should be DataSource

tristaZero (approved):

tristaZero (changes requested):
test

tristaZero (changes requested):
test

tristaZero (approved):

PR #18522, opened 2022-06-23
add allowDMLWithoutShardingKey in ruleConfig

### LINKED ISSUE #18430: Sharding Audit
I want to add a configuration in server.yaml called
allow-broadcast-without-sharding-key, it indicates whether to allow broadcasting without sharding key, the default is true, which is consistent with the existing logic, and can be changed to false if necessary.

The broadcast behavior without the shard key will cause a large area of access to the DB, causing timeouts and other problems. If it is an update, the risk is even greater, so I think it is necessary to support it.

I checked the source code, and it seems that as long as I add this logical judgment to the ShardingStandardRoutingEngine, I have already implemented it.

I would like to consult my requirements, parameter naming and whether the location is reasonable.

Hope for your reply!

natehuangting:
Fixes #18430 .

- add allowDMLWithoutShardingKey in ruleConfig


terrymanu (changes requested):
It is better use SQL cheker rule to handle SQL audit behaviour.

PR #18530, opened 2022-06-23
Refactor ShardingScalingRulesQueryResultSet

terrymanu:


shardingsphere-features/shardingsphere-sharding/shardingsphere-sharding-distsql/shardingsphere-sharding-distsql-handler/src/main/java/org/apache/shardingsphere/scaling/distsql/handler/query/ShardingScalingRulesQueryResultSet.java
```diff
@@ -42,9 +43,8 @@ public final class ShardingScalingRulesQueryResultSet implements DistSQLResultSe
     
     @Override
     public void init(final ShardingSphereDatabase database, final SQLStatement sqlStatement) {
-        Optional<ShardingRuleConfiguration> ruleConfig = database.getRuleMetaData().getConfigurations()
-                .stream().filter(each -> each instanceof ShardingRuleConfiguration).map(each -> (ShardingRuleConfiguration) each).findAny();
-        data = ruleConfig.map(ShardingRuleConfiguration::getScaling).orElse(Collections.emptyMap()).entrySet().iterator();
+        Optional<ShardingRule> rule = database.getRuleMetaData().findSingleRule(ShardingRule.class);
+        data = rule.map(shardingRule -> ((ShardingRuleConfiguration) shardingRule.getConfiguration()).getScaling().entrySet().iterator()).orElse(Collections.emptyIterator());
```
totalo, line 47: `orElseGet` mybe better

totalo (changes requested):

shardingsphere-features/shardingsphere-sharding/shardingsphere-sharding-distsql/shardingsphere-sharding-distsql-handler/src/main/java/org/apache/shardingsphere/scaling/distsql/handler/query/ShardingScalingRulesQueryResultSet.java
terrymanu: Not true, `orElseGet` can only use in the direct value. In this situation,  we need get configuration in the optional value, it is unnecessary to mock the configuration when rule absent.

totalo (approved):

PR #18592, opened 2022-06-25
Support force route db or table use hint

gavenpeng:
Increment feature: support forced route use hint not influenced with sharding strategy

- support force routing by hint with in code mode,as long as the thread context has the force-routing flag,may be route
to config database or config table,not influenced by sharding strategy config.

demo：

```java
public void demoFunc() {

HintManager hintManager = HintManager.getInstance();
//force route to config database 
hintManager.addDatabaseShardingValue("logic_database_name", "real_database_name");
// force route to 
hintManager.addTableShardingValue("logic_table_name", "real_table_name");
try {     
    //your code
} catch (Throwable e) {
    xxxxxxx   
} finally {
    // this is must
    hintManager.close();   
}
```


terrymanu (approved):

terrymanu:
The design maybe incorrect. 
`hintManager.addDatabaseShardingValue("logic_database_name", xxx)`

The second param `xxx` should be sharding value, not actual table name.

gavenpeng:
me set the actual table name or actual db name for the second param xxx is not do sharding with strategy,may be direct return the actual value,this maybe exist incorrect use hint

PR #18593, opened 2022-06-25
Improve ExportDatabaseConfigurationHandler

### LINKED ISSUE #18437: ExportDatabaseConfigurationHandler need to be improved
The export data in ExportDatabaseConfigurationHandler is in file: `shardingsphere-proxy/shardingsphere-proxy-backend/src/test/resources/expected/export-database-configuration.yaml`. The content of the export data is not good engough.

1. `!!set` should to be removed in line 33 
2. `?` should to be removed in line 34
3. `''` value of none should be removed in line 40
4. `null` value of default_scaling should be removed in line 45
5. : `null` should be removed in line 59

mabaiwan:
Fixes #18437.

- Improve ExportDatabaseConfigurationHandler.


terrymanu (commented):
I am not sure whether replace with String is the only solution.
It is better to use original yaml engine to generate the content.

mabaiwan:
> I am not sure whether replace with String is the only solution. It is better to use original yaml engine to generate the content.

I'll try another solution.

terrymanu (approved):

terrymanu:
Great, we can create a new issue to keep improving it.

PR #18677, opened 2022-06-29
Refactor ShardingSphereMetaData.databases to ConcurrentHashMap

terrymanu:


zjcnb:
I reproduced this problem in my local environment, we have only `server.yaml` and `ShrdingSphere-Proxy `start up success. 

When I connect `Proxy` and excute `DQL` statement. the `databaseName` is null.

image

Because `databases` are `ConcurrentHashMap`, so it will throw exception when `key` is null.

terrymanu:
Duplicate with #18819

Issue #18584, opened 2022-06-25
SQL route can optimize


wt-better:
### Describe the feature you would like.

Eg：`select * from tbl_order where order_id in (1,2)`

`tbl_order` is sharding  table, two database: database01, database02; two tables: tbl_order_01, tbl_order_02 .

**shardingKey** is `order_id`，database and table sharding rule all `order_id%2` .

The current  will produce four `shardingsphere.RouteUnit`, respectively: <databse01:tbl_order_01>, <databse01:tbl_order_02>, <databse02:tbl_order_01>, <databse02:tbl_order_02> .

I think two routeUnit `shardingsphere.RouteUnit`：<databse01:tbl_order_01> 、<databse02:tbl_order_02>  ok,this optimize can reduce sql execute count and result merge cost.

version: shardingsphere 5.1.1, The following code:

`org.apache.shardingsphere.sharding.route.engine.type.standard.ShardingStandardRoutingEngine#route0`

Further, the binding algorithm is proposed，when datbase and table sharing rule and value consistent，we can reduce once route 
calculate cost。

terrymanu:
Yes, it is a good idea, could you provide the pseudo code or draw a diagram to make your idea more clear?

cheese8:
I am try to improve this.

wt-better:
Thanks a lot, expect !

cheese8:
> Yes, it is a good idea, could you provide the pseudo code or draw a diagram to make your idea more clear?

Please review the InlineShardingAlgorithm proposal. @terrymanu

terrymanu:
It better to design first

cheese8:
![image](https://user-images.githubusercontent.com/25882819/177003372-1711f468-a6df-40a0-accb-4dca116e9296.png)

wt-better:
please attent sharding values, not only sharding strategy.



---- Replied Message ----
| From | ***@***.***> |
| Date | 07/02/2022 21:41 |
| To | ***@***.***> |
| Cc | ***@***.******@***.***> |
| Subject | Re: [apache/shardingsphere] SQL route can optimize (Issue #18584) |

cheese8:
Can you describe your scene in detail to show us why database and table sharding rule are the same, for example all order_id%2. 
@wt-better

wt-better:
database and table rule in same is very common on my company.

---- Replied Message ----
[...]

PR #18710, opened 2022-06-29
reduce sharding dataNodes on the same shardingStrategy - InlineShardingAlgorithm

cheese8:
For #18584.

-reduce sharding dataNodes on the same shardingStrategy - InlineShardingAlgorithm

### DIFF
--- a/shardingsphere-features/shardingsphere-sharding/shardingsphere-sharding-api/src/main/java/org/apache/shardingsphere/sharding/spi/ShardingAlgorithm.java
+++ b/shardingsphere-features/shardingsphere-sharding/shardingsphere-sharding-api/src/main/java/org/apache/shardingsphere/sharding/spi/ShardingAlgorithm.java
@@ -22,5 +22,13 @@ import org.apache.shardingsphere.infra.config.algorithm.ShardingSphereAlgorithm;
 /**
  * Sharding algorithm.
  */
-public interface ShardingAlgorithm extends ShardingSphereAlgorithm {
+public interface ShardingAlgorithm extends ShardingSphereAlgorithm, Comparable {
+    
+    @Override
+    default int compareTo(final Object other) {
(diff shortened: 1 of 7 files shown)


terrymanu (changes requested):
The route logic is too important, we need more design.
Why add Comparable for sharding algorithm, I can not get the point

cheese8:
> The route logic is too important, we need more design. Why add Comparable for sharding algorithm, the proposal of code maybe need more clear.

Yeah, the route logic indeed is the most important. 
I reorganized the idea and logic of this proposal were listed as below:
1、Now, route0 returned the Cartesian Product on the routedDataSources X the routedTables, then we got 4 dataNodes(ds0, table0 / ds0, table1 / ds1, table0/ ds1, table1) on routedDataSources(ds0, ds1) and routedTables(table0, table1) .

2、As the original issue mentioned that, while the databaseShardingStrategy the same with the tableShardingStrategy, no matter what the shardingConditionValue provided, the return of the Cartesian Product on the routedDataSources X the routedTables should be broken and reduced, then we got 2 dataNodes(ds0, table0 / ds1, table1) on routedDataSources(ds0, ds1) and routedTables(table0, table1) in this case, in other words, the database and the table should route the same suffix index on the specified shardingConditionValue.

3、how to judge the databaseShardingStrategy the same with the tableShardingStrategy?
We should compare the databaseShardingStrategy to the tableShardingStrategy, only if they are(shardingColumn and shardingAlgorithm) not null, and shardingColumn and shardingAlgorithm are both equals, especially the algorithmExpression without prefix of the shardingAlgorithm are the same, the same judged, otherwise not the same judged.

cheese8:
> add Comparable for sharding algorithm

add Comparable for sharding algorithm and set default implementation
1、we should judge the sharding algorithm are the same or not.
2、we can take InlineShardingAlgorithm for proposal.

PR #18724, opened 2022-06-30
use statementContext replace statement in method check

natehuangting:
- use statementContext replace statement in method check


natehuangting:
For #18430

terrymanu (approved):

PR #18751, opened 2022-07-01
add hint disableAuditNames

### LINKED ISSUE #18705: Add hint disableAuditNames

The shardingsphere proxy already supports the auditing function, but in some cases the auditing function needs to be turned off. For example, the auditing function is to prohibit queries without sharding keys, but the business needs to perform batch operation. I am going to add the disableAuditNames hint by adding the hint in the SQL Auditing can be avoided. If the proxy wants to allow users to use the hint, the variable allowHintDisable needs to be set to true in the audit policy.

The specific hint usage is as follows.
```sql
/* ShardingSphere hint: disableAuditNames=auditname1, auditname2 */ SELECT * FROM t_order;
```

natehuangting:
Fixes #18705 .

- add hint disableAuditName


natehuangting:
For #18430

strongduanmu (approved):

Issue #18389, opened 2022-06-16
[ OSPP 2022 ] Solve unsupported Oracle SQL about Alter Statement for ShardingSphere Parser


chakkk309:
This issue is created to better track my PRs for OSPP 2022, you can view more about my project [here](https://summer-ospp.ac.cn/#/org/prodetail/22ee10103).

## Background
ShardingSphere Parser Engine helps users parse SQL statements into AST and generate corresponding SQL Statement objects from it.

## Process
1. Compare SQL definitions in [Oficial SQL Doc](https://docs.oracle.com/en/database/oracle/oracle-database/18/lnpls/plsql-language-elements.html#GUID-285DFAEE-1A34-4248-BA4B-2591476A0C3C) and [ShardingSphere SQL Doc](https://github.com/apache/shardingsphere/tree/master/shardingsphere-sql-parser/shardingsphere-sql-parser-dialect/shardingsphere-sql-parser-oracle/src/main/antlr4/imports/oracle).
2. If there is any difference in ShardingSphere SQL Doc, please correct them by referring to the Official SQL Doc.
3. Run `mvn install the current_file_module`.
4. Check whether there are any exceptions. If indeed, please fix them. (`Especially xxxVisitor.class`).
5. Add new corresponding SQL case in SQL Cases and expected parsed result in Expected Statment XML.
6. Run SQLParserParameterizedTest to make sure no exceptions.

## Tasks
- [x] ALTER INDEXTYPE [doc](https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/ALTER-INDEXTYPE.html#GUID-BFA7E29C-7905-4811-9119-B20FD8EA18F2) 
- [x] ALTER INMEMORY JOIN GROUP [doc](https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/ALTER-INMEMORY-JOIN-GROUP.html#GUID-AF24F413-BB14-4B5D-93BF-9EB31ACFEBEC) 
- [x] ALTER JAVA [doc](https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/ALTER-JAVA.html#GUID-6B211750-3247-4D71-9533-3DD8F66640CD) 
- [x] ALTER LIBRARY [doc](https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/ALTER-LIBRARY.html#GUID-BB90AF66-3B1F-46C4-9716-4578DE0AE1F3) 
[...]

PR #18890, opened 2022-07-06
Support `ALTER JAVA` statement parse for Oracle

chakkk309:
Ref https://github.com/apache/shardingsphere/issues/6480 & https://github.com/apache/shardingsphere/issues/18389

- Add `ALTER JAVA` grammar
- Support parsing `ALTER JAVA` in Oracle
- Add test


chakkk309:
I got a parse error when I used the Antrl4 plugin for validation:

image

It seems like **public** in testing SQL is parsed as the keyword **PUBLIC** in the database. But according to the syntax of `alter-java`, the public here should resolve to the corresponding **schemaName**.

tuichenchuxin:
@chakkk309 I have added `public` to unreservedWord, Please check if it's ok.

tuichenchuxin (approved):

PR #18892, opened 2022-07-06
feat(integration-test): create new run mode

JiekerTime:
After this PR, we will have three run modes.

1. Memory: Execute integration tests via jdbc+H2.
2. Standalone: Perform single ShardingSphere(JDBC,Proxy) integration tests based on database (MySQL, PG) containers.
3. Cluster: Add more ShardingSphere(JDBC,Proxy) nodes based on standalone and import registry services such as zookeeper and ETCD.

In fact, the PostgreSQL database scenario test always had problems, but the single-rule integration test had not been opened before.
image
image

So i close the PG single-rule integration test

### DIFF
--- a/.github/workflows/it.yml
+++ b/.github/workflows/it.yml
@@ -123,7 +123,7 @@ jobs:
         if: matrix.adapter == 'proxy'
         run: docker load -i /tmp/shardingsphere-proxy-test.tar
       - name: Run Integration Test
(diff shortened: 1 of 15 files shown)


terrymanu (changes requested):
Singleton is not ShardingSphere mode, why add it in IT only?

JiekerTime:
> Singleton is not ShardingSphere mode, why add it in IT only?

Because I think in the integration test we have memory mode and cluster mode, no corresponding definition is given for the mode of single node operation.
Just like the single-scenario mode in our scenario test, I want to refine the architecture mode to make it easy for subsequent customizability extensions.

PR #18944, opened 2022-07-07
[bug]databaseDiscovery can not find master

### LINKED ISSUE #18943: After switchover, databaseDiscovery cannot discover new master
master

ShardingSphere-Proxy

### Actual behavior
Two nodes ds_0, ds_1, originally ds_0 is the host, after the switch ds_1 is the host. But the Proxy cannot find that ds_1 is the master.

When switching, the replication status of ds_1 is cleared, the show slave status is empty, and the proxy adds ds_1 to disableDatasourceNames. Since the disabled node does not participate in the logic of judging whether it is the master, and the show slave status is always empty, the state cannot be changed to enable.

企业微信截图_aa59b243-52b8-4979-bdce-47e6f3f304a3
企业微信截图_349ac266-ff9e-49cd-bed9-db8a21483e3d

natehuangting:
Fixes #18943 

- [bug]databaseDiscovery can not find master


zjcnb:
@natehuangting Hi , i don't think this is handled well. Let's think about other idea.

zhaojinchao95 (approved):

PR #18979, opened 2022-07-09
Handling of star in select statement in postgresql

### LINKED ISSUE #18078: Handling of star in select statement
The postgresql parser seems to parse this without giving any error

`select * as some_col from some_table`

A * is not allowed to have aliases in postgresql.

Weirdly, postgresql allows an alias after a dot star. So a query like this is allowed `select t.* as some_col from some_table t`. The alias is ignored in this case.

zrsaber:
Fixes #18078.

- fix handling of star in select statement
-
-


terrymanu (changes requested):
Handle in G4 file is not enough, can you complete the visitor and SQLStatement?

zrsaber:
Visitor and statement are implemented in the SELECT statement. The error is not reported because the G4 file is not parsed

terrymanu (approved):

PR #19053, opened 2022-07-12
Rename indextype with camelcase

chakkk309:
- Rename indextype with camelcase


chakkk309:
need to fix some errors.

PR #19096, opened 2022-07-13
Improve stop scaling when running increment task.

azexcy:
- Scaling IT case add stop/start case
- Temporary remove `create index ... on ...` sql, until support `create index if not exists`


shardingsphere-test/shardingsphere-integration-test/shardingsphere-integration-test-scaling/src/test/java/org/apache/shardingsphere/integration/data/pipeline/env/IntegrationTestEnvironment.java
```diff
@@ -175,4 +176,17 @@ public String getActualDataSourcePassword(final DatabaseType databaseType) {
...
+     * @param databaseType database type.
+     * @return if matched return version, otherwise return empty.
+     */
+    public List<String> getNativeDatabaseVersions(final String databaseType) {
+        if (StringUtils.equalsIgnoreCase(databaseType, getNativeDatabaseType())) {
+            // Just to occupy the space, will not really use.
+            return Collections.singletonList("keep");
+        }
+        return Collections.emptyList();
+    }
```
sandynz, line 191: For `getNativeDatabaseVersions`, it's better to rename it.

And return `Collections.singletonList("keep")` is not readable enough.

Is it possible to return `boolean`?

sandynz (changes requested):
Thanks. There's some code snippets need to be double checked.

PR #19197, opened 2022-07-15
Fix alter local transaction rule

FlyingZC:
Fix alter local transaction rule.

- Fix alter local transaction rule without TYPE param;
- TransactionRule does not need to modify the databases;
- Create empty `ShardingSphereTransactionManagerEngine` may occured NPE when closed.


FlyingZC:
For second change item:
The caller must handle the `database` param itself, and the caller has already handled the database itself, such as removing or adding
The caller of the method `addResource` is as follows:
image
image
The caller has already added `database` itself.

The caller of the method `org.apache.shardingsphere.transaction.rule.TransactionRule#closeStaleResource(java.lang.String)` is as follows:
image
The caller has already removed `database` itself.

The caller of the method `org.apache.shardingsphere.transaction.rule.TransactionRule#closeStaleResource()` is as follows:
image
The caller has already closed and rebuilt the `TransactionRule`,so we needn't to handle the `database`.

FlyingZC:
For third change item:
When we create empty `ShardingSphereTransactionManagerEngine`,the param of `xaTransactionManagerProvider` in `XAShardingSphereTransactionManager` is also be empty.So I need to determine whether it is null.
image
image

jingshanglu (approved):

PR #19239, opened 2022-07-15
fix create table generator with no primary key.

tuichenchuxin:
For #19181.

- fix create table generator with no primary key.


strongduanmu (approved):
Good job, merged.

PR #19404, opened 2022-07-21
Fix MySQL syntax error cannot be thrown to client

### LINKED ISSUE #11955: The Proxy executes the select date() command and the result is wrong

master

 ShardingSphere-Proxy

### Mysql or Postgresql

Mysql 8.x

### Expected behavior

show error syntax

```
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ')' at line 1
```

### Actual behavior

result 1
```
+--------+
| date() |
+--------+
| 1      |
+--------+
```

TeslaCN:
Fixes #11955.


tristaZero (changes requested):
It's not so clear to add ex para into such interface

TeslaCN:
> It's not so clear to add ex para into such interface

SaneQueryResult is used for handling SQLException. We can hardly return a proper result if we don't know what happened.

tristaZero (approved):

PR #19518, opened 2022-07-25
Dynamic config use command instead of using cnf,conf file at IT

azexcy:
- Use command instead of database cnf,conf file
- Remove unused file


azexcy:
image
Because used `my.cnf` config file, although server-id can use command to config, but still need set at `my.cnf`

terrymanu (approved):

Issue #18897, opened 2022-07-06
Why DriverJDBCExecutor#executor  do not handle the error result of branch thread execution？


mayfsac:
## Question
Why DriverJDBCExecutor#executor  do not handle the error result of branch thread execution？
 DriverJDBCExecutor#executor：

![image](https://user-images.githubusercontent.com/43204877/177512854-6437ce45-89f5-431a-9365-e899755c8555.png)
JDBCExecutorCallback#execute：
![image](https://user-images.githubusercontent.com/43204877/177512621-edf3fba5-1801-4d0d-846a-eece36fbb9dc.png)

In this way, the result of the overall sql execution is inconsistent with the expected result.

eg:
insert into xx_table (id,xxx,xxx) values (1,xx,xxx),(2,xxx,xxx);

This statement will route to two datasources, ds0, ds1.

If there is a row with primary key 1 in ds1, it will throw exception: Duplicate entry '1' for key 'PRIMARY';
The final result is: the sql routed to ds0 is successfully executed, and the sql execution of ds1 fails.

Why is it not processed as all failures, which leads to data inconsistency, and distributed transactions will not be rolled back

terrymanu:
What is your version?

mayfsac:
sharding-jdbc  5.1.1

mayfsac:
> What is your version?
sharding-jdbc 5.1.1
This question does not involve distributed transactions, but just asks why sql is handled like this when it is executed

TeslaCN:
The null check was added in #8760.

mayfsac:
> moved to top in #9073.

Is this bug a high priority?

TeslaCN:
`return null` when `!isTruckThread` is because we didn't think about Exception cause by data. We will finish this issue before next release.

TeslaCN:
## Description

The root cause of this issue: **When handling exceptions, exceptions caused by data problems are not considered.**

This logic is added in PR <https://github.com/apache/shardingsphere/pull/8760>, the original intention is to deal with the inconsistency of the front and back database types of ShardingSphere-Proxy.

For example, ShardingSphere-Proxy uses the MySQL protocol to provide external services, and the database that Proxy actually connects to is the PostgreSQL database.

Use MySQL Connector/J to connect to ShardingSphere-Proxy. During the connection establishment phase, some special SQL will be sent, such as querying MySQL-specific variables. If MySQL-specific SQL such as `show tables` is directly executed in PostgreSQL, an error will occur.

[...]

PR #19525, opened 2022-07-25
Fix error from database is not thrown as expected

TeslaCN:
Related to #18897.


TeslaCN:
The scaling test is already failed on master branch. License check will be fixed in https://github.com/apache/shardingsphere/pull/19540.

terrymanu (approved):

PR #19563, opened 2022-07-26
Optimizing worker-id generation by in-memory computing in cluster mode

soulasuna:
## Optimizing worker-id generation by in-memory computing in cluster mode

- Optimizing worker-id generation by in-memory computing in cluster mode


shardingsphere-infra/shardingsphere-infra-common/src/main/java/org/apache/shardingsphere/infra/instance/ComputeNodeInstance.java
```diff
@@ -41,6 +41,8 @@ public final class ComputeNodeInstance {
     
     private Collection<String> labels = new ArrayList<>();
     
+    private volatile Long workerId;
```
TeslaCN, line 44: Could we change this to primitive type?

soulasuna: Of course i will fix it.

menghaoranss (approved):

Issue #19224, opened 2022-07-15
Use DistSQL add resource have unexpected behavior


wsm12138:
## Bug Report

**For English only**, other languages will not accept.

Before report a bug, make sure you have:

- Searched open and closed [GitHub issues](https://github.com/apache/shardingsphere/issues).
- Read documentation: [ShardingSphere Doc](https://shardingsphere.apache.org/document/current/en/overview).

Please pay attention on issues you submitted, because we maybe need more details. 
If no response anymore and we cannot reproduce it on current information, we will **close it**.

Please answer these questions before submitting your issue. Thanks!

### Which version of ShardingSphere did you use?

`e6328cb989cfb789189227892cb892740095d2d6`

### Which project did you use? ShardingSphere-JDBC or ShardingSphere-Proxy?
ShardingSphere-Proxy

### Expected behavior
add resource have error .so `show database resources` should't have result
### Actual behavior
add resource have error . excute `show database resources` have result
<img width="1855" alt="image" src="https://user-images.githubusercontent.com/86462784/180726407-73712d41-05f5-480a-a8f0-73de8e5dd525.png">

### Steps to reproduce the behavior, such as: SQL to execute, sharding rule configuration, when exception occur etc.
1. use server.yaml start proxy
server.yaml
```
mode:
  type: Cluster
  repository:
    type: ZooKeeper
    props:
      namespace: 0715distsql
      server-lists: 192.168.10.21:2181
      retryIntervalMilliseconds: 500
      timeToLiveSeconds: 60
      maxRetries: 3
      operationTimeoutMilliseconds: 500
  overwrite: true

rules:
- !AUTHORITY
  users:
    - root@%:root
    - sharding@:sharding
  provider:
    type: ALL_PERMITTED
- !SQL_PARSER
  sqlCommentParseEnabled: true
  sqlStatementCache:
    initialCapacity: 2000
    maximumSize: 65535
  parseTreeCache:
    initialCapacity: 128
    maximumSize: 1024

props:
  proxy-frontend-database-protocol-type: "PostgreSQL"
```
2. psql -U root  -h 127.0.0.1 -p 3333(proxy  port) -d postgres
3. execute sql
```
ADD RESOURCE resource_0 (
 URL="jdbc:postgresql://192.168.10.21:5432/postgres",
 USER=gaussdb,
 PASSWORD=passwd
);
```
```
show database resources;
```

RaigorJiang:
@wsm12138 
[...]

PR #19595, opened 2022-07-27
Multiple types of databases are not supported

jingshanglu:
ref #19224


shardingsphere-infra/shardingsphere-infra-common/src/main/java/org/apache/shardingsphere/infra/datasource/props/DataSourcePropertiesValidator.java
```diff
@@ -72,7 +74,8 @@ private void validate(final String dataSourceName, final DataSourceProperties da
     }
     
     private void checkFailFast(final DataSource dataSource, final DatabaseType databaseType) throws SQLException {
-        if (!dataSource.getConnection().getMetaData().getDatabaseProductName().equals(databaseType.getType())) {
+        Connection connection = dataSource.getConnection();
```
TeslaCN, line 77: Shall we close connection after used?

TeslaCN (approved):

PR #19657, opened 2022-07-28
Use assertThat to instead of assertEquals.

yx9o:
Use assertThat to instead of assertEquals.


terrymanu:
Please fix checkstyle error

terrymanu:
It is caused by another issue, I will fix it and re-run the test cases

terrymanu (approved):
LGTM. I just merge it

PR #19685, opened 2022-07-29
Refactor : refactor the transaction it filter, drop the other rule

taojintianxia:


TeslaCN (changes requested):
You may commit a changes to md file in docs and git reset --hard after check passed.

TeslaCN (approved):

Issue #19513, opened 2022-07-25
Restructure ShardingSphere Docs


Mike0601:
1. Restructure Chapter 1 Overview into Project Introduction: 
---What is Apache ShardingSphere
---Core Concepts
---Architecture
---Advantages
---Application Scenarios
2. Remove Chapter 3 Concepts.
3. Move out FAQ from Reference.
4. Restructure Reference into Test Manual and How it works.

PR #19731, opened 2022-08-01
newly added the principle of gateway

Mike0601:
For #19513 .

-
-
-


terrymanu (changes requested):
Is the picture need to add `Proxy`? `JDBC` has same capability.
How about remove  the word `Proxy`?

terrymanu (approved):

PR #19761, opened 2022-08-01
Standalone mode run IT case

zjcnb:


zjcnb:
I will create a new pull request when i fix ci

PR #19900, opened 2022-08-05
fix PostgreSQL/openGauss RESET statement

### LINKED ISSUE #19810: PostgreSQL/openGauss RESET statement cannot be parsed caused by #15309
## Bug Report

Related to #18738.

master - 73674d5ef5de2adce8a4a881e79baf4c401c352c

SQL parser.

### Expected behavior

PostgreSQL / openGauss RESET statement could be parsed correctly.

### Actual behavior

```
bmsql_sharding=> reset all;
ERROR:  You have an error in your SQL syntax
```

#15309 add an `EOF` which broke the rule of RESET statement.

runqi-zhao:
Fixes #19810.

- According to the discussion, only one EOF needs to be removed


shardingsphere-sql-parser/shardingsphere-sql-parser-dialect/shardingsphere-sql-parser-postgresql/src/main/antlr4/imports/postgresql/DALStatement.g4
```diff
@@ -53,7 +53,7 @@ configurationParameterClause
     ;
 
 resetParameter
-    : RESET (ALL | identifier) EOF
+    : RESET (ALL | identifier)
```
strongduanmu, line 56: @runqi-zhao Can you add some sql parse test case?

strongduanmu (changes requested):

runqi-zhao:
Hi @strongduanmu, After carefully studying the syntax of reset and looking at it, I found that the corresponding test cases have all been implemented, what is the specific use case that you need to add here?

strongduanmu (approved):

strongduanmu:
> Hi @strongduanmu, After carefully studying the syntax of reset and looking at it, I found that the corresponding test cases have all been implemented, what is the specific use case that you need to add here?

@runqi-zhao If there are already test cases, then there is no need to add additional test cases.

PR #20039, opened 2022-08-09
Split DistSQLResultSet to DatabaseDistSQLResultSet and GlobalRuleDistSQLResultSet.

yx9o:
In the later stage, the syntax of Global Rule will be migrated to specific modules. In order to ensure that the ResultSet can be returned normally after the migration, the `DistSQLResultSet` will be split to `DatabaseDistSQLResultSet` and `GlobalRuleDistSQLResultSet`. The Global Rule syntax will use `GlobalRuleDistSQLResultSet`.


RaigorJiang:
LGTM

Hi @terrymanu, Do you agree with this plan?

RaigorJiang (approved):

PR #20008, opened 2022-08-09
Update antlr4 to 4.10.1 close #19990

shanrenxj:
Fixes #19990.
Update antlr4 to 4.10.1
Changes proposed in this pull request:
-
-
-

PR #20119, opened 2022-08-12
Update Anltr to 4.10 to support Hibernate 6 integration

### LINKED ISSUE #19990: Update antlr4 to 4.10+
Hibernate 6.0.1+ was ues antlr4: 4.10+
when I force to use  antlr4: 4.10+，I got error：
org.antlr.v4.runtime.atn.ATN; Could not deserialize ATN with version **4** (expected 3).
then when I force to use  antlr4: 4.9.x，I got error
org.antlr.v4.runtime.atn.ATN; Could not deserialize ATN with version **3** (expected 4).

linghengqian:
Fixes #19990.

- Update pom.xml and LICENSE.
- Drop support for Hibrenate 5 in favor of Hibernate 6 integration.
- Also fixes #20008 .

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -64,7 +64,7 @@
         <commons-collections4.version>4.4</commons-collections4.version>
         <commons-codec.version>1.15</commons-codec.version>
         
-        <antlr4.version>4.9.2</antlr4.version>
+        <antlr4.version>4.10</antlr4.version>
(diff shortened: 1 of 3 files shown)


linghengqian:
Oh, my sdkman has always been JDK17, so no compilation errors were found, this problem should actually be put on hold.

PR #20189, opened 2022-08-15
Refactor Refresh TABLE METADATA document

zjcnb:


RaigorJiang (changes requested):
After communication with @zjcnb , we think that the Chinese description is not accurate enough and needs to be updated.

zjcnb:
> After communication with @zjcnb , we think that the Chinese description is not accurate enough and needs to be updated.

Yes, i will adjustment it.

PR #20197, opened 2022-08-15
Improve DriverDataSourceCache 

wuwen5:
Use ```computeIfAbsent``` avoid duplicate creation of data sources.

### DIFF
--- a/shardingsphere-jdbc/shardingsphere-jdbc-core/src/main/java/org/apache/shardingsphere/driver/jdbc/core/driver/DriverDataSourceCache.java
+++ b/shardingsphere-jdbc/shardingsphere-jdbc-core/src/main/java/org/apache/shardingsphere/driver/jdbc/core/driver/DriverDataSourceCache.java
@@ -18,7 +18,6 @@
 package org.apache.shardingsphere.driver.jdbc.core.driver;
 
 import org.apache.shardingsphere.driver.api.yaml.YamlShardingSphereDataSourceFactory;
-import org.apache.shardingsphere.driver.jdbc.core.datasource.ShardingSphereDataSource;
 
 import javax.sql.DataSource;
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


shardingsphere-jdbc/shardingsphere-jdbc-core/src/main/java/org/apache/shardingsphere/driver/jdbc/core/driver/DriverDataSourceCache.java
```diff
@@ -44,23 +43,17 @@ public DataSource get(final String url) throws SQLException {
         if (dataSourceMap.containsKey(url)) {
             return dataSourceMap.get(url);
         }
-        DataSource dataSource;
+        return dataSourceMap.computeIfAbsent(url, DriverDataSourceCache::createDataSource);
```
TeslaCN, line 46: This may impact concurrent performance in Java 8. https://bugs.openjdk.org/browse/JDK-8161372

TeslaCN (changes requested):

shardingsphere-jdbc/shardingsphere-jdbc-core/src/main/java/org/apache/shardingsphere/driver/jdbc/core/driver/DriverDataSourceCache.java
wuwen5: Yep, this problem is that when the key present，Therefore, keeping the above code can avoid this problem。

```
if (dataSourceMap.containsKey(url)) {
    return dataSourceMap.get(url);
}
```
Similar solutions
https://github.com/mybatis/mybatis-3/pull/1934/commits/beee143f72b0c396591aa0c59739a1e64fd41867

wuwen5: Generally, we don't need this code ``` if (dataSourceMap.containsKey(url)) ```, but because of this problem [JDK-8161372](https://bugs.openjdk.org/browse/JDK-8161372), we keep it.

TeslaCN: Got it. I didn't saw the logic above.

shardingsphere-jdbc/shardingsphere-jdbc-core/src/main/java/org/apache/shardingsphere/driver/jdbc/core/driver/DriverDataSourceCache.java
```diff
@@ -44,23 +43,17 @@ public DataSource get(final String url) throws SQLException {
...
+    }
+    
+    @SuppressWarnings("unchecked")
+    static <T extends Throwable> DataSource createDataSource(final String url) throws T {
```
TeslaCN, line 50: Consider adding `private` to method.

TeslaCN (approved):

PR #20269, opened 2022-08-18
Add : add a new action only focus on checkstyle check

taojintianxia:


.github/workflows/checkstyle-check.yml
```diff
@@ -0,0 +1,64 @@
...
+  checkstyle-check:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v3
+      - name: Cache Maven Repos
+        uses: actions/cache@v2
+        with:
+          path: ~/.m2/repository
+          key: shardingsphere-it-cache-${{ github.sha }}
```
TeslaCN, line 55: Why the key is IT?

.github/workflows/checkstyle-check.yml
```diff
@@ -0,0 +1,64 @@
...
+  checkstyle-check:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v3
+      - name: Cache Maven Repos
+        uses: actions/cache@v2
+        with:
+          path: ~/.m2/repository
+          key: shardingsphere-checkstyle-check-cache-${{ github.sha }}
```
TeslaCN, line 55: The cache will always be absent if using `github.sha` here. It will always use the restore key.

.github/workflows/checkstyle-check.yml
taojintianxia: I was thinking about how to caching the plugin related but not the rest dependencies.
this is a good refactor point.

.github/workflows/checkstyle-check.yml
taojintianxia: how about followings key. this key is better than the previewous key+restore key combination.
```
key: ${{ runner.os }}-checkstyle-check-${{ hashFiles('.github/.workflows/checkstyle-check.yml') }}
```

terrymanu (changes requested):
checkstyle-and-apache-rat in it.yml has already existed

taojintianxia:
test on my own repository as followings :
1. checkout a new branch `checkstyle-test` and set the on push branch includ `checkstyle-test`
2. remove a period in java doc in a Java class
3. push the change to branch `checkstyle-test`

image

terrymanu (approved):

Issue #19937, opened 2022-08-07
Optimize Federation-Query


boyjoy1127:
## Feature Request

Federation Query should be optimized.

### Is your feature request related to a problem?
No

### Describe the feature you would like.
- [X] Add TranslateTable to support projection and filter push down to the table rel-node.
- [X] Add optimization rules to support projection and filter push down to the table rel-node.
- [x] Optimize rules to support projection and filter push down well.
- [ ] Add cache for SQL and query plan.
- [ ] Add calcite statistics-related API in the future infra-meta module.
- [ ] Optimize rules to support the join well. Collect calcite statistics of the schema and table, to support optimization well.
- [ ] Optimize rules to support the sub-query well. Collect calcite statistics of the schema and table, to support optimization well.
- [ ] Optimize the procedure of the execution to accelerate the query.

strongduanmu:
Hi @boyjoy1127, Could we finish this issue in 5.1.3?

strongduanmu:
Since the design of sql federation is not consistent with this issue now, I will close this issue. If there is a new solution, please discuss it in the community first.

PR #20426, opened 2022-08-22
19937 remove filterable table

boyjoy1127:
ref #19937.

- remove the filterable table.
- add rules to optimize the projection and filter push down.
- fix related unit tests.


shardingsphere-infra/shardingsphere-infra-federation/shardingsphere-infra-federation-optimizer/src/main/java/org/apache/shardingsphere/infra/federation/optimizer/ShardingSphereOptimizer.java
```diff
@@ -34,7 +34,9 @@ public final class ShardingSphereOptimizer {
     
     private final SqlToRelConverter converter;
     
-    private final RelOptPlanner hepPlanner;
+    private final RelOptPlanner hepPlannerWithoutCalc;
```
strongduanmu, line 37: What's the difference between hepPlannerWithoutCalc and hepPlannerWithCalc?

strongduanmu (changes requested):

strongduanmu (approved):

PR #20479, opened 2022-08-24
Basic support migration to different database type

sandynz:
Local test migration from Oracle to PostgreSQL.

Include stages:
- Data source check on preparation
- Inventory dump and import
- Data consistency check (Column type: varchar2, timestamp, number. More column types will be added later)

Un-included stages:
- Create schema and table
- Incremental dump and import

- Compatible with data source checker; case-sensitive table name metadata loading
- And Oracle SQL builder impl; Improve Oracle query by page with ROWNUM
- Compatible with inventory dumper and importer
- Extract ColumnValueReader for common usage to resolve different database column java type mismatch
- Compatible with data consistency check
- Compatible with MySQL integration test


yx9o (approved):

sandynz:
On column value match, some special cases notes:

1, For timestamp column type, Oracle JDBC driver getObject and getTimestamp return different type of java object.

resultSet.getObject(i).getClass().getName() return oracle.sql.TIMESTAMP.
resultSet.getTimestamp(i).getClass().getName() return java.sql.Timestamp.

2, java.math.BigDecimal value doesn't match if their scale doesn't match, even if values are equals.

e.g. value1=332.2, value2=332.20, value1.class=java.math.BigDecimal, value2.class=java.math.BigDecimal
They're not equals by BigDecimal.equals

sandynz:
Quote on SQL builder is disabled for now, it need to do some improvement to support case-sensitive and case-insensitive databases in the same migration job.
Table name and column name should be quoted only if they're sepcial keywords in this database.

e.g. For column name case-insensitive database, real column name is `id`, `id` and `ID` in SQL are equivalent, but if it's quoted (e.g. `"ID"`) then column name is not matched.

PR #20493, opened 2022-08-24
Intercept statements that modify the sharding rule

azexcy:
Avoid incorrect data due to unsupported operations

- Intercept statements that modify the sharding rule


shardingsphere-proxy/shardingsphere-proxy-backend/src/main/java/org/apache/shardingsphere/proxy/backend/handler/distsql/rdl/rule/RuleDefinitionBackendHandler.java
```diff
@@ -70,6 +70,9 @@ protected ResponseHeader execute(final String databaseName, final T sqlStatement
         ShardingSphereDatabase database = ProxyContext.getInstance().getDatabase(databaseName);
         RuleConfiguration currentRuleConfig = findCurrentRuleConfiguration(database, ruleConfigClass).orElse(null);
         ruleDefinitionUpdater.checkSQLStatement(database, sqlStatement, currentRuleConfig);
+        if (NOT_SUPPORT_ALTERED_ACTIONS.contains(sqlStatement.getClass().getCanonicalName())) {
+            throw new UnsupportedOperationException("not support altered sharding config, it's may cause data incorrectly");
```
zjcnb, line 74: Maybe use `Not` instead of `not` is better.

azexcy:
After discussion, this isn't a good way to handle it, so  I close the pr.

PR #20570, opened 2022-08-26
offsetDateTime convert failed bug fix

### LINKED ISSUE #20568: OffsetDateTime type in model got ClassCastException

5.1.2

ShardingSphere-JDBC

### Expected behavior

when I use the OffsetDateTime in model, I hope that query can works

```java
public class Order {
    private Long orderId;
    private Long userId;
    private Integer status;
    private OffsetDateTime createTime;
}
```

### Actual behavior

Got exception:

```java
Caused by: java.lang.ClassCastException: class java.sql.Timestamp cannot be cast to class java.time.OffsetDateTime (java.sql.Timestamp is in module java.sql of loader 'platform'; java.time.OffsetDateTime is in module java.base of loader 'bootstrap')
	at org.apache.ibatis.type.OffsetDateTimeTypeHandler.getNullableResult(OffsetDateTimeTypeHandler.java:38)
	at org.apache.ibatis.type.OffsetDateTimeTypeHandler.getNullableResult(OffsetDateTimeTypeHandler.java:28)
	at org.apache.ibatis.type.BaseTypeHandler.getResult(BaseTypeHandler.java:85)
	... 99 more
```

ShardingSphereResultSet handle the type

```java
@Override
    public <T> T getObject(final int columnIndex, final Class<T> type) throws SQLException {
        if (BigInteger.class.equals(type)) {
            return (T) BigInteger.valueOf(getLong(columnIndex));
        } else if (Blob.class.equals(type)) {
            return (T) getBlob(columnIndex);
        } else if (Clob.class.equals(type)) {
            return (T) getClob(columnIndex);
        } else if (LocalDateTime.class.equals(type) || LocalDate.class.equals(type) || LocalTime.class.equals(type) || OffsetDateTime.class.equals(type)) { // to handle the OffsetDateTime type
            return (T) ResultSetUtil.convertValue(mergeResultSet.getValue(columnIndex, Timestamp.class), type);
        } else {
            return (T) ResultSetUtil.convertValue(mergeResultSet.getValue(columnIndex, type), type);
        }
    }
```

ResultSetUtil also do this

```java
  private static Object convertTimestampValue(final Object value, final Class<?> convertType) {
        Timestamp timestamp = (Timestamp) value;
        if (LocalDateTime.class.equals(convertType)) {
            return timestamp.toInstant().atZone(ZoneId.systemDefault()).toLocalDateTime();
        }
        if (LocalDate.class.equals(convertType)) {
            return timestamp.toInstant().atZone(ZoneId.systemDefault()).toLocalDate();
        }
        if (LocalTime.class.equals(convertType)) {
            return timestamp.toInstant().atZone(ZoneId.systemDefault()).toLocalTime();
        }
       // need to handle the OffsetDateTime
        if (OffsetDateTime.class.equals(convertType)) {
            return timestamp.toInstant().atZone(ZoneId.systemDefault()).toOffsetDateTime();
        }
        return value;
    }
```

1. define your model with date type OffsetDateTime

   ```java
   public class Order {
       private Long orderId;
   
       private Long userId;
   
       private Integer status;
   
       private OffsetDateTime createTime;
   }
   ```

2. just use mybatis and execute query like:`select * from order where order_id=xx`

3. then you will be got the exception

   

I had fixed the bug in my project.

irwinai:
Fixes #20568.

- Fix bugs when variable type is OffsetDateTime

### DIFF
--- a/shardingsphere-infra/shardingsphere-infra-executor/src/main/java/org/apache/shardingsphere/infra/executor/sql/execute/result/query/impl/driver/jdbc/type/util/ResultSetUtil.java
+++ b/shardingsphere-infra/shardingsphere-infra-executor/src/main/java/org/apache/shardingsphere/infra/executor/sql/execute/result/query/impl/driver/jdbc/type/util/ResultSetUtil.java
@@ -31,10 +31,7 @@ import java.sql.SQLException;
 import java.sql.SQLFeatureNotSupportedException;
 import java.sql.Time;
 import java.sql.Timestamp;
-import java.time.LocalDate;
-import java.time.LocalDateTime;
-import java.time.LocalTime;
-import java.time.ZoneId;
+import java.time.*;
 import java.util.Date;
 
 /**
@@ -147,6 +144,9 @@ public final class ResultSetUtil {
         }
         if (LocalTime.class.equals(convertType)) {
             return timestamp.toInstant().atZone(ZoneId.systemDefault()).toLocalTime();
+        }
+        if(OffsetDateTime.class.equals(convertType)){
+            return timestamp.toInstant().atZone(ZoneId.systemDefault()).toOffsetDateTime();
         }
         return value;
     }
--- a/shardingsphere-infra/shardingsphere-infra-executor/src/main/java/org/apache/shardingsphere/infra/executor/sql/execute/result/query/impl/driver/jdbc/type/util/ResultSetUtil.java
+++ b/shardingsphere-infra/shardingsphere-infra-executor/src/main/java/org/apache/shardingsphere/infra/executor/sql/execute/result/query/impl/driver/jdbc/type/util/ResultSetUtil.java
@@ -31,7 +31,11 @@ import java.sql.SQLException;
 import java.sql.SQLFeatureNotSupportedException;
 import java.sql.Time;
 import java.sql.Timestamp;
-import java.time.*;
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.LocalTime;
+import java.time.OffsetDateTime;
+import java.time.ZoneId;
 import java.util.Date;
 
 /**
@@ -145,7 +149,7 @@ public final class ResultSetUtil {
         if (LocalTime.class.equals(convertType)) {
             return timestamp.toInstant().atZone(ZoneId.systemDefault()).toLocalTime();
         }
-        if(OffsetDateTime.class.equals(convertType)){
+        if (OffsetDateTime.class.equals(convertType)) {
             return timestamp.toInstant().atZone(ZoneId.systemDefault()).toOffsetDateTime();
         }
         return value;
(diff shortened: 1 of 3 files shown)


shardingsphere-infra/shardingsphere-infra-executor/src/main/java/org/apache/shardingsphere/infra/executor/sql/execute/result/query/impl/driver/jdbc/type/util/ResultSetUtil.java
```diff
@@ -31,10 +31,7 @@
 import java.sql.SQLFeatureNotSupportedException;
 import java.sql.Time;
 import java.sql.Timestamp;
-import java.time.LocalDate;
-import java.time.LocalDateTime;
-import java.time.LocalTime;
-import java.time.ZoneId;
+import java.time.*;
```
linghengqian, line 34: Should be changed to single class import, which breaks CI.

shardingsphere-jdbc/shardingsphere-jdbc-core/src/main/java/org/apache/shardingsphere/driver/jdbc/core/resultset/ShardingSphereResultSet.java
```diff
@@ -407,7 +408,7 @@ public <T> T getObject(final int columnIndex, final Class<T> type) throws SQLExc
             return (T) getBlob(columnIndex);
         } else if (Clob.class.equals(type)) {
             return (T) getClob(columnIndex);
-        } else if (LocalDateTime.class.equals(type) || LocalDate.class.equals(type) || LocalTime.class.equals(type)) {
```
linghengqian, line 410: The judgment condition of this piece is too complicated. I think it is better to judge only `java.time.temporal.TemporalAccessor`.

shardingsphere-infra/shardingsphere-infra-executor/src/main/java/org/apache/shardingsphere/infra/executor/sql/execute/result/query/impl/driver/jdbc/type/util/ResultSetUtil.java
```diff
@@ -148,6 +145,9 @@ private static Object convertTimestampValue(final Object value, final Class<?> c
         if (LocalTime.class.equals(convertType)) {
             return timestamp.toInstant().atZone(ZoneId.systemDefault()).toLocalTime();
         }
+        if(OffsetDateTime.class.equals(convertType)){
+            return timestamp.toInstant().atZone(ZoneId.systemDefault()).toOffsetDateTime();
+        }
```
linghengqian, line 150: Add a unit test?

strongduanmu (approved):
@irwinai Thank you for your contribution. 👍

PR #16923, opened 2022-04-19
Add authority configuration in server.yaml

lanchengx:
Changes proposed in this pull request:
- Add authority configuration

PR #20620, opened 2022-08-28
add md5 support for openGauss

### LINKED ISSUE #20619: support MD5 authentication for proxy frontend.
## Feature Request

I want use pg driver to connect ss-proxy, but in backend database type is openGauss.

No.

the reason is pg driver default authentication is  MD5, but when backend database type is openGauss, only support SHA256.
I want configure ss-proxy support both MD5 and SHA256 authentication。

justbk2015:
Fixes #20619.

- This PR is both  support MD5 and SHA256 authentication  in ss-proxy.
- In default configure, it is same as before SHA256 method.
- To suport MD5 authentication  method, you just need config server.yaml to add ":md5" in Authrity Rule.

example:
rules:
  - !AUTHORITY
  users:
    - root@%:root
    - sharding@:sharding
    - sharding1@%:sharding1:sha256 # use sha256 in default
    - sharding2@%:sharding2:md5 # use md5 in default
    -  sharding2@%:sharding2:aaa   # use sha256 in default
   
notice:
1. this feature effective only in opengauss database on backend.
2. only support ":md5" for md5 method(ignore case), other use default SHA256.

now, you can use postgresql jdbc driver in frontend and backend database type is openGauss!

### DIFF
--- a/shardingsphere-infra/shardingsphere-infra-common/src/main/java/org/apache/shardingsphere/infra/metadata/user/ShardingSphereUser.java
+++ b/shardingsphere-infra/shardingsphere-infra-common/src/main/java/org/apache/shardingsphere/infra/metadata/user/ShardingSphereUser.java
@@ -31,8 +31,15 @@ public final class ShardingSphereUser {
     
     private final String password;
     
+    private final String authType;
+    
     public ShardingSphereUser(final String username, final String password, final String hostname) {
+        this(username, password, hostname, null);
+    }
+    
+    public ShardingSphereUser(final String username, final String password, final String hostname, final String authType) {
         grantee = new Grantee(username, hostname);
         this.password = password;
+        this.authType = authType;
     }
 }
--- a/shardingsphere-infra/shardingsphere-infra-common/src/main/java/org/apache/shardingsphere/infra/metadata/user/yaml/config/YamlUserConfiguration.java
+++ b/shardingsphere-infra/shardingsphere-infra-common/src/main/java/org/apache/shardingsphere/infra/metadata/user/yaml/config/YamlUserConfiguration.java
@@ -34,8 +34,14 @@ public final class YamlUserConfiguration implements YamlConfiguration {
     
     private String password;
     
+    private String authorityType;
+    
     @Override
     public String toString() {
-        return username + "@" + (null == hostname ? "%" : hostname) + ":" + password;
+        String result = username + "@" + (null == hostname ? "%" : hostname) + ":" + password;
+        if (authorityType != null && !"".equals(authorityType.trim())) {
+            result += ":" + authorityType;
+        }
+        return result;
     }
 }
--- a/shardingsphere-infra/shardingsphere-infra-common/src/main/java/org/apache/shardingsphere/infra/metadata/user/yaml/config/YamlUsersConfigurationConverter.java
+++ b/shardingsphere-infra/shardingsphere-infra-common/src/main/java/org/apache/shardingsphere/infra/metadata/user/yaml/config/YamlUsersConfigurationConverter.java
@@ -79,13 +79,30 @@ public final class YamlUsersConfigurationConverter {
                 "user configuration `%s` is invalid, the configuration format should be like `username@hostname:password`", yamlUser);
         Preconditions.checkArgument(yamlUser.indexOf("@") < yamlUser.indexOf(":"),
                 "user configuration `%s` is invalid, the configuration format should be like `username@hostname:password`", yamlUser);
-        String username = yamlUser.substring(0, yamlUser.indexOf("@"));
-        String hostname = yamlUser.substring(yamlUser.indexOf("@") + 1, yamlUser.indexOf(":"));
-        String password = yamlUser.substring(yamlUser.indexOf(":") + 1);
+        // Match for user@host:pwd:authType, but authType is optional.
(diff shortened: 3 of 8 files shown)


TeslaCN:
Hi @justbk2015 Thanks for your contribution.
This PR made a little change to configuration format. Please @terrymanu help check this.

TeslaCN:
Hi @justbk2015 
The CI failed. Please take a look.

RaigorJiang:
Hi @justbk2015 @TeslaCN 
We have reserved a new YAML configuration format to support enabling different authentication methods for different users. 
In addition, the new API is placed outside the rule in server.yaml.
The reference configuration format is:
````yaml
mode:
  type: Cluster
  repository:
    type: ZooKeeper
    props:
      ...
  overwrite: true

authority:
  users:
    - user: root@%
      password: root
    - user: sharding@%
      password: sharding
      auth:  # MD5, TEXT, SHA256 ...
  privilege:
    type: ALL_PRIVILEGES_PERMITTED

rules:
  - !TRANSACTION
    ...
  - !SQL_PARSER
    ...
props:
````
The configuration class is [YamlAuthorityUserConfiguration](https://github.com/apache/shardingsphere/blob/master/shardingsphere-kernel/shardingsphere-authority/shardingsphere-authority-core/src/main/java/org/apache/shardingsphere/authority/yaml/config/YamlAuthorityUserConfiguration.java)
Original PR was #16923 (converted to current format for compatibility)
I'm also planning to completely convert the old configuration to the new way in the near future, maybe some discussion is needed.

TeslaCN:
@RaigorJiang 
Thanks for your reply. Actually I searched issues about authority API yesterday but found no issue or discussion about the changes in #16932. It's better to file an issue or discussion about any API changes next time.

justbk2015:
> Hi @justbk2015 @TeslaCN We have reserved a new YAML configuration format to support enabling different authentication methods for different users. In addition, the new API is placed outside the rule in server.yaml. The reference configuration format is:
> 
> ```yaml
> mode:
>   type: Cluster
>   repository:
>     type: ZooKeeper
>     props:
>       ...
>   overwrite: true
> 
> authority:
>   users:
>     - user: root@%
>       password: root
>     - user: sharding@%
>       password: sharding
>       auth:  # MD5, TEXT, SHA256 ...
>   privilege:
>     type: ALL_PRIVILEGES_PERMITTED
> 
> rules:
>   - !TRANSACTION
>     ...
>   - !SQL_PARSER
>     ...
> props:
> ```
> 
> The configuration class is [YamlAuthorityUserConfiguration](https://github.com/apache/shardingsphere/blob/master/shardingsphere-kernel/shardingsphere-authority/shardingsphere-authority-core/src/main/java/org/apache/shardingsphere/authority/yaml/config/YamlAuthorityUserConfiguration.java) Original PR was #16923 (converted to current format for compatibility) I'm also planning to completely convert the old configuration to the new way in the near future, maybe some discussion is needed.

you PR seem not compatibility in old version,I don't quite agree with this; in additidion, more complex configuration in you PR, it's hard for users.

RaigorJiang:
> It's better to file an issue or discussion about any API changes next time.

@TeslaCN  Yes, you are right, there should be a description of the API.

RaigorJiang:
> you PR seem not compatibility in old version,I don't quite agree with this; in additidion, more complex configuration in you PR, it's hard for users.

@justbk2015 In fact, `auth` is an optional configuration, for ordinary users, just split username and password.
Another reason for split is that in some cases `password` is not necessary, such as accessing a third-party authentication system.

TeslaCN:
Hi @RaigorJiang 
Could you provide some suggestion about this PR?

RaigorJiang:
> Hi @RaigorJiang Could you provide some suggestion about this PR?

@TeslaCN 
Authority is a kernel configuration, and I do agree with what @justbk2015 said, when changing API, should consider compatibility with older version.
But there are several factors that drive us to consider new APIs:
1. Specify different authentication methods for users
2. Users can have no password
When the second scenario occurs, the configuration of this PR will not support it.

So I propose that the PMCs discuss this and give advice?

justbk2015:
> > Hi @RaigorJiang Could you provide some suggestion about this PR?
> 
> @TeslaCN Authority is a kernel configuration, and I do agree with what @justbk2015 said, when changing API, should consider compatibility with older version. But there are several factors that drive us to consider new APIs:
> 
> 1. Specify different authentication methods for users
> 2. Users can have no password
>    When the second scenario occurs, the configuration of this PR will not support it.
> 
> So I propose that the PMCs discuss this and give advice?

is there some PMCs involves? any suggestion?

TeslaCN:
> > Hi @RaigorJiang Could you provide some suggestion about this PR?
> 
> @TeslaCN Authority is a kernel configuration, and I do agree with what @justbk2015 said, when changing API, should consider compatibility with older version. But there are several factors that drive us to consider new APIs:
> 
> 1. Specify different authentication methods for users
> 2. Users can have no password
>    When the second scenario occurs, the configuration of this PR will not support it.
> 
> So I propose that the PMCs discuss this and give advice?

Technical proposal could be discuss in dev mailing list. Every one could start or participant in discussion.

RaigorJiang:
@justbk2015 
Sorry, I will close this PR because the authority configuration has been modified.
The new configuration reserves the option of auth, which will be beneficial to achieve the desired functionality of this PR.

PR #20651, opened 2022-08-30
Add `CREATE DB_DISCOVERY TYPE` docs.

JiekerTime:

PR #20671, opened 2022-08-31
Change railroad diagram's identifier to string

JiekerTime:


RaigorJiang:
Hi @JiekerTime 
Why we should change them to string?

JiekerTime:
> Hi @JiekerTime Why we should change them to string?

I'm not sure, maybe you can look at the conversation of https://github.com/apache/shardingsphere/pull/20651

RaigorJiang (approved):

PR #20771, opened 2022-09-04
add resource bug

### LINKED ISSUE #20769: DistSQL 'add resource' will delete all resources in the original database
master

ShardingSphere-Proxy

image

natehuangting:
Fixes #20769 .


zjcnb:
@natehuangting Hi, this pr has some problem, The createNewDataSource method are return an incremental data source.

natehuangting:
> @natehuangting Hi, this pr has some problem, The createNewDataSource method are return an incremental data source.

Thanks! But I think this modification is more intuitive and convenient, and there are less unnecessary calculations. Maybe we can refactor the method name?

Issue #17939, opened 2022-05-25
[GSOC 2022] Develop an external tool to convert YAML configuration DistSQL script


isHuangXin:
This issue is created to better track my PRs for GSoC 2022.
You can find more details about my project [here](https://summerofcode.withgoogle.com/programs/2022/projects/4PnSweAf)

## Background
Apache ShardingSphere is positioned as a Database Plus, and aims at building a standard layer and ecosystem above heterogeneous databases. It focuses on how to reuse existing databases and their respective upper layer, rather than creating a new database. The goal is to minimize or eliminate the challenges caused by underlying databases fragmentation.

Below is my completion plan 

step 1: Parsing YAML files.
Step2: Figure out what we need to complete the output DistSQL RDL  is 'ADD' and 'CREATE' syntax from parsing YAML files.
Step3: Wrap a command line tool to complete the conversion of YAML to DistSQL without starting a proxy.

- [x] ADD RESOURCE dataSource
- [x] CREATE SHARDING TABLE RULE 
- [x] Sharding Binding Table Rule
- [x] Sharding Broadcast Table Rule
- [x] Sharding Scaling Rule
- [x] CREATE READWRITE_SPLITTING RULE
- [x] CREATE DB_DISCOVERY RULE
- [x] CREATE ENCRYPT RULE
- [x] CREATE SHADOW RULE

RaigorJiang:
Hi @isHuangXin ,
Nice to see you here!  But the link you provided cannot be accessed (access denied).

isHuangXin:
> Hi @isHuangXin , Nice to see you here! But the link you provided cannot be accessed (access denied).

Sry for late reply. The link has been updated, is it accessible now?

RaigorJiang:
> Sry for late reply. The link has been updated, is it accessible now?

Yes, the new link is accessible.

TeslaCN:
This issue is not finished yet.

isHuangXin:
Hi, @RaigorJiang 
This issue about GSoC 2022 has been completed 🎉.
Thank you for the great guidance, which has made me learn a lot. 

I will continue to contribute code to ShardingSphere, here is my GSoC 2022 final report link: https://github.com/isHuangXin/GSoC-2022-DistSQL-Convert-Tool

RaigorJiang:
@isHuangXin 
Sorry, this CI failure has not been completed
https://github.com/apache/shardingsphere/pull/20843

RaigorJiang:
[...]

PR #20843, opened 2022-09-07
Add convert shadow distsql and test case

isHuangXin:
For #17939

- Add convert shadow distsql
- Add test case


isHuangXin:
This is so weird, I have adjusted the order of the parameters in the shadow tables properties.
The picture is the result of the printout of my code.

Screen Shot 2022-09-08 at 09 04 10

RaigorJiang:
> About CI error, this is so weird. I have adjusted the order of the parameters in the shadow tables properties. The picture is the result of the printout of my code.
> 
> Screen Shot 2022-09-08 at 09 04 10
> 
> Screen Shot 2022-09-08 at 09 23 05

Hi @isHuangXin 
Because Properties is an inherited HashTable, keys are not guaranteed in order.
Can we compare line by line here, and then use contains to make an assertion on the Properties line

isHuangXin:
Hi, @RaigorJiang. 
Thanks for your timely code review comments. I have updated the code according to your comments on the slack channel.

RaigorJiang (approved):

PR #20847, opened 2022-09-07
Modify transaction integration test trigger conditions

FlyingZC:
- Modify transaction integration test trigger conditions

### DIFF
--- a/.github/workflows/it-transaction.yml
+++ b/.github/workflows/it-transaction.yml
@@ -22,6 +22,10 @@ on:
     branches: [ master, dev ]
     paths:
       - '.github/workflows/it-transaction.yml'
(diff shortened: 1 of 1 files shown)


.github/workflows/it-transaction.yml
```diff
@@ -42,6 +50,8 @@ on:
       - 'shardingsphere-test/shardingsphere-integration-test/shardingsphere-integration-test-env/**'
       - 'shardingsphere-test/shardingsphere-integration-test/shardingsphere-integration-test-transaction/**'
       - '!*.md'
+  schedule:
```
taojintianxia, line 53: schedule is not that useful here.

FlyingZC: You are right.

strongduanmu (approved):

PR #20956, opened 2022-09-13
Fix mounted wrong server.yaml path and distinguish the assertion results of different mode.

### LINKED ISSUE #20932: ProxyClusterContainerConfigurationFactory not get cluster server.yaml configuration
I found not persist metadata to governance center when i run integration test case with cluster mode. And i see `ProxyClusterContainerConfigurationFactory.class` get `server.yaml` configuration are not right. when i adjustment path to `/env/common/cluster/proxy/conf/`, there are many integration test case are run failed.  

image

Relate pull request #20103

`master` https://github.com/apache/shardingsphere/commit/f7e511be1b976124e327a3a0f93315552b08d895

### Which project did you use? 

`shardingsphere-test`

### Expected behavior

`shardingsphere-integration-test-suite` test case can run really in cluster mode.

### Actual behavior

image

```
public final class ProxyClusterContainerConfigurationFactory {
    
    /**
     * Create instance of adaptor container configuration.
     * 
     * @param scenario scenario
     * @param databaseType database type
     * @return created instance
     */
    public static AdaptorContainerConfiguration newInstance(final String scenario, final DatabaseType databaseType) {
        return new AdaptorContainerConfiguration(scenario, getMountedResources(scenario, databaseType));
    }
    
    private static Map<String, String> getMountedResources(final String scenario, final DatabaseType databaseType) {
        Map<String, String> result = new HashMap<>(2, 1);
        String pathInContainer = "/opt/shardingsphere-proxy/conf";
        result.put("/env/common/standalone/proxy/conf/", pathInContainer);
        result.put("/env/scenario/" + scenario + "/proxy/conf/" + databaseType.getType().toLowerCase(), pathInContainer);
        return result;
    }
}
```

`/env/common/standalone/proxy/conf/` should be change to `/env/common/cluster/proxy/conf/`

zhaojinchao95:
For #20932


zhaojinchao95:
image

tuichenchuxin (approved):

PR #20995, opened 2022-09-15
Add refresh data source states(#20738)

### LINKED ISSUE #20738: Support using disabled data source at startup
Now we can use DistSQL to disable the datasource's state in read-write splitting rules.
For example,proxy A has disabled the data source.Then proxy B is started with the same configuration, and the status of the disabled data source will still be checked, resulting in startup failure.

In order to ensure that the proxy can startup normally, we need to do the following tasks:

- [ ] disabled data sources will not be checked when startup
- [ ] use `--force` program parameters to ensure that the proxy starts normally regardless of whether it can be connected or not

FlyingZC:
Fixes #20738.

  - Refresh the data source state when modifying the read-write splitting data source state


FlyingZC:
image

strongduanmu (approved):

PR #21120, opened 2022-09-21
Bump Java from 18 to 19 in CI

TeslaCN:


TeslaCN:
Java 19 is not available in actions/setup-java yet.

PR #17712, opened 2022-05-16
Use caffeine instead of guava cache

longtai-cn:
Fixes #17684.

Changes proposed in this pull request:
- Use caffeine instead of guava cache.

PR #21149, opened 2022-09-23
Add optimized plan cache.

boyjoy1127:
Ref #19937

  - add plan cache.
  - fix conflicts when merging code from master.


shardingsphere-kernel/shardingsphere-sql-federation/shardingsphere-sql-federation-optimizer/src/main/java/org/apache/shardingsphere/sqlfederation/optimizer/OptimizedPlanManagement.java
```diff
@@ -0,0 +1,75 @@
...
+    
+    public OptimizedPlanManagement(final SQLOptimizeEngine optimizer, final SqlValidator validator, final SqlToRelConverter converter) {
+        this.cache = buildCache();
+        this.optimizer = optimizer;
+        this.validator = validator;
+        this.converter = converter;
+    }
+    
+    private Cache<String, SQLOptimizeContext> buildCache() {
+        return CacheBuilder.newBuilder()
```
TeslaCN, line 52: How about using Caffeine cache?

boyjoy1127: Why use Caffine cache, does it have any benefits? @TeslaCN

TeslaCN: * The performance of Caffeine is better than Guava in most scenario. 
* And ShardingSphere is reducing usage of Guava. 

The SQL parser cache has been migrated to Caffeine in https://github.com/apache/shardingsphere/pull/17712.

PR #21207, opened 2022-09-26
Fix max-connections-size-per-query in transaction(#21099)

### LINKED ISSUE #21099: Local transaction will be broken while max-connections-size-per-query greater than 1
## Bug Report

- 5.1.0

- ShardingSphere-JDBC

### Expected behavior
- `DriverExecutionPrepareEngine` return one connection if statement in a local transaction

### Actual behavior
- `DriverExecutionPrepareEngine` return more than one connection if max-connections-size-per-query greater than 1

image

- Absence of LocalTransactionManager

#### Demo project
- **repo url:** https://github.com/I-Love-China/shardingsphere
- **branch:** `comment-global`
- **main entry:** `org.apache.shardingsphere.shardingspheredemo.service.HealthRecordServiceTest#test`
- **DDL:** https://github.com/I-Love-China/shardingsphere/blob/comment-global/shardingsphere-demo/src/main/resources/sql/DDL.sql
- There will be a warning while running this unit test: `expected connections: 1, actual connections: 3`;

image

FlyingZC:
Fixes #21099.

  - Fix max-connections-size-per-query in transaction
---

- [x] I have passed maven check: `mvn clean install -B -T2C -DskipTests -Dmaven.javadoc.skip=true -e`.


shardingsphere-jdbc/shardingsphere-jdbc-core/src/main/java/org/apache/shardingsphere/driver/jdbc/core/statement/ShardingSpherePreparedStatement.java
```diff
@@ -307,7 +307,9 @@ private ResultSet executeFederationQuery(final QueryContext queryContext) throws
     }
     
     private DriverExecutionPrepareEngine<JDBCExecutionUnit, Connection> createDriverExecutionPrepareEngine() {
-        int maxConnectionsSizePerQuery = metaDataContexts.getMetaData().getProps().<Integer>getValue(ConfigurationPropertyKey.MAX_CONNECTIONS_SIZE_PER_QUERY);
+        int maxConnectionsSizePerQuery = connection.isHoldTransaction()
```
strongduanmu, line 310: Did you consider the possible extra memory usage when maxConnectionsSizePerQuery = 1?

strongduanmu (changes requested):

strongduanmu:
Hi @FlyingZC, can you solve code conflict?

FlyingZC:
By the pr https://github.com/apache/shardingsphere/pull/21207. It guarantees that one transaction on a dataSource use the same connection, but may cause more memory consumption, because due to the limit of the number of connections, it may cause that streaming loading cannot be used, and memory loading is used.
It is difficult to choose between the two(one connection in trasaction to enable ACID or more connections to avoid memory consumption), which may require more design, but perhaps users can make their own trade-offs. When user configures this parameter `max-connections-size-per-query`, he should be aware of the possible advantages and disadvantages.

shardingsphere-jdbc/shardingsphere-jdbc-core/src/main/java/org/apache/shardingsphere/driver/jdbc/core/statement/ShardingSpherePreparedStatement.java
I-Love-China: > Did you consider the possible extra memory usage when maxConnectionsSizePerQuery = 1?

Correctness is the most fundamental requirement.

strongduanmu: The ShardingSphere local transaction manager does not guarantee transaction correctness when routing to multiple data nodes. If required, use XA transactions to guarantee transactional correctness across multiple data nodes.

natehuangting: @I-Love-China 
In the ss design, local does not guarantee any features. If you need to ensure transactionality, I think you need to develop some functions yourself. For example, if multiple data sources are of the same instance, you can use only one connection, and it is forbidden to perform cross-shard queries in transactions (I don't think it is necessary to perform a large batch of queries in a transaction, because you can already determine what shards you want to update, which needs to be acceptable to your business). If your data is small, you can also consider loading all into memory.

I-Love-China: > The ShardingSphere local transaction manager does not guarantee transaction correctness when routing to multiple data nodes. If required, use XA transactions to guarantee transactional correctness across multiple data nodes.

But these data nodes are located at the same database instance.

I-Love-China: > @I-Love-China In the ss design, local does not guarantee any features. If you need to ensure transactionality, I think you need to develop some functions yourself. For example, if multiple data sources are of the same instance, you can use only one connection, and it is forbidden to perform cross-shard queries in transactions (I don't think it is necessary to perform a large batch of queries in a transaction, because you can already determine what shards you want to update, which needs to be acceptable to your business). If your data is small, you can also consider loading all into memory.

I am so confused. According to this, local transactions will be totally destroyed.

Issue #20736, opened 2022-09-02
Support SQL HINT


tuichenchuxin:
## Feature Request

We have `HintManager` for users to use for hint route, But it's depends on `ThreadLocal`.

We want to use SQL hint to replace HintManager for more convenient to use.

## replace `HintManager.setDatabaseShardingValue`
```
/* SHARDINGSPHERE_HINT: SHARDING_DATABASE_VALUE=1 */
SELECT * FROM t_order,t_order_item;
```
## replace `HintManager.addDatabaseShardingValue` `HintManager.addTableShardingValue`
```
/* SHARDINGSPHERE_HINT: t_order.SHARDING_DATABASE_VALUE=100,t_order.SHARDING_TABLE_VALUE=100 */
SELECT * FROM t_order,t_order_item;
```
## replace `HintManager.setWriteRouteOnly` mix sharding
```
/* SHARDINGSPHERE_HINT: t_order.SHARDING_DATABASE_VALUE=100,t_order.SHARDING_TABLE_VALUE=100,t_order_item.SHARDING_DATABASE_VALUE=100,WRITE_ROUTE_ONLY=true */
SELECT * FROM t_order,t_order_item;
```

## for better performance

Since we have SQL parse cache, shardingSphere have a good performance. But if we have same SQL with different comment, we can not use the parse cache. So it's better to extract hint in proxy and jdbc, and remove the hint message from sql before parsing sql.
 
## tasks

- [x] support sql hint for sharding
- [x] refactor sql hint for readwrite
- [x] move shadow comment extracted to sql hint
- [ ] better performance for sql hint

dongzl:
I want to try it. If others are interested, we can do it together.

strongduanmu:
Hi @tuichenchuxin @dongzl , could we finish this issue in 5.2.1?

dongzl:
> Hi @tuichenchuxin @dongzl , could we finish this issue in 5.2.1?

Hi @strongduanmu ，what's the deadline for Release 5.2.1? I think only part of the work can be done.

strongduanmu:
Since this task still has some subtasks unfinished, I will move it to 5.2.2.

lanrenbulan:
Hi, In proxy5.2.1, I tested SQL hint, but it didn't work. #21437

TeslaCN:
Hi @tuichenchuxin @dongzl 
Have we finished this?

sandynz:
[...]

PR #21346, opened 2022-10-05
Support sharding SQL hint.

dongzl:
ref #20736 .

- support sharding sql hint.
- add test case.


tuichenchuxin (commented):
I think SQL hint is same as HintManager, the only difference between them is just setting or getting the value way is different.

tuichenchuxin (commented):
LGTM. By the way, why databaseValue is only int? It should be comparable is ok.

dongzl:
> LGTM. By the way, why databaseValue is only int? It should be comparable is ok.

OK, fixes it.

tuichenchuxin (approved):

features/sharding/core/src/main/java/org/apache/shardingsphere/sharding/route/engine/type/standard/ShardingStandardRoutingEngine.java
```diff
@@ -94,6 +98,17 @@ private boolean isRoutingByHint(final ShardingRule shardingRule, final TableRule
                 && shardingRule.getTableShardingStrategyConfiguration(tableRule) instanceof HintShardingStrategyConfiguration;
     }
     
+    private boolean isRoutingBySQLHint() {
+        boolean result = false;
+        if (sqlStatementContext instanceof CommonSQLStatementContext) {
+            Collection<String> tableNames = sqlStatementContext.getTablesContext().getTableNames();
+            for (String each : tableNames) {
+                result = result || ((CommonSQLStatementContext<?>) sqlStatementContext).containsHintShardingValue(each);
```
strongduanmu, line 106: Can you use guard clause here? Just like:

```java
if (CommonSQLStatementContext<?>) sqlStatementContext).containsHintShardingValue(each) {
return true;
}
```

strongduanmu (changes requested):

strongduanmu (approved):
Good job, merged.

PR #21431, opened 2022-10-09
Dynamic Loading SQL Parser Parameterized Test

### LINKED ISSUE #16383: Dynamic load SQL parser test cases from MySQL and PostgreSQL
The current SQL parser test case for unsupported SQL is `org.apache.shardingsphere.test.sql.parser.parameterized.engine.UnsupportedSQLParserParameterizedTest`. 
The test data is in `src/main/resources/sql/unsupported/unsupported.xml` which composed by static and dynamic.

image

We need to keep the static test cases in `UnsupportedSQLParserParameterizedTest` and create a new class, which name maybe `DynamicLoadingSQLParserParameterizedTest`.

The class should load SQL cases from official test cases from MySQL[1] and PostgreSQL[2]. 
The step of `DynamicLoadingSQLParserParameterizedTest`:

1. Load official test cases from GitHub
2. Generate to arrays by parameterized test
3. Run `SQLParserEngine.parse()` try to parse SQL
4. Assert result as `SQLStatement`
5. Catch `SQLParserException` for unsupported SQL
6. Catch `CastException` for parse error
7. Record and print exception number, the final objective is zero exception, please do not assert fail if exception number is not zero, just let the test cases can be done.
8. Please define the test case as IT, it is better to run it in GitHub Action only, the normal `mvn install` should not depend on network environment.

[1] https://github.com/mysql/mysql-server/tree/8.0/mysql-test/t
[2] https://github.com/postgres/postgres/tree/master/src/test/regress/sql

Jacob953:
Fixes #16383.

  - create DynamicLoadingSQLParserParameterizedTest
  - define the test case as IT, and run it in GitHub Action only.

---

- [x] I have passed maven check: `mvn clean install -B -T2C -DskipTests -Dmaven.javadoc.skip=true -e`.


.github/workflows/it.yml
```diff
@@ -226,3 +226,34 @@ jobs:
...
+            ${{ runner.os }}-maven-
+      - uses: actions/setup-java@v2
+        with:
+          distribution: 'temurin'
+          java-version: 8
+      - name: Build Project
+        run: ./mvnw -B clean install -DskipITs -DskipTests
+      - name: Run Parameterized Test
+        run: |
+          ./mvnw -B clean install -f test/parser/pom.xml -Pit.dynamic.loading -Dspotless.apply.skip=true -Ddatabase=${{ matrix.database }} -Durl=${{ matrix.url }}
```
terrymanu, line 259: It is better to put business logic into java codes with junit, just keep GitHub action simple.
It could be better to run in local `mvn install` too.

test/parser/src/main/java/org/apache/shardingsphere/test/sql/parser/parameterized/engine/DynamicLoadingSQLParserParameterizedTest.java
```diff
@@ -0,0 +1,145 @@
...
+        sqlCaseURL = System.getProperty("url");
+        databaseType = System.getProperty("database");
+    }
+    
+    private static String[] getSqlCaseFiles(final String sqlCaseURL) throws IOException {
+        Document document = Jsoup.connect(sqlCaseURL).get();
+        return document.getElementsByClass(SQL_CASE_FILE_MARKER).text().split("\\s+");
+    }
+    
+    private static Collection<Object[]> getTestParameters(final String databaseType) throws IOException {
```
terrymanu, line 67: What is the useful of parameter `databaseType`?

pom.xml
```diff
@@ -691,6 +693,19 @@
...
+                <artifactId>jsoup</artifactId>
+                <version>${jsoup.version}</version>
+            </dependency>
+
+            <dependency>
+                <groupId>org.testng</groupId>
+                <artifactId>testng</artifactId>
+                <version>${testng.version}</version>
+                <scope>compile</scope>
+            </dependency>
```
terrymanu, line 708: Is it necessary to import new testing framework?

pom.xml
```diff
@@ -691,6 +693,19 @@
                 <artifactId>caffeine</artifactId>
                 <version>${caffeine.version}</version>
             </dependency>
+
+            <dependency>
+                <groupId>org.jsoup</groupId>
+                <artifactId>jsoup</artifactId>
+                <version>${jsoup.version}</version>
+            </dependency>
```
terrymanu, line 701: How about using GitHub API to instead of jsoup to fetch sql text?

terrymanu (changes requested):

Jacob953:
Implement steps:
- Get all `download_url`s of SQL case files via github-api at once through getTestParameters.
- Append each SQL ended up with `;` and ID, which is generated by the SQL case file and case number, to an array.
- Pass the array to assertDynamicLoadingSQL at DynamicLoading(MySQL/PostgreSQL)ParserParameterizedTest.
- Each SQL will go through the parsing engine
    - if successful, the case would pass at `new SQLVisitorEngine`.
    - else, print id, value, and db-type of SQL and throw non-RuntimeException to make the test case pass with @SingletonSPI
image

.mvn/jvm.config
```diff
@@ -1 +1 @@
--Xmx2g -XX:MaxMetaspaceSize=512m
```
terrymanu, line 1: Why need to increase so much mem here?

pom.xml
```diff
@@ -153,7 +153,8 @@
         <git-commit-id-plugin.version>4.9.10</git-commit-id-plugin.version>
         <freemarker.version>2.3.31</freemarker.version>
         <spotless-maven-plugin.version>2.22.1</spotless-maven-plugin.version>
-        <caffeine.version>2.9.3</caffeine.version>
+        <caffeine.version>2.9.2</caffeine.version>
```
terrymanu, line 156: Is it necessary to downgrade caffeine?

pom.xml
```diff
@@ -688,6 +689,11 @@
                 <version>${springframework.version}</version>
                 <scope>test</scope>
             </dependency>
+            <dependency>
+                <groupId>org.springframework</groupId>
+                <artifactId>spring-web</artifactId>
+                <version>${springframework.version}</version>
+            </dependency>
```
terrymanu, line 696: ShardingSphere is a backend project, please do not involve spring-web dependencies

test/parser/src/main/java/org/apache/shardingsphere/test/sql/parser/parameterized/engine/DynamicLoadingSQLParserParameterizedTest.java
```diff
@@ -0,0 +1,97 @@
...
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.nio.charset.StandardCharsets;
+import java.util.Collection;
+import java.util.Map;
+import java.util.Properties;
+import java.util.ArrayList;
+
+@RequiredArgsConstructor
+@SingletonSPI
```
terrymanu, line 40: What is the useful?

test/parser/src/main/java/org/apache/shardingsphere/test/sql/parser/parameterized/engine/DynamicLoadingSQLParserParameterizedTest.java
```diff
@@ -0,0 +1,97 @@
...
+        String sqlCasesRepo = patches[4];
+        String sqlCasesDirectory = patches[7];
+        return new RestTemplate().getForObject(
+                "https://api.github.com/repos/{owner}/{repo}/contents/{directory}", ArrayList.class,
+                sqlCasesOwner, sqlCasesRepo, sqlCasesDirectory);
+    }
+    
+    protected static Collection<Object[]> getTestParameters(final String sqlCaseURL) throws IOException, URISyntaxException {
+        Collection<Object[]> result = new ArrayList<>();
+        ArrayList<Map<String, String>> response = getResponse(sqlCaseURL);
```
terrymanu, line 61: Please use List to instead of ArrayList here

test/parser/src/main/java/org/apache/shardingsphere/test/sql/parser/parameterized/engine/DynamicLoadingSQLParserParameterizedTest.java
```diff
@@ -0,0 +1,97 @@
...
+        String sqlCasesOwner = patches[3];
+        String sqlCasesRepo = patches[4];
+        String sqlCasesDirectory = patches[7];
+        return new RestTemplate().getForObject(
+                "https://api.github.com/repos/{owner}/{repo}/contents/{directory}", ArrayList.class,
+                sqlCasesOwner, sqlCasesRepo, sqlCasesDirectory);
+    }
+    
+    protected static Collection<Object[]> getTestParameters(final String sqlCaseURL) throws IOException, URISyntaxException {
+        Collection<Object[]> result = new ArrayList<>();
```
terrymanu, line 60: Please use LinkedList to instead of ArrayList here

test/parser/src/main/java/org/apache/shardingsphere/test/sql/parser/parameterized/engine/DynamicLoadingSQLParserParameterizedTest.java
```diff
@@ -0,0 +1,97 @@
...
+        return result;
+    }
+    
+    @Test(expected = Exception.class)
+    public final void assertDynamicLoadingSQL() {
+        CacheOption cacheOption = new CacheOption(128, 1024L);
+        ParseASTNode parseContext = new SQLParserEngine(databaseType, cacheOption).parse(sqlCaseValue, false);
+        new SQLVisitorEngine(databaseType, "STATEMENT", true, new Properties()).visit(parseContext);
+        System.out.println("ParserError: " + sqlCaseId + " value: " + sqlCaseValue + " db-type: " + databaseType);
+        throw new RuntimeException();
```
terrymanu, line 95: Why always throw exception?

terrymanu (changes requested):
How to distinguish the success SQL and failed SQL?

test/parser/pom.xml
```diff
@@ -117,5 +117,17 @@
...
+            <artifactId>spring-boot-starter</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>commons-io</groupId>
+            <artifactId>commons-io</artifactId>
+        </dependency>
```
terrymanu, line 131: Is it necessary to import the dependencies?

terrymanu (changes requested):

test/parser/src/main/java/org/apache/shardingsphere/test/sql/parser/parameterized/engine/DynamicLoadingSQLParserParameterizedTest.java
```diff
@@ -0,0 +1,97 @@
...
+    private final String sqlCaseId;
+    
+    private final String sqlCaseValue;
+    
+    private final String databaseType;
+    
+    private static LinkedList<Map<String, Object>> getResponse(final String sqlCaseURL) throws IOException {
+        String[] patches = sqlCaseURL.split("/", 8);
+        String url = "https://api.github.com/repos/" + patches[3] + "/" + patches[4] + "/contents/" + patches[7];
+        return new JsonMapper().readValue(new URL(url), new TypeReference<LinkedList<Map<String, Object>>>() {
```
terrymanu, line 55: Why need to spilt url, how about use the whole url?

test/parser/src/main/java/org/apache/shardingsphere/test/sql/parser/parameterized/engine/DynamicLoadingSQLParserParameterizedTest.java
```diff
@@ -0,0 +1,97 @@
...
+        });
+    }
+    
+    protected static Collection<Object[]> getTestParameters(final String sqlCaseURL) throws IOException, URISyntaxException {
+        Collection<Object[]> result = new LinkedList<>();
+        List<Map<String, Object>> response = getResponse(sqlCaseURL);
```
terrymanu, line 61: What the content in `Map<String, Object>` ?

test/parser/pom.xml
```diff
@@ -117,5 +117,9 @@
             <artifactId>shardingsphere-test-common</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+        </dependency>
```
terrymanu, line 123: Is it necessary to import it？

Jacob953:
test comment

.github/workflows/it.yml
Jacob953: done

test/parser/src/main/java/org/apache/shardingsphere/test/sql/parser/parameterized/engine/DynamicLoadingSQLParserParameterizedTest.java
Jacob953: fixed

pom.xml
Jacob953: fixed

test/parser/src/main/java/org/apache/shardingsphere/test/sql/parser/parameterized/engine/DynamicLoadingSQLParserParameterizedTest.java
terrymanu: @SingletonSPI is for SPI only. obviously, DynamicLoadingSQLParserParameterizedTest is not SPI. So the annotation is useless here.

test/parser/src/main/java/org/apache/shardingsphere/test/sql/parser/parameterized/engine/DynamicLoadingSQLParserParameterizedTest.java
```diff
@@ -0,0 +1,118 @@
...
+    
+    private final String databaseType;
+    
+    private static String[] getContentLines(final String url) throws IOException {
+        InputStreamReader in = new InputStreamReader(new URL(url).openStream());
+        String content = new BufferedReader(in).lines().collect(Collectors.joining(System.lineSeparator()));
+        return content.split("\n");
+    }
+    
+    private static LinkedList<Map<String, String>> getResponse(final String sqlCaseURL) throws IOException {
```
terrymanu, line 56: Please use interface (List) for return type declaration

test/parser/src/main/java/org/apache/shardingsphere/test/sql/parser/parameterized/engine/DynamicLoadingSQLParserParameterizedTest.java
terrymanu: Maybe you misunderstand the usage of @SingletonSPI.
@SingletonSPI is for create SPI with singleton way only.
I still wonder why always throw exception, it should be record failed SQL without any exception thrown

terrymanu (approved):

PR #21508, opened 2022-10-11
Migration job stop cascade latest consistency check job

sandynz:
  - Migration job stop cascade latest check job
  - Start consistency check job do not throw exception when check job is finished
  - Fix wrong column of remaining_seconds
  - Ignore related consistency check job error on start/stop/commit/rollback

---

- [ ] I have passed maven check: `mvn clean install -B -T2C -DskipTests -Dmaven.javadoc.skip=true -e`.

### DIFF
--- a/kernel/data-pipeline/core/src/main/java/org/apache/shardingsphere/data/pipeline/scenario/migration/MigrationJobAPIImpl.java
+++ b/kernel/data-pipeline/core/src/main/java/org/apache/shardingsphere/data/pipeline/scenario/migration/MigrationJobAPIImpl.java
@@ -250,12 +250,28 @@ public final class MigrationJobAPIImpl extends AbstractInventoryIncrementalJobAP
     @Override
     public void startDisabledJob(final String jobId) {
         super.startDisabledJob(jobId);
-        PipelineAPIFactory.getGovernanceRepositoryAPI().getCheckLatestJobId(jobId).ifPresent(optional -> ConsistencyCheckJobAPIFactory.getInstance().startDisabledJob(optional));
+        PipelineAPIFactory.getGovernanceRepositoryAPI().getCheckLatestJobId(jobId).ifPresent(optional -> {
+            try {
+                ConsistencyCheckJobAPIFactory.getInstance().startDisabledJob(optional);
+                // CHECKSTYLE:OFF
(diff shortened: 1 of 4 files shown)


kernel/data-pipeline/core/src/main/java/org/apache/shardingsphere/data/pipeline/scenario/migration/MigrationJobAPIImpl.java
```diff
@@ -266,7 +294,13 @@ private void dropCheckJobs(final String jobId) {
         log.info("dropCheckJobs start...");
         long startTimeMillis = System.currentTimeMillis();
         for (String each : checkJobIds) {
-            dropJob(each);
+            try {
+                dropJob(each);
+                // CHECKSTYLE:OFF
+            } catch (final RuntimeException ex) {
+                // CHECKSTYLE:ON
+                log.info("drop check job failed, check job id: {}, error: {}", each, ex.getMessage());
```
taojintianxia, line 302: why we use info level to print out the catched exception ?

yx9o (approved):

kernel/data-pipeline/core/src/main/java/org/apache/shardingsphere/data/pipeline/scenario/migration/MigrationJobAPIImpl.java
sandynz: Since it could be ignored, just record it here without stack trace

taojintianxia: > Since it could be ignored, just record it here without stack trace

I see. then that exception instance ex should be called `ignored`.

> [code of conduct](https://shardingsphere.apache.org/community/en/involved/conduct/code/) `Exceptions when catch are named with ex; Exceptions when catch but do nothing are named with ignored`

sandynz: But it's used in logging, if `ignored` is used, it shows warning in IDEA:
图片
The ignored case:
图片

Issue #21574, opened 2022-10-15
Refactor PreparedStatement in ShardingSphere-Proxy


TeslaCN:
## Refactor

https://github.com/apache/shardingsphere/blob/00f5553fa3b5ba2b772e9fecb58fe72f6ee4119e/proxy/backend/src/main/java/org/apache/shardingsphere/proxy/backend/session/PreparedStatement.java#L28

The current `PreparedStatement` naming coincides with the JDBC `PreparedStatement`. Consider adjusting the naming and improving the immutability of the implementation class.

- [x] Rename `PreparedStatement` to `ServerPreparedStatement`. #21577
- [x] Make all fields of implementations of PreparedStatement final.
- [x] Wrapping `EmptyStatement` and `DistSQLStatement` into `SQLStatementContext`.
- [ ] Consider moving `String sql` into SQLStatementContext.

TeslaCN:
Step 4 is a big change and is not for ShardingSphere-Proxy only, it may require further discussion. This issue has been finished.

PR #21577, opened 2022-10-15
Rename PreparedStatement to ServerPreparedStatement

TeslaCN:
Step 1 of #21574.


proxy/backend/src/main/java/org/apache/shardingsphere/proxy/backend/session/ServerPreparedStatement.java
```diff
@@ -25,7 +25,7 @@
 /**
  * Logic prepared statement for clients of ShardingSphere-Proxy.
```
terrymanu, line 26: Please revise javadoc

terrymanu (approved):

PR #21919, opened 2022-11-03
Use ConcurrentHashMap to instead of sync in DistributedLockHolder

terrymanu:


soulasuna (dismissed):

mode/type/cluster/repository/api/src/main/java/org/apache/shardingsphere/mode/repository/cluster/lock/holder/DistributedLockHolder.java
```diff
@@ -33,14 +35,14 @@ public final class DistributedLockHolder {
     
     private final TypedProperties<?> props;
     
-    private final Map<String, DistributedLock> locks;
+    private final ConcurrentHashMap<String, DistributedLock> locks;
```
TeslaCN, line 38: Should we use `ConcurrentMap` or `Map` here?

terrymanu: sure

yx9o (approved):

Issue #21518, opened 2022-10-12
process blob data error 


jitawangzi:
ShardingSphere-Proxy 5.2.0

It is a same issue with #18627  #3024

I test it in version 5.2.0, this problem is still not resolved.

The blob data stored is not the same as read from mysql.

TeslaCN:
Could you provide more information about this issue? Such as configurations, DDL, DML...

jitawangzi:
> #3024

Let me talk about the general process, i use ShardingSphere-Proxy 5.2.0, configure 5 data sources for sharding, then  create a table (without configuring sharding rules) including blob fields, insert  a record  into this table, specify the content of the blob field, and then query this data from the database, and find that the value of the blob of the queried data is different from the value when i inserted it.

More specific description and configuration 、test code, etc, can be found in the reply of @xriqyu in #3024. I hava  the same problem .

jitawangzi:
I reproduced the complete test code and running environment, test project address:
https://github.com/jitawangzi/ShardingSphereProxyBlobTest.git

Just run Main directly, and you can see data errors in the output：

```binary data equals result : false```

The test project also contains all the configuration of ShardingSphere,sql of create table,

I have modified the configuration  "config-sharding.yaml"


``` 
schemaName: test_sharding

dataSources:
  ds_0:
    url: jdbc:mysql://127.0.0.1:51234/test?serverTimezone=UTC&useSSL=false
    username: sharding
    password: sharding
    connectionTimeoutMilliseconds: 30000
    idleTimeoutMilliseconds: 60000
    maxLifetimeMilliseconds: 1800000
    maxPoolSize: 50
    minPoolSize: 1
```

and "server.yaml"

``` 

rules:
  - !AUTHORITY
     users:
      - sharding@:sharding
     provider:
      type: ALL_PRIVILEGES_PERMITTED
  - !TRANSACTION
     defaultType: LOCAL
  - !SQL_PARSER
     sqlCommentParseEnabled: true
     sqlStatementCache:
      initialCapacity: 2000
      maximumSize: 65535
     parseTreeCache:
      initialCapacity: 128
      maximumSize: 1024

props:
  max-connections-size-per-query: 3
  kernel-executor-size: 16
  proxy-frontend-flush-threshold: 128
  proxy-opentracing-enabled: false
  proxy-hint-enabled: false
  sql-show: true
  check-table-metadata-enabled: false
  show-process-list-enabled: false
[...]

Issue #3024, opened 2019-09-11
proxy-4.0.0-RC2 process blob data error


xriqyu:
writedata:
byte[] oriArray = new byte[]{
                1,0,99,111,109,46,108,115,46,114,112,99,46,115,101,114,118,101,114,46,82,112,99,82,101,113,117,101,115,-12,49,48,46,48,46,52,46,49,-75,75,74,65,104,115,100,97,107,100,104,97,115,107,-28,115,101,116,68,97,116,-31,1,1,91,76,106,97,118,97,46,108,97,110,103,46,79,98,106,101,99,116,-69,6,9,-64,-39,-86,20,3,115,100,104,98,115,97,104,100,98,115,-31,1,2,99,111,109,46,108,115,46,114,112,99,46,115,101,114,118,101,114,46,82,112,99,82,101,115,112,111,110,115,-27,0,0,0,3,115,107,104,100,98,115,107,100,117,104,97,108,100,115,106,102,100,107,106,97,102,110,97,100,102,-22,10,66,61,-8,-5,108,25,0,0,3,2,0,3,0,0
        };

but read data is
99,111,109,46,109,121,115,113,108,46,106,100,98,99,46,66,108,111,98,64,52,48,102,101,53,50,53,54

equals is false

table sql:
CREATE TABLE `data_record` (
  `uuid` varchar(40) COLLATE utf8_unicode_ci NOT NULL DEFAULT '',
  `byteData` blob,
  `type` int(11) NOT NULL,
  PRIMARY KEY (`uuid`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci;

Simple Test code:
git@github.com:xriqyu/TestProxyTest.git

TestApplicationTests,java

thanks

terrymanu:
Is it same issue with #3017?

xriqyu:
different . not has exception，only process data error .
use mysql ok, use proxy error.

code:
[...]

Issue #17892, opened 2022-05-24
MyBatis INSERT blob cause SQLSyntaxErrorException when blob length is 0


wingkingbobo:
### Which version of ShardingSphere did you use?
5.1.1

### Which project did you use? ShardingSphere-JDBC or ShardingSphere-Proxy?
ShardingSphere-Proxy

### Expected behavior
execute insert by mybatis with a blob value.
```java
@Insert("INSERT INTO remark(id, userid, content, time, blobValue) " +
            "VALUES(null, #{userId}, #{content}, #{time}, #{blobValue})"
    )
    boolean insert2(RemarkPO remark);
```
### Actual behavior
```
### Error updating database.  Cause: java.sql.SQLSyntaxErrorException: This version of ShardingSphere-Proxy doesn't yet support this SQL. 'You have an error in your SQL syntax'
### The error may exist in rivermap/cn/wemapsyncserver/dao/UploadShapeDao.java (best guess)
### The error may involve rivermap.cn.wemapsyncserver.dao.UploadShapeDao.insertBatchProperty-Inline
### The error occurred while setting parameters
### SQL: insert into dt_shape_property_simple (shapeid, userid, groupid, create_userid, name, type, name_alias, stringvalue, doublevalue, intvalue, blobvalue, version, datasize, createtime, modifytime, deleted) values  (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?) , (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
### Cause: java.sql.SQLSyntaxErrorException: This version of ShardingSphere-Proxy doesn't yet support this SQL. 'You have an error in your SQL syntax'
```
### Reason analyze (If you can)
* Use the configuration mentioned in #17888 and add a `blobvalue` field of `blob` type in the `remark` table.
```java
 public class RemarkPO {

    private byte[] blobValue;

    public byte[] getBlobValue() {
        return blobValue;
    }

    public void setBlobValue(byte[] blobValue) {
        this.blobValue = blobValue;
    }
}
```
* `insert(remark)`
* The execution conditions and results are as follows:
  * `remark.setBlobValue(null)`: OK
  * `remark.setBlobValue(new byte[5])`: OK
  * `remark.setBlobValue(new byte[0])`: throw SQLSyntaxErrorException

TeslaCN:
I found the SQL was not rewritten correctly.

```
[...]

PR #21922, opened 2022-11-03
support MySQLComQuery Blob

### LINKED ISSUE #21627: sharding-proxy insert blob size is not correct

1. apache-shardingsphere-5.0.0-shardingsphere-proxy-bin
2. apache-shardingsphere-5.2.0-shardingsphere-proxy-bin

ShardingSphere-Proxy

### Behavior

image

above it, `id==1` insert by proxy  and `id==2` insert by direct，maybe you think this tool(dbeaver) not support fine, i read this record and write to file，record（`id==1` ）open fail

config(seriver.yml) show sql, can see `insert into blob_clob(id,b_blob,c_clob) values(1,_binary'......','........')`，maybe get sql from packet ？

* **config-shadow.yml**
```yml
#....
######################################################################################################
#
# If you want to connect to MySQL, you should manually copy MySQL driver to lib directory.
#
######################################################################################################
#
databaseName: shadow_db
#
dataSources:
  ds:
    url: jdbc:mysql://192.168.1.253:3306/test?useUnicode=true&characterEncoding=UTF-8
    username: test
    password: test
#    connectionTimeoutMilliseconds: 30000
#    idleTimeoutMilliseconds: 60000
#    maxLifetimeMilliseconds: 1800000
#    maxPoolSize: 50
#    minPoolSize: 1
#...
```
* **server.yml**
```yml
#...
#mode:
#  type: Cluster
#  repository:
#    type: ZooKeeper
#    props:
#      namespace: governance_ds
#      server-lists: localhost:2181
#      retryIntervalMilliseconds: 500
#      timeToLiveSeconds: 60
#      maxRetries: 3
#      operationTimeoutMilliseconds: 500
#  overwrite: false
#
rules:
  - !AUTHORITY
    users:
      - root@%:root
      - sharding@:sharding
    provider:
      type: ALL_PERMITTED
  - !TRANSACTION
    defaultType: XA
    providerType: Atomikos
  - !SQL_PARSER
    sqlCommentParseEnabled: true
    sqlStatementCache:
      initialCapacity: 2000
      maximumSize: 65535
    parseTreeCache:
      initialCapacity: 128
      maximumSize: 1024

props:
#  max-connections-size-per-query: 1
#  kernel-executor-size: 16  # Infinite by default.
#  proxy-frontend-flush-threshold: 128  # The default value is 128.
#  proxy-hint-enabled: false
  sql-show: true
#  check-table-metadata-enabled: false
#...
```
* **${shardingsphere-proxy}/bin/start.sh**
```java
public class BlobClobTest {

    private String url;
    private Properties properties;

    public static void main(String[] args) throws Exception {
        BlobClobTest test = new BlobClobTest();
        test.setUp();
//        test.createTable();
        test.writeStream();
    }

    public void setUp() throws ClassNotFoundException {
        Class.forName("com.mysql.cj.jdbc.Driver");
        url = "jdbc:mysql://localhost:3307/shadow_db?useUnicode=true&characterEncoding=UTF-8";
        properties = new Properties();
        properties.put("user", "root");
        properties.put("password", "root");
    }

    public void createTable() {
        String sql = "CREATE TABLE `blob_clob` (\n" +
                "  `id` int(11) DEFAULT NULL,\n" +
                "  `b_blob` mediumblob,\n" +
                "  `c_clob` mediumtext\n" +
                ") ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;"; //mysql
        try (Connection conn = DriverManager.getConnection(url, properties);
             PreparedStatement stmt = conn.prepareStatement(sql)) {
            System.out.println(stmt.executeUpdate());
        } catch (SQLException e) {
            e.printStackTrace();
        }
    }

    public void writeStream() {
        String sql = "insert into blob_clob(id,b_blob,c_clob) values(?,?,?)";
        try (Connection conn = DriverManager.getConnection(url, properties);
             PreparedStatement stmt = conn.prepareStatement(sql)) {
            stmt.setInt(1, 1);
            stmt.setBlob(2,
                    Files.newInputStream(Paths.get(System.getProperty("user.home"), "test", "test.png")));
            stmt.setClob(3,
                    Files.newBufferedReader(Paths.get(System.getProperty("user.home"), "test", "test.txt")));
            System.out.println(stmt.executeUpdate());
        } catch (SQLException | IOException e) {
            e.printStackTrace();
        }
    }

}
```

yoloz:
Fixes #21627 #3024 #21518.

  - revert stream bytes;
  - position of blob are replaced by parameter;

---

- [ x] My code follows the [code of conduct](https://shardingsphere.apache.org/community/en/involved/conduct/code/) of this project.
- [ x] I have self-reviewed the commit code.
- [ x] I have (or in comment I request) added corresponding labels for the pull request.
- [ x] I have passed maven check locally : `mvn clean install -B -T2C -DskipTests -Dmaven.javadoc.skip=true -e`.


TeslaCN (changes requested):
Hi @yoloz 
ShardingSphere-Proxy didn't set `CLIENT_QUERY_ATTRIBUTES`, so there is only a SQL in COM_QUERY packet. Could you describe your logic?

image

https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_com_query.html

yoloz:
Hi @TeslaCN 
java-connector will append blob(stream) and bytes  in sql to send, which you can see [https://github.com/mysql/mysql-connector-j/blob/release/8.0/src/main/user-impl/java/com/mysql/cj/jdbc/ClientPreparedStatement.java](https://github.com/mysql/mysql-connector-j/blob/release/8.0/src/main/user-impl/java/com/mysql/cj/jdbc/ClientPreparedStatement.java) 1047行：
` Message sendPacket = ((PreparedQuery<?>) this.query).fillSendPacket(bindings);`
```java
 for (int i = 0; i < bindValues.length; i++) {
                bindings.checkParameterSet(i);

                sendPacket.writeBytes(StringLengthDataType.STRING_FIXED, staticSqlStrings[i]);

                if (bindValues[i].isStream()) {
                    streamToBytes(sendPacket, bindValues[i].getStreamValue(), true, bindValues[i].getStreamLength(), useStreamLengths);
                } else {
                    sendPacket.writeBytes(StringLengthDataType.STRING_FIXED, bindValues[i].getByteValue());
                }
            }
```
if use hex, begin with `x`, default `type bytes` use this, while no other set , `type blob` default use escape ,begin with `_binary` and:
* `0` to `\0` , can see ascii ;
* `\` and "`"  to double;

TeslaCN:
If this is a SQL, we should send it directly instead of extracting values from it. Maybe what we need to figure out is why the SQL cannot be sent to database directly.

yoloz:
> If this is a SQL, we should send it directly instead of extracting values from it. Maybe we need to figure out why the SQL cannot be sent to database directly.

yeah, i think direct read by packet.charset to string will lost some thing

TeslaCN:
> > If this is a SQL, we should send it directly instead of extracting values from it. Maybe we need to figure out why the SQL cannot be sent to database directly.
> 
> yeah, i think direct read by packet.charset to string will lost some thing, also i found from `mysql server`, limit by my poor `c` and time ,no more thing get

You may try enabling `sql-show` and check whether the SQL is rewrote correctly. We have met some issues about blob before. Such as https://github.com/apache/shardingsphere/issues/17892

yoloz:
> > > If this is a SQL, we should send it directly instead of extracting values from it. Maybe we need to figure out why the SQL cannot be sent to database directly.
> > 
> > 
> > yeah, i think direct read by packet.charset to string will lost some thing, also i found from `mysql server`, limit by my poor `c` and time ,no more thing get
> 
> You may try enabling `sql-show` and check whether the SQL is rewrote correctly. We have met some issues about blob before. Such as #17892

ok, i try it again, at #21922  i have checked sql

TeslaCN:
We need to figure out the root cause of blob problem. I'm closing this for now.

yoloz:
> We need to figure out the root cause of blob problem. I'm closing this for now.

👌

PR #21966, opened 2022-11-04
Dynamic load SQL parser test cases from MySQL and PostgreSQL

### LINKED ISSUE #16383: Dynamic load SQL parser test cases from MySQL and PostgreSQL

Jacob953:
Fixes #16383.

  - Accessing files using jsonpath
  - Remove duplicate test cases that can be automatically overwritten

---

- [x] I have passed maven check locally : `mvn clean install -B -T2C -DskipTests -Dmaven.javadoc.skip=true -e`.


test/parser/src/main/java/org/apache/shardingsphere/test/sql/parser/parameterized/engine/DynamicLoadingSQLParserParameterizedTest.java
```diff
@@ -61,16 +64,11 @@ private static Collection<Map<String, String>> getResponse(final String sqlCaseU
...
-        String nameItem = null;
-        for (String each : lines) {
-            if (each.contains("name")) {
-                nameItem = each.split("\"")[3];
-            } else if (each.contains("download_url")) {
-                String downloadURLItem = each.split("\"")[3];
-                result.add(ImmutableMap.of("name", nameItem, "download_url", downloadURLItem));
-            }
-        }
+        String casesGitHubApoContent = getContent(casesGitHubApiURL);
```
terrymanu, line 67: what's mean of `apo` ?

terrymanu (changes requested):
The unsupported.xml includes lots cases from databases.
Can not remove it directly.
Please pick up and remove the duplicated SQL cases only.

test/parser/src/main/java/org/apache/shardingsphere/test/sql/parser/parameterized/engine/DynamicLoadingSQLParserParameterizedTest.java
Jacob953: Sorry about that. It should be `Api`.

pom.xml
```diff
@@ -726,6 +727,12 @@
                 <artifactId>caffeine</artifactId>
                 <version>${caffeine.version}</version>
             </dependency>
+            
+            <dependency>
+                <groupId>com.jayway.jsonpath</groupId>
+                <artifactId>json-path</artifactId>
+                <version>${json-path.version}</version>
```
terrymanu, line 734: Please add test scope

Jacob953: fixed

test/parser/src/main/java/org/apache/shardingsphere/test/sql/parser/parameterized/engine/DynamicLoadingSQLParserParameterizedTest.java
Jacob953: fixed

terrymanu (approved):

PR #21992, opened 2022-11-07
Improve AbstractMemoryQueryResult.wasNull

sandynz:
  - Improve AbstractMemoryQueryResult.wasNull

---

- [ ] I have passed maven check locally : `mvn clean install -B -T2C -DskipTests -Dmaven.javadoc.skip=true -e`.

### DIFF
--- a/infra/executor/src/main/java/org/apache/shardingsphere/infra/executor/sql/execute/result/query/type/memory/AbstractMemoryQueryResult.java
+++ b/infra/executor/src/main/java/org/apache/shardingsphere/infra/executor/sql/execute/result/query/type/memory/AbstractMemoryQueryResult.java
@@ -47,6 +47,8 @@ public abstract class AbstractMemoryQueryResult implements QueryResult {
     @Getter
     private long rowCount;
     
+    private boolean wasNull;
+    
     protected AbstractMemoryQueryResult(final QueryResultMetaData metaData, final Collection<MemoryQueryResultDataRow> rows) {
         this.metaData = metaData;
         this.rows = rows.iterator();
@@ -66,17 +68,23 @@ public abstract class AbstractMemoryQueryResult implements QueryResult {
     
     @Override
     public final Object getValue(final int columnIndex, final Class<?> type) {
-        return currentRow.getValue().get(columnIndex - 1);
+        Object result = currentRow.getValue().get(columnIndex - 1);
+        wasNull = null == result;
+        return result;
     }
     
     @Override
     public final Object getCalendarValue(final int columnIndex, final Class<?> type, final Calendar calendar) {
-        return currentRow.getValue().get(columnIndex - 1);
+        Object result = currentRow.getValue().get(columnIndex - 1);
+        wasNull = null == result;
+        return result;
     }
     
     @Override
     public final InputStream getInputStream(final int columnIndex, final String type) {
-        return getInputStream(currentRow.getValue().get(columnIndex - 1));
+        Object value = currentRow.getValue().get(columnIndex - 1);
+        wasNull = null == value;
+        return getInputStream(value);
     }
     
     @SneakyThrows(IOException.class)
@@ -91,7 +99,7 @@ public abstract class AbstractMemoryQueryResult implements QueryResult {
     
     @Override
     public final boolean wasNull() {
-        return null == currentRow;
+        return wasNull;
     }
     
     @Override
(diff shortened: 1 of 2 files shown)


terrymanu (approved):

terrymanu (approved):

FlyingZC:
When call `next` method, 
```java
public final boolean next() {
    if (rows.hasNext()) {
        currentRow = rows.next();
        rowCount--;
        return true;
    }
   // when rows.hasNext() is false, this line will set currentRow = null and `wasNull()` method will return true before,but return false now.
    currentRow = null;
    return false;
}
```

when rows.hasNext() is false, this line will set currentRow = null and `wasNull()` method will return true before,but return false now.

wasNull method before:
```java
  public final boolean wasNull() {
     // return true;
      return null == currentRow;
  }
```

wasNull method now:
```java
  public final boolean wasNull() {
          // return false;
      return wasNull;
  }
```

sandynz:
> When call `next` method,
> 
> ```java
> public final boolean next() {
>     if (rows.hasNext()) {
>         currentRow = rows.next();
>         rowCount--;
>         return true;
>     }
>    // when rows.hasNext() is false, this line will set currentRow = null and `wasNull()` method will return true before,but return false now.
>     currentRow = null;
>     return false;
> }
> ```
> 
> when rows.hasNext() is false, this line will set currentRow = null and `wasNull()` method will return true before,but return false now.
> 
> wasNull method before:
> 
> ```java
>   public final boolean wasNull() {
>      // return true;
>       return null == currentRow;
>   }
> ```
> 
> wasNull method now:
> 
> ```java
>   public final boolean wasNull() {
>           // return false;
>       return wasNull;
>   }
> ```

Hi @FlyingZC , some investigation as following.

The definition in JDBC specification of `wasNull`:
```
    /**
     * Reports whether
     * the last column read had a value of SQL <code>NULL</code>.
     * Note that you must first call one of the getter methods
     * on a column to try to read its value and then call
     * the method <code>wasNull</code> to see if the value read was
     * SQL <code>NULL</code>.
     *
     * @return <code>true</code> if the last column value read was SQL
     *         <code>NULL</code> and <code>false</code> otherwise
     * @exception SQLException if a database access error occurs or this method is
     *            called on a closed result set
     */
    boolean wasNull() throws SQLException;
```

MySQL JDBC implementation of `wasNull` (mysql-connector-java:5.1.47):
ResultSetImpl.java
```
    /** Did the previous value retrieval find a NULL? */
    protected boolean wasNullFlag = false;

    public boolean wasNull() throws SQLException {
        return this.wasNullFlag;
    }
```
The default value of `wasNullFlag` is `false`.

PostgreSQL JDBC implementation of `wasNull` (postgresql:42.4.1):
PgResultSet.java
```
  /**
   * True if the last obtained column value was SQL NULL as specified by {@link #wasNull}. The value
   * is always updated by the {@link #getRawValue} method.
   */
  protected boolean wasNullFlag = false;

  public boolean wasNull() throws SQLException {
    checkClosed();
    return wasNullFlag;
  }
```
The default value of `wasNullFlag` is `false`.

So when `ResultSet.hasNext()` return `false` (Invoke ResultSet.getXxx() method will throw exception, `wasNullFlag` won't be set with value), `wasNull` should not be invoked, else it might return old value of `wasNullFlag` or default value `false` (if there's no any record in ResultSet).

But I didn't do real test on JDBC driver.

FlyingZC:
```java
Connection connection = getConnection();
String query = "select * from t_order;";
ResultSet resultSet = connection.createStatement().executeQuery(query);
while (resultSet.next()) {
    // null value
    String status = resultSet.getString(3);
    // will be true
    boolean wasNull = resultSet.wasNull();
    if (wasNull) {
        System.out.print("status: " + status);
    }
}
```
By my test in mysql JDBC, when `resultSet.getXxx()` return null, `resultSet.wasNull()` will be true.It's the same with your change.But maybe we should check the usage of these methods,because the logic has changed.

sandynz:
OK, thanks. I've updated some unit test cases in `JDBCMemoryQueryResultTest` before, could you help to double check it? It might need more test cases

FlyingZC:
Maybe next time we could first raise an issue to describe the issue, or describe the reason for the repair in the PR. This pr is easy to make others think it is a simple refactoring...

sandynz:
OK, I'll update PR description and labels

PR #22003, opened 2022-11-08
Refactor logback config by Java spi

Pace2Car:
  - Add Java spi logback config
  - Remove default config file logback.xml

---

- [x] I have passed maven check locally : `mvn clean install -B -T2C -DskipTests -Dmaven.javadoc.skip=true -e`.


terrymanu (changes requested):
Why not use xml file? java code is hard to understand, is it?

Pace2Car:
@terrymanu Use xml file is still the preferred option, add java code to provide the default configuration.

terrymanu:
> @terrymanu Use xml file is still the preferred option, add java code to provide the default configuration.

Maybe it is better to use xml as default log format.

TeslaCN (changes requested):
I don't think these changes are reasonable.

RaigorJiang:
Hi @terrymanu @TeslaCN 
We think that Proxy acts as a database, however there is no database that requires the user to configure the log level to start working properly.

**This PR provides a possibility:** 
Proxy can work according to the default log level without `logback.xml` configuration.
For example, when we deploy Proxy to a container, we need to mount a `conf/` directory, and then the `conf/` directory can have no `logback.xml`.

Of course, according to the configuration order of logback, users can use xml to meet their individual needs.

At the same time, I think there is room for improvement, for example, the configuration of logger may be stored in global rule, so that we can modify the log configuration of the whole cluster at once.

TeslaCN:
> Hi @terrymanu @TeslaCN We think that Proxy acts as a database, however there is no database that requires the user to configure the log level to start working properly.
> 
About that, could Proxy start without `server.yaml`?

RaigorJiang:
> > Hi @terrymanu @TeslaCN We think that Proxy acts as a database, however there is no database that requires the user to configure the log level to start working properly.
> 
> About that, could Proxy start without `server.yaml`?

If it is a standalone Proxy, it is now possible to have no server.yaml.
In the case of cluster mode, the user needs a configuration to connect to the governance center.

TeslaCN:
If server.yaml is not required in Standalone mode. They don't have to mount anything. There is already a logback.xml in conf. If users don't intend to change log configuration, they could just keep the origin one.

On one hand, if they want to mount something into container, users can also keep the origin logback.xml.
On the other hand, if they want to change log configuration, users can copy the origin one from containers instead of creating one from scratch.

terrymanu (approved):

TeslaCN (approved):
Approve without reason.

RaigorJiang (approved):

Issue #20935, opened 2022-09-12
Discuss the time to increase the minimum JDK compilation version of the ShardingSphere project to 11


linghengqian:
## Feature Request

**For English only**, other languages will not accept.

Please pay attention on issues you submitted, because we maybe need more details. 
If no response anymore and we cannot make decision by current information, we will **close it**.

Please answer these questions before submitting your issue. Thanks!

### Is your feature request related to a problem?

- Yes, refer to https://github.com/apache/shardingsphere/pull/20008 and https://github.com/apache/shardingsphere/pull/20119 .

### Describe the feature you would like.

- Starting from Antlr `4.10`, it requires users to have a minimum JDK version of 11. Refer to https://github.com/antlr/antlr4/releases/tag/4.10.1 .
- And for Hibernate which relies on Antlr, it uses Antlr `4.10` from Hibernate `6.0.1`, Hibernate `6.1.x` works on SpringBoot OSS `3.0.0-M4`, for ShardingSphere JDBC `5.2.0` which uses Antlr `4.9.2`, this breaks ShardingSphere JDBC Use of hybrid deployment form with zookeeper/etcd.

- I think it's time to consider the point in time when the minimum compiled version of ShardingSphere is raised to JDK11. I think this event should appear later than November 2022, when SpringBoot OSS `3.0.0` releases the first Release version. For the master branch of ShardingSphere, in which version should it raise the minimum build version? Is it ShardingSphere `5.3.0` or ShardingSphere `6.0.0`?

- Refer to https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-3.0.0-M4-Release-Notes#hibernate-61 .

linghengqian:
- I'm reopening this issue because no one has responded to the time point, and I'll submit a PR directly if it's ok.

TeslaCN:
> * I'm reopening this issue because no one has responded to the time point, and I'll submit a PR directly if it's ok.

This may impact a lot of users. It's better to discuss breaking changes in mailing list.

linghengqian:
- Considering the process at https://github.com/apache/shardingsphere/issues/1157, maybe it would be better for PMC to start a new email thread?  👀

zhfeng:
Vote on github issue? Is it an official process? I think it should happen on the mailing list although it seems few peope take care of it.

linghengqian:
[...]

PR #23550, opened 2023-01-13
Force to build on JDK 11+

zhfeng:
It sould be helpful to fix https://github.com/apache/shardingsphere/issues/19990 and https://github.com/apache/shardingsphere/issues/20935

  - Force to use `JDK 11` above to build shardingsphere
  - We still run all the e2e tests on `JDK 8` to make sure there is no compatitable issues.
  - https://github.com/apache/shardingsphere/issues/19990 which needs to run `antlr-maven-plugin` to generate the parser classes. The antlr Tool are only for `JDK 11` after upgrading to 4.10.1. So if upgrading `antlr`, it will force user to build shardingsphere with `JDK 11`

It seems that we need to support  `JDK 8` for a long time. There is no such plan to drop it, see the relevant discussion https://lists.apache.org/thread/kgtfx5vnw73o61lkvdshyfvjfvvzk3js

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1193,6 +1193,9 @@
             <activation>
                 <jdk>[11,)</jdk>
             </activation>
+            <properties>
+                <maven.compiler.release>8</maven.compiler.release>
+            </properties>
             <build>
                 <pluginManagement>
                     <plugins>
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,7 +61,7 @@ jobs:
     timeout-minutes: 60
     strategy:
       matrix:
-        java-version: [ 8, 19 ]
+        java-version: [ 11, 19 ]
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-java@v3
@@ -78,6 +78,13 @@ jobs:
         run: ./mvnw -T1C -B -ntp clean install
       - name: Build examples with Maven
         run: ./mvnw -T1C -B -f examples/pom.xml clean package -DskipTests
+      - name: Setup JDK 8 for Test
+        uses: actions/setup-java@v3
+        with:
+          distribution: 'temurin'
+          java-version: 8
+      - name: Run tests with JDK 8
+        run: ./mvnw -T1C -B -ntp -fae test
 
   test-coverage:
     if: github.repository == 'apache/shardingsphere'
@@ -95,8 +102,15 @@ jobs:
       - uses: actions/setup-java@v3
         with:
           distribution: 'temurin'
+          java-version: 11
+      - name: Build with Maven
+        run: ./mvnw -T1C -B -ntp clean install -DskipTests
+      - name: Setup JDK 8 for Test
+        uses: actions/setup-java@v3
+        with:
+          distribution: 'temurin'
           java-version: 8
       - name: Test with Maven
-        run: ./mvnw -T1C -B -ntp clean install cobertura:cobertura -Djacoco.skip=false
+        run: ./mvnw -T1C -B -ntp test cobertura:cobertura -Djacoco.skip=false
       - name: Upload to Codecov
         run: bash <(curl -s https://codecov.io/bash)
--- a/.github/workflows/docker-release.yml
+++ b/.github/workflows/docker-release.yml
@@ -48,7 +48,7 @@ jobs:
       - uses: actions/setup-java@v3
         with:
           distribution: 'temurin'
-          java-version: 8
+          java-version: 11
       - name: Set environment
         run: export MAVEN_OPTS=' -Dmaven.javadoc.skip=true -Djacoco.skip=true $MAVEN_OPTS'
       - name: Build docker image
--- a/.github/workflows/e2e-agent.yml
+++ b/.github/workflows/e2e-agent.yml
@@ -70,7 +70,7 @@ jobs:
       - uses: actions/setup-java@v3
         with:
           distribution: 'temurin'
-          java-version: '8'
+          java-version: '11'
       - uses: actions/cache@v3
         with:
           path: ~/.m2/repository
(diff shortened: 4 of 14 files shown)


linghengqian (commented):
- I checked CI.
```shell
Failed to execute goal org.codehaus.mojo:cobertura-maven-plugin:2.7:instrument (default-cli) on project shardingsphere: Execution default-cli of goal org.codehaus.mojo:cobertura-maven-plugin:2.7:instrument failed : Plugin org.codehaus.mojo:cobertura-maven-plugin:2.7 or one of its dependencies could not be resolved: Could not find artifact com.sun:tools:jar:0 at specified path /usr/lib/jvm/temurin- 11-jdk-amd64/../lib/tools.jar -> [Help 1]
```
- It looks like `tools.jar` is being called somewhere, which is a unique built-in JAR on JDK 8.

zhfeng:
Yeah, I'm taking a look at it.

linghengqian:
- I'm a bit confused after looking at http://cobertura.github.io/cobertura/, shouldn't there be a more modern maven plugin for uploading to codecov? This even goes back to 2017, https://github.com/apache/shardingsphere/commit/e727b71f7e6ecacae12487299d5f15b908b7a28b. It looks like @haocao has withdrawn from the github, maybe it can be replaced by `jacoco-maven-plugin`? Like https://github.com/codecov/example-java-maven/blob/main/pom.xml .

zhfeng:
@terrymanu @TeslaCN can you take a review? Thanks!

linghengqian (dismissed):
LGTM.

terrymanu (changes requested):
Please hold on, I need some time to double check

TeslaCN:
Hi @zhfeng @linghengqian 
If we upgrade antlr to `4.10.+`, even though ShardingSphere is Java 8 bytecode, but Java 8 runtime still cannot using antlr `4.10+`.
Is it possible to building ShardingSphere parser modules with different classifiers?

TeslaCN:
> Hi @zhfeng @linghengqian If we upgrade antlr to `4.10.+`, even though ShardingSphere is Java 8 bytecode, but Java 8 runtime still cannot using antlr `4.10+`. Is it possible to build ShardingSphere parser modules with different classifiers?

Sorry, I thought the antlr `4.10.+` was compiled as Java 11 bytecode. It seems that only antlr tools require Java 11. The runtime is still Java 8 bytecode.

pom.xml
```diff
@@ -944,7 +944,7 @@
                                     <version>${maven.version.range}</version>
                                 </requireMavenVersion>
                                 <requireJavaVersion>
-                                    <version>${java.version}</version>
+                                    <version>11</version>
```
TeslaCN, line 947: Could we just keep the origin version here, otherwise we have to add `-Denforcer.skip=true` to do any other operations with Java 8.

zhfeng: Yeah, we can. But It cuould confuse people that building with Java 8 will fail after upgrading antlr to 4.10.1. Anyway, we need to add a NOTE in README to mention that `Building from source` of shardingsphere must use `JDK 11` above. @terrymanu WDYT?

terrymanu: I prefer to create `compile java version` and `release java version` properties to separate the different versions.

linghengqian (commented):
- I think the problem that CI had were solved.  You can try to rebase the branch.

zhfeng:
Thanks @linghengqian and I will take care of the conflicts after the Chinese New Year holiday.

pom.xml
```diff
@@ -944,7 +945,7 @@
                                     <version>${maven.version.range}</version>
                                 </requireMavenVersion>
                                 <requireJavaVersion>
-                                    <version>${java.version}</version>
+                                    <version>${compile.java.version}</version>
```
TeslaCN, line 948: Hi @zhfeng 
I prefer to set `8` as `requireJavaVersion`. Java 11 is required only when compiling parser modules.
Sometimes we just need to compile a part of modules (such as protocol, agent), build an image, do some tests, which could be done using Java 8. Adding `-Denforcer.skip=true` is not elegant.
And the error of compiling antlr 4.10+ with Java 8 is clear enough for developers.
We could clarify Java requirement in [Wiki](https://github.com/apache/shardingsphere/wiki) and [docs](https://shardingsphere.apache.org/community/en/involved/contribute/dev-env/).

TeslaCN: Or is it possible to add `maven-enforcer-plugin` with Java 11 requirement to those modules require Java 11? The other modules keep using Java 8

zhfeng: Yeah, I think it makes sense and I will investigate.

zhfeng:
Hi guys,

I just update to run CI build on JDK 11 but test it on JDK 8. `<requireJavaVersion>` is still with `<java.version>`. After upgrading to antlr 4.10+, the CI should work because it uses JDK 11 but if you build sharding-sphere from source with JDK 8 the following modules which use `antlr4-maven-plugin` will fail.

- shardingsphere-distsql-parser
- shardingsphere-db-discovery-distsql-parser
- shardingsphere-encrypt-distsql-parser
- shardingsphere-readwrite-splitting-distsql-parser
- shardingsphere-shadow-distsql-parser
- shardingsphere-sharding-distsql-parser
- shardingsphere-mask-distsql-parser
- shardingsphere-authority-distsql-parser
- shardingsphere-data-pipeline-distsql-parser
- shardingsphere-parser-distsql-parser
- shardingsphere-sql-federation-optimizer
- shardingsphere-sql-translator-distsql-parser
- shardingsphere-traffic-distsql-parser
- shardingsphere-transaction-distsql-parser
- shardingsphere-sql-parser-mysql
- shardingsphere-sql-parser-opengauss
- shardingsphere-sql-parser-oracle
- shardingsphere-sql-parser-postgresql
- shardingsphere-sql-parser-sql92
- shardingsphere-sql-parser-sqlserver

zhfeng:
@TeslaCN @terrymanu can you take a review of these changes? I agree to clarify JDK requirement on wiki and docs. Only the modules which uses `antlr-maven-plugin` will be impact. If the community is happy to accept this PR, I'm going to prepare for the antlr 4.10+ upgrade which could be

- Upgarde `<antlr4.version>` to `4.10.1`
- Run `antlr-maven-plugin` only in `jdk11+` profile
- In `jdk8` profile, try to add the antlr generation directoy as source codes directory. if not, fail the building with the message `Please use JDK 11 to re-build artifact to generate the anltr codes`

any thought?

TeslaCN (approved):
LGTM

terrymanu (approved):

terrymanu:
@zhfeng LGTM, thank you!

darren-wang:
@zhfeng hi, I'm new to sharding's source code, could you have a look at question https://github.com/apache/shardingsphere/issues/31379 thanks a lot!

Issue #23486, opened 2023-01-11
Migrating base Docker image from openjdk to other distribution


TeslaCN:
## Refactor

Since the [openjdk in Docker Hub](https://hub.docker.com/_/openjdk) has been officially deprecated. We should migrate base JDK image in all of our Dockerfile to other distribution.

https://github.com/docker-library/openjdk/issues/505
https://github.com/docker-library/openjdk/pull/495

![image](https://user-images.githubusercontent.com/20503072/211696217-91ea2446-6d40-4c54-98ff-f2bc9c65b848.png)

PR #23720, opened 2023-01-24
Use multi-platform image in E2E Dockerfile

TeslaCN:
We have migrated E2E image from `openjdk` to `eclipse-temurin` in https://github.com/apache/shardingsphere/issues/23486. But the alpine variant image only support `amd64`.
Change the image to `8-jdk` to support non-amd64 environment.

image

image


TeslaCN:
The transaction E2E keep failing. #23727.

terrymanu (approved):

PR #24221, opened 2023-02-17
password support special characters for DistSql

### LINKED ISSUE #24178: DB Password cannot contain special  characters with distSQL
## Bug Report

DistSQL like that:

···
REGISTER STORAGE UNIT test_db (
    URL="jdbc:mysql://127.0.0.1:3306/test_db",
    USER="root",
    PASSWORD="abc[123"
);
···

BUT Parsed the password is ”abc123“ 

image

check the code：

```
public final class KernelDistSQLStatementVisitor extends KernelDistSQLStatementBaseVisitor<ASTNode> implements SQLVisitor {
   ...
   private String getPassword(final PasswordContext ctx) {
        return getIdentifierValue(ctx);
    }
   ...
    private String getIdentifierValue(final ParseTree context) {
        return null == context ? null : new IdentifierValue(context.getText()).getValue();
    }
}


public final class IdentifierValue implements ValueASTNode<String> {
   ...
    public IdentifierValue(final String text) {
        this(SQLUtil.getExactlyValue(text), QuoteCharacter.getQuoteCharacter(text));
    }
}

public final class SQLUtil {
   ...
   private static final String EXCLUDED_CHARACTERS = "[]`'\"";
   public static String getExactlyValue(final String value) {
        //Filtered  here 
        return null == value ? null : CharMatcher.anyOf(EXCLUDED_CHARACTERS).removeFrom(value);
    }
}
  
```

sunkai-cai:
Fixes #24178.

### DIFF
--- a/sql-parser/statement/src/main/java/org/apache/shardingsphere/sql/parser/sql/common/value/literal/impl/StringLiteralValue.java
+++ b/sql-parser/statement/src/main/java/org/apache/shardingsphere/sql/parser/sql/common/value/literal/impl/StringLiteralValue.java
@@ -18,6 +18,7 @@
 package org.apache.shardingsphere.sql.parser.sql.common.value.literal.impl;
 
 import lombok.Getter;
+import org.apache.groovy.parser.antlr4.util.StringUtils;
 import org.apache.shardingsphere.sql.parser.sql.common.value.literal.LiteralValue;
 
 /**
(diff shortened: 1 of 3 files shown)


sql-parser/statement/src/main/java/org/apache/shardingsphere/sql/parser/sql/common/value/literal/impl/StringLiteralValue.java
```diff
@@ -31,4 +32,23 @@ public final class StringLiteralValue implements LiteralValue<String> {
     public StringLiteralValue(final String value) {
         this.value = value.substring(1, value.length() - 1);
     }
+    
+    /**
+     * get special escape sequences string, do this:
```
RaigorJiang, line 37: The first letter should be uppercase.

Also, I think we should not keep so many comments, since that's not a feature we provide, it's the ability of `StringUtils.replaceStandardEscapes`.

sunkai-cai: ok

RaigorJiang: Thank you!

RaigorJiang (approved):

PR #24423, opened 2023-03-02
Fix grammar for mysql begin statement

jiangtao69039:
according to official documentation, **BEGIN** statement has an optional keyword '**WORK**'

syntax:
```
BEGIN [WORK]
```


strongduanmu (approved):
@jiangtao69039 Thank you for your contribution, merged.

PR #24635, opened 2023-03-16
Refactor : refactor the ci workflow with corresponding JDK version

taojintianxia:

PR #24638, opened 2023-03-16
Refactor : remove JDK 8 from nightly CI

taojintianxia:
remove JDK 8 from nightly CI


TeslaCN:
My suggestion is revert https://github.com/apache/shardingsphere/pull/24635.

PR #24797, opened 2023-03-24
Avoid repeat call findRules method

zhaojinchao95:
Fixes #24690 .


jdbc/core/src/main/java/org/apache/shardingsphere/driver/jdbc/adapter/AbstractResultSetAdapter.java
```diff
@@ -46,23 +46,26 @@ public abstract class AbstractResultSetAdapter extends AbstractUnsupportedOperat
     @Getter
     private final Statement statement;
     
+    private final boolean singleTable;
```
strongduanmu, line 49: Do you think `transparentStatement` is better?

strongduanmu (changes requested):

strongduanmu (approved):

PR #24970, opened 2023-04-03
support for select substring

kanha-gupta:
@strongduanmu Hello, while adding support for SQL's having identifiers, Numbers & Symbols I have came across a problem.
In the following code, I have debugged MySQLStatementSQLVisitor.java & the result of method VisitSubstringFunction have the required Parameters/OperandList. But the same do not shows up in SQLNodeConverterEngineIT.java ACTUAL parsing while running.
Can you please guide me regarding what I am doing wrong in parsing these type of queries ?
Thank you

### DIFF
--- a/sql-parser/dialect/mysql/src/main/java/org/apache/shardingsphere/sql/parser/mysql/visitor/statement/impl/MySQLStatementSQLVisitor.java
+++ b/sql-parser/dialect/mysql/src/main/java/org/apache/shardingsphere/sql/parser/mysql/visitor/statement/impl/MySQLStatementSQLVisitor.java
@@ -942,7 +942,14 @@ public abstract class MySQLStatementSQLVisitor extends MySQLStatementBaseVisitor
     @Override
     public final ASTNode visitSubstringFunction(final SubstringFunctionContext ctx) {
         calculateParameterCount(Collections.singleton(ctx.expr()));
-        return new FunctionSegment(ctx.getStart().getStartIndex(), ctx.getStop().getStopIndex(), null == ctx.SUBSTR() ? ctx.SUBSTRING().getText() : ctx.SUBSTR().getText(), getOriginalText(ctx));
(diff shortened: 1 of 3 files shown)


kanha-gupta:
@strongduanmu Hello, I request you to review the code.
Actual parsing of Select_extract gives "SELECT EXTRACT(YEAR() FROM '2019-07-02') "
How to deal with parenthesis after YEAR keyword ?

strongduanmu:
> @strongduanmu Hello, I request you to review the code. Actual parsing of Select_extract gives "SELECT EXTRACT(YEAR() FROM '2019-07-02') " How to deal with parenthesis after YEAR keyword ?

Hi @kanha-gupta, can you show the result we expected in Calcite?

strongduanmu:
You can refer calcite parse result for `SELECT SUBSTRING('foobarbar' FROM 4)` sql.

image

strongduanmu:
And `SELECT SUBSTRING('foobarbar' FROM 4)`, you can refer this result. You just need to extract the function name and constants as arguments to the outer function.

image

kanha-gupta:
@strongduanmu Hello
for sql query "SELECT EXTRACT(YEAR FROM '2019-07-02')"
I tried many ways but I cant exactly figure out how to just get the word "YEAR" instead of "YEAR()". Extracting FunctionName is giving errors on my machine therefore Please suggest the approach.
Thank you

kanha-gupta:
@strongduanmu Thanks a lot for guidance.
select_substring task is done.  2 more doubts :
1- for select_trim, There is a symbol BOTH which needs to parsed, I tried doing it using Expression segment & also LiteralExpressionSegment but it is being Parsed as String instead of Symbol therefore it results in "BOTH" instead of BOTH.
2 - for select_extract, YEAR is a function & Extracting only FunctionName leads to Errors & parsing it as a function leads to parenthesis ( YEAR() )
I request you to please provide your guidance regarding this Because rest of the 5 special function tasks have these type of parsing.
Thank you

PR #25168, opened 2023-04-14
add message args check.

JiekerTime:


terrymanu (changes requested):
Is there any problem before change?

JiekerTime:
> Is there any problem before change?

Not found yet, just want to add a strategy to prevent the incoming parameters are empty leading to the system error return problems can not be troubleshooting.

terrymanu:
messageArgs should not permit null. I just close the PR

PR #25189, opened 2023-04-17
Trim functions support

kanha-gupta:
Fixes #ISSUSE_ID.

@strongduanmu Respected Sir, Please review this. Tests are being passed successfully now in SQLNodeConverterEngineIT & InternalSQLParserIT


test/it/parser/src/main/resources/sql/supported/dml/select-special-function.xml
```diff
@@ -26,9 +26,9 @@
     <sql-case id="select_substring" value="SELECT SUBSTRING('foobarbar' from 4)" db-types="MySQL" />
     <sql-case id="select_extract" value="SELECT EXTRACT(YEAR FROM '2019-07-02')" db-types="MySQL" />
     <sql-case id="select_char" value="SELECT CHAR(77,121,83,81,'76')" db-types="MySQL" />
-    <sql-case id="select_trim" value="SELECT TRIM('  bar   ')" db-types="MySQL" />
-    <sql-case id="select_with_trim_expr" value="select trim('#' from `name`) from t_order" db-types="MySQL" />
-    <sql-case id="select_with_trim_expr_from_expr" value="select trim(remove_name from name) from t_order" db-types="MySQL" />
+    <sql-case id="select_trim" value="SELECT TRIM(BOTH ' ' from ' bar ')" db-types="MySQL" />
```
strongduanmu, line 29: Why add BOTH keyword to these sql cases? Are they the wrong cases? If they are not wrong cases, I suggest adding cases to cover the logic we want to test.

strongduanmu (changes requested):

strongduanmu (approved):

Issue #25170, opened 2023-04-14
When using version 5.12 to query the left join table, the routing failed.


lmandy:
The SQL statement is as follows, where two tables are binding relationships:
--------------------------------------------------------------------------------------
SELECT
pd.id,pde.ewallet_order_id
FROM	payment_dynamicaccountinfo pd
LEFT JOIN payment_dynamicaccountinfo_extra pde ON pd.id = pde.payment_dynamicaccountinfo_id
WHERE 1 = 1 AND pd.dynamic_date >= '2023-01-01 00:00:00'
--------------------------------------------------------------------------------------
The current version is 5.12, and I have checked the latest version, which also has this problem.
I looked at the source code because the tableNames attribute in the TablesContext class is of type HashSet, causing the order of the tables to change
![image](https://user-images.githubusercontent.com/16572623/232008418-5af70148-29a4-4a87-bc9c-d8d98a0500c7.png)
--------------------------------------------------------------------------------------
The version I used before was 4.1.1, but it has been solved in this version.
![image](https://user-images.githubusercontent.com/16572623/232008822-2c8d64e2-cbe3-46c2-8e75-57b5b9acfcf8.png)

I found the previous related issues #866

RaigorJiang:
Hi @lmandy ,
Are you interested in fixing this issue?

lmandy:
Yes, I'd love to.

PR #25194, opened 2023-04-17
Routing failure when solving left join associated query. Change hashset to linkedHashSet

lmandy:
I have changed the Hashset to LinkedHashSet, This mysql #25170


lmandy:
CI failed, as fllows
image
Please deal with this pr

strongduanmu:
Hi @lmandy, thank you for your contribution. Can you add unit test for this logic?

lmandy:
Ok, which module should I add? Is there an example for reference?

terrymanu:
Close because of staled.

PR #25312, opened 2023-04-24
Add more error prone rules for pmd

terrymanu:
For #25174


RaigorJiang (changes requested):
Please fix CI failure.

RaigorJiang (approved):

PR #25377, opened 2023-04-27
show table supports the full keyword

totalo:


sql-parser/statement/src/main/java/org/apache/shardingsphere/sql/parser/sql/dialect/statement/mysql/dal/MySQLShowTablesStatement.java
```diff
@@ -36,6 +36,8 @@ public final class MySQLShowTablesStatement extends AbstractSQLStatement impleme
     
     private ShowFilterSegment filter;
     
+    private Boolean hasFull;
```
strongduanmu, line 39: Why not use boolean? And what about constainsFull?

strongduanmu (changes requested):

sql-parser/statement/src/main/java/org/apache/shardingsphere/sql/parser/sql/dialect/statement/mysql/dal/MySQLShowTablesStatement.java
totalo: done.

strongduanmu (approved):
LGTM, merged.

Issue #24200, opened 2023-02-16
[GSOC 2023] Enhance SQLNodeConverterEngine to support more MySQL SQL statements


strongduanmu:
# Apache ShardingSphere

Apache ShardingSphere is positioned as a Database Plus, and aims at building a standard layer and ecosystem above heterogeneous databases. It focuses on how to reuse existing databases and their respective upper layer, rather than creating a new database. The goal is to minimize or eliminate the challenges caused by underlying databases fragmentation.

Page: https://shardingsphere.apache.org/
Github: https://github.com/apache/shardingsphere 

# Background

The ShardingSphere SQL federation engine provides support for complex SQL statements, and it can well support cross-database join queries, subqueries, aggregation queries and other statements. An important part of SQL federation engine is to convert the SQL statement parsed by ShardingSphere into SqlNode, so that Calcite can be used to implement SQL optimization and federated query.

# Task

This issue is to solve the MySQL exception that occurs during SQLNodeConverterEngine conversion. The specific case list is as follows.

- [x] select_char
- [x] select_extract
- [x] select_from_dual
- [x] select_from_with_table
- [x] select_group_by_with_having_and_window
- [x] select_not_between_with_single_table
- [x] select_not_in_with_single_table
- [x] select_substring
- [x] select_trim
- [x] select_weight_string
- [x] select_where_with_bit_expr_with_ampersand
- [x] select_where_with_bit_expr_with_caret
- [x] select_where_with_bit_expr_with_div
- [x] select_where_with_bit_expr_with_minus_interval
- [x] select_where_with_bit_expr_with_mod
- [x] select_where_with_bit_expr_with_mod_sign
- [x] select_where_with_bit_expr_with_plus_interval
- [x] select_where_with_bit_expr_with_signed_left_shift
- [x] select_where_with_bit_expr_with_signed_right_shift
- [x] select_where_with_bit_expr_with_vertical_bar
- [x] select_where_with_boolean_primary_with_comparison_subquery
- [x] select_where_with_boolean_primary_with_is
- [x] select_where_with_boolean_primary_with_is_not
- [ ] select_where_with_boolean_primary_with_null_safe
- [ ] select_where_with_expr_with_and_sign
- [x] select_where_with_expr_with_is
- [x] select_where_with_expr_with_is_not
- [x] select_where_with_expr_with_not
- [x] select_where_with_expr_with_not_sign
- [x] select_where_with_expr_with_or_sign
[...]

PR #25876, opened 2023-05-24
Support for AMPERSAND,CARET AND DIV

kanha-gupta:
Ref #24200 
Adds support for :
1.select_where_with_bit_expr_with_ampersand
2.select_where_with_bit_expr_with_caret
3.select_where_with_bit_expr_with_div

Screenshot (324)
Screenshot (325)
Screenshot (326)


kanha-gupta:
@strongduanmu Since calcite do not support these operators natively, I have added support for it through method call.
Please review :)
(Images are attached in the above message thread)

kernel/sql-federation/optimizer/src/main/java/org/apache/shardingsphere/sqlfederation/optimizer/converter/segment/expression/impl/BinaryOperationExpressionConverter.java
```diff
@@ -107,6 +109,15 @@ private SqlOperator convertOperator(final BinaryOperationExpression segment) {
                 operator = "IS NOT FALSE";
             }
         }
+        if ("&".equalsIgnoreCase(operator)) {
```
strongduanmu, line 112: Hi @kanha-gupta, can you show sql node toString result?

strongduanmu (changes requested):

strongduanmu:
Hi @kanha-gupta, you can refer pr https://github.com/apache/shardingsphere/pull/25882 to add sql case.

PR #25942, opened 2023-05-29
support pg bit and bool type parsed from/to java&jdbc

### LINKED ISSUE #25933: proxy query pg bool, column description's typeOid is not match with actual value
5.3.0 

proxy 

### Expected behavior
node typeorm, query pg boolean field  display correct value

### Actual behavior
typeorm reflect always true for boolean field 

pg boolean is mapped to  jdbc java.sql.Types.BIT(-7)   related to java boolean

so, the actual value in  proxy(java) is  boolean value
but  PostgreSQLColumnType's  mapping form   java.sql.Types.BIT(-7)  is   POSTGRESQL_TYPE_BIT(1560)  now 
the cliean (typeorm)  cannot convert  'true' or 'false' text  according to typeOid(1560)

t_test.bool_field   type:  boolean    value: false

```java
 String sql = "select bool_field from t_test where id = ?";
        try (Connection conn = dataSource.getConnection()) {
            PreparedStatement preparedStatement = conn.prepareStatement(sql);
            preparedStatement.setString(1, "1");
            ResultSet resultSet = preparedStatement.executeQuery();
            if (resultSet.next()) {
                return String.valueOf(resultSet.getBoolean("bool_field"));
            }
        } catch (Exception ex) {
            throw ex;
        }
```

JDBCPortal#bind   queryHeader  
image

PostgreSQLColumnDescription typeOid =  POSTGRESQL_TYPE_BIT(1560)
image

but  JdbcPortal#execute  row value is  Boolean(false)
image

susongyan:
Fixes #25933.

- parse pg bit parameter
- set correct typeOid for PG bit and bool type  for proxy

### DIFF
--- a/proxy/backend/core/src/main/java/org/apache/shardingsphere/proxy/backend/response/data/QueryResponseCell.java
+++ b/proxy/backend/core/src/main/java/org/apache/shardingsphere/proxy/backend/response/data/QueryResponseCell.java
@@ -31,7 +31,7 @@ public final class QueryResponseCell {
     
     private final Object data;
     
-    private final Optional<String> columnTypeName;
+    private final String columnTypeName;
     
     public QueryResponseCell(final int jdbcType, final Object data) {
         this(jdbcType, data, null);
@@ -40,6 +40,10 @@ public final class QueryResponseCell {
     public QueryResponseCell(final int jdbcType, final Object data, final String columnTypeName) {
         this.jdbcType = jdbcType;
         this.data = data;
-        this.columnTypeName = Optional.ofNullable(columnTypeName);
+        this.columnTypeName = columnTypeName;
+    }
+    
+    public Optional<String> getColumnTypeName() {
+        return Optional.ofNullable(columnTypeName);
     }
 }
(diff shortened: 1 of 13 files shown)


proxy/backend/core/src/main/java/org/apache/shardingsphere/proxy/backend/response/data/QueryResponseCell.java
```diff
@@ -18,16 +18,26 @@
...
  */
-@RequiredArgsConstructor
 @Getter
 public final class QueryResponseCell {
     
     private final int jdbcType;
     
     private final Object data;
+    
+    private String columnTypeName;
```
TeslaCN, line 32: If this field is nullable, is it possible to make it final and add an Optional getter?

susongyan: ok,

proxy/backend/core/src/main/java/org/apache/shardingsphere/proxy/backend/response/data/QueryResponseCell.java
```diff
@@ -18,16 +18,28 @@
...
  */
-@RequiredArgsConstructor
 @Getter
 public final class QueryResponseCell {
     
     private final int jdbcType;
     
     private final Object data;
+    
+    private final Optional<String> columnTypeName;
```
TeslaCN, line 34:
```suggestion
    private final String columnTypeName;
```

Add a method
```java
public Optional<String> getColumnTypeName() { return Optional.ofNullable(columnTypeName); }
```

TeslaCN (approved):

PR #26079, opened 2023-06-06
Support table name backquote parse

### LINKED ISSUE #26035: About parsing table names that contain backticks
## About parsing table names that contain backticks

create table ```t_order` (i int)

Hi community, currently Shardingsphere cannot parse the SQL statement that contains a table name with backticks. 
I have made changes to the corresponding parsing rules locally and now the SQL can be parsed correctly. 
However, when writing tests, I found that Shardingsphere processes the table name as "t_order" instead of "`t_order" (with backticks) after calling the constructor of IdentifierValue, which invokes SQLUtils.getExactlyValue(text). 
Is this correct behavior?

============== Test out
java.lang.AssertionError: 
SQL Case ID : create_table_with_backtick
SQL         : CREATE TABLE ```t_order` (i int)
Table name assertion error: 

Expected: is "`t_order"
     but: was "t_order"

zihaoAK47:
Fixes #26035.

  - Support table name backquote parse
  - Add test sql

---

- [ ✓] My code follows the [code of conduct](https://shardingsphere.apache.org/community/en/involved/conduct/code/) of this project.
- [ ✓] I have self-reviewed the commit code.
- [ ✓] I have passed maven check locally : `./mvnw clean install -B -T1C -Dmaven.javadoc.skip -Dmaven.jacoco.skip -e`.
- [ ✓] I have added corresponding unit tests for my changes.

### DIFF
--- a/test/it/parser/src/main/resources/case/ddl/create-table.xml
+++ b/test/it/parser/src/main/resources/case/ddl/create-table.xml
@@ -17,6 +17,12 @@
   -->
 
 <sql-parser-test-cases>
+    <create-table sql-case-id="create_table_with_backtick">
+        <table name="t_order" start-delimiter="`" end-delimiter="`" start-index="13" stop-index="23" />
+        <column-definition type="int" start-index="26" stop-index="30">
+            <column name="i" />
+        </column-definition>
+    </create-table>
     <create-table sql-case-id="create_table_with_like">
         <table name="t_log" start-index="13" stop-index="17" />
         <like-table name="t_old_log" start-index="24" stop-index="32" />
--- a/test/it/parser/src/main/resources/case/ddl/create-table.xml
+++ b/test/it/parser/src/main/resources/case/ddl/create-table.xml
@@ -18,9 +18,15 @@
 
 <sql-parser-test-cases>
     <create-table sql-case-id="create_table_with_backtick">
-        <table name="t_order" start-delimiter="`" end-delimiter="`" start-index="13" stop-index="23" />
+        <table name="`t_order" start-delimiter="`" end-delimiter="`" start-index="13" stop-index="23" />
         <column-definition type="int" start-index="26" stop-index="30">
             <column name="i" />
+        </column-definition>
+    </create-table>
+    <create-table sql-case-id="create_table_column_with_backtick">
+        <table name="``t_o`r`d`e`r`" start-delimiter="`" end-delimiter="`" start-index="13" stop-index="35" />
+        <column-definition type="int" start-delimiter="`" end-delimiter="`" start-index="38" stop-index="46">
+            <column name="`i" />
         </column-definition>
     </create-table>
     <create-table sql-case-id="create_table_with_like">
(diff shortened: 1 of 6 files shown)


zihaoAK47:
@RaigorJiang Hi mentor, although the backticks have been removed from IdentifierValue, it can still correctly create tables. I tested it using shardingsphere-proxy on my local machine and the table names were correct. The test statements have been modified and the code has passed the tests.

test/it/parser/src/main/resources/case/ddl/create-table.xml
```diff
@@ -17,6 +17,12 @@
   -->
 
 <sql-parser-test-cases>
+    <create-table sql-case-id="create_table_with_backtick">
+        <table name="t_order" start-delimiter="`" end-delimiter="`" start-index="13" stop-index="23" />
```
RaigorJiang, line 21: Hi @zihaoAK47 
In #26035, We know that the expected table name should be `t_order, but now the result of the test is t_order, is this not as expected?


The following is my test in native mysql:
image

RaigorJiang (changes requested):

test/it/parser/src/main/resources/case/ddl/create-table.xml
zihaoAK47: @RaigorJiang Sorry, I misunderstood the purpose of the start-delimiter and end-delimiter attributes for processing backticks. I have now created a new method to retrieve the correct table name enclosed with backticks. Please check it, mentor.

RaigorJiang: Hi @zihaoAK47 
There is a method `getQuotedContent` in `IdentifierValue`, which is used to obtain the complete content in the quotation marks. Do you think this method can meet the requirements?

```java
public static String getQuotedContent(final String text) {
        if (Strings.isNullOrEmpty(text)) {
            return text;
        }
        QuoteCharacter quoteCharacter = QuoteCharacter.getQuoteCharacter(text);
        if (QuoteCharacter.NONE == quoteCharacter) {
            return text.trim();
        }
        return text.substring(1, text.length() - 1);
    }
```

zihaoAK47: @RaigorJiang Tutor, I think the IdentifierValue.getQuotedContent(final String text) method is sufficient if we only need to retrieve the content that is enclosed by special characters, but this method cannot obtain the actual content, such as the real table name of \`t_order\` , which is t_order.
\`\`\`t_order\`  real table name is  \`t_order

I am confused about whether the IdentifierValue property in the TableNameSegment class should represent the real table name. After processing with IdentifierValue.getQuotedContent, \`\`\`t_orde\` is transformed into \`\`t_order. However, I believe that the IdentifierValue property should represent the real table name, which should be \`t_order instead of \`\`t_order.

tuichenchuxin (approved):

RaigorJiang (approved):

RaigorJiang:
Thank you guys!

PR #26739, opened 2023-07-03
Update create-readwrite-splitting-rule.en.md

yy2so:


docs/document/content/user-manual/shardingsphere-proxy/distsql/syntax/rdl/rule-definition/readwrite-splitting/create-readwrite-splitting-rule.en.md
```diff
@@ -91,12 +91,12 @@ CREATE READWRITE_SPLITTING RULE IF NOT EXISTS ms_group_0 (
 );
 ```
 
-### Reserved words
+### Syntax
```
terrymanu, line 94: The other related pages 、 use `Reserved words` too (for example: create-encrypt-rule.en.md).
So change it alone maybe not good practice.

PR #26756, opened 2023-07-04
Add method isEmpty() to DatabaseRuleConfiguration

### LINKED ISSUE #26751: Consider add method isEmpty() to DatabaseRuleConfiguration
Hi community,

After the new structure of metadata is launched, the subscriber and refresh method of rule has been completely refactored. Currently, only the modified single rule will be rebuilt, but the mechanism for cleaning up empty rules is missing.

Therefore, consider add method isEmpty() to DatabaseRuleConfiguration to clean up empty rules in the refresh mechanism.

Pace2Car:
Fixes #26751.

### DIFF
--- a/features/sharding/api/src/main/java/org/apache/shardingsphere/sharding/api/config/ShardingRuleConfiguration.java
+++ b/features/sharding/api/src/main/java/org/apache/shardingsphere/sharding/api/config/ShardingRuleConfiguration.java
@@ -65,4 +65,9 @@ public final class ShardingRuleConfiguration implements DatabaseRuleConfiguratio
     private Map<String, AlgorithmConfiguration> auditors = new LinkedHashMap<>();
     
     private ShardingCacheConfiguration shardingCache;
+    
+    @Override
+    public boolean isEmpty() {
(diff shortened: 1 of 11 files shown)


features/sharding/api/src/main/java/org/apache/shardingsphere/sharding/api/config/ShardingRuleConfiguration.java
```diff
@@ -65,4 +65,9 @@ public final class ShardingRuleConfiguration implements DatabaseRuleConfiguratio
     private Map<String, AlgorithmConfiguration> auditors = new LinkedHashMap<>();
     
     private ShardingCacheConfiguration shardingCache;
+    
+    @Override
+    public boolean isEmpty() {
+        return tables.isEmpty() && autoTables.isEmpty() && bindingTableGroups.isEmpty() && null == defaultDatabaseShardingStrategy && null == defaultTableShardingStrategy;
```
RaigorJiang, line 71: if tables and autoTables are empty, then bindingTableGroups must be empty

Pace2Car: Fixed, THX

RaigorJiang (approved):

Issue #27569, opened 2023-07-30
[Oracal SQL] Improve compatibility with Oracle SQL


lancelly:
Hi, community,
This issue is for my OSPP project(https://summer-ospp.ac.cn/org/prodetail/235930377?list=org&navpage=org).
I'm currently working on improving compatibility with Oracle SQL in ShardingSphere.
There are about 50 known unsupported Oracel SQLs in ShardingSphere as listed in the OSPP project. I will refer to Offical Guide of Oracel(https://docs.oracle.com/en/database/oracle/oracle-database/21/sqlrf/index.html) and try to fix most of the unsupported SQLs in the next couple of months. 
@RaigorJiang is my mentor for the OSPP project. I'm looking forward to contribute to the community and get this work done.

lancelly:
Hi, I fixed HR-5-19 in this PR: https://github.com/apache/shardingsphere/pull/27572
It is my first PR in the community. Please take a look~ @RaigorJiang

RaigorJiang:
Hi @lancelly, welcome here!

lancelly:
Hi, I fixed HR-5-158 in this PR: https://github.com/apache/shardingsphere/pull/27603

lancelly:
Hi, I fixed HR-7-1 in this PR: https://github.com/apache/shardingsphere/pull/27604

lancelly:
Hi, I fixed OE-4-6/OE-4-7/OE-4-8 in this PR: https://github.com/apache/shardingsphere/pull/27614

PR #27872, opened 2023-08-02
[Oracle SQL] Support parsing create type for Oracle

lancelly:
In this PR, I referred to the Official guide of Oracle: https://docs.oracle.com/en/database/oracle/oracle-database/21/lnpls/CREATE-TYPE-statement.html#GUID-389D603D-FBD0-452A-8414-240BBBC57034 and supported parsing basic create type clause for Oracle. Since the Create Type clause is rather complex, this pr does not take all the branches of this clause into account.
Also, in this PR, I supported MultisetExpression in CastFunction as it is related to one of the cases in this issue: https://github.com/apache/shardingsphere/issues/27569
The related cases are OE-4-5, OE-5-17, OE-5-21, OE-5-186, OE-6-7, OE-8-2, SH-5-108.


lancelly:
The test cases may seem to be simple, but I find that most of the cases in ddl package look like this.

RaigorJiang (approved):

RaigorJiang:
Great work, thank you! @lancelly

PR #27904, opened 2023-08-04
fix issue-27424 when the table contains case-sensitive fields

### LINKED ISSUE #27424: In sharing-jdbc 5.4.0, insert stament all columnNames to lowerCase
## Bug Report
Version: 5.4.0

I'm using sharding-jdbc, in the insert statement has a case-sensitive field, there will be more than one sharding condition, I debug the source code to find out, because, the source code in the insert statement to get the field, all converted to lowercase, This results in fields that exist being judged as non-existent, for example, accountNo and accountno, which will be judged as accountNo being a non-existent field, and then there will be a field that has two slice conditions, one normal and one with the value null, and then when the value is assigned, the null value overwrites the normal value
image

bpzhang:
Fixes #27424.

  -
when the table contains case-sensitive fields, in ```appendMissingShardingConditions``` SS will judge a same field is not equal
---


strongduanmu:
Hi @bpzhang, thank you for your pr, can you add unit test and e2e test to cover this change?

bpzhang:
> Hi @bpzhang, thank you for your pr, can you add unit test and e2e test to cover this change?

Yeah already add some test cases

RaigorJiang:
Hi @bpzhang 
There is a ci failed, could you fix it?

bpzhang:
> Hi @bpzhang There is a ci failed, could you fix it?

this seems absent a jar in the repository, and should I deploy it to remote repository manual?

RaigorJiang:
> > Hi @bpzhang There is a ci failed, could you fix it?
> 
> this seems absent a jar[shardingsphere-test-e2e-env] in the repository, and should I deploy it to remote repository manual?

That was caused by network problems. I have re-run it.

features/sharding/core/src/main/java/org/apache/shardingsphere/sharding/route/engine/condition/engine/InsertClauseShardingConditionEngine.java
```diff
@@ -100,7 +100,7 @@ private void appendMissingShardingConditions(final InsertStatementContext sqlSta
             return;
         }
         for (String each : allColumnNames) {
-            if (!columnNames.contains(each) && !shardingRule.isGenerateKeyColumn(each, tableName) && shardingRule.findShardingColumn(each, tableName).isPresent()) {
+            if (!columnNames.contains(each.toLowerCase()) && !shardingRule.isGenerateKeyColumn(each, tableName) && shardingRule.findShardingColumn(each, tableName).isPresent()) {
```
strongduanmu, line 103: Why convert each to lowerCase?

infra/binder/src/main/java/org/apache/shardingsphere/infra/binder/context/statement/dml/InsertStatementContext.java
```diff
@@ -93,7 +93,7 @@ public InsertStatementContext(final ShardingSphereMetaData metaData, final List<
         onDuplicateKeyUpdateValueContext = getOnDuplicateKeyUpdateValueContext(params, parametersOffset).orElse(null);
         tablesContext = new TablesContext(getAllSimpleTableSegments(), getDatabaseType());
         ShardingSphereSchema schema = getSchema(metaData, defaultDatabaseName);
-        columnNames = containsInsertColumns() ? insertColumnNames : schema.getVisibleColumnNames(sqlStatement.getTable().getTableName().getIdentifier().getValue());
+        columnNames = containsInsertColumns() ? insertColumnNames : schema.getVisibleColumnNames(sqlStatement.getTable().getTableName().getIdentifier().getValue().toLowerCase());
```
strongduanmu, line 96: The result of getVisibleColumnNames may be orginal column in DBMS.

strongduanmu (changes requested):

features/sharding/core/src/main/java/org/apache/shardingsphere/sharding/route/engine/condition/engine/InsertClauseShardingConditionEngine.java
bpzhang: https://user-images.githubusercontent.com/1734396/255540089-b905d899-c0e7-4053-abc3-f9f7df27054a.png
this method make all field to lowercase, but SS not process the field of table, so there is a problem when table contain case-sensitive fields, even though the field is same one  ,such as 'accountNo'

infra/binder/src/main/java/org/apache/shardingsphere/infra/binder/context/statement/dml/InsertStatementContext.java
bpzhang: this modify  for consistency with the insert statement have fields

features/sharding/core/src/main/java/org/apache/shardingsphere/sharding/route/engine/condition/engine/InsertClauseShardingConditionEngine.java
134130: In my opinion, adapting column identifier to `lowercase` is not good choice. It looks just avoiding this issue.
We need to fix `getColumnNamesForInsertColumns` returns it's own identifier, not `lowercase`.
When just fixing with this way, it will be appeared on other case later.

bpzhang: Yes, all modifications are to fix this problem, not avoid it. I asked this question in the previous issue, and the answer at that time was "ShardingSphere adopts a case-insensitive strategy, which is matched in lowe Rcase"

134130: Aha, Thank you!

strongduanmu (approved):
LGTM, merged.

PR #28327, opened 2023-08-31
Fixes 26754

### LINKED ISSUE #26754: Regarding Distributed Transaction
image
I believe the instructions in this document are not very accurate.  In the context of shardingsphere and single-service cross-database transactions, the XA mode is expected to have better performance than Seata's AT mode.  This is because, in the absence of cross-service scenarios, the XA mode does not require long waits for transaction execution;  it only needs to handle commit and rollback operations.  On the other hand, the AT mode requires querying and writing data copies before and after the write operations, as well as writing undo logs, which clearly consumes more time.

The AT mode outperforms the XA mode only in scenarios involving cross-service transactions.  This is because cross-service transactions result in the inability to process the business logic of each transaction branch before a unified transaction commit, leading to longer waiting times.

Swastyy:
Fixes #26754

### DIFF
--- a/docs/document/content/features/transaction/_index.en.md
+++ b/docs/document/content/features/transaction/_index.en.md
@@ -78,7 +78,7 @@ The following table can be used for comparison to help developers choose the sui
 | Business transformation  | None            | None               | Seata server needed|
 | Consistency    | Not supported         | Not supported             | Final consistency       |
 | Isolation    | Not supported        | Supported           | Business side guaranteed  |
-| Concurrent performance | no loss        | severe loss          | slight loss       |
+| Concurrent performance | no loss        | slight loss          | severe loss       |
 | Applied scenarios  | Inconsistent processing by the business side | short transaction & low-level concurrency | long transaction & high concurrency |
 
 ## Application Scenarios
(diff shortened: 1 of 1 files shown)


docs/document/content/features/transaction/_index.en.md
```diff
@@ -78,7 +78,7 @@ The following table can be used for comparison to help developers choose the sui
 | Business transformation  | None            | None               | Seata server needed|
 | Consistency    | Not supported         | Not supported             | Final consistency       |
 | Isolation    | Not supported        | Supported           | Business side guaranteed  |
-| Concurrent performance | no loss        | severe loss          | slight loss       |
+| Concurrent performance | no loss        | slight loss          | severe loss       |
```
strongduanmu, line 81: Why change this description?

strongduanmu (changes requested):

docs/document/content/features/transaction/_index.en.md
Swastyy: Hi @strongduanmu , is it not required but the information is also not to the point. Let me know if this part of change is not required. Thanks.

strongduanmu: Thank you very much, but this change looks wrong, so I will close this pr.

PR #28465, opened 2023-09-19
Made some optimizations for the isColumnNameNotConflicts method

daleyzou:
In the process of reading the code, some optimizations were made for the isColumnNameNotConflicts method.

Use stream streaming to replace multiple ifs.


Pace2Car:
Thank you!

PR #28625, opened 2023-10-02
[Oracle SQL] Enhancing CreateFunction of Oracle by adding several kinds of statement 

lancelly:
Related issue: https://github.com/apache/shardingsphere/issues/27569
In this pr, I added several kinds of statement for createFunctionStatement of Oracle. I referred to this doc where you can find the detailed definition: https://docs.oracle.com/en/database/oracle/oracle-database/21/lnpls/block.html#GUID-9ACEB9ED-567E-4E1A-A16A-B8B35214FC9D
截屏2023-10-02 22 34 15

Now the following SQL can be parsed:
"CREATE FUNCTION f(cur SYS_REFCURSOR, mgr_hiredate DATE) RETURN NUMBER IS emp_hiredate DATE; before number :=0; after number:=0; begin loop fetch cur into emp_hiredate; exit when cur%NOTFOUND; if emp_hiredate > mgr_hiredate then after:=after+1; else before:=before+1; end if; end loop; close cur; if before > after then return 1; else return 0; end if; end;
"
截屏2023-10-02 22 36 16
截屏2023-10-02 22 41 22


RaigorJiang (approved):

RaigorJiang:
@lancelly Good fix, thank you!

PR #29755, opened 2024-01-17
Add GraalVM Reachability Metadata and corresponding nativeTest for Etcd integration

linghengqian:
For #29052.

Changes proposed in this pull request:
  - Add GraalVM Reachability Metadata and corresponding nativeTest for Etcd integration.
  - Temporarily actively hosting GraalVM Reachability Metadata for `io.netty:netty-transport-classes-epoll:4.1.99.Final`.
  - Further remove unnecessary metadata JSON entries.
  - Change the port used by MySQL's nativeTest. On Windows 11 Professional with WSL2 enabled, `65107` is occupied, which is different from the behavior of Windows 11 Home and Student with WSL2 enabled. Dynamically defining YAML is one of the issues I'm tracking in #29657.
  - Turn on quick build mode for the Maven Profile of `nativeTestInShardingSphere`. This helps development by sacrificing the peak performance of Native Image for faster build speeds. Refer to https://medium.com/graalvm/graalvm-22-1-developer-experience-improvements-apple-silicon-builds-and-more-b7ac9a0f6066 .
  - Set the `buildArgs` of `-J-Xmx7g` for the Maven Profile of `nativeTestInShardingSphere` to prevent Native Image build failures on low-performance devices with less than 10GB of available memory like Github Actions.
  - Renames `graalvm.version` to `graal-sdk.version`. This property actually refers to the Maven module of `org.graalvm.sdk:graal-sdk:21.2.0`, not to the GraalVM CE version used by the developer.

---

Before committing this PR, I'm sure that I have checked the following options:
- [x] My code follows the [code of conduct](https://shardingsphere.apache.org/community/en/involved/conduct/code/) of this project.
- [x] I have self-reviewed the commit code.
- [x] I have (or in comment I request) added corresponding labels for the pull request.
- [x] I have passed maven check locally : `./mvnw clean install -B -T1C -Dmaven.javadoc.skip -Dmaven.jacoco.skip -e`.
- [ ] I have made corresponding changes to the documentation.
- [x] I have added corresponding unit tests for my changes.

Issue #29052, opened 2023-11-16
Improve GraalVM Reachability Metadata and corresponding nativeTest related unit tests


linghengqian:
## Feature Request

**For English only**, other languages will not be accepted.

Please pay attention on issues you submitted, because we maybe need more details. 
If no response anymore and we cannot make decision by current information, we will **close it**.

Please answer these questions before submitting your issue. Thanks!

### Is your feature request related to a problem?

- #21347.

### Describe the feature you would like.

- In the current master branch, for the `sharding`, `readwrite-splitting`, `mask`, `shadow` and `encrypt` features in `standalone` mode, there are already corresponding GraalVM Reachability Metadata and corresponding nativeTest-related unit tests located under GraalVM Native Image.
- The previous PR did not cover further situations. We need to add additional unit tests and corresponding GraalVM Reachability Metadata for the following functions. Some unit tests will require the GraalVM Reachability Metadata of testcontainers.
    - [ ] Cluster mode - Zookeeper, Etcd, Consul
    - [ ] XA Distributed Transactions - Atomikos, Narayana
    - [ ] BASE Distributed Transactions - Seata
    - [ ] DistSQL
    - [ ] Databases other than H2database - MySQL, PostgreSQL, OpenGauss

- In the current master branch, ShardingSphere temporarily hosts the GraalVM Reachability Metadata of some third-party libraries. These irrelevant json entries should be submitted to https://github.com/oracle/graalvm-reachability-metadata and then deleted from the ShardingSphere side.
    - [ ] `com.github.ben-manes.caffeine:caffeine:2.9.3`
    - [ ] `com.zaxxer:HikariCP:4.0.3`
    - [ ] `org.apache.calcite:calcite-core:1.35.0`
    - [ ] `org.apache.groovy:groovy:4.0.10`
    - [ ] `org.apache.shardingsphere.elasticjob:elasticjob-lite-core:3.0.4`, this is addressed at https://github.com/apache/shardingsphere-elasticjob/issues/2258 .
    - [ ] `org.apache.zookeeper:zookeeper:3.9.0`
    - [ ] `org.glassfish.jaxb:jaxb-runtime:2.3.1`
    - [ ] `org.slf4j:slf4j-api:1.7.36`

- The current issue involves several user-side experience issues.
[...]

PR #29995, opened 2024-02-04
Add ClasspathURLProvider implementation that reads system properties

linghengqian:
For #29052.

  - Add ClasspathURLProvider implementation that reads system properties. For nativeTest, we need a way to dynamically define properties in a YAML file. To cover the JDBC Driver scenario, I don't think the lower level Java API should be used in nativeTest.
  - This PR simultaneously modifies some nativeTest, especially to further prevent the problem discovered by #29755 that the ports involved in MySQL's nativeTest are occupied by unknown applications.

---

### DIFF
--- a/jdbc/core/src/main/java/org/apache/shardingsphere/driver/jdbc/core/driver/spi/classpath/ClasspathWithSystemPropsURLProvider.java
+++ b/jdbc/core/src/main/java/org/apache/shardingsphere/driver/jdbc/core/driver/spi/classpath/ClasspathWithSystemPropsURLProvider.java
@@ -56,7 +56,7 @@ public final class ClasspathWithSystemPropsURLProvider extends AbstractClasspath
             while (null != (line = reader.readLine())) {
                 if (!line.startsWith("#")) {
                     line = replaceSystemProperties(line);
-                    builder.append(line).append('\n');
+                    builder.append(line).append(System.lineSeparator());
                 }
             }
             return builder.toString().getBytes(StandardCharsets.UTF_8);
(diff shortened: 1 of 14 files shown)


jdbc/core/src/main/java/org/apache/shardingsphere/driver/jdbc/core/driver/spi/classpath/ClasspathWithSystemPropsURLProvider.java
```diff
@@ -0,0 +1,80 @@
...
+        String file = getConfigurationFile(url, urlPrefix, PATH_TYPE);
+        try (
+                InputStream stream = getResourceAsStream(file);
+                BufferedReader reader = new BufferedReader(new InputStreamReader(stream, StandardCharsets.UTF_8))) {
+            StringBuilder builder = new StringBuilder();
+            String line;
+            while (null != (line = reader.readLine())) {
+                if (!line.startsWith("#")) {
+                    line = replaceSystemProperties(line);
+                    builder.append(line).append('\n');
```
RaigorJiang, line 59: Should we use lineSeparator instead '\n'?

linghengqian: - I made corrections in the second commit. But using `\n` is actually an attempt to be compatible with the previous design, let me see if any E2E CI is broken.

linghengqian: - It looks like CI is normal. I don't have a Windows Native Dev environment, so I'm not sure if this will break anything.

terrymanu (approved):

PR #30145, opened 2024-02-16
Fixes ClassPathURLLoader cannot find files under GraalVM Native Image

linghengqian:
For #30139.

Changes proposed in this pull request:
  - Fixes ClassPathURLLoader cannot find files under GraalVM Native Image. ~~If we can't use `java.io.InputStream` in this class, then obviously we need to manually open `NativeImageResourceFileSystem` under GraalVM Native Image.~~ Split this PR from the #30138 side for clarification.
  - Creating a `java.io.File` instance will break usage in both JAR(`ZipFileSystemProvider`) and GraalVM Native Image(`NativeImageResourceFileSystemProvider`). Refer to https://github.com/oracle/graal/issues/7804 .
  - Obviously, developers cannot create `java.io.File` instances via URLs like `resource:file:/resources/{entry}` or `jar:file:/resources/{entry}`, because creating `java.io.File` instances is only supported by the File System corresponding to the URL of `file:/resources/{entry}`.

---

Before committing this PR, I'm sure that I have checked the following options:
- [x] My code follows the [code of conduct](https://shardingsphere.apache.org/community/en/involved/conduct/code/) of this project.
- [x] I have self-reviewed the commit code.
- [x] I have (or in comment I request) added corresponding labels for the pull request.
- [x] I have passed maven check locally : `./mvnw clean install -B -T1C -Dmaven.javadoc.skip -Dmaven.jacoco.skip -e`.
- [ ] I have made corresponding changes to the documentation.
- [ ] I have added corresponding unit tests for my changes.

PR #30159, opened 2024-02-17
Use @RequiredArgsConstructor instead of @AllArgsConstructor

terrymanu:


linghengqian (approved):
- The fix for NativeTest CI is actually in #30145. I don't think this PR needs to focus on NativeTest.

PR #31687, opened 2024-06-13
Fix Sharding column not tracked through aliases

### LINKED ISSUE #31656: Sharding column not tracked through aliases
5.5.1-SNAPSHOT

Through unit tests

### Expected behavior
I'd expect that aliases would not have an impact on the logic of rewriting a query to run against each shard. The actual alias name doesn't matter - it's the original column that it references that matters (see repro steps below)

### Actual behavior
The column originally aliased in a query is not tracked, but instead only the name is considered.

When rewriting a query by replacing the logical table name with the physical table name, aliases must be tracked so that you know which alias represents the sharding column. Otherwise, the query will not be correctly rewritten.

Add the following to src/test/resources/scenario/sharding/case/dml/select.xml and run ShardingSQLRewriterIT. This unit test should pass:

    <rewrite-assertion id="select_over_aliased_shard_column">
        <input sql="SELECT * FROM (SELECT account_id aliased_account_id FROM t_account) WHERE aliased_account_id  = ?" parameters="100" />
        <output sql="SELECT * FROM (SELECT account_id aliased_account_id FROM t_account_0) WHERE aliased_account_id  = ?" parameters="100" />
    </rewrite-assertion>

Instead ShardingSphere outputs two queries against t_account_0 and t_account_1 not realizing that the sharding column is being constrained.

The reverse, when a different column is aliased with the name of the sharding column. ShardingSphere thinks that the sharding column is being constrained while this is not the case. This unit test should pass:

    <rewrite-assertion id="select_over_alias_renamed_as_shard_column">
        <input sql="SELECT * FROM (SELECT amount account_id FROM t_account) WHERE account_id  = ?" parameters="100" />
        <output sql="SELECT * FROM (SELECT amount account_id FROM t_account_0) WHERE account_id  = ?" parameters="100" />
        <output sql="SELECT * FROM (SELECT amount account_id FROM t_account_1) WHERE account_id  = ?" parameters="100" />
    </rewrite-assertion>

Instead ShardingSphere thinks that the sharding column is being constrained.

I can put together a PR that adds these tests if that's helpful

TherChenYang:
Fixes #31656.

  - Fix Sharding column not tracked through aliases
  - When there is an alias for our sharded column, errors may occur in judging the shard condition, leading to routing all tables in the database.
  - When the alias of our regular column happens to be the same as the partitioned column, it will also incorrectly judge the slice condition, resulting in an incorrect route.

``` sql
-- alias for sharding column
SELECT * FROM (SELECT account_id aliased_account_id FROM t_account) t1 WHERE aliased_account_id  = 100

-- Incorrect routing result
SELECT * FROM (SELECT account_id aliased_account_id FROM t_account_0) t1 WHERE aliased_account_id  = 100
SELECT * FROM (SELECT account_id aliased_account_id FROM t_account_1) t1 WHERE aliased_account_id  = 100

-- Correct routing result
SELECT * FROM (SELECT account_id aliased_account_id FROM t_account_0) t1 WHERE aliased_account_id  = 100
```
     
``` sql
--- sharding column with alias 
SELECT * FROM (SELECT amount account_id FROM t_account) t1 WHERE account_id  = 100

-- Incorrect routing result
SELECT * FROM (SELECT amount account_id FROM t_account_0) t1 WHERE account_id  = 100

-- Correct routing result
SELECT * FROM (SELECT amount account_id FROM t_account_0) t1 WHERE account_id  = 100
SELECT * FROM (SELECT amount account_id FROM t_account_1) t1 WHERE account_id  = 100

```

---

### DIFF
--- a/features/sharding/core/src/main/java/org/apache/shardingsphere/sharding/route/engine/condition/engine/WhereClauseShardingConditionEngine.java
+++ b/features/sharding/core/src/main/java/org/apache/shardingsphere/sharding/route/engine/condition/engine/WhereClauseShardingConditionEngine.java
@@ -120,7 +120,7 @@ public final class WhereClauseShardingConditionEngine {
         for (ExpressionSegment each : predicates) {
             for (ColumnSegment columnSegment : ColumnExtractor.extract(each)) {
                 Optional<String> tableName = Optional.ofNullable(columnTableNames.get(columnSegment.getExpression()));
-                Optional<String> shardingColumn = tableName.flatMap(optional -> shardingRule.findShardingColumn(columnSegment.getIdentifier().getValue(), optional));
+                Optional<String> shardingColumn = tableName.flatMap(optional -> shardingRule.findShardingColumn(columnSegment.getColumnBoundedInfo().getOriginalColumn().getValue(), optional));
                 if (!tableName.isPresent() || !shardingColumn.isPresent()) {
                     continue;
                 }
--- a/test/it/rewriter/src/test/resources/scenario/sharding/case/dml/select.xml
+++ b/test/it/rewriter/src/test/resources/scenario/sharding/case/dml/select.xml
@@ -17,6 +17,28 @@
   -->
 
 <rewrite-assertions yaml-rule="scenario/sharding/config/sharding-rule.yaml">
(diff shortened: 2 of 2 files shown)


strongduanmu (approved):
@TherChenYang Thank you for your work, merged.

Issue #31912, opened 2024-06-28
Improve the federated query MySQL function


zihaoAK47:
# Background

Hi Community,

Federated queries help users perform cross-database join queries, subqueries, and aggregate queries. Currently, we plan to enhance MySQL functions in the federated query environment to support more SQL execution scenarios in federated query contexts.

For more details, please refer to: [Apache ShardingSphere Documentation on SQL Federation](https://shardingsphere.apache.org/document/current/cn/features/sql-federation/).

# Task

- [x] NOT [link](https://dev.mysql.com/doc/refman/8.4/en/logical-operators.html#operator_not)
- [x] ATAN [link](https://dev.mysql.com/doc/refman/8.4/en/mathematical-functions.html#function_atan)  @zihaoAK47 
- [ ] BENCHMARK [link](https://dev.mysql.com/doc/refman/8.4/en/information-functions.html#function_benchmark)
- [ ] BIN [link](https://dev.mysql.com/doc/refman/8.4/en/string-functions.html#function_bin) @zihaoAK47 
- [ ] CRC32 [link](https://dev.mysql.com/doc/refman/8.4/en/mathematical-functions.html#function_crc32)
- [ ] SUBSTRING_INDEX [link](https://dev.mysql.com/doc/refman/8.4/en/string-functions.html#function_substring-index)
- [ ] FORMAT [link](https://dev.mysql.com/doc/refman/8.4/en/string-functions.html#function_format)
- [ ] CONV [link](https://dev.mysql.com/doc/refman/8.4/en/mathematical-functions.html#function_conv)
- [ ] SUBSTRING [link](https://dev.mysql.com/doc/refman/8.4/en/string-functions.html#function_substring)
- [ ] CHAR [link](https://dev.mysql.com/doc/refman/8.4/en/string-functions.html#function_char)
- [ ] ELT [link](https://dev.mysql.com/doc/refman/8.4/en/string-functions.html#function_elt)
- [ ] EXPORT_SET [link](https://dev.mysql.com/doc/refman/8.4/en/string-functions.html#function_export-set)
- [ ] FIELD [link](https://dev.mysql.com/doc/refman/8.4/en/string-functions.html#function_field)
- [ ] FIND_IN_SET [link](https://dev.mysql.com/doc/refman/8.4/en/string-functions.html#function_find-in-set)
- [ ] HEX [link](https://dev.mysql.com/doc/refman/8.4/en/string-functions.html#function_hex)
[...]

PR #31924, opened 2024-06-29
Improve atan and anan2 sql function for mysql

zihaoAK47:
Ref #31912.

  - Improve atan and anan2 sql function for mysql

mysql is compatible with calls to different numbers of parameters

截屏2024-06-30 00 04 22

---

### DIFF
--- a/kernel/sql-federation/optimizer/src/main/java/org/apache/shardingsphere/sqlfederation/optimizer/planner/util/SQLFederationFunctionUtils.java
+++ b/kernel/sql-federation/optimizer/src/main/java/org/apache/shardingsphere/sqlfederation/optimizer/planner/util/SQLFederationFunctionUtils.java
@@ -19,6 +19,7 @@ package org.apache.shardingsphere.sqlfederation.optimizer.planner.util;
 
 import lombok.AccessLevel;
 import lombok.NoArgsConstructor;
+import org.apache.calcite.runtime.SqlFunctions;
 import org.apache.calcite.schema.SchemaPlus;
 import org.apache.calcite.schema.impl.ScalarFunctionImpl;
 import org.apache.shardingsphere.infra.autogen.version.ShardingSphereVersion;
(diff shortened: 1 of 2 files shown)


kernel/sql-federation/optimizer/src/main/java/org/apache/shardingsphere/sqlfederation/optimizer/planner/util/SQLFederationFunctionUtils.java
```diff
@@ -50,6 +51,8 @@ public static void registryUserDefinedFunction(final String schemaName, final Sc
         schemaPlus.add("pg_catalog.intervaltonum", ScalarFunctionImpl.create(SQLFederationFunctionUtils.class, "intervalToNum"));
         schemaPlus.add("pg_catalog.gs_password_notifyTime", ScalarFunctionImpl.create(SQLFederationFunctionUtils.class, "gsPasswordNotifyTime"));
         schemaPlus.add("bit_count", ScalarFunctionImpl.create(MySQLBitCountFunction.class, "bitCount"));
+        schemaPlus.add("atan", ScalarFunctionImpl.create(SqlFunctions.class, "atan2"));
+        schemaPlus.add("atan2", ScalarFunctionImpl.create(SqlFunctions.class, "atan"));
```
strongduanmu, line 55: Is SqlFunctions a built-in class in Caclite? Why doesn't Calcite use the functions in this class by default?

strongduanmu (changes requested):

kernel/sql-federation/optimizer/src/main/java/org/apache/shardingsphere/sqlfederation/optimizer/planner/util/SQLFederationFunctionUtils.java
zihaoAK47: Yes, SqlFunctions is a built-in class in Calcite that implements the execution logic for atan and atan2 functions.  However, in Calcite, the atan function method signature only allows one parameter, while the atan2 function method signature allows two parameters.

MySQL handles the atan function differently depending on the number of parameters: it uses a different execution logic for one parameter (atan) versus two parameters (atan2).

When Calcite encounters a call to the atan function with two parameters, it can find the implementation based on the function name but fails during parameter matching because in Calcite, the atan function can only accept a single parameter.

To resolve this, I added an implementation for the atan2 function alongside atan.  Now, at runtime, Calcite can find both implementations of these functions.  It will match based on the number of parameters and invoke the appropriate method, thereby calling the atan2 function implementation when two parameters are provided.

In MySQL, the atan2 function can indeed be executed with just one parameter.

Ref: https://github.com/mysql/mysql-server/blob/trunk/sql/item_func.cc#L2906
image

strongduanmu: That is, they are equivalent in the MySQL dialect. Thanks for clarifying that, this is a great PR.

strongduanmu (approved):
LGTM, merged.

PR #33095, opened 2024-10-02
Support running DistSQL under Proxy Native in the form of GraalVM Native Image

linghengqian:
For #29052.

  - Support running DistSQL under Proxy Native in the form of GraalVM Native Image.
  - In order to directly collect the GraalVM Reachability Metadata of ShardingSphere Proxy through unit testing, we must directly start ShardingSphere Proxy without using the Docker Image.
  - The most embarrassing part is that ShardingSphere Proxy cannot actually connect to H2database, which requires starting a MySQL container.
  - The most awkward part of this PR is the splicing of DistSQL in `org.apache.shardingsphere.test.natived.proxy.features.ShardingTest`. Since we need to use the MySQL port of the MySQL container obtained from the testcontainers-java API, we always need to manually splice DistSQL instead of directly reading the `.sql` file. But the problem is that JDK8 does not have the concept of multi-line strings, which makes the related unit tests quite difficult to maintain. Maybe there is a way to write these unit tests under JDK 21.
  - `java.util.LinkedHashSet` is a magical exception. Running ShardingSphere Proxy under GraalVM Native Image requires its constructor. Add it to `extra-filter.json`.

---

### DIFF
--- a/.github/workflows/nightly-build.yml
+++ b/.github/workflows/nightly-build.yml
@@ -133,7 +133,7 @@ jobs:
           password: ${{ secrets.GITHUB_TOKEN }}
       - name: Push Docker Image
         run: |
-          ./mvnw -am -pl distribution/proxy-native -Prelease.native,docker.buildx.push.native -B -T1C -DskipTests -Dproxy.image.repository=${{ env.PROXY_NATIVE }} -Dproxy.image.tag=${{ github.sha }} clean package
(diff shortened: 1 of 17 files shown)


linghengqian (commented):
- The focus of the current PR is a bit complicated, so it is necessary to temporarily split some irrelevant parts.

taojintianxia (approved):

PR #34243, opened 2025-01-04
Refactor argument placeholder to support SPI

JoshuaChen:
In some cloud compliance scenarios, using environment properties or system properties to manage secrets is not allowed. Instead, we need to use approved property sources to handle these sensitive values.

To meet this requirement, I refactored the argument placeholder so it can support reading from other property sources through a plugin module. This ensures compliance with the rules.

---

### DIFF
--- a/infra/argument/core/pom.xml
+++ b/infra/argument/core/pom.xml
@@ -20,34 +20,24 @@
     <modelVersion>4.0.0</modelVersion>
     <parent>
         <groupId>org.apache.shardingsphere</groupId>
-        <artifactId>shardingsphere</artifactId>
+        <artifactId>shardingsphere-infra-argument</artifactId>
         <version>5.5.2-SNAPSHOT</version>
     </parent>
-    <artifactId>shardingsphere-infra</artifactId>
-    <packaging>pom</packaging>
+    <artifactId>shardingsphere-infra-argument-core</artifactId>
     <name>${project.artifactId}</name>
     
-    <modules>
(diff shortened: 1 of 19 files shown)


strongduanmu (changes requested):
The scenario you introduced does not seem to be something ShardingSphere should consider. ShardingSphere should be as simple and standard as possible instead of considering personal scenarios.

JoshuaChen:
Um~I believe compliance requirements are not personal preferences but mandatory rules. In China, we have security classification protection requirements, while multinational companies must comply with various regulations in different countries.

Here’s a simple example: In AWS, sensitive keys, like database passwords, should be stored in AWS Secrets Manager. They should not appear in environment variables or system properties because this could expose the database password to unauthorized users. In private infrastructure, such keys should be stored in tools like HashiCorp Vault or similar products.

Additionally, for an open-source product, I agree that it should remain simple enough to use. However, I also believe it needs a certain level of flexibility so that others can extend it based on their specific needs. Within the scope of this PR, I am only providing the ability for others to extend properties with additional data sources.

terrymanu:
I believe that security requirements are important, but ShardingSphere is an infrastructure component, and it should not include more things that are not directly related to the ShardingSphere project itself. 

Integrating Secrets Manager with ShardingSphere is the responsibility of the business application (or another project that integrates Secrets Manager and ShardingSphere), rather than embedding Secrets Manager into ShardingSphere.

PR #34332, opened 2025-01-13
Fix issue:Add field type judgment in PaginationContext

### LINKED ISSUE #34289: Class MySQLComStmtExecuteExecutor throws a NullPointerException when select with limit by Golang
5.5.1

proxy

### Expected behavior
select success

### Actual behavior
Class MySQLComStmtExecuteExecutor throws NullPointerException
image

For some prepared query SQL statements with limit condition, class PaginationContext will throw a ClassCastException when init creating, causing the query fail, and the final manifestation is that class MySQLComStmtExecuteExecutor throws a NullPointerException

1、use Golang to execute the SQL statement
image
2、sql statement example：
select id, content from test_table where version=? and create_time>? order by create_time desc limit ?, ?

Class:PaginationContext
the obj is instanceof byte[], if it is forcefully converted to int, a ClassCastException will be thrown
image

strawberry1314:
Fixes #34289.
Add type judgment for the limit parameter

  -

---


infra/binder/src/main/java/org/apache/shardingsphere/infra/binder/context/segment/select/pagination/PaginationContext.java
```diff
@@ -58,7 +58,7 @@ private Long getValue(final PaginationValueSegment paginationValueSegment, final
             if (null == obj) {
                 return null;
             }
-            return obj instanceof Long ? (long) obj : (int) obj;
+            return obj instanceof byte[] ? (long) (((byte[]) obj)[0] - '0') : obj instanceof Long ? (long) obj : (int) obj;
```
strongduanmu, line 61: Why is the LIMIT parameter a byte[]? Is this a special treatment of the MySQL protocol for the Go language?

strongduanmu: If possible, please add an E2E SQL that can reproduce this exception. You can add a new case in https://github.com/apache/shardingsphere/blob/master/test/e2e/sql/src/test/resources/cases/dql/e2e-dql-select.xml.

strongduanmu (changes requested):

infra/binder/src/main/java/org/apache/shardingsphere/infra/binder/context/segment/select/pagination/PaginationContext.java
strawberry1314: I am not particularly clear about the processing method of the Go language, I just found that the LIMIT parameter will be resolved to byte[] when I use Golang to execute prepare sql.
such as : SELECT order_id FROM t_order LIMIT ?

strawberry1314: > If possible, please add an E2E SQL that can reproduce this exception. You can add a new case in https://github.com/apache/shardingsphere/blob/master/test/e2e/sql/src/test/resources/cases/dql/e2e-dql-select.xml.

I have already added an E2E SQL case

terrymanu:
Thanks for the fix — the direction of addressing the byte[] pagination parameter issue is absolutely correct. The original problem (Go driver / binary protocol producing a byte[] for LIMIT/OFFSET and causing a ClassCastException → NPE in MySQLComStmtExecuteExecutor) is real, and adding an extra branch for byte[] makes sense.

One small concern:
The current implementation extracts the numeric value by using only the first byte:

((byte[]) obj)[0] - '0'


This works for single-digit limits, but may lead to incorrect pagination values for multi-digit inputs like 10, 100, etc. For such cases the method would return 1 instead of the full number. This silently changes query results, which could be more dangerous than the original explicit failure.

To make the behavior more robust, it would be safer to parse the full value, e.g.:

Treat the byte[] as a UTF-8 numeric string (new String(bytes) + Long.parseLong), or

Decode according to the upstream protocol rules if the bytes represent a binary-encoded integer.

This keeps the fix correct while avoiding potential functional regressions.

Everything else in the PR looks good — thanks again for working on this! If needed, I’m happy to help with additional test coverage (e.g., LIMIT ≥ 10).

strawberry1314:
> Thanks for the fix — the direction of addressing the byte[] pagination parameter issue is absolutely correct. The original problem (Go driver / binary protocol producing a byte[] for LIMIT/OFFSET and causing a ClassCastException → NPE in MySQLComStmtExecuteExecutor) is real, and adding an extra branch for byte[] makes sense.
> 
> One small concern: The current implementation extracts the numeric value by using only the first byte:
> 
> ((byte[]) obj)[0] - '0'
> 
> This works for single-digit limits, but may lead to incorrect pagination values for multi-digit inputs like 10, 100, etc. For such cases the method would return 1 instead of the full number. This silently changes query results, which could be more dangerous than the original explicit failure.
> 
> To make the behavior more robust, it would be safer to parse the full value, e.g.:
> 
> Treat the byte[] as a UTF-8 numeric string (new String(bytes) + Long.parseLong), or
> 
> Decode according to the upstream protocol rules if the bytes represent a binary-encoded integer.
> 
> This keeps the fix correct while avoiding potential functional regressions.
> 
> Everything else in the PR looks good — thanks again for working on this! If needed, I’m happy to help with additional test coverage (e.g., LIMIT ≥ 10).



> Thanks for the fix — the direction of addressing the byte[] pagination parameter issue is absolutely correct. The original problem (Go driver / binary protocol producing a byte[] for LIMIT/OFFSET and causing a ClassCastException → NPE in MySQLComStmtExecuteExecutor) is real, and adding an extra branch for byte[] makes sense.
> 
> One small concern: The current implementation extracts the numeric value by using only the first byte:
> 
> ((byte[]) obj)[0] - '0'
> 
> This works for single-digit limits, but may lead to incorrect pagination values for multi-digit inputs like 10, 100, etc. For such cases the method would return 1 instead of the full number. This silently changes query results, which could be more dangerous than the original explicit failure.
> 
> To make the behavior more robust, it would be safer to parse the full value, e.g.:
> 
> Treat the byte[] as a UTF-8 numeric string (new String(bytes) + Long.parseLong), or
> 
> Decode according to the upstream protocol rules if the bytes represent a binary-encoded integer.
> 
> This keeps the fix correct while avoiding potential functional regressions.
> 
> Everything else in the PR looks good — thanks again for working on this! If needed, I’m happy to help with additional test coverage (e.g., LIMIT ≥ 10).

Thanks for the thorough review and for catching this important issue with multi-digit values. You're absolutely right, I have made the modifications.
Besides, thank you for helping to add additional test coverage .

terrymanu (changes requested):
1. Rebase and paths
     The branch modifies both a removed path (infra/binder/src/.../PaginationContext.java) and the current one (infra/binder/core/.../PaginationContext.java). Please rebase onto the latest master and keep only changes under the existing core path to avoid conflicts and
        dead code.
  2. Byte[] handling for pagination params is fragile
     Currently byte[] is treated as an ASCII string; if a client sends integers via the binary protocol, this can still fail or parse incorrectly. Please investigate how pagination params arrive under the MySQL binary protocol and adjust PaginationContext’s conversion
        to handle Number and binary/string byte[] robustly, with clear errors on invalid input, and add unit tests covering multi-digit, negative, and non-ASCII cases.
  3. Tests and E2E
     The existing E2E tests use the Java driver and cannot verify Golang/binary cases. Please revert the E2E changes in this PR, add unit tests for the byte[] compatibility above, and add a unit test for the MySQL parameter column-definition bounds check.
  4. Evidence
     Please provide a screenshot or logs showing a Golang driver executing a query with LIMIT/OFFSET placeholders successfully after the fix, to confirm the target scenario is covered.

terrymanu:

PR #34307, opened 2025-01-11
Add GraalVM Reachability Metadata and corresponding nativeTest for Firebird

linghengqian:
For #29052.

Changes proposed in this pull request:
  - Add GraalVM Reachability Metadata and corresponding nativeTest for Firebird.
  - Non-essential parts of this PR have been split into other PRs, including,
    - #34333
    - #34335
    - ~#34358~
  - The current PR is affected by multiple external issues.
    - According to https://firebirdsql.org/en/news/official-firebird-docker-images , there is no official Docker Image for Firebird. Use the recommended https://github.com/fdcastel/firebird-docker instead
    -  Cannot use `create table if not exists` for Docker Image `ghcr.io/fdcastel/firebird:5.0.1`, see https://github.com/FirebirdSQL/firebird/issues/8062
    - Docker Image `ghcr.io/fdcastel/firebird:5.0.1` does not work with `DROP TABLE IF EXISTS`, see https://github.com/FirebirdSQL/firebird/issues/4203
    - Docker Image `ghcr.io/fdcastel/firebird:5.0.1` cannot use `TRUNCATE TABLE`, see https://github.com/FirebirdSQL/firebird/issues/2892
    - Due to https://github.com/FirebirdSQL/jaybird/issues/629 and https://github.com/FirebirdSQL/jaybird/blob/master/devdoc/jdp/jdp-2024-02-create-database-through-jdbc-url.adoc , the SQL statement `Create Database` cannot be executed on the Firebird JDBC driver. Unit testing requires the use of `org.firebirdsql.management.FBManager`.
  - **Temporarily disable unit testing of Seata integration of Proxy Native. It's strange that starting Seata Client causes connection leak in Firebird's NativeTest. This requires a separate investigation.**
---

Before committing this PR, I'm sure that I have checked the following options:
- [x] My code follows the [code of conduct](https://shardingsphere.apache.org/community/en/involved/conduct/code/) of this project.
- [x] I have self-reviewed the commit code.
- [x] I have (or in comment I request) added corresponding labels for the pull request.
- [x] I have passed maven check locally : `./mvnw clean install -B -T1C -Dmaven.javadoc.skip -Dmaven.jacoco.skip -e`.
- [x] I have made corresponding changes to the documentation.
- [x] I have added corresponding unit tests for my changes.
- [x] I have updated the Release Notes of the current development version. For more details, see [Update Release Note](https://shardingsphere.apache.org/community/en/involved/contribute/contributor/)

PR #33773, opened 2024-11-23
Added firebird database type and parser module

TedCraft:
Fixes #33699.

Changes proposed in this pull request:
  - Added Firebird related database type
  - Added Firebird related parser module
  - Added corresponding tests for Firebird database type and parser module

---

Before committing this PR, I'm sure that I have checked the following options:
- [x] My code follows the [code of conduct](https://shardingsphere.apache.org/community/en/involved/conduct/code/) of this project.
- [x] I have self-reviewed the commit code.
- [ ] I have (or in comment I request) added corresponding labels for the pull request.
- [x] I have passed maven check locally : `./mvnw clean install -B -T1C -Dmaven.javadoc.skip -Dmaven.jacoco.skip -e`.
- [ ] I have made corresponding changes to the documentation.
- [x] I have added corresponding unit tests for my changes.
- [x] I have updated the Release Notes of the current development version. For more details, see [Update Release Note](https://shardingsphere.apache.org/community/en/involved/contribute/contributor/)

PR #34335, opened 2025-01-13
Support connecting to Firebird via jdbcUrl containing the absolute path to fdb

linghengqian:
For #29052.

  - Support connecting to Firebird via jdbcUrl containing the absolute path to fdb. The current PR was split from #34307 to facilitate review.
  - This is due to #33773 which does not support resolving jdbcUrls like `jdbc:firebird://localhost:32783//var/lib/firebird/data/demo_ds_2.fdb`.
  - The current PR is affected by multiple external issues.
    - Done discussion around LICENSE for `org.firebirdsql.jdbc:jaybird:5.0.6.java8` not being distributed under the ASF license at https://github.com/apache/gravitino/issues/4352#issuecomment-2453295083 .
    - Due to the change of the minimum supported JDK version to JDK17 as reported by https://firebirdsql.org/en/news/jaybird-6-0-0-released , it is not possible to use `org.firebirdsql.jdbc:jaybird:6.0.0` in unit tests of shardingsphere. We still need JDK8 to run unit tests.
    - According to some verification in https://github.com/FirebirdSQL/jaybird/blob/v6.0.0/src/test/org/firebirdsql/jdbc/FBDriverTest.java , it seems that the exception should not be thrown although a jdbcUrl like `jdbc:firebirdsql:xxxxxxxx` does not make sense.

---

### DIFF
--- a/infra/database/type/firebird/src/test/java/org/apache/shardingsphere/infra/database/firebird/connector/FirebirdConnectionPropertiesParserTest.java
+++ b/infra/database/type/firebird/src/test/java/org/apache/shardingsphere/infra/database/firebird/connector/FirebirdConnectionPropertiesParserTest.java
@@ -19,10 +19,10 @@ package org.apache.shardingsphere.infra.database.firebird.connector;
 
 import org.apache.shardingsphere.infra.database.core.connector.ConnectionProperties;
 import org.apache.shardingsphere.infra.database.core.connector.ConnectionPropertiesParser;
-import org.apache.shardingsphere.infra.database.core.exception.UnrecognizedDatabaseURLException;
 import org.apache.shardingsphere.infra.database.core.spi.DatabaseTypedSPILoader;
 import org.apache.shardingsphere.infra.database.core.type.DatabaseType;
 import org.apache.shardingsphere.infra.spi.type.typed.TypedSPILoader;
+import org.apache.shardingsphere.test.util.PropertiesBuilder;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtensionContext;
 import org.junit.jupiter.params.ParameterizedTest;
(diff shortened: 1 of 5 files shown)


infra/database/type/firebird/src/test/java/org/apache/shardingsphere/infra/database/firebird/connector/FirebirdConnectionPropertiesParserTest.java
```diff
@@ -54,14 +56,29 @@ void assertNewConstructor(final String name, final String url, final String host
     
     @Test
     void assertNewConstructorFailure() {
-        assertThrows(UnrecognizedDatabaseURLException.class, () -> parser.parse("jdbc:firebirdsql:xxxxxxxx", null, null));
+        assertDoesNotThrow(() -> parser.parse("jdbc:firebirdsql:xxxxxxxx", null, null));
+        assertThrows(SQLNonTransientConnectionException.class, () -> parser.parse("jdbc:firebirdsql://localhost:c:/data/db/test.fdb", null, null));
```
linghengqian, line 60: - - According to some verification in https://github.com/FirebirdSQL/jaybird/blob/v6.0.0/src/test/org/firebirdsql/jdbc/FBDriverTest.java , it seems that the exception should not be thrown although a jdbcUrl like `jdbc:firebirdsql:xxxxxxxx` does not make sense.
- What do you think? @TedCraft

TedCraft: According to the manual (https://github.com/apache/shardingsphere/pull/34335#discussion_r1912879440) the jdbcUrl format jdbc:firebirdsql:xxxxxxxxxxxx kinda makes sense, since it uses the legacy url format used in firebird. In this case, we just specify the database alias (or path) and the host and port are used by default (localhost:3050). But even though this url type is supported, according to the same manual, it is recommended to use the new url format.

linghengqian: - My understanding is that I should respect the performance of the Firebird JDBC Driver's classes. 

- There seems to be only the LGPL LICENSE for the Firebird JDBC Driver, so I would assume that everyone who uses both shardingsphere and the Firebird JDBC Driver would be affected by the LGPL. Maven's `optional` property is enough to comply with the ASF's requirements when distributing. I think there is no problem with PR at the moment.

strongduanmu (approved):

PR #34859, opened 2025-03-04
Include shell script in repo to start proxy container with local changes

omkar-shitole:
Support/FIX: Start the container instantly just by running `sh start.sh.`

Added a `start.sh `file in the repo to help contributors/developers start a docker proxy container in a completely ready state.

Start the container including all the latest changes in the local repo. instantly just by running the start.sh script.

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -1,10 +1,24 @@
 #!/bin/bash
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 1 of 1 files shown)


terrymanu:
Thanks a lot for the contribution and for sharing this helper script. We really appreciate the intention to improve local development experience.

However, after reviewing the PR in detail, we are unable to merge it at this time, mainly for the following reasons:

1. The script duplicates existing build & Docker logic

Most of the steps in this script (building the proxy Docker image, obtaining the project version, running the container, mounting configuration directories) already exist in our current Maven profiles and Docker setup.
Introducing a separate root-level script creates parallel logic that will need to be maintained independently.

2. Strong assumptions about environment

The script assumes:

Bash availability,

Linux container support,

Specific terminal programs (gnome-terminal / konsole),

PostgreSQL client (psql),

Hard-coded ports and user names.

These assumptions do not hold for many contributors (e.g., Windows users, non-PSQL users), so the script cannot serve as a general-purpose development tool.

3. Tight coupling to repository structure

The script directly depends on current directory layout (distribution/proxy) and image-building conventions.
Any future structural changes would silently break it, adding maintenance burden.

4. Not aligned with the project’s tooling direction

We try to keep ad-hoc developer scripts out of the repository root unless they are:

cross-platform,

integrated with existing tooling,

supported long-term.

This script is very helpful for personal/local use, but not yet suitable as an official tool shipped in the repository.

### Summary

The idea is valuable, and we do want to improve development workflows, but this specific script cannot be merged in its current form due to maintainability, portability, and duplication issues.

If you are interested, we are happy to discuss integrating this into a unified tooling module (e.g., tools/ or a docker-compose based workflow) where we can design something reusable and cross-platform.

Thanks again for the proposal and for contributing!

PR #35348, opened 2025-05-09
Fix `show processlist` not wait for all nodes

### LINKED ISSUE #35299: The results of show processlist may be incomplete
5.5.3-SNAPSHOT 99aef45f4fe7eafbbca1ca7092f15fdb883fd460

ShardingSphere-Proxy

### Expected behavior
show processlist displays all sessions

### Actual behavior
Sometimes all sessions are shown, sometimes some are missing

When waiting for each node to report, it may exit early

1. Start multiple ShardingSphere-Proxy nodes and join the same cluster
2. Start multiple mysql clients and connect to different proxy nodes
3. Execute `select sleep(200)` on each client
4. Connect to any proxy and execute `show processlist`  and observe multiple times

#### Case1
Image

#### Case2
Image

RaigorJiang:
Fixes #35299.

Change ReentrantLock to CountDownLatch and wait for all nodes to respond.

This adjustment affects both the `show processlist` and `kill query` functions.


jiangML (approved):

RaigorJiang:
### After fix
image

PR #36370, opened 2025-08-20
support IN clause optimization for sharding key

Fottas:
Related to #36454

### DIFF
--- a/features/sharding/api/src/main/java/org/apache/shardingsphere/sharding/spi/ShardingAlgorithm.java
+++ b/features/sharding/api/src/main/java/org/apache/shardingsphere/sharding/spi/ShardingAlgorithm.java
@@ -18,7 +18,6 @@
 package org.apache.shardingsphere.sharding.spi;
 
 import org.apache.shardingsphere.infra.algorithm.core.ShardingSphereAlgorithm;
-
 import java.util.Optional;
 
 /**
--- a/features/sharding/core/src/main/java/org/apache/shardingsphere/sharding/rewrite/token/generator/impl/ShardingInPredicateTokenGenerator.java
+++ b/features/sharding/core/src/main/java/org/apache/shardingsphere/sharding/rewrite/token/generator/impl/ShardingInPredicateTokenGenerator.java
@@ -49,6 +49,7 @@ import org.apache.shardingsphere.sql.parser.statement.core.segment.dml.expr.InEx
 import org.apache.shardingsphere.sql.parser.statement.core.segment.dml.expr.ListExpression;
 import org.apache.shardingsphere.sql.parser.statement.core.segment.dml.expr.simple.LiteralExpressionSegment;
 import org.apache.shardingsphere.sql.parser.statement.core.segment.dml.expr.simple.ParameterMarkerExpressionSegment;
+import org.apache.shardingsphere.sql.parser.statement.core.segment.dml.predicate.WhereSegment;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
@@ -66,17 +67,21 @@ import java.util.stream.Stream;
  * Refactored for enhanced readability, maintainability, and performance with Java 8 compatibility.
  */
 @RequiredArgsConstructor
-public final class ShardingInPredicateTokenGenerator implements CollectionSQLTokenGenerator<SQLStatementContext>,
-        RouteContextAware, IgnoreForSingleRoute, ParametersAware {
-
+public final class ShardingInPredicateTokenGenerator
+        implements
+            CollectionSQLTokenGenerator<SQLStatementContext>,
+            RouteContextAware,
+            IgnoreForSingleRoute,
+            ParametersAware {
+    
     private final ShardingRule shardingRule;
-
+    
     @Setter
     private RouteContext routeContext;
-
+    
     @Setter
     private List<Object> parameters = Collections.emptyList();
-
+    
     /**
      * Check if SQL token generation is needed for the given statement context.
      */
@@ -85,7 +90,7 @@ public final class ShardingInPredicateTokenGenerator implements CollectionSQLTok
         return sqlStatementContext instanceof SelectStatementContext
                 && hasOptimizableInConditions((SelectStatementContext) sqlStatementContext);
     }
-
+    
     /**
      * Generate SQL tokens for optimizing IN predicates in sharding scenarios.
      */
@@ -97,7 +102,7 @@ public final class ShardingInPredicateTokenGenerator implements CollectionSQLTok
                 .filter(Objects::nonNull)
                 .collect(Collectors.toList());
     }
-
+    
     /**
      * Check if there are optimizable IN conditions in the SELECT statement.
      *
@@ -107,7 +112,7 @@ public final class ShardingInPredicateTokenGenerator implements CollectionSQLTok
     private boolean hasOptimizableInConditions(final SelectStatementContext selectContext) {
         return !extractOptimizableInExpressions(selectContext).isEmpty();
     }
-
+    
     /**
      * Extract all IN expressions that can be optimized for sharding.
      *
@@ -115,24 +120,18 @@ public final class ShardingInPredicateTokenGenerator implements CollectionSQLTok
      * @return List of optimizable IN expressions
      */
     private List<InExpression> extractOptimizableInExpressions(final SelectStatementContext selectContext) {
-        return extractAllInExpressions(selectContext)
+        Collection<WhereSegment> whereSegments = selectContext.getWhereSegments();
+        if (whereSegments == null || whereSegments.isEmpty()) {
+            return Collections.emptyList();
+        }
+        return whereSegments
+                .stream()
+                .flatMap(segment -> extractInExpressionsRecursively(segment.getExpr()))
                 .filter(this::isOptimizable)
                 .filter(inExpr -> isShardingRelated(inExpr, selectContext))
                 .collect(Collectors.toList());
     }
-
-    /**
-     * Extract all IN expressions from WHERE clauses recursively.
-     *
-     * @param selectContext Select statement context
-     * @return Stream of all IN expressions
-     */
-    private Stream<InExpression> extractAllInExpressions(final SelectStatementContext selectContext) {
-        return selectContext.getWhereSegments()
-                .stream()
-                .flatMap(segment -> extractInExpressionsRecursively(segment.getExpr()));
-    }
-
+    
     /**
      * Recursively extract IN expressions from nested binary operations.
      *
@@ -151,7 +150,7 @@ public final class ShardingInPredicateTokenGenerator implements CollectionSQLTok
         }
         return Stream.empty();
     }
-
+    
     /**
      * Check if an IN expression is optimizable (has list with multiple items).
      *
@@ -165,7 +164,7 @@ public final class ShardingInPredicateTokenGenerator implements CollectionSQLTok
         ListExpression listExpr = (ListExpression) inExpression.getRight();
         return listExpr.getItems().size() > 1;
     }
-
+    
     /**
      * Check if an IN expression is related to sharding (involves sharded table and column).
      *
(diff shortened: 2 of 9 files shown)


Fottas:
> Hi @Fottas, this PR looks very complicated. Could you please submit an issue first to describe your design?

Hi @strongduanmu:
Issue Description: IN Predicate Sharding Optimization

Problem: Currently ShardingSphere sends all values in IN predicates to every shard, causing unnecessary data transfer and computation. For example, WHERE id IN (1,2,3,4) is sent as-is to all shards, even though based on sharding keys we can determine that values 1,3 belong to shard A while values 2,4 belong to shard B.
Solution: Route-aware IN predicate splitting that distributes values only to their target shards based on sharding algorithms.

Technical Design
Implementation Principle: Parse each value in IN expressions, invoke existing sharding algorithms to calculate target shards, then rewrite IN clauses for each RouteUnit to contain only values belonging to that shard.

Core Components:
1. ShardingInPredicateValue - Encapsulates IN values with their target route information
2. ShardingInPredicateToken - SQL token that rewrites IN clauses per route, filtering values and parameters
3. ShardingInPredicateTokenGenerator - Analyzes IN expressions and calculates shard distribution using existing sharding 
algorithms
4. ParameterFilterable - Cross-module decoupling interface that enables parameter filtering capability between modules, since ShardingInPredicateToken is in the sharding module while RouteSQLRewriteEngine is in the infra module

Key Features:
- Supports both standard and complex sharding strategies
- Handles parameter markers and literals
- Optimizes single-value cases to equality comparisons
- Achieves module decoupling through interface abstraction, allowing core engine to work without direct dependencies on sharding token implementations

Fottas:
Hi @strongduanmu, I've also created issue #36454 to track this design discussion as requested.

terrymanu (changes requested):
Thank you very much for the contribution and for providing the detailed design explanation in the issue.
The optimization goal (sharding-aware IN predicate rewrite) is valuable, and the high-level design direction is reasonable.

However, after reviewing the current PR, the overall change set is still too large and crosses several core modules:

1. The modification scope is very wide

This PR touches multiple layers at the same time, including:

Sharding algorithm SPI

Sharding rewrite token system

RouteSQLRewriteEngine in infra rewrite core

New interface (ParameterFilterable)

SQL rewrite pipeline integration

Many new types + large generator logic

IT tests + binder interactions

Any issue in these areas can introduce subtle problems such as wrong routing, parameter index mismatches, inconsistent rewrite results, or silent data deviation.

2. The PR introduces structural changes that need architectural discussion

The approach of filtering parameters per-route using a new interface and extending the rewrite engine is conceptually correct, but it changes the core rewrite flow.
Such changes require agreement on:

How parameter filtering should be modeled

Whether this should be a generic rewrite capability or sharding-specific

How to guarantee compatibility with existing token generators

How to avoid breaking custom sharding algorithms

How to handle complex SQL (subqueries, nested IN, multi-sharding-keys)

These points still need further design refinement before implementation can safely move forward.

3. The change is too large to review and validate safely in one PR

This PR includes 1300+ lines across 10+ files.
Given the sensitivity of the modules involved, it is not suitable to merge such a large change at once.
We risk introducing regressions that are difficult to detect through unit tests alone.

4. Suggested next steps

To proceed safely, we recommend:

Continue design discussion in the issue (e.g., #36454)
Clarify module boundaries, SPI behavior, rewrite engine responsibilities, and parameter-filter design.

Split the implementation into smaller, incremental PRs, for example:

Introduce the new rewrite extension point in isolation.

Add minimal sharding IN optimization for simple cases.

Gradually extend to complex sharding, placeholders, subqueries, etc.

Add more targeted tests per step so review becomes manageable and regressions easier to detect.

Conclusion

The problem and the proposed direction are reasonable, but the current PR is too large and involves core architecture changes. We need further design iteration and smaller, incremental PRs before it can be safely accepted.

Thank you again for the effort—happy to continue the discussion and help refine the approach.

Fottas:
### 1. Parameter filtering modeling

Current tokens can only rewrite SQL via `toString(RouteUnit)`, but cannot tell the rewrite engine which parameters to remove. For IN optimization, I need both capabilities: rewrite SQL per route AND filter parameters per route.

That's why I added `ParameterFilterable` interface in `infra.rewrite.core` - it's an extension point that lets tokens communicate their parameter filtering needs to the engine. Currently designed for IN optimization, but the mechanism is generic enough for other scenarios if needed.

### 2. Compatibility with existing token generators

The interface is optional. Existing tokens don't implement it, so they're completely unaffected. When no filterable tokens exist, the engine returns original parameters unchanged.

### 3. Custom algorithm compatibility

I don't modify any SPI interfaces. For each IN value, I call the algorithm's existing `doSharding()` method (same as routing does) to determine target tables, then map them to route units. If the algorithm returns null or empty, the value is marked as orphan and excluded from all routes.

Since I'm using the same SPI that routing uses, any algorithm that works with routing will work with this optimization.

### 4. Complex SQL support

The optimization works at the parameter level, not SQL structure level. It applies to any query using `StandardParameterBuilder` (all non-batch queries including subqueries and JOINs), regardless of SQL complexity.

In the token generator, I currently support `StandardShardingAlgorithm` and `ComplexKeysShardingAlgorithm`. For each IN value, I invoke the algorithm to determine target tables. For other strategy types, the value is included in all routes (original behavior), ensuring compatibility without requiring changes.

---

## PR Split Plan

### PR 1: Add ParameterFilterable interface

Introduce the parameter filtering extension point in `infra.rewrite.core`. This interface allows tokens to declare which parameters should be removed for each route, enabling route-aware parameter filtering in the rewrite engine.

---

### PR 2: Integrate parameter filtering into RouteSQLRewriteEngine

Implement the parameter filtering mechanism in the rewrite engine. When processing each route, the engine checks for filterable tokens and applies parameter filtering accordingly. When no filterable tokens exist, it maintains original behavior with zero overhead.

---

### PR 3: Add data structures (ShardingInPredicateValue and Token)

Introduce the data model for IN predicate optimization. `ShardingInPredicateValue` represents a single IN value with its target route information. `ShardingInPredicateToken` generates optimized SQL per route and implements parameter filtering, with single-value optimization (IN → =) for better database performance.

---

### PR 4: Simple optimization - literal values only

First end-to-end working version. The token generator detects IN predicates on sharding keys, calls sharding algorithms to determine target routes for each value, and generates tokens with route distribution information. This PR handles only literal values to keep the change manageable.

---

### PR 5: Add parameter marker support

Extend the generator to support parameter markers (`?`) in IN predicates. Implement `ParametersAware` interface to access bound parameters, enabling optimization for prepared statements.

---

### PR 6: Complex algorithm support

Extend support to `ComplexKeysShardingAlgorithm` for multi-column sharding keys. The generator builds appropriate sharding values and invokes the algorithm to determine target routes.

---

### PR 7: Nested expression support

Support IN predicates nested in AND/OR expressions. The generator recursively extracts IN predicates from binary operations, handling arbitrary nesting depth and multiple IN predicates in the same query.

---

@terrymanu  Thanks again for the detailed feedback. I've tried to address the architectural concerns you raised and broken down the work into smaller PRs. 

I'm not entirely sure if this is what you're looking for - if I've misunderstood anything or if you'd like me to approach it differently, please let me know. Happy to adjust the plan based on your guidance.

terrymanu:
Thanks for the contribution, please merge them into a single PR to help us review the codes. 
Thank you again!

terrymanu:
> ### 1. Parameter filtering modeling
> Current tokens can only rewrite SQL via `toString(RouteUnit)`, but cannot tell the rewrite engine which parameters to remove. For IN optimization, I need both capabilities: rewrite SQL per route AND filter parameters per route.
> 
> That's why I added `ParameterFilterable` interface in `infra.rewrite.core` - it's an extension point that lets tokens communicate their parameter filtering needs to the engine. Currently designed for IN optimization, but the mechanism is generic enough for other scenarios if needed.
> 
> ### 2. Compatibility with existing token generators
> The interface is optional. Existing tokens don't implement it, so they're completely unaffected. When no filterable tokens exist, the engine returns original parameters unchanged.
> 
> ### 3. Custom algorithm compatibility
> I don't modify any SPI interfaces. For each IN value, I call the algorithm's existing `doSharding()` method (same as routing does) to determine target tables, then map them to route units. If the algorithm returns null or empty, the value is marked as orphan and excluded from all routes.
> 
> Since I'm using the same SPI that routing uses, any algorithm that works with routing will work with this optimization.
> 
> ### 4. Complex SQL support
> The optimization works at the parameter level, not SQL structure level. It applies to any query using `StandardParameterBuilder` (all non-batch queries including subqueries and JOINs), regardless of SQL complexity.
> 
> In the token generator, I currently support `StandardShardingAlgorithm` and `ComplexKeysShardingAlgorithm`. For each IN value, I invoke the algorithm to determine target tables. For other strategy types, the value is included in all routes (original behavior), ensuring compatibility without requiring changes.
> 
> ## PR Split Plan
> ### PR 1: Add ParameterFilterable interface
> Introduce the parameter filtering extension point in `infra.rewrite.core`. This interface allows tokens to declare which parameters should be removed for each route, enabling route-aware parameter filtering in the rewrite engine.
> 
> ### PR 2: Integrate parameter filtering into RouteSQLRewriteEngine
> Implement the parameter filtering mechanism in the rewrite engine. When processing each route, the engine checks for filterable tokens and applies parameter filtering accordingly. When no filterable tokens exist, it maintains original behavior with zero overhead.
> 
> ### PR 3: Add data structures (ShardingInPredicateValue and Token)
> Introduce the data model for IN predicate optimization. `ShardingInPredicateValue` represents a single IN value with its target route information. `ShardingInPredicateToken` generates optimized SQL per route and implements parameter filtering, with single-value optimization (IN → =) for better database performance.
> 
> ### PR 4: Simple optimization - literal values only
> First end-to-end working version. The token generator detects IN predicates on sharding keys, calls sharding algorithms to determine target routes for each value, and generates tokens with route distribution information. This PR handles only literal values to keep the change manageable.
> 
> ### PR 5: Add parameter marker support
> Extend the generator to support parameter markers (`?`) in IN predicates. Implement `ParametersAware` interface to access bound parameters, enabling optimization for prepared statements.
> 
> ### PR 6: Complex algorithm support
> Extend support to `ComplexKeysShardingAlgorithm` for multi-column sharding keys. The generator builds appropriate sharding values and invokes the algorithm to determine target routes.
> 
> ### PR 7: Nested expression support
> Support IN predicates nested in AND/OR expressions. The generator recursively extracts IN predicates from binary operations, handling arbitrary nesting depth and multiple IN predicates in the same query.
> 
> @terrymanu Thanks again for the detailed feedback. I've tried to address the architectural concerns you raised and broken down the work into smaller PRs.
> 
> I'm not entirely sure if this is what you're looking for - if I've misunderstood anything or if you'd like me to approach it differently, please let me know. Happy to adjust the plan based on your guidance.


Some problems need to be re-design:

  - Responsibility mixing: pushing parameter filtering into SQL tokens breaks the current separation where ParameterBuilder/ParameterRewriter handle parameters and tokens handle SQL text; this risks index/order conflicts with other parameter
    rewriters (encryption, pagination, shadow, etc.).
  - Routing consistency risk: re-running doSharding per value inside tokens may diverge from the existing RouteContext (which already reflects hints, binding tables, merged conditions). Routing-stage condition merges/AlwaysFalse handling aren’t
    reused, so misrouting or missed routes are possible.
  - SQL/parameter alignment: RouteSQLRewriteEngine currently reuses the same parameter list in multi-route aggregation (UNION ALL); if each RouteUnit filters parameters differently, the existing concatenation can leave placeholder counts and
    parameter counts misaligned.
  - Incomplete coverage: the plan only mentions StandardParameterBuilder; GroupedParameterBuilder / batch / multi-value scenarios are unspecified, so behavior may be inconsistent or fail outright there.
  - Behavior change: when an algorithm returns empty/null or no route, values are silently dropped from IN, effectively weakening user predicates without fallback or warning—risking data loss.
  - Complex SQL / multi-table joins: the optimization assumes single-column distribution; it doesn’t explain how to keep routes consistent across binding tables, joins, or subqueries sharing parameters, so partial filtering and incomplete
    results are possible.
  - New interface surface risk: introducing a cross-module ParameterFilterable for a single scenario invites more tokens to take on parameter logic, increasing coupling and eroding the current parameter-vs-syntax separation.

terrymanu:

PR #36502, opened 2025-09-08
Change system table check logic

### LINKED ISSUE #36462: filters out user tables with "$" in names as system tables
## Bug Report

5.5.2

ShardingSphere-Proxy

----------------------------

Hi everyone.

While working with Proxy, I ran into an issue: tables with a `$` in their name do not work correctly — they are not loaded into the schema during metadata loading, and any queries to them fail with an error (`Table or view 'R$TABLE' does not exist`). I reproduced this problem on both `Firebird` and `PostgreSQL`.

Looking at the code, such tables are filtered out during metadata loading in [SchemaMetaDataLoader](https://github.com/apache/shardingsphere/blob/789ee8ca36981737d3a9d1e98e08b2ec1806467e/database/connector/core/src/main/java/org/apache/shardingsphere/database/connector/core/metadata/data/loader/type/SchemaMetaDataLoader.java#L41) by the following rule:

```Java
private boolean isSystemTable(final String table) {
    return table.contains("/") || table.contains("##") || table.contains("$");
}
```
This filter is used in the table loading method:

```Java
private Collection<String> loadTableNames(final Connection connection, final String schemaName, final Collection<String> excludedTables) throws SQLException {
    Collection<String> result = new CaseInsensitiveSet<>();
    String[] tableTypes = new String[]{TABLE_TYPE, PARTITIONED_TABLE_TYPE, VIEW_TYPE, SYSTEM_TABLE_TYPE, SYSTEM_VIEW_TYPE};
    try (ResultSet resultSet = connection.getMetaData().getTables(connection.getCatalog(), schemaName, null, tableTypes)) {
        while (resultSet.next()) {
            String table = resultSet.getString(TABLE_NAME);
            if (!isSystemTable(table) && !excludedTables.contains(table)) {
                result.add(table);
            }
        }
    }
    return result;
}
```
This means that any table whose name contains `$` is automatically treated as “system” and excluded from the user schema. The problem is that in both `PostgreSQL` and `Firebird`, `$` is perfectly valid in user identifiers and does not imply a system object. As a result, valid user tables with `$` are filtered out at Proxy startup and become unusable.

At the same time, system tables for each database are already explicitly defined in `system/schema (YAML)`. This mechanism allows Proxy to recognize which tables are truly system ones and handle them separately. Given that, it’s not clear why we still need a global heuristic like `contains("$")`. Wouldn’t it be more reliable to check against the system tables list for the current `databaseType` (e.g., for `Firebird`: `RDB$*`) rather than applying generic character-based rules?

I understand that this rule was originally introduced to prevent loading system tables that are not yet fully implemented, but it would be better if it were refined so that it does not also break ordinary user tables.

Right now the global `$` rule creates a contradiction: system tables are already covered by `system/schema`, but ordinary user tables containing `$` are also excluded and lost. Should this behavior be reconsidered so that user tables with `$` remain available after Proxy restarts? Otherwise, it becomes impossible to work with non-system tables whose names happen to include $.

P.S. I also noticed the [DriverQuerySystemCatalogOption](https://github.com/apache/shardingsphere/blob/789ee8ca36981737d3a9d1e98e08b2ec1806467e/database/connector/core/src/main/java/org/apache/shardingsphere/database/connector/core/metadata/database/metadata/option/table/DialectDriverQuerySystemCatalogOption.java#L23) interface and its `isSystemTable(...)` method. At first I thought I could override it to define system tables in a dialect-specific way (as done for openGauss). But it turns out this interface is not used during metadata loading in `SchemaMetaDataLoader`; instead, it is used later during query execution (SQL federation) to recognize system catalog queries and choose a special execution path (e.g., in-memory results / avoiding backend scans). That makes sense for offloading Proxy and handling system queries, but at startup the `contains("$")` rule still applies and filters out such tables.

makssent:
Fixes #36462

- Moved system-table detection to SystemTableManager and relocated it under shardingsphere/database/core.  
- Introduced a dialect-driven SPI: DialectSystemTableManager to provide per-dialect system table lists and checks.  
- Updated SchemaMetaDataLoader to delegate to the dialect implementation’s isSystemTable, falling back to the legacy heuristic when no dialect is available.  
- Added unit tests.  

---

- [X] My code follows the [code of conduct](https://shardingsphere.apache.org/community/en/involved/conduct/code/) of this project.
- [X] I have self-reviewed the commit code.
- [X] I have (or in comment I request) added corresponding labels for the pull request.


makssent:
There are two points I’m not fully satisfied with:

1. In `SystemTableManagerTest` I had to implement a custom schema loading mechanism, because the schemas are not available during tests before the program starts. At runtime everything works correctly, the issue only exists in the test environment. I couldn’t find another solution: adding `dialect-connector-*` dependencies would be a more natural option, but it breaks the build.

2. The TODOs I left. I think they should probably be removed, but I wanted to highlight one point: in `SimpleTableSegmentBinder` we currently use `SystemTableManager.isSystemTable()` without `databaseType`, even though `databaseType` is available and this would not change the program logic. The code could be rewritten like this, and the implementation of `isSystemTable` without `databaseType` could be removed from `SystemTableManager`:

    ```java
    if (SystemTableManager.isSystemTable(binderContext.getSqlStatement().getDatabaseType().getType(),
            schemaName, tableName)) {
        return;
    }
    ```

    This would make the method that only accepts `schema` unnecessary, since it’s not used anywhere else. I didn’t remove it in this PR — maybe you have future plans for it — but I wanted to draw attention to this point.

I verified it with tests. I also checked Firebird manually — everything works as intended.

Looking forward to your review and suggestions for improvement.

database/connector/core/src/main/java/org/apache/shardingsphere/database/connector/core/metadata/manager/DialectSystemTableManager.java
```diff
@@ -0,0 +1,38 @@
...
+package org.apache.shardingsphere.database.connector.core.metadata.manager;
+
+import org.apache.shardingsphere.database.connector.core.spi.DatabaseTypedSPI;
+import org.apache.shardingsphere.infra.spi.annotation.SingletonSPI;
+
+/**
+ * Dialect system table manager.
+ */
+@SingletonSPI
+public interface DialectSystemTableManager extends DatabaseTypedSPI {
```
strongduanmu: Do you think `DialectSystemTableDetector` is better? The meaning of Manager is broader, and we only want to add a class to detect whether a table is a system table.

database/connector/core/src/main/java/org/apache/shardingsphere/database/connector/core/metadata/data/loader/type/SchemaMetaDataLoader.java
```diff
@@ -124,6 +126,7 @@ private Collection<String> loadTableNames(final Connection connection, final Str
     }
     
     private boolean isSystemTable(final String table) {
-        return table.contains("$") || table.contains("/") || table.contains("##");
+        return DatabaseTypedSPILoader.findService(DialectSystemTableManager.class, databaseType)
```
strongduanmu: Maybe you can add a engine class to hide the details of SPI loading so that other scenarios can directly call the Engine. You can refer to the logic of SQLRouteEngine.

database/connector/core/src/main/java/org/apache/shardingsphere/database/connector/core/metadata/manager/SystemTableManager.java
```diff
@@ -0,0 +1,122 @@
...
+import java.util.Map;
+import java.util.Optional;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/**
+ * System table manager.
+ */
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public final class SystemTableManager {
```
strongduanmu: SystemTableDetector maybe better.

database/connector/core/src/main/java/org/apache/shardingsphere/database/connector/core/metadata/manager/SystemTableManager.java
```diff
@@ -0,0 +1,122 @@
...
+    private static final Map<String, Map<String, Collection<String>>> DATABASE_TYPE_SCHEMA_TABLE_MAP;
+    
+    private static final String COMMON = "common";
+    
+    static {
+        List<String> resourceNames;
+        try (Stream<String> resourceNameStream = ClasspathResourceDirectoryReader.read("schema")) {
+            resourceNames = resourceNameStream.filter(each -> each.endsWith(".yaml")).collect(Collectors.toList());
+        }
+        DATABASE_TYPE_SCHEMA_TABLE_MAP = resourceNames.stream().map(resourceName -> resourceName.split("/")).filter(each -> each.length == 4)
```
strongduanmu: Please put 4 costant to the left hand of `each.length == 4`

database/connector/core/src/main/java/org/apache/shardingsphere/database/connector/core/metadata/manager/SystemTableManager.java
```diff
@@ -0,0 +1,122 @@
...
+    /**
+     * Judge whether current table is system table or not.
+     *
+     * @param schema schema
+     * @param tableName table name
+     * @return whether current table is system table or not
+     */
+    // TODO check if this function is needed, since it's only used in SimpleTableSegmentBinder and could be replaced by SystemTableManager.isSystemTable(databaseType). (issues/36462)
+    public static boolean isSystemTable(final String schema, final String tableName) {
+        for (Map.Entry<String, Map<String, Collection<String>>> entry : DATABASE_TYPE_SCHEMA_TABLE_MAP.entrySet()) {
```
strongduanmu: Please use `Entry` to replace `Map.Entry`

database/connector/core/src/main/java/org/apache/shardingsphere/database/connector/core/metadata/manager/SystemTableManager.java
```diff
@@ -0,0 +1,122 @@
...
+    
+    /**
+     * Judge whether current table is system table or not.
+     *
+     * @param databaseType database type
+     * @param schema       schema
+     * @param tableName    table name
+     * @return whether current table is system table or not
+     */
+    public static boolean isSystemTable(final String databaseType, final String schema, final String tableName) {
```
strongduanmu, line 74: These logics seem very common. As long as the system table is configured under resource, they can work normally?

strongduanmu: If this is the case, perhaps we can refactor it without SPI, as long as we ensure that the database without configuring the system table YAML file can go to the default logic.

database/connector/core/src/test/java/org/apache/shardingsphere/database/connector/core/metadata/manager/SystemTableManagerTest.java
```diff
@@ -0,0 +1,120 @@
...
+    
+    private static ClassLoader originalClassLoader;
+    
+    private static URLClassLoader schemaClassLoader;
+    
+    @BeforeAll
+    static void setUp() throws Exception {
+        originalClassLoader = Thread.currentThread().getContextClassLoader();
+        URL[] urls = {
+                Paths.get("../dialect/mysql/src/main/resources").toUri().toURL(),
```
strongduanmu: You can write some FIXTUREs and test configurations specifically for testing to test the logic of SystemTableManager without actually loading the dialect module.

database/connector/dialect/firebird/src/main/java/org/apache/shardingsphere/database/connector/firebird/metadata/manager/FirebirdSystemTableManager.java
```diff
@@ -0,0 +1,31 @@
...
+
+import org.apache.shardingsphere.database.connector.core.metadata.manager.DialectSystemTableManager;
+
+/**
+ * Firebird system table manager.
+ */
+public final class FirebirdSystemTableManager implements DialectSystemTableManager {
+    
+    @Override
+    public String getDatabaseType() {
```
strongduanmu: It seems that the implementation classes of these SPIs are empty. Based on this, I suggest removing the SPI interface and implementation class and encapsulating these logics through a SystemTableDetector.

strongduanmu (changes requested):
Finally, I would like to remind you that to determine whether a table is a true system table, you also need to look at the path to which the table belongs (db - schema - table). If the path belongs to a user-defined schema, perhaps this table should be identified as a non-system table.

makssent:
Hi, sorry for the late reply — I’m working on another task in parallel. I’ve reviewed your comments and will apply the fixes. Thanks!

makssent:
I hope I’ve finished and done everything correctly. I think my problem was also that I just copied the old implementation from `SystemSchemaManager` and didn’t even try to change it, trying not to deviate too much from the base code, not assuming that it was already outdated and not fully aligned with current standards.

I created a new SPI — `SystemTableRule` (now optional, so each database can define its own rules). I removed the unnecessary override of `isSystemTable()` because everywhere it is used we already know the `databaseType`. I also deleted the unused `getTables()` function, renamed `*Manager` to `*Detector`, and did some refactoring. I think it will be better to implement specific logic later, when it is actually needed, instead of keeping it around without use.

I’m looking forward to your next comments, and I hope I didn’t miss anything. It was interesting for me to dig into the core.

makssent:
I see that the build is failing, but I noticed the same error appears in other PRs as well. I’ll wait until the issue is resolved on your side and then rerun the CI.

makssent:
Hi @strongduanmu, please take a look at the PR when you have time

terrymanu:
Due to the large number of conflicts, I've closed it for now. If you wish to continue modifying, you can start fresh from the latest branch.

PR #36942, opened 2025-10-26
Try to use mvnd for SQL E2E GitHub action

terrymanu:


.github/workflows/nightly-e2e-sql.yml
```diff
@@ -53,8 +53,11 @@ jobs:
           restore-keys: |
             ${{ needs.global-environment.outputs.GLOBAL_CACHE_PREFIX }}-maven-third-party-e2e-cache-
             ${{ needs.global-environment.outputs.GLOBAL_CACHE_PREFIX }}-maven-third-party-
+      - uses: gauthamchandra/mvnd-action@v0.2.1
+        with:
+          version: '1.0.3'
       - name: Build E2E Image
-        run: ./mvnw -B clean install -am -pl test/e2e/sql -Pe2e.env.docker -DskipTests -Dspotless.apply.skip=true
+        run: mvnd -B clean install -am -pl test/e2e/sql -Pe2e.env.docker -DskipTests -Dspotless.apply.skip=true
```
linghengqian, line 60: I don't understand why we need to use `mvnd` this way, which essentially throws away the existing `mvnw` script. Can't we just execute the following command directly in `PowerShell 7`/`Bash` to have `mvnw` update its configuration file to download the `mvnd` binary,

```shell
mvn wrapper:wrapper "-Dtype=only-script" "-Dmvnd=1.0.3"
```
This is already documented at https://github.com/apache/maven-wrapper/blob/master/maven-wrapper-plugin/src/it/projects/test_mvnd/test.properties .

PR #37181, opened 2025-11-26
Fix Oracle TIMESTAMP WITH TIME ZONE ORDER BY exception

### LINKED ISSUE #33678: The following error message appears when executing an SQL statement with an ORDER BY clause in Oracle: "Order by value must implement Comparable."
shardingsphere Ver. shardingsphere, 5.4.1
oracle Ver. oracle12c
ojdbc Ver. ojdbc8

ShardingSphere-JDBC
### Expected behavior
config：
actualDataNodes: ds0.dmg_kettle_conf, ds0.dmg_kettle_config_dmgtmp
logicTable: dmg_kettle_config
behavior：
when executing SQL "select * from dmg_kettle_config order by create_timstmp desc", except the ordered result, the dataTye of **create_timstmp is TimeStamp With Time Zone**, (doesn't support this dataType?)
### Actual behavior
when executing SQL "select * from dmg_kettle_config order by create_timstmp desc", the error message  "Order by value must implement Comparable." occurs

When executing query, in `oracle.jdbc.driver.OracleResultSetMetaData.getColumnType`, the create_timstmp dataType is 181, and mappings to code -101
image
and in shardingJdbc, when merging resut, it only transfer the code 93 to java.sql.Timestamp, but the code -101 convert to Object:
`org.apache.shardingsphere.sharding.execute.sql.execute.result.MemoryQueryResult.getRows(ResultSet resultSet)`
image
so, when code running the order executor: `org.apache.shardingsphere.sharding.merge.dql.orderby.OrderByValue.getOrderValues()`, cause this error.
image
btw: Setting the following parameter: System.getProperties().setProperty("oracle.jdbc.J2EE13Compliant", "true") is ineffective.

[https://shardingsphere.apache.org/document/4.1.1/en/faq/#10-why-does-oracle-database-throw-order-by-value-must-implements-comparable-exception-when-using-timestamp-order-by](https://shardingsphere.apache.org/document/4.1.1/en/faq/#10-why-does-oracle-database-throw-order-by-value-must-implements-comparable-exception-when-using-timestamp-order-by)
  
SQL "select * from dmg_kettle_config order by create_timstmp desc", **create_timstmp is TimeStamp With Time Zone**, (doesn't support this dataType?)

zhan7236:
### Changes

This PR fixes the "Order by value must implement Comparable" error when executing `ORDER BY` operations on Oracle `TIMESTAMP WITH TIME ZONE` type columns.

### Root Cause

The Oracle JDBC driver uses non-standard type codes:
- `-101` for `TIMESTAMP WITH TIME ZONE`
- `-102` for `TIMESTAMP WITH LOCAL TIME ZONE`

The `ResultSetMapper.load()` method's `default` branch was calling `resultSet.getObject()`, which returns Oracle-specific objects that don't implement `Comparable`.

### Solution

Based on maintainer feedback, this PR takes a targeted approach:

1. **Added `getDefaultValue()` default method to `DialectResultSetMapper` interface**:
   - Default implementation returns `resultSet.getObject()` to preserve backward compatibility.
   - Allows dialect-specific handling of unrecognized column types.

2. **Modified `ResultSetMapper.load()` default branch**:
   - Delegates to `getDefaultValue()` when a dialect mapper is present.
   - Falls back to `getObject()` otherwise.

3. **Created `OracleResultSetMapper` to handle Oracle-specific type codes**:
   - Only converts Oracle's `-101` and `-102` type codes to `java.sql.Timestamp`.
   - All other unhandled types continue to use `getObject()` (no side effects for JSON, UUID, BIT, etc.).

### Key Design Decisions

| Concern | Solution |
| :--- | :--- |
| Avoid side effects for other dialects | Default `getDefaultValue()` returns `getObject()` |
| Oracle-specific handling | Only `-101` and `-102` are specially handled in `OracleResultSetMapper` |
| Backward compatibility | Non-Oracle dialects behavior unchanged |

### File Changes

| File | Type | Description |
| :--- | :--- | :--- |
| `DialectResultSetMapper.java` | Modified | Added `getDefaultValue()` default method returning `getObject()` |
| `ResultSetMapper.java` | Modified | Delegate default branch to dialect mapper's `getDefaultValue()` |
| `OracleResultSetMapper.java` | New | Handle Oracle timestamp type codes (-101, -102) |
| `DialectResultSetMapper` SPI | New | Register Oracle mapper |
| `OracleResultSetMapperTest.java` | New | Unit tests (5 test cases) |

### Testing

- All existing tests pass.
- Added 5 new test cases for `OracleResultSetMapper`:
  - `assertGetSmallintValue`
  - `assertGetDateValue`
  - `assertGetDefaultValueWithOracleTimestampWithTimeZone`
  - `assertGetDefaultValueWithOracleTimestampWithLocalTimeZone`
  - `assertGetDefaultValueWithUnhandledType`

Fixes #33678

### DIFF
--- a/database/connector/core/src/main/java/org/apache/shardingsphere/database/connector/core/resultset/DialectResultSetMapper.java
+++ b/database/connector/core/src/main/java/org/apache/shardingsphere/database/connector/core/resultset/DialectResultSetMapper.java
@@ -50,15 +50,15 @@ public interface DialectResultSetMapper extends DatabaseTypedSPI {
     Object getDateValue(ResultSet resultSet, int columnIndex) throws SQLException;
     
     /**
-     * Get timestamp value from result set.
+     * Get default value from result set for unhandled column types.
      *
      * @param resultSet result set
      * @param columnIndex column index
      * @param columnType column type from metadata
-     * @return timestamp value
+     * @return value for unhandled column type
      * @throws SQLException sql exception
      */
-    default Object getTimestampValue(ResultSet resultSet, int columnIndex, int columnType) throws SQLException {
-        return resultSet.getTimestamp(columnIndex);
+    default Object getDefaultValue(ResultSet resultSet, int columnIndex, int columnType) throws SQLException {
+        return resultSet.getObject(columnIndex);
     }
 }
--- a/database/connector/core/src/main/java/org/apache/shardingsphere/database/connector/core/resultset/ResultSetMapper.java
+++ b/database/connector/core/src/main/java/org/apache/shardingsphere/database/connector/core/resultset/ResultSetMapper.java
@@ -92,7 +92,7 @@ public final class ResultSetMapper {
                 return resultSet.getArray(columnIndex);
             default:
                 return dialectResultSetMapper.isPresent()
-                        ? dialectResultSetMapper.get().getTimestampValue(resultSet, columnIndex, metaData.getColumnType(columnIndex))
+                        ? dialectResultSetMapper.get().getDefaultValue(resultSet, columnIndex, metaData.getColumnType(columnIndex))
                         : resultSet.getObject(columnIndex);
         }
     }
--- a/database/connector/dialect/oracle/src/main/java/org/apache/shardingsphere/database/connector/oracle/resultset/OracleResultSetMapper.java
+++ b/database/connector/dialect/oracle/src/main/java/org/apache/shardingsphere/database/connector/oracle/resultset/OracleResultSetMapper.java
@@ -50,7 +50,7 @@ public final class OracleResultSetMapper implements DialectResultSetMapper {
     }
     
     @Override
-    public Object getTimestampValue(final ResultSet resultSet, final int columnIndex, final int columnType) throws SQLException {
+    public Object getDefaultValue(final ResultSet resultSet, final int columnIndex, final int columnType) throws SQLException {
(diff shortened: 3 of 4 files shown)


terrymanu (changes requested):
The direction of this PR for handling Oracle TIMESTAMPTZ is correct — thank you for the contribution!
However, the current change to the ResultSetMapper default branch is a bit too broad and may introduce unintended side effects for other dialects (MySQL / PostgreSQL / OpenGauss, etc.).

Specifically:
When a dialect-level mapper exists, all unhandled JDBC types now go through getTimestampValue(), which by default calls getTimestamp().
This means many non-timestamp types (JSON, UUID, BIT, SQLXML, ROWID, etc.) will switch from the previous getObject() behavior to getTimestamp(), potentially causing exceptions or incorrect type conversions.

Recommendation:

Restrict the new timestamp handling logic only to timestamp / timestamp-with-timezone types.

Keep getObject() as the default for all other types.

Handle Oracle’s vendor-specific type codes (-101 / -102) solely inside the Oracle dialect mapper.

This preserves the Oracle fix while avoiding compatibility issues for other databases.
Happy to help with further review if needed.

zhan7236:
@terrymanu Fixed, thanks!

terrymanu:
Great!

Issue #36454, opened 2025-09-02
[Feature Request] IN Predicate Sharding Optimization


Fottas:
Problem: Currently ShardingSphere sends all values in IN predicates to every shard, causing unnecessary data transfer and computation. For example, WHERE id IN (1,2,3,4) is sent as-is to all shards, even though based on sharding keys we can determine that values 1,3 belong to shard A while values 2,4 belong to shard B.
Solution: Route-aware IN predicate splitting that distributes values only to their target shards based on sharding algorithms.

Technical Design
Implementation Principle: Parse each value in IN expressions, invoke existing sharding algorithms to calculate target shards, then rewrite IN clauses for each RouteUnit to contain only values belonging to that shard.

Core Components:
1. ShardingInPredicateValue - Encapsulates IN values with their target route information
2. ShardingInPredicateToken - SQL token that rewrites IN clauses per route, filtering values and parameters
3. ShardingInPredicateTokenGenerator - Analyzes IN expressions and calculates shard distribution using existing sharding algorithms
4. ParameterFilterable - Cross-module decoupling interface that enables parameter filtering capability between modules, since ShardingInPredicateToken is in the sharding module while RouteSQLRewriteEngine is in the infra module

Key Features:
- Supports both standard and complex sharding strategies
- Handles parameter markers and literals
- Optimizes single-value cases to equality comparisons
- Achieves module decoupling through interface abstraction, allowing core engine to work without direct  dependencies on sharding token implementations

Fottas:
Hi @RaigorJiang @strongduanmu,

Thanks for your attention and support! I'd like to check the review status:
1. Are there any issues with the current technical approach that need improvement?
2. Do I need to supplement more documentation or tests?

Looking forward to your feedback, thanks!

terrymanu:
Thank you very much for the contribution and for providing the detailed design explanation in the issue.
The optimization goal (sharding-aware IN predicate rewrite) is valuable, and the high-level design direction is reasonable.

However, after reviewing the current PR, the overall change set is still too large and crosses several core modules:

1. The modification scope is very wide

This PR touches multiple layers at the same time, including:

Sharding algorithm SPI

Sharding rewrite token system

RouteSQLRewriteEngine in infra rewrite core

New interface (ParameterFilterable)

SQL rewrite pipeline integration

Many new types + large generator logic

IT tests + binder interactions

[...]

PR #37255, opened 2025-12-02
[Part 1] Add ParameterFilterable interface

Fottas:
Overview
This is the first PR for the IN predicate optimization feature, introducing the foundational interface for parameter filtering.
Changes

Add ParameterFilterable interface to define parameter filtering capability
Provide getRemovedParameterIndices(RouteUnit) method to declare parameter indices that should be removed
Provide isParameterFilterable() default method to allow conditional filtering enablement

Design Rationale

Interface is placed in infra.rewrite.core module as a generic extension point for the rewrite engine
Currently designed specifically for sharding IN optimization; may potentially support other parameter filtering scenarios in the future if such needs arise
Fully backward compatible - does not affect existing token generators

Related Issue
Part of #36454 (1)

Next Steps
The next PR will integrate parameter filtering logic into RouteSQLRewriteEngine.


terrymanu:

PR #37376, opened 2025-12-13
Support for the explain and do keywords in Stored Procedures

ccxxxyy:
Fixes https://github.com/apache/shardingsphere/issues/35088

  -

---

### DIFF
--- a/parser/sql/engine/dialect/mysql/src/main/java/org/apache/shardingsphere/sql/parser/engine/mysql/visitor/statement/type/MySQLDDLStatementVisitor.java
+++ b/parser/sql/engine/dialect/mysql/src/main/java/org/apache/shardingsphere/sql/parser/engine/mysql/visitor/statement/type/MySQLDDLStatementVisitor.java
@@ -199,6 +199,7 @@ import org.apache.shardingsphere.sql.parser.statement.mysql.ddl.instance.MySQLAl
 import org.apache.shardingsphere.sql.parser.statement.mysql.ddl.logfile.MySQLAlterLogfileGroupStatement;
 import org.apache.shardingsphere.sql.parser.statement.mysql.ddl.logfile.MySQLCreateLogfileGroupStatement;
 import org.apache.shardingsphere.sql.parser.statement.mysql.ddl.logfile.MySQLDropLogfileGroupStatement;
+import org.apache.shardingsphere.sql.parser.statement.mysql.dal.show.column.MySQLDescribeStatement;
 
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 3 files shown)


terrymanu (changes requested):
- parser/sql/engine/dialect/mysql/src/main/antlr4/imports/mysql/DDLStatement.g4:680-687: EXPLAIN is added as a bare keyword to validStatement instead of using the explain rule. Inside BEGIN ... END, EXPLAIN SELECT ... will be parsed as two statements (EXPLAIN + SELECT), allowing an EXPLAIN with no target and losing the explain semantics. Downstream rewriting/analysis will treat it as a plain query — a behavior regression.

- parser/sql/engine/dialect/mysql/src/main/java/org/apache/shardingsphere/sql/parser/engine/mysql/visitor/statement/type/MySQLDDLStatementVisitor.java:818-844: createValidStatementSegment still only emits table/DML/SELECT nodes. Newly allowed DO/EXPLAIN would return null and be dropped, so the routine body AST will not contain DoStatement/ExplainStatement; the feature isn’t actually wired through.

- Tests are insufficient: test/it/parser/src/main/resources/sql/supported/ddl/create-procedure.xml and .../case/ddl/create-procedure.xml only add SQL cases, with no assertions on routine-body contents, so the above semantic gaps would not be caught.

  Recommendation: wire validStatement to the explain rule and include doStatement, extend the visitor to return ExplainStatement/DoStatement, and add parser assertions ensuring routine bodies contain the correct ValidStatementSegment entries.

ccxxxyy:
@terrymanu Hi, I've already made the revisions as requested. Hope that you can review it when you have some free time.

terrymanu (changes requested):
Nice direction adding EXPLAIN/DO in stored procedures! Two small tweaks to keep behavior consistent and avoid a regression:

  - MySQLDDLStatementVisitor.visitExplain (parser/sql/engine/dialect/mysql/src/main/java/org/apache/shardingsphere/sql/parser/engine/mysql/visitor/statement/type/MySQLDDLStatementVisitor.java): when ctx.tableName()!=null it currently returns visitChildren(ctx), yielding
    a SimpleTableSegment while createValidStatementSegment expects an ExplainStatement. This will throw ClassCastException and drop semantics for EXPLAIN/DESC table inside procedures. Please mirror the DAL visitor and build a MySQLDescribeStatement (including column wildcard handling) for the table-name branch.
  - Coverage: add a stored-proc case with EXPLAIN/DESC table (with table name) in test/it/parser/src/main/resources/sql/supported/ddl/create-procedure.xml, and annotate the corresponding sql-statement position/type in test/it/parser/src/main/resources/case/ddl/create-procedure.xml (expect MySQLDescribeStatement), so this branch is asserted and won’t regress.

  Thanks for pushing this forward—looking forward to the update!

ccxxxyy:
@terrymanu Hi, I have made the revisions as requested. Hope that you could take a look at it when you have a moment.

Issue #31509, opened 2024-05-31
Support parsing Doris sql


iamhucong:
## Background

Hi community.\
The ShardingSphere SQL parser engine helps users to parse SQL to create the AST (Abstract Syntax Tree) and visit the AST to get SQLStatement (Java Object). Currently, we are planning to enhance the support for Doris SQL parsing in ShardingSphere.

More details:
<https://shardingsphere.apache.org/document/current/en/reference/sharding/parse/>  

## Issue Background Explanation

The current issue involves using a custom web scraping script to fetch SQL cases from the official website of a corresponding database. These SQL cases are then passed to ShardingSphere's parsing engine for analysis. For SQL cases that fail to be parsed successfully, every 3 to 5 SQL cases are grouped together as an issue.
1. Since  SQL cases are obtained through web scraping, it cannot be guaranteed that all SQL cases are correct. Please follow the following process to handle this pull request (PR).
2. Some SQL cases may have already been fixed in other PRs. For cases that can already be executed successfully, simply leave a comment to ignore them.
3. If a SQL case can be executed successfully without any code changes, there is no need to add a corresponding test assert file.  

## Task

This issue is to support more Doris sql parsing, as follows:

```
ALTER RESOURCE "showPolicy_1_resource" PROPERTIES("s3.connection.maximum" = "1111")
```

- [ ] [link](https://doris.apache.org/docs/dev/sql-manual/sql-statements/Data-Definition-Statements/Alter/ALTER-RESOURCE/)

```
ALTER, RESOURCE
```

- [ ] [link](https://doris.apache.org/docs/dev/sql-manual/sql-statements/Data-Definition-Statements/Alter/ALTER-RESOURCE/)

```
SELECT split_by_string('1, 2 3, 4,5, abcde', ', ')
```

- [ ] [link](https://doris.apache.org/docs/2.0/sql-manual/sql-functions/string-functions/split-by-string/)

```
SUBSTR
```

- [ ] [link](https://doris.apache.org/docs/2.0/sql-manual/sql-functions/string-functions/substr/)

```
BITOR
```

- [ ] [link](https://doris.apache.org/docs/1.2/sql-manual/sql-functions/bitwise-functions/bitor/)

```
ADMIN COPY TABLET tablet_id PROPERTIES("xxx")
```

- [ ] [link](https://doris.apache.org/zh-CN/docs/dev/sql-manual/sql-statements/Database-Administration-Statements/ADMIN-COPY-TABLET/)

```
[...]

PR #37392, opened 2025-12-15
Support for the ALTER RESOURCE statement

ccxxxyy:
for https://github.com/apache/shardingsphere/issues/31509

  -

---

### DIFF
--- a/parser/sql/engine/dialect/doris/src/main/java/org/apache/shardingsphere/sql/parser/engine/doris/visitor/statement/type/DorisDALStatementVisitor.java
+++ b/parser/sql/engine/dialect/doris/src/main/java/org/apache/shardingsphere/sql/parser/engine/doris/visitor/statement/type/DorisDALStatementVisitor.java
@@ -158,6 +158,12 @@ import org.apache.shardingsphere.sql.parser.statement.core.value.collection.Coll
 import org.apache.shardingsphere.sql.parser.statement.core.value.identifier.IdentifierValue;
 import org.apache.shardingsphere.sql.parser.statement.core.value.literal.impl.NumberLiteralValue;
 import org.apache.shardingsphere.sql.parser.statement.core.value.literal.impl.StringLiteralValue;
+import org.apache.shardingsphere.sql.parser.statement.core.value.literal.LiteralValue;
+import org.apache.shardingsphere.sql.parser.statement.core.value.literal.impl.BooleanLiteralValue;
+import org.apache.shardingsphere.sql.parser.statement.core.value.literal.impl.DateTimeLiteralValue;
+import org.apache.shardingsphere.sql.parser.statement.core.value.literal.impl.NullLiteralValue;
+import org.apache.shardingsphere.sql.parser.statement.core.value.literal.impl.OtherLiteralValue;
+import org.apache.shardingsphere.sql.parser.statement.core.value.literal.impl.TemporalLiteralValue;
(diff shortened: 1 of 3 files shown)


terrymanu (changes requested):
##  Root Cause

  - DorisAlterResourceStatement redeclares databaseType, which already exists in the parent SQLStatement, creating duplicate state and potential divergence.
  - Assertions only check the resource name and non-null properties, never validating property key/value parsing, so visitor errors can slip through.
  - Grammar and visitor accept only string resource names and property keys/values; Doris syntax allows identifier-style keys and values beyond strings (official reference: https://doris.apache.org/docs/2.1/sql-manual/sql-statements/cluster-management/compute-management/ALTER-RESOURCE/).

##  Analysis

  - Redundant field: the parent already holds databaseType; keeping another copy in the subclass risks inconsistency on future changes/serialization.
  - Assertion gap: DorisAlterResourceStatementAssert doesn’t verify property key/values, so misparsed or missing properties still pass.
  - Syntax/visitor coverage: official syntax shows ALTER RESOURCE '<resource_name>' PROPERTIES(\key` = value, ...), where the key is an identifier (backticked) and the value is not restricted to strings. The current rule using only string_` would reject valid statements with unquoted identifiers or non-string values.

##  Conclusion (Change Request)

  - Remove the duplicate databaseType field and assignment in DorisAlterResourceStatement; rely on the parent field. Keep constructor calling super(databaseType) and storing only resourceName and properties:

```java
  public final class DorisAlterResourceStatement extends DALStatement {
      private final String resourceName;
      private final Properties properties;

      public DorisAlterResourceStatement(final DatabaseType databaseType, final String resourceName, final Properties properties) {
          super(databaseType);
          this.resourceName = resourceName;
          this.properties = properties;
      }
  }
```

  - Strengthen assertions: add expected property key/values in test cases (e.g., list or map) and validate size plus specific key/values in DorisAlterResourceStatementAssert to ensure visitor parsing is correct.
  - Align grammar and visitor with the official doc: allow resource names/property keys as identifiers or strings, and property values as literals (string/number/boolean). Update the visitor to handle IdentifierValue, StringLiteralValue, NumberLiteralValue, and booleans accordingly, and add SQL cases plus assertions covering these forms. After these adjustments, merging will be safer.

ccxxxyy:
@terrymanu Hi, I have made the revisions as required. Hope you can review them when you have some time.

terrymanu (changes requested):
Here’s what needs to be fixed before we can merge—thanks for the solid start, please keep going:

  - parser/sql/engine/dialect/doris/src/main/java/org/apache/shardingsphere/sql/parser/engine/doris/visitor/statement/type/DorisDALStatementVisitor.java:909: getPropertyValue only handles string/number literals; boolean/NULL/hex/etc. fall through to result.toString(), so
    those property values are parsed as object ids, not the literal values. Please decode all LiteralValue types (BooleanLiteralValue, NullLiteralValue, hex/bit, etc.) via getValue(), mirroring other dialect handlers.
  - Tests: test/it/parser/src/main/resources/case/dal/alter.xml and .../sql/supported/dal/alter.xml only cover string/number properties. Add cases for boolean/NULL (and other literal forms) to exercise the fixed paths.

  Once these are addressed, the PR should be in good shape to re-review.

ccxxxyy:
@terrymanu  Hi, Hope you could take a look when you have a moment!

ccxxxyy:
> Nice work on the comprehensive coverage and tidy wiring—thanks for the thorough contribution!

It's my pleasure!

PR #37509, opened 2025-12-25
Support for the REGEXP function and the SELECT INTO OUTFILE statement and the CREATE FUNCTION statement

ccxxxyy:
for https://github.com/apache/shardingsphere/issues/31498

  -

---


terrymanu (changes requested):
The direction is good—adding Doris REGEXP and SELECT … INTO OUTFILE syntax plus tests is valuable, and the new outfile segment extraction/assertions are timely. Thanks for the work so far.

  Two important issues still need refinement:

  1. Keyword compatibility: In Doris Nereids grammar, ALIAS and PARAMETER are non-reserved, and SYMBOL isn’t a keyword; only INTERMEDIATE is reserved. This change makes ALIAS/PARAMETER/SYMBOL keywords without allowing them as identifiers, so unquoted tables/columns/ aliases with those names will fail to parse—a regression. Suggested fix: keep INTERMEDIATE as a keyword; treat ALIAS and PARAMETER as non-reserved (add back to the identifier set or only match them where needed); do not treat SYMBOL as a keyword—handle it as a normal identifier/property.
  2. CREATE FUNCTION semantics gap: The extended Doris syntax is added, but the visitor only sets functionName and optional routineBody. Return type, parameter list, PROPERTIES, alias expression, etc., are not mapped into CreateFunctionStatement, so parsing results remain incomplete. Please add the corresponding segments/fields and visitor logic (or justify not keeping them and narrow the grammar); otherwise, semantics are missing.

  After fixing these, check if any new cases need to be adjusted accordingly. Looking forward to your continued improvements—this capability is very valuable.

ccxxxyy:
@terrymanu Hi, I have made the revisions as required. Hope you can review them when you have time.

terrymanu (changes requested):
The direction is solid: enhancing Doris support for REGEXP, SELECT INTO OUTFILE, and CREATE FUNCTION with grammar, visitor, segment, and tests. Coverage is noticeably better than before—nice work, and thanks for adding test cases.

  However, a few issues need fixing before merge:

  - test/it/parser/.../impl/outfile/ExpectedProperty.java is newly added but unused, while ExpectedOutfileClause still relies on the distsql ExpectedProperty. This is dead code/wrong import. Please remove this class or switch to a genuinely referenced type to avoid leaving unused files in the codebase.
  - Promoting ALIAS and PARAMETER to keywords may break existing identifiers (e.g., column/table names) if Doris doesn’t treat them as reserved. Please confirm against official syntax and add evidence there’s no compatibility regression; if uncertain, revert the keyword promotion.
  - For SELECT INTO OUTFILE, the visitor only captures filePath/format/properties; COLUMNS/LINES/CHARACTER SET are still ignored, so the AST is incomplete and rewrite/routing can’t access these options. Please add corresponding segments/assignments or explain why Doris doesn’t need them.
  - The linked issue mentions array_filter, but grammar/visitor/tests don’t cover it. If this PR is meant to include that syntax, please add it; if out of scope, clarify in the description to avoid confusion.
  - Please check for any changes unrelated to this PR’s goal and revert them to keep the diff focused.

  Once these are addressed, please resubmit—I’d be happy to review again. Keep it up

ccxxxyy:
@terrymanu Hi! I hope my revisions can meet the requirements. Hope you could take a look when you have a moment.

- Regarding ALIAS and PARAMETER: Doris does not treat them as reserved, although they are still kept in the DorisKeyword file. However, in the previous commit, I have added ALIAS and PARAMETER to the identifierKeywordsUnambiguous rule in the BaseRule file. I hope this can resolve any potential conflicts.
- Regarding the SELECT INTO OUTFILE statement: I haven't found any description of COLUMNS/LINES/CHARACTER in all the official [documentation ](https://doris.apache.org/zh-CN/docs/4.x/data-operate/export/outfile) related to this statement. The content I can only find about it is <query_stmt>
INTO OUTFILE "<file_path>"
[ FORMAT AS <format_as> ]
[ <properties> ], and this is also what I want to support and fix in the PR. Therefore, I didn't consider COLUMNS/LINES/CHARACTER before.

- Regarding `array_filter`: For every unsupported syntax mentioned in the issue, I will do my best to fix it. Before proposing a PR, I attempted to support the parsing of the `array_filter` syntax, but my several attempts all ended in failure. The difficulty currently exceeds my ability to solve it. Therefore, I will try to clearly express in the PR title which syntaxes from the issue I have fixed. Meanwhile, I use "for" instead of "fix" when referring to the issue in the PR. I hope that when reviewing the issue in the future, I can successfully propose parsing solutions for these syntaxes that are currently difficult for me, or that more excellent contributors can complete the support. If there are any places where I need to provide further explanations or descriptions about the fixes, please let me know.

PR #37510, opened 2025-12-25
support oracle blob type.(https://github.com/apache/shardingsphere/is…

### LINKED ISSUE #37504: Insertion Failure with BLOB Columns
## Bug Report

v5.5.1

ShardingSphere-JDBC

### Expected behavior
insert success

### Actual behavior
insert failure

I know it. I can fix this bug, so create this issue. Actually, I already fixed it in my project.

SQL like this:
insert into my_user(id,user_name,addr) values(?,?,?) ::: [122401, 测试blob, java.io.ByteArrayInputStream@685f56ef]

addr is blob type in oracle

I will describe it in detail in the following section

daiqiang:
Fixes #37504 

  -
AbstractPreparedStatementAdapter， support blob type with ojdbc6
---

- [ √] My code follows the [code of conduct](https://shardingsphere.apache.org/community/en/involved/conduct/code/) of this project.
- [ √] I have self-reviewed the commit code.
- [ √] I have (or in comment I request) added corresponding labels for the pull request.
- [ √] I have passed maven check locally : `./mvnw clean install -B -T1C -Dmaven.javadoc.skip -Dmaven.jacoco.skip -e`.

### DIFF
--- a/jdbc/src/main/java/org/apache/shardingsphere/driver/jdbc/adapter/AbstractPreparedStatementAdapter.java
+++ b/jdbc/src/main/java/org/apache/shardingsphere/driver/jdbc/adapter/AbstractPreparedStatementAdapter.java
@@ -300,7 +300,7 @@ public abstract class AbstractPreparedStatementAdapter extends AbstractUnsupport
             }
         }
     }
-
+    
     private void handleOracleInputStreamSetBlob(final PreparedStatement preparedStatement, final int index, final Object each) throws SQLException {
         try {
             long length = -1;


terrymanu (changes requested):
Here’s a gentle change request. The direction is right—thanks for tackling the Oracle BLOB issue—but we need more safety and compatibility before merging.

  - First, kudos: identifying that ojdbc6 throws “Invalid column type” when using setObject with InputStream and trying setBlob is on the right track for issue #37504. Thanks for the contribution.
  - Gaps to address:
      1. The replay now forces every InputStream through setBlob(index, stream), losing the original setter semantics (setAsciiStream, setBinaryStream, setObject) and length information. That can change behavior for non-BLOB streams and may throw SQLFeatureNotSupportedException on drivers that don’t support setBlob(InputStream) (e.g., some PostgreSQL/JDBC variants). Suggestion: record which setter was originally called and replay with the same method/signature (including length); only use setBlob when it’s definitively a BLOB scenario.
      2. Some Oracle driver versions/configurations prefer setBlob(index, stream, length); the current change doesn’t pass length, so failures can still occur. Please capture and replay the length when the original setter provided it.
      3. There are no tests covering Oracle BLOB via PreparedStatement with an InputStream, so we can’t validate the fix or guard against regressions. Please add a ShardingSphere-JDBC test that uses InputStream to write a BLOB, ensuring no “Invalid column type,” and verify that non-Oracle stream parameters retain their prior behavior.
  - Newly introduced risk: stream parameters for other databases are now treated as BLOB, which can alter data types or cause execution errors. Please narrow the impact by replaying based on the original setter/database type, keeping the previous behavior for non-Oracle where appropriate.
  - There are no unrelated changes to revert. However, the above issues should be fixed before we merge. Looking forward to the updated version and happy to review again.

daiqiang:
Thanks for your detailed review! I've updated the code based on your feedback, and here's the summary of changes & known constraints:

#### 1. Narrowed the impact scope (addressed "all InputStream forced to setBlob" issue)
- Only apply setBlob logic to **Oracle database** (judged by `DatabaseMetaData.getDatabaseProductName()`);
- Non-Oracle databases retain the original `setObject` logic for all stream parameters (no behavior change).

#### 2. Added length parameter for setBlob (addressed "missing length" issue)
- Prioritize calling `setBlob(index, stream, length)` when length is available;
- For `ByteArrayInputStream`, auto get length via `available()`;
- Fallback to `setBlob(index, stream)` if length is unavailable.

#### 3. Known constraint (ojdbc6 limitation)
Due to ojdbc6's incomplete JDBC compliance:
- `ParameterMetaData.getParameterTypeName()`/`getParameterType()` throw `SQLFeatureNotSupportedException` (ojdbc6's known issue);
- Cannot judge whether the parameter is a BLOB column, so all Oracle InputStream parameters (even non-BLOB streams) will use setBlob temporarily.
- Mitigation: Added fallback logic — if `setBlob` fails (throws SQLException), auto revert to original `setObject` to avoid new exceptions.

#### 4. Test case explanation
- Verified the fix in my local project (ShardingSphere 5.5.1 + Oracle + ojdbc6): Insert BLOB with InputStream no longer throws "Invalid column type";
- Cannot add official test cases to master branch (5.5.3-SNAPSHOT) for 2 reasons:
  1. I'm not familiar with the test structure of ShardingSphere-JDBC (could you provide guidance/docs about where to write Oracle-related test cases?);
  2. The master branch (5.5.3-SNAPSHOT) lacks Oracle-related parser code — basic Oracle CRUD cannot run normally (e.g., missing files in `parser->sql->engine->dialect->oracle`), making it impossible to run Oracle tests.

#### 5. Impact scope confirmation
- Non-Oracle databases: 100% retain original logic (no impact);
- Oracle databases: Only InputStream parameters use setBlob (with fallback to setObject on failure), other parameter types remain unchanged.

Looking forward to your further feedback! If there's any way to mitigate the ojdbc6 limitation (e.g., alternative ways to judge BLOB columns), I'm happy to adjust the code.

terrymanu (changes requested):
Changes needed before acceptance:

  - Design placement: ShardingSphere is pluggable; Oracle is a pluggable dialect. Avoid hardcoding dbProductName=="Oracle" in the main flow (jdbc/.../AbstractPreparedStatementAdapter). Add an SPI hook in the main flow and implement the Oracle InputStream→BLOB handling in the database dialect module.
  - Root cause not fully addressed: The true issue is losing the original setter semantics during parameter replay. Guessing via product name + instanceof InputStream is narrow and can mis-handle non-BLOB InputStream parameters. Please preserve the original setter (or use column metadata) via the SPI hook rather than relying solely on stream type.
  - Compatibility risk: Exact match on DatabaseProductName to "Oracle" is too narrow (common values include “Oracle Database”), so the fix may be ineffective in practice.
  - Driver handling: Only ByteArrayInputStream passes length; other streams still use length-less setBlob, which may continue to throw “invalid column type” on older drivers, and falling back to setObject reintroduces the original issue. Please harden this path or bound the scope.
  - Missing validation: No new unit/integration tests cover Oracle BLOB insert replay. Add unit tests (Mockito mock PreparedStatement to verify InputStream calls setBlob) and, if possible, an integration case.

  Please rework per above (move to dialect SPI, add tests, broaden matching/semantics preservation) and resubmit. Happy to discuss further.

daiqiang:
This is too difficult, I give up.

PR #37750, opened 2026-01-15
Fix metadata not refreshed after executing DDL within PostgreSQL transaction 

vudsen:


vudsen:
Sry for the noise, I just want to try weather the workflows can run in my fork but I mis click the wrong branch.

PR #38036, opened 2026-02-13
Use docker v28.0.4 to run E2E

terrymanu:


linghengqian (commented):
Considering the changes claimed in https://github.com/actions/runner-images/issues/13474#issuecomment-3928452506 , the current PR appears to introduce redundant code.

PR #38057, opened 2026-02-16
handle byte[] pagination parameters in PaginationContext(#34289)

### LINKED ISSUE #34289: Class MySQLComStmtExecuteExecutor throws a NullPointerException when select with limit by Golang

aMetric:
## What is the purpose of the change

Fixes #34289.

When executing MySQL prepared statements with pagination placeholders (for example, `LIMIT ?` or `LIMIT ?, ?`), some clients (such as Go MySQL clients) may send parameter values as `byte[]` in binary protocol.
`PaginationContext` previously assumed pagination marker values were only `Long` or `Integer`, which could cause runtime failures in the proxy execution path.

This PR makes pagination parameter conversion robust for binary-protocol inputs without changing normal numeric behavior.

## Brief changelog

- Refactor pagination marker conversion in:
  - `PaginationContext#getValue`
  - add `PaginationContext#getLongValue`
- Support pagination marker values from:
  - `Number`
  - `byte[]` (parsed as UTF-8 full numeric string)
  - string-like objects via `toString()`
- Add dedicated unit tests:
  - `PaginationContextTest#assertGetActualOffsetWithByteArrayParameters`
  - `PaginationContextTest#assertGetActualRowCountWithByteArrayParameters`
  - `SelectStatementContextTest#assertBindParametersPopulatePaginationContextWithByteArrayParameters`


terrymanu:
Please fix spotless for code format

aMetric:
Hi @terrymanu ,

Happy New Year! 🎉

I've fixed the code format with Spotless. Please review when you have time.

Thanks!

terrymanu (approved):

terrymanu:
@aMetric  Happy New Year! 🎉
Thank you for contribution!

Issue #31472, opened 2024-05-31
Support parsing Doris sql


iamhucong:
## Background

Hi community.\
The ShardingSphere SQL parser engine helps users to parse SQL to create the AST (Abstract Syntax Tree) and visit the AST to get SQLStatement (Java Object). Currently, we are planning to enhance the support for Doris SQL parsing in ShardingSphere.

More details:
<https://shardingsphere.apache.org/document/current/en/reference/sharding/parse/>  

## Issue Background Explanation

The current issue involves using a custom web scraping script to fetch SQL cases from the official website of a corresponding database. These SQL cases are then passed to ShardingSphere's parsing engine for analysis. For SQL cases that fail to be parsed successfully, every 3 to 5 SQL cases are grouped together as an issue.
1. Since  SQL cases are obtained through web scraping, it cannot be guaranteed that all SQL cases are correct. Please follow the following process to handle this pull request (PR).
2. Some SQL cases may have already been fixed in other PRs. For cases that can already be executed successfully, simply leave a comment to ignore them.
3. If a SQL case can be executed successfully without any code changes, there is no need to add a corresponding test assert file.  

## Task

This issue is to support more Doris sql parsing, as follows:

```
CREATE ROUTINE LOAD example_db.test1 PROPERTIES (     "desired_concurrent_number"="3",     "max_batch_interval" = "20",     "max_batch_rows" = "300000",     "max_batch_size" = "209715200",     "strict_mode" = "false" ) FROM KAFKA (     "kafka_broker_list" = "broker1:9092,broker2:9092,broker3:9092",     "kafka_topic" = "my_topic",     "property.group.id" = "xxx",     "property.client.id" = "xxx",     "property.kafka_default_offsets" = "OFFSET_BEGINNING" )
```

- [x] [link](https://doris.apache.org/docs/dev/sql-manual/sql-statements/Data-Manipulation-Statements/Load/CREATE-ROUTINE-LOAD/)

```
[...]

PR #38137, opened 2026-02-22
Support parsing Doris CREATE ROUTINE LOAD syntax

ccxxxyy:
for https://github.com/apache/shardingsphere/issues/31472

official docuemnt:

- https://doris.apache.org/docs/4.x/sql-manual/sql-statements/data-modification/load-and-export/CREATE-ROUTINE-LOAD

  -

---


ccxxxyy:
Regarding **KAFKA**, [Doris ](https://doris.apache.org/docs/4.x/sql-manual/basic-element/reserved-keywords)does not treat it as reserved keyword. Therefore, I have added it to `identifierKeywordsAmbiguous3Roles` in `BaseRule.g4`, so that it can be used in places where identifiers are expected.

My reasoning is as follows:

The identifier usage chain is:

tableName → name → identifier
columnName → identifier

and identifier includes identifierKeywordsAmbiguous3Roles.

This means that the keyword can be used as identifiers, such as table names, column names, and database names, without requiring quotation marks.

terrymanu (changes requested):
This is a good direction, and the grammar/visitor/statement/IT wiring for CREATE ROUTINE LOAD is mostly in place.
  There are still a few blocking items before merge:

  1. Root-cause fix is still incomplete: COLUMNS mapping expressions are accepted by grammar but dropped in AST

  - Grammar supports columnName (EQ_ expr)? in columnMapping at parser/sql/engine/dialect/doris/src/main/antlr4/imports/doris/DMLStatement.g4:235.
  - Visitor only collects columnName and does not extract mapping expressions at parser/sql/engine/dialect/doris/src/main/java/org/apache/shardingsphere/sql/parser/engine/doris/visitor/statement/type/DorisDMLStatementVisitor.java:146.
  - Statement model has no field/segment to keep mapping expressions at parser/sql/statement/dialect/doris/src/main/java/org/apache/shardingsphere/sql/parser/statement/doris/dml/DorisCreateRoutineLoadStatement.java:55.
  - This causes silent semantic loss for inputs like v3 = k1 * 100 or dt=from_unixtime(...). Please add segment modeling + visitor extraction + assertions.

  2. Please add tests mapped one-to-one with the fix points

  - Add at least one CREATE ROUTINE LOAD case containing column = expr.
  - Assert the mapped expression in parser case assertions, not only parse success.

ccxxxyy:
@terrymanu Hi! Thanks very much for your correction and guidance. I have made the requested changes in this commit and would greatly appreciate it if you could review them when you have time.

- I have fixed the issues in `DorisDMLStatementVisitor.java`, as well as the related `Statement`, `Assert`, and `TestCase` classes.  

- I have added corresponding test cases and assertions to:

`src/main/resources/sql/supported/dml/create-routine-load.xml`

`src/main/resources/case/dml/create-routine-load.xml`

The expressions are now properly asserted, and all tests have passed successfully.

Thanks again for your reply and for the time.

terrymanu (approved):

PR #38388, opened 2026-03-10
Add core Oracle system schema tables and set SYS as default

### LINKED ISSUE #37766: ShardingSphere-JDBC 5.5.2 encounters the error under Oracle: Table or view 'all_sequences' does not exist
ShardingSphere 5.5.2

ShardingSphere-JDBC

When I use ShardingSphere 5.5.2 in development environment with JDK11 + Oracle 11g + hibernate 5.6.15.Final + spring 5.2.20.RELEASE, I received the following error: "nested exception is org.apache.shardingsphere.infra.exception.kernel.metadata.TableNotFoundException: **Table or view 'all_sequences' does not exist**."

or "nested exception is org.apache.shardingsphere.infra.exception.kernel.metadata.TableNotFoundException: **Table or view 'all_synonyms' does not exist.**"

somiljain2006:
Fixes #37766

Set SYS as the default system schema in OracleSchemaOption so that Oracle dictionary views like ALL_SEQUENCES and ALL_SYNONYMS can be recognized as system tables.

Add YAML definitions for commonly used Oracle dictionary views:
- ALL_SEQUENCES
- ALL_SYNONYMS
- ALL_TABLES
- ALL_VIEWS
- ALL_TAB_COLUMNS
- USER_TABLES

Update unit tests to verify the default system schema behavior.

---

- [x] I have passed maven check locally: `./mvnw clean install -B -T1C -Dmaven.javadoc.skip -Dmaven.jacoco.skip -e`.


somiljain2006:
@terrymanu @linghengqian @strongduanmu Can you please review this pr?

somiljain2006:
Thank you, @terrymanu, for the detailed review and for pointing out the missing assembly chain. You were absolutely right that pointing to SYS wasn't enough without the metadata assembly.

I have already pushed new commits that address all of these concerns end-to-end:

1. Added DialectSystemDatabase Implementation [P0]:
I created OracleSystemDatabase.java, which returns SYS for the system schemas, and registered it in database/connector/dialect/oracle/src/main/resources/META-INF/services/org.apache.shardingsphere.database.connector.core.metadata.database.system.DialectSystemDatabase. The SYS schema is now properly assembled into the metadata.

2. Added Java Unit Test:
Added OracleSystemDatabaseTest.java to verify the SPI loading and ensure code coverage is maintained.

3. Added Binder Regression Test [P1]:
I added a regression test for an unqualified dictionary view (SELECT * FROM ALL_TABLES) in the parser/binder test suite (select.xml). The test explicitly verifies that the ALL_TABLES table-bound original schema resolves to SYS.

somiljain2006:
Thank you, @terrymanu, for the detailed follow-up review.

I have addressed both of the [P1] issues in the latest commits:

1. End-to-End Test Activated:
I added the missing <sql-case> for SELECT * FROM ALL_TABLES to sqls/dml/select.xml, so the test case select_oracle_all_tables_unqualified now has both the SQL input and the expected binder output. The binder IT now executes this scenario and validates that the unqualified dictionary view resolves through the SYS schema path.

2. SPI Validation in Unit Test Fixed:
I updated OracleSystemDatabaseTest to use DatabaseTypedSPILoader.getService() instead of directly instantiating the class (matching the pattern used in MySQLSystemDatabaseTest). The test now correctly validates that the framework successfully discovers the SPI service registration in META-INF/services.

somiljain2006:
@terrymanu Thanks for the thorough review and clarification.

I have addressed the remaining P1 issues mentioned in the previous feedback:

1. I moved the select_oracle_all_tables_unqualified case into the binder's own local resources (test/it/binder/src/test/resources/sqls/dml/select.xml). The binder IT now successfully reads and processes the new SQL.
2. I updated the mockSchemas() method in SQLBinderIT to properly assemble and inject the system schemas using the DialectSystemDatabase SPI. This ensures the mock metadata used by the binder tests now contains the SYS schema.
3. I updated the binder's expected XML (test/it/binder/src/test/resources/cases/dml/select.xml) to explicitly assert that ALL_TABLES successfully binds to the SYS schema.

somiljain2006:
Thank you, @terrymanu, for pointing out the semantic compatibility risk and the need for stronger column-level validation. I have updated the PR to address both issues:

1. Compatibility Risk Resolved (P1):
I updated SimpleTableSegmentBinder so that it performs standard schema first name resolution. The SYS dictionary routing is now strictly used as a fallback only when normal resolution fails. This preserves Oracle's original lookup priority and ensures same name objects in the current schema won't be incorrectly rebound.

2. Real Assembly Path & Stronger Regression (P2):
I removed the empty schema injection in SQLBinderIT and replaced it with the real SystemSchemaBuilder execution path, meaning the testing fixture now loads the actual YAML dictionary-view definitions.
I also added the requested SELECT OWNER FROM ALL_TABLES regression case. The binder IT now successfully validates column-level binding (OWNER) against the real dictionary-view metadata, proving both the YAML assembly and the fallback routing work end-to-end.

somiljain2006:
@terrymanu I have addressed all three of the remaining issues in my latest commit:

1. Semantic Regression Fixed [P0]:
I updated SimpleTableSegmentBinder.getSchemaName() so that the "default schema first, system schema fallback" precedence is explicitly scoped to "Oracle". For PostgreSQL, openGauss, and generic dialects, the code preserves the original precedence, searching the system schema first to prevent disruption to pg_catalog.

2. Assembly Disabled Path Fixed [P1]:
In the Oracle fallback logic, I added a safeguard that explicitly checks if the table metadata exists in SYS (sysSchema.containsTable(tableName)). If system-schema-metadata-assembly-enabled=false, the Oracle YAML isn't loaded, containsTable() yields false, and the binder gracefully drops back to the default schema instead of degrading into an empty temporary-table context.

3. Coverage for ALL_SEQUENCES and ALL_SYNONYMS [P2]:
I have added end-to-end regression tests to the binder IT for both ALL_SEQUENCES and ALL_SYNONYMS to ensure the original issue's failure surface is fully covered.

menghaoranss:
@somiljain2006 Hi, thank you very much for your contribution and effort on this PR!

We have reverted this change(https://github.com/apache/shardingsphere/pull/38485) because some parts of the implementation do not align with ShardingSphere’s overall design, particularly breaking the SPI-based pluggable architecture, which is a key principle of the project.

We truly appreciate your work and the time you invested. Looking forward to your next contribution!

somiljain2006:
@menghaoranss No worries, I learned a lot while working on this PR. Looking forward to contributing more meaningful improvements in the future.

PR #38405, opened 2026-03-12
Fix correlated project pushdown for SQL federation IN subqueries

### LINKED ISSUE #37439: [SQL Federation] Correlated IN-subquery fails in a non-join subquery but succeed in a more complex subquery with join.

shardingsphere-proxy 5.5.2
postgres 17.6 as backend

### Expected behavior

The following two query should all succeed.

### Actual behavior
```sql
testdb=> select  
*
from 
  t0 as ref_0
where (ref_0.c2) in (select  
        ref_0.c6 as c_0
      from t23 as ref_3
);
ERROR:  SQL federation does not support SQL 'select  
*
from 
  t0 as ref_0
where (ref_0.c2) in (select  
        ref_0.c6 as c_0
      from t23 as ref_3
);'.
More details: java.lang.RuntimeException: Error while applying rule EnumerableScanConverterRule(in:NONE,out:ENUMERABLE), args [rel#7036:LogicalScan.NONE.[](table=[public, t23],pushDownRelBuilder=LogicalProject(c_0=[$cor0.c6])
  LogicalTableScan(table=[[public, t23]])
)]

testdb=> select  
*
from 
  t0 as ref_0
where (ref_0.c2) in (select  
        ref_0.c6 as c_0
      from 
        (t9 as ref_1
          full outer join t23 as ref_3
          on (ref_1.vkey = ref_3.vkey ))
);
 vkey | pkey | c0 | c1 | c2 | c3 | c4 | c5 | c6 | c7 | c8 
------+------+----+----+----+----+----+----+----+----+----
(0 rows)
```

The second query is similar to the first one, but it is more complex (it has a join in subquery). We consider the second query need a more complex optimization to execute. If the second query is valid, the first one should be valid as well.

We consider it a **vulnerability** of SQL Federation Planner.

[ss_postgres.zip](https://github.com/user-attachments/files/24253131/ss_postgres.zip)

run `docker compose up -d` in this file.

run following query to initialize db.
```sql
SET DEFAULT SINGLE TABLE STORAGE UNIT = ds_0;

DROP TABLE IF EXISTS t0;

create table t0 ( 
vkey int4 ,
pkey int4 ,
c0 text ,
c1 int4 ,
c2 text ,
c3 text ,
c4 int4 ,
c5 text ,
c6 text ,
c7 text ,
c8 numeric 
);

CREATE BROADCAST TABLE RULE t9;

DROP TABLE IF EXISTS t9;

create table t9 ( 
vkey int4 ,
pkey int4 ,
c17 text ,
c18 numeric ,
c19 text ,
c20 int4 ,
c21 text ,
c22 numeric ,
c23 numeric ,
c24 numeric 
);

SET DEFAULT SINGLE TABLE STORAGE UNIT = ds_1;

DROP TABLE IF EXISTS t23;

create table t23 ( 
vkey int4 ,
pkey int4 ,
c29 int4 ,
c30 numeric 
);
```

SQL to execute are shown above.

ym0506:
### Summary

This PR fixes correlated `IN` subqueries that fail when the subquery projects an outer reference from a non-join subquery.

### Root Cause

`PushProjectIntoScanRule` pushed project expressions into `LogicalScan` without checking whether the projection contained correlated references.

For queries like:

    SELECT o.order_id
    FROM t_order o
    WHERE o.user_id IN (
        SELECT o.user_id
        FROM t_order_item i
    )

the projected expression inside the subquery contains an outer reference. After it was pushed into `LogicalScan`, the scan conversion path failed because the pushed-down scan tree still contained correlated expressions.

The more complex join case succeeded because it did not go through the same simple project-to-scan pushdown path.

### Fix

This change makes `PushProjectIntoScanRule` skip pushdown when a projected expression contains correlated references.

Specifically:
- detect `RexCorrelVariable`
- detect `RexFieldAccess` that references a correlated expression
- recursively inspect nested `RexCall` operands

### Tests

Added regression coverage for:
- `PushProjectIntoScanRuleTest`
- `SQLStatementCompilerIT#assertCompileWhenCorrelatedInSubqueryProjectsOuterColumn`

Fixes #37439

### DIFF
--- a/kernel/sql-federation/compiler/src/main/java/org/apache/shardingsphere/sqlfederation/compiler/planner/rule/transformation/PushProjectIntoScanRule.java
+++ b/kernel/sql-federation/compiler/src/main/java/org/apache/shardingsphere/sqlfederation/compiler/planner/rule/transformation/PushProjectIntoScanRule.java
@@ -23,6 +23,8 @@ import org.apache.calcite.plan.RelRule;
 import org.apache.calcite.rel.logical.LogicalProject;
 import org.apache.calcite.rel.rules.TransformationRule;
 import org.apache.calcite.rex.RexCall;
+import org.apache.calcite.rex.RexCorrelVariable;
+import org.apache.calcite.rex.RexFieldAccess;
 import org.apache.calcite.rex.RexNode;
 import org.apache.calcite.rex.RexSubQuery;
 import org.apache.shardingsphere.sqlfederation.compiler.rel.operator.logical.LogicalScan;
@@ -55,11 +57,28 @@ public final class PushProjectIntoScanRule extends RelRule<PushProjectIntoScanRu
         }
         LogicalProject logicalProject = call.rel(0);
         for (RexNode each : logicalProject.getProjects()) {
-            if (each instanceof RexSubQuery || containsCastFunction(each)) {
+            if (each instanceof RexSubQuery || containsCorrelate(each) || containsCastFunction(each)) {
                 return false;
             }
         }
         return true;
+    }
+    
+    private boolean containsCorrelate(final RexNode rexNode) {
+        if (rexNode instanceof RexCorrelVariable) {
+            return true;
+        }
+        if (rexNode instanceof RexFieldAccess) {
+            return containsCorrelate(((RexFieldAccess) rexNode).getReferenceExpr());
+        }
+        if (rexNode instanceof RexCall) {
+            for (RexNode each : ((RexCall) rexNode).getOperands()) {
+                if (containsCorrelate(each)) {
+                    return true;
+                }
+            }
+        }
+        return false;
     }
     
     private boolean containsCastFunction(final RexNode rexNode) {
--- a/kernel/sql-federation/compiler/src/test/java/org/apache/shardingsphere/sqlfederation/compiler/compiler/it/SQLStatementCompilerIT.java
+++ b/kernel/sql-federation/compiler/src/test/java/org/apache/shardingsphere/sqlfederation/compiler/compiler/it/SQLStatementCompilerIT.java
@@ -37,6 +37,7 @@ import org.apache.shardingsphere.sqlfederation.compiler.metadata.schema.SQLFeder
 import org.apache.shardingsphere.sqlfederation.compiler.rel.converter.SQLFederationRelConverter;
 import org.apache.shardingsphere.sqlfederation.compiler.sql.function.mysql.MySQLOperatorTable;
 import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtensionContext;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.Arguments;
@@ -54,7 +55,9 @@ import java.util.Properties;
 import java.util.stream.Stream;
 
 import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.containsString;
 import static org.hamcrest.Matchers.is;
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 import static org.mockito.Mockito.mock;
 
 class SQLStatementCompilerIT {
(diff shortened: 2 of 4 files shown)


strongduanmu:
Hi @ym0506, this pr looks great, can you add an e2e sql test in `db_tbl_sql_federation` scenario?

ym0506:
> Hi @ym0506, this pr looks great, can you add an e2e sql test in `db_tbl_sql_federation` scenario?

thank you
Added an e2e SQL regression case in the db_tbl_sql_federation scenario.

The new case covers the correlated non-join IN subquery path that triggered this issue:
SELECT o.order_id FROM t_order o WHERE o.order_id < ? AND o.user_id IN (SELECT o.user_id FROM t_order_item i) ORDER BY order_id

I limited it with order_id < 1010 so the expected result stays small and deterministic.

ym0506:
I adjusted the e2e regression case to keep the outer side selective while preserving the correlated non-join IN-subquery path.

The new SQL still covers the original issue pattern, but wraps the outer table in a filtered derived table so the outer scan can keep `order_id = ?` in the pushed-down SQL.

terrymanu (commented):
### Decision

- **Merge Verdict: Mergeable**
- **Reviewed Scope:** Latest PR head `7c7f5db2f4d6ecc24f477a1ed9d93386bf7922e8`; PR base metadata SHA and local merge-base `4b18e3ece196fc784c43c9d47491edca24c4c35d`; fetched latest `master` `1e934a2f99e9ff5d9d4d2ebedda05ef8a13e852e`.
  The local triple-dot file list matched GitHub REST `/pulls/38405/files` exactly: 4 modified files, +47/-1.
  Reviewed files: `kernel/sql-federation/compiler/src/main/java/org/apache/shardingsphere/sqlfederation/compiler/planner/rule/transformation/PushProjectIntoScanRule.java:60`, `kernel/sql-federation/compiler/src/test/java/org/apache/shardingsphere/sqlfederation/compiler/planner/rule/transformation/PushProjectIntoScanRuleTest.java:82`, `kernel/sql-federation/compiler/src/test/java/org/apache/shardingsphere/sqlfederation/compiler/compiler/it/SQLStatementCompilerIT.java:248`, `test/e2e/sql/src/test/resources/cases/dql/cases/db_tbl_sql_federation/e2e-dql-select-sub-query-for-db-tbl-sql-federation.xml:28`.
- **Not Reviewed Scope:** Full Docker e2e execution with live MySQL/PostgreSQL/openGauss containers; GitHub Actions/check-run status; unrelated local workspace changes.
- **Need Expert Review:** No dedicated expert review required beyond normal SQL Federation planner maintainer review.

### Basis

- The linked issue #37439 fails when a simple correlated `IN` subquery projects an outer reference into a single-table subquery scan, leaving `LogicalProject(c_0=[$cor0.c6])` inside `LogicalScan` and failing during `EnumerableScanConverterRule`.
- The fix addresses that trigger point directly: `PushProjectIntoScanRule` now refuses project pushdown when a project expression contains a correlation reference, so the correlated expression stays above the scan instead of being embedded in the physical scan tree.
- Existing adjacent behavior is preserved: system-schema, top-level subquery, and `CAST` guards remain; non-correlated project pushdown still matches and still calls `logicalScan.pushDown(logicalProject)`.
- Shared-path review looks acceptable. This is a generic Calcite/SQL Federation planner guard, not PostgreSQL-specific logic, and the e2e case covers MySQL, PostgreSQL, and openGauss for the shared `db_tbl_sql_federation` path.
- The previously requested `db_tbl_sql_federation` e2e coverage is present in the latest diff and exercises the non-join correlated `IN` subquery path.
- No substantive unrelated changes were found.

### Verification

- Boundary check: `curl ... /repos/apache/shardingsphere/pulls/38405/files` exited 0 and matched `git diff --name-status 4b18e3ece196fc784c43c9d47491edca24c4c35d..refs/codex-review/38405/head`.
- Targeted tests on an isolated PR-head worktree exited 0:
  `./mvnw -pl kernel/sql-federation/compiler -am -DskipITs -Dspotless.skip=true -Dcheckstyle.skip=true -Drat.skip=true -Djacoco.skip=true -Dtest=PushProjectIntoScanRuleTest,SQLStatementCompilerIT -Dsurefire.failIfNoSpecifiedTests=false test`
  Result: 114 tests run, 0 failures, 0 errors.
- Repository style gates on the same isolated PR-head worktree exited 0:
  `./mvnw -pl kernel/sql-federation/compiler -am -Pcheck -T1C -DskipTests -Djacoco.skip=true spotless:check checkstyle:check`
- Fresh adversarial pass did not find unresolved cross-dialect, feature-disabled, original-symptom, performance, implicit-state, or unrelated-scope blockers.

strongduanmu (approved):
Looks great, merged.

PR #38444, opened 2026-03-15
jdbc: add MySQL-compatible typed string conversion

### LINKED ISSUE #37757: Suggest The value conversion from database to JAVA object in ResultSetUtils refers to the native MySQL compatibility
https://github.com/apache/shardingsphere/blob/57ac720ec058386e35fdd760e6627177cb12ca50/infra/executor/src/main/java/org/apache/shardingsphere/infra/executor/sql/execute/result/query/impl/driver/jdbc/type/util/ResultSetUtils.java#L67

issue:
有些老项目数据库定义与java的对象定义不是严格匹配导致数据源切换shardingJDBC后报错。
version: (项目使用的5.3.2) - latest
目前master分支对于String也只支持转换Timestamp,java.sql.Date,Time三种。

原因：
项目中还会存在String对于Integer类似这种没有兼容: （实际项目中使用了mybatis-plus的@EnumValue 修饰的java对象类型为Integer，数据库定义为char。在com.baomidou.mybatisplus.core.handlers.MybatisEnumTypeHandler#getNullableResult(java.sql.ResultSet, int)中获取值时是调用了Object value = rs.getObject(columnIndex, this.propertyType);）
对于原生的com.mysql.cj.jdbc.result.ResultSetImpl#getObject(int, java.lang.Class<T>)中是支持的。

建议：
所以建议是否考虑使用原生的方法兼容（目前做法是把数据库中的类型修改了）感谢

issue:
The database definitions of some old projects do not strictly match the object definitions in Java, resulting in errors when switching the data source to shardingJDBC.
version: (5.3.2 used in the project) - latest
Currently, the master branch only supports converting String to Timestamp, java.sql.Date, and Time.
Reason:
In the project, there may be incompatibilities like String versus Integer: (In the actual project, the Java object type decorated with @EnumValue from mybatis-plus is Integer, while the database definition is char. When obtaining the value in com.baomidou.mybatisplus.core.handlers.MybatisEnumTypeHandler#getNullableResult(java.sql.ResultSet, int), Object value = rs.getObject(columnIndex, this.propertyType) is called.); ）
It is supported in the native com.mysql.cj.jdbc.result.ResultSetImpl#getObject(int, java.lang.Class).
Suggestion:
So, I suggest considering using native methods for compatibility (the current approach is to modify the types in the database) Thanks.

aviralgarg05:
Fixes #37757.

  - Add MySQL-compatible typed string conversion for JDBC `getObject(int, Class<T>)` path while keeping strict default behavior for non-MySQL dialects.
  - Add focused unit tests for MySQL conversion success/invalid input and non-MySQL unchanged behavior.

---

### DIFF
--- a/.github/workflows/e2e-sql.yml
+++ b/.github/workflows/e2e-sql.yml
@@ -34,6 +34,7 @@ permissions:
 env:
   MAVEN_OPTS: -Dhttp.keepAlive=false -Dmaven.wagon.http.pool=false -Dmaven.wagon.http.retryHandler.class=standard -Dmaven.wagon.http.retryHandler.count=3 -Dspotless.apply.skip=true
   SEGMENT_DOWNLOAD_TIMEOUT_MINS: 1
+  TESTCONTAINERS_RYUK_DISABLED: true
 
 jobs:
   global-environment:
--- a/.github/workflows/e2e-sql.yml
+++ b/.github/workflows/e2e-sql.yml
@@ -34,7 +34,6 @@ permissions:
 env:
   MAVEN_OPTS: -Dhttp.keepAlive=false -Dmaven.wagon.http.pool=false -Dmaven.wagon.http.retryHandler.class=standard -Dmaven.wagon.http.retryHandler.count=3 -Dspotless.apply.skip=true
   SEGMENT_DOWNLOAD_TIMEOUT_MINS: 1
-  TESTCONTAINERS_RYUK_DISABLED: true
 
 jobs:
   global-environment:
--- a/infra/executor/src/main/java/org/apache/shardingsphere/infra/executor/sql/execute/result/query/impl/driver/jdbc/type/util/ResultSetUtils.java
+++ b/infra/executor/src/main/java/org/apache/shardingsphere/infra/executor/sql/execute/result/query/impl/driver/jdbc/type/util/ResultSetUtils.java
@@ -133,7 +133,7 @@ public final class ResultSetUtils {
     }
     
     private static boolean isMySQLCompatibleConversion(final DatabaseType protocolType) {
-        return null != protocolType && ("MySQL".equals(protocolType.getType()) || protocolType.getTrunkDatabaseType().map(optional -> "MySQL".equals(optional.getType())).orElse(false));
+        return null != protocolType && "MySQL".equals(protocolType.getType());
     }
     
     private static Optional<Object> convertStringValue(final String value, final Class<?> convertType, final boolean isMySQLCompatibleConversion) {
@@ -160,7 +160,7 @@ public final class ResultSetUtils {
     private static Optional<Object> convertMySQLStringValue(final String value, final Class<?> convertType) {
         String trimmedValue = value.trim();
         if (trimmedValue.isEmpty()) {
-            throw new UnsupportedDataTypeConversionException(convertType, value);
+            return Optional.of(convertMySQLEmptyValue(convertType, value));
         }
         try {
             switch (convertType.getName()) {
@@ -169,16 +169,16 @@ public final class ResultSetUtils {
                     return Optional.of(convertMySQLBooleanValue(trimmedValue, convertType, value));
                 case "byte":
                 case "java.lang.Byte":
-                    return Optional.of(new BigDecimal(trimmedValue).byteValueExact());
+                    return Optional.of(convertMySQLIntegerValue(trimmedValue).byteValueExact());
                 case "short":
                 case "java.lang.Short":
-                    return Optional.of(new BigDecimal(trimmedValue).shortValueExact());
+                    return Optional.of(convertMySQLIntegerValue(trimmedValue).shortValueExact());
                 case "int":
                 case "java.lang.Integer":
-                    return Optional.of(new BigDecimal(trimmedValue).intValueExact());
+                    return Optional.of(convertMySQLIntegerValue(trimmedValue).intValueExact());
                 case "long":
                 case "java.lang.Long":
-                    return Optional.of(new BigDecimal(trimmedValue).longValueExact());
+                    return Optional.of(convertMySQLIntegerValue(trimmedValue).longValueExact());
                 case "double":
                 case "java.lang.Double":
                     return Optional.of(Double.parseDouble(trimmedValue));
@@ -192,6 +192,44 @@ public final class ResultSetUtils {
             }
         } catch (final NumberFormatException | ArithmeticException ignored) {
             throw new UnsupportedDataTypeConversionException(convertType, value);
+        }
+    }
+    
+    private static BigDecimal convertMySQLIntegerValue(final String value) {
+        if (value.indexOf('.') > -1 || value.indexOf('e') > -1 || value.indexOf('E') > -1) {
+            double doubleValue = Double.parseDouble(value);
+            return BigDecimal.valueOf(doubleValue).setScale(0, RoundingMode.DOWN);
+        }
+        return new BigDecimal(value);
+    }
+    
+    private static Object convertMySQLEmptyValue(final Class<?> convertType, final String originalValue) {
+        switch (convertType.getName()) {
+            case "boolean":
+            case "java.lang.Boolean":
+                return false;
+            case "byte":
+            case "java.lang.Byte":
+                return (byte) 0;
+            case "short":
+            case "java.lang.Short":
+                return (short) 0;
+            case "int":
+            case "java.lang.Integer":
+                return 0;
+            case "long":
+            case "java.lang.Long":
+                return 0L;
+            case "float":
+            case "java.lang.Float":
+                return 0.0F;
+            case "double":
+            case "java.lang.Double":
+                return 0.0D;
+            case "java.math.BigDecimal":
+                return BigDecimal.ZERO;
+            default:
+                throw new UnsupportedDataTypeConversionException(convertType, originalValue);
         }
     }
     
--- a/infra/executor/src/main/java/org/apache/shardingsphere/infra/executor/sql/execute/result/query/impl/driver/jdbc/type/util/ResultSetUtils.java
+++ b/infra/executor/src/main/java/org/apache/shardingsphere/infra/executor/sql/execute/result/query/impl/driver/jdbc/type/util/ResultSetUtils.java
@@ -20,6 +20,7 @@ package org.apache.shardingsphere.infra.executor.sql.execute.result.query.impl.d
 import com.google.common.primitives.Ints;
 import com.google.common.primitives.Longs;
 import com.google.common.primitives.Shorts;
+import org.apache.shardingsphere.database.connector.core.type.DatabaseType;
 import lombok.AccessLevel;
 import lombok.NoArgsConstructor;
 import org.apache.shardingsphere.infra.exception.ShardingSpherePreconditions;
@@ -75,11 +76,15 @@ public final class ResultSetUtils {
      *
      * @param value original value
      * @param convertType expected class type
-     * @param isMySQLCompatibleConversion is MySQL compatible conversion or not
+     * @param protocolType protocol database type
      * @return converted value
      * @throws SQLFeatureNotSupportedException SQL feature not supported exception
      */
-    public static Object convertValue(final Object value, final Class<?> convertType, final boolean isMySQLCompatibleConversion) throws SQLFeatureNotSupportedException {
+    public static Object convertValue(final Object value, final Class<?> convertType, final DatabaseType protocolType) throws SQLFeatureNotSupportedException {
+        return convertValue(value, convertType, isMySQLCompatibleConversion(protocolType));
+    }
+    
+    private static Object convertValue(final Object value, final Class<?> convertType, final boolean isMySQLCompatibleConversion) throws SQLFeatureNotSupportedException {
         ShardingSpherePreconditions.checkNotNull(convertType, () -> new SQLFeatureNotSupportedException("Type can not be null"));
         if (null == value) {
             return convertNullValue(convertType);
@@ -125,6 +130,10 @@ public final class ResultSetUtils {
         } catch (final ClassCastException ignored) {
             throw new SQLFeatureNotSupportedException("getObject with type, cannot convert " + value.getClass().getName() + ":" + value + " to " + convertType.getName());
         }
+    }
+    
+    private static boolean isMySQLCompatibleConversion(final DatabaseType protocolType) {
+        return null != protocolType && ("MySQL".equals(protocolType.getType()) || protocolType.getTrunkDatabaseType().map(optional -> "MySQL".equals(optional.getType())).orElse(false));
     }
     
     private static Optional<Object> convertStringValue(final String value, final Class<?> convertType, final boolean isMySQLCompatibleConversion) {
--- a/infra/executor/src/main/java/org/apache/shardingsphere/infra/executor/sql/execute/result/query/impl/driver/jdbc/type/util/ResultSetUtils.java
+++ b/infra/executor/src/main/java/org/apache/shardingsphere/infra/executor/sql/execute/result/query/impl/driver/jdbc/type/util/ResultSetUtils.java
@@ -27,6 +27,7 @@ import org.apache.shardingsphere.infra.exception.ShardingSpherePreconditions;
 import org.apache.shardingsphere.infra.exception.kernel.data.UnsupportedDataTypeConversionException;
 
 import java.math.BigDecimal;
+import java.math.BigInteger;
 import java.math.RoundingMode;
 import java.net.MalformedURLException;
 import java.net.URL;
@@ -42,8 +43,8 @@ import java.time.ZoneId;
 import java.time.format.DateTimeFormatter;
 import java.time.temporal.TemporalAccessor;
 import java.util.Date;
-import java.util.Locale;
 import java.util.Optional;
+import java.util.regex.Pattern;
 
 /**
  * Result set utility class.
@@ -58,6 +59,10 @@ public final class ResultSetUtils {
                     + "[.SSSSSSSSS][.SSSSSSSS][.SSSSSSS][.SSSSSS][.SSSSS][.SSSS][.SSS][.SS][.S]"
                     + "[ ]"
                     + "[XXXXX][XXXX][XXX][XX][X]");
+    
+    private static final Pattern MYSQL_INTEGER_PATTERN = Pattern.compile("-?\\d+");
+    
+    private static final Pattern MYSQL_FLOATING_PATTERN = Pattern.compile("-?\\d*\\.\\d*");
     
     /**
      * Convert value via expected class type.
@@ -158,35 +163,34 @@ public final class ResultSetUtils {
     }
     
     private static Optional<Object> convertMySQLStringValue(final String value, final Class<?> convertType) {
-        String trimmedValue = value.trim();
-        if (trimmedValue.isEmpty()) {
+        if (value.isEmpty()) {
             return Optional.of(convertMySQLEmptyValue(convertType, value));
         }
         try {
             switch (convertType.getName()) {
                 case "boolean":
                 case "java.lang.Boolean":
-                    return Optional.of(convertMySQLBooleanValue(trimmedValue, convertType, value));
+                    return Optional.of(convertMySQLBooleanValue(value, convertType));
                 case "byte":
                 case "java.lang.Byte":
-                    return Optional.of(convertMySQLIntegerValue(trimmedValue).byteValueExact());
+                    return Optional.of(convertMySQLByteValue(value, convertType));
                 case "short":
                 case "java.lang.Short":
-                    return Optional.of(convertMySQLIntegerValue(trimmedValue).shortValueExact());
+                    return Optional.of(convertMySQLIntegerValue(value).shortValueExact());
                 case "int":
                 case "java.lang.Integer":
-                    return Optional.of(convertMySQLIntegerValue(trimmedValue).intValueExact());
+                    return Optional.of(convertMySQLIntegerValue(value).intValueExact());
                 case "long":
                 case "java.lang.Long":
-                    return Optional.of(convertMySQLIntegerValue(trimmedValue).longValueExact());
+                    return Optional.of(convertMySQLIntegerValue(value).longValueExact());
                 case "double":
                 case "java.lang.Double":
-                    return Optional.of(Double.parseDouble(trimmedValue));
+                    return Optional.of(Double.parseDouble(value));
                 case "float":
                 case "java.lang.Float":
-                    return Optional.of(Float.parseFloat(trimmedValue));
+                    return Optional.of(Float.parseFloat(value));
                 case "java.math.BigDecimal":
-                    return Optional.of(new BigDecimal(trimmedValue));
+                    return Optional.of(new BigDecimal(value));
                 default:
                     return Optional.empty();
             }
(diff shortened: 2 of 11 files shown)


terrymanu (changes requested):
Decision

Merge Verdict: Not Mergeable

Reviewed Scope: PR head 6016348a6cc for [PR 38444](https://github.com/apache/shardingsphere/pull/38444), including the 4 changed files in infra/executor and jdbc, plus the related execution-path context in ShardingSphereStatement.java (line 265) and ShardingSpherePreparedStatement.java (line 147).

Not Reviewed Scope: no live JDBC reproduction against a real multi-database setup, and no Proxy runtime validation.

Need Expert Review: JDBC/domain review is sufficient after a fix.

Positive Feedback

The overall direction is aligned with [issue 37757](https://github.com/apache/shardingsphere/issues/37757): strict behavior is preserved for non-MySQL paths, while MySQL-compatible typed string conversion is added for getObject(int, Class<T>).
The PR also adds focused tests in ResultSetUtilsTest.java (line 127) and ShardingSphereResultSetTest.java (line 578), which is the right validation direction.
Major Issues

[P1] Protocol detection uses the connection database instead of the executed database. In ShardingSphereResultSet.java (line 406), MySQL-compatible conversion is enabled based on connection.getCurrentDatabaseName(). However, the actual execution target can differ from the connection’s current database:
ShardingSphereStatement.java (line 269) updates usedDatabaseName from the SQL context, and ShardingSpherePreparedStatement.java (line 149) stores the actual usedDatabase.
In those cases, this PR can still read the wrong protocol and either skip MySQL conversion when it should apply, or apply it to a non-MySQL target. That means the root-cause chain is not fully fixed yet, so the PR is not mergeable.
Newly Introduced Issues

The new protocol-selection logic introduces a regression surface for cross-database execution, because it couples result-set conversion to the connection’s current database instead of the database actually used for the query.
The current tests do not cover this boundary case. Under the BCDE and precise-assertion expectations in CODE_OF_CONDUCT.md (line 90) and CODE_OF_CONDUCT.md (line 99), this scenario should have a dedicated regression test.
Next-Step Suggestions

Please source the protocol type from the actual executed database for the result set, not from connection.getCurrentDatabaseName().
Please add a regression test for Statement where the SQL target database differs from the connection’s current database.
Please add the same kind of regression test for PreparedStatement.
After that, I can do another focused review on the protocol-selection path.

terrymanu (changes requested):
Decision

Merge Verdict: Not Mergeable

Reviewed Scope: latest head [5d90bef](https://github.com/apache/shardingsphere/commit/5d90befc9c594f4e2b48f3d731c3279f28f36597) of [PR 38444](https://github.com/apache/shardingsphere/pull/38444), all 6 changed files in infra/executor and jdbc, the maintainer design note on [issue 37757](https://github.com/apache/shardingsphere/issues/37757#issuecomment-3769144014), and the prior blocking review on [PR 38444](https://github.com/apache/shardingsphere/pull/38444#pullrequestreview-3950088395).

Not Reviewed Scope: no live JDBC reproduction against real MySQL/MariaDB/Doris, no Proxy runtime validation, and no full ./mvnw clean install -B -T1C -Pcheck.

Need Expert Review: JDBC/infra maintainer review is needed after the architectural path is adjusted; no separate security review is indicated.

The overall direction is correct, especially because the blocker from the previous round has been addressed: this version no longer determines the protocol based on connection.getCurrentDatabaseName(), and it adds regression tests for both Statement and PreparedStatement in [ShardingSphereResultSetTest](https://github.com/apache/shardingsphere/blob/5d90befc9c594f4e2b48f3d731c3279f28f36597/jdbc/src/test/java/org/apache/shardingsphere/driver/jdbc/core/resultset/ShardingSphereResultSetTest.java#L587-L627).

Major Issues

[P1] [ResultSetUtils.java](https://github.com/apache/shardingsphere/blob/5d90befc9c594f4e2b48f3d731c3279f28f36597/infra/executor/src/main/java/org/apache/shardingsphere/infra/executor/sql/execute/result/query/impl/driver/jdbc/type/util/ResultSetUtils.java#L68-L199) now models dialect compatibility as a new public boolean switch instead of the DatabaseType-driven conversion boundary discussed in the issue. This exposes vendor-specific semantics in a shared utility API and does not align well with the consistency, simplicity, and abstraction-level requirements reflected in CODE_OF_CONDUCT.md (line 10), CODE_OF_CONDUCT.md (line 11), and CODE_OF_CONDUCT.md (line 12). Before merge, this should be changed to a DatabaseType-driven conversion abstraction rather than a public isMySQLCompatibleConversion flag.
[P2] [ResultSetUtilsTest.java](https://github.com/apache/shardingsphere/blob/5d90befc9c594f4e2b48f3d731c3279f28f36597/infra/executor/src/test/java/org/apache/shardingsphere/infra/executor/sql/execute/result/query/impl/driver/jdbc/type/util/ResultSetUtilsTest.java#L127-L145) only covers successful Integer, successful Boolean, and one invalid Integer case, but the added branches actually include Boolean, Byte, Short, Integer, Long, Float, Double, BigDecimal, and empty/invalid input. For a shared utility change, that coverage is still too thin and does not adequately satisfy the coverage expectations reflected in CODE_OF_CONDUCT.md (line 20), CODE_OF_CONDUCT.md (line 90), CODE_OF_CONDUCT.md (line 95), and CODE_OF_CONDUCT.md (line 99).
Newly Introduced Issues

[ShardingSphereResultSet.java](https://github.com/apache/shardingsphere/blob/5d90befc9c594f4e2b48f3d731c3279f28f36597/jdbc/src/main/java/org/apache/shardingsphere/driver/jdbc/core/resultset/ShardingSphereResultSet.java#L378-L417) places protocol detection inside the typed getObject hot path of a class annotated with @HighFrequencyInvocation, and each call walks back through the statement into metadata to re-check the protocol. That increases overhead on a hot path and tightens coupling between result-set logic and concrete statement implementations. This does not align well with the performance constraints around CODE_OF_CONDUCT.md (line 70). The executed database protocol should be carried down or cached during execution/merge rather than re-derived on every typed access.
Next Steps

Refactor the dialect selection into a DatabaseType-driven conversion abstraction instead of exposing MySQL-compatible semantics through a public boolean flag.
Make ShardingSphereResultSet reuse already-known protocol information rather than looking metadata up on each typed getObject call.
Add dedicated tests for the remaining numeric branches and the empty-string failure path, then proceed with another focused review.
Multi-round Comparison

Fixed: the previous review’s blocker, using the connection’s current database rather than the actual executed database, has been fixed in the latest head.
Unresolved: dialect compatibility is still exposed as a MySQL boolean switch on the shared utility rather than a DatabaseType-driven abstraction.
New: there is now hot-path protocol lookup/coupling in ShardingSphereResultSet, along with one-to-one test gaps for the newly added conversion branches.

terrymanu (changes requested):
Decision
Merge Verdict: Not Mergeable

Reviewed Scope: latest PR head 9acc335, all 6 changed files, plus issue #37757, prior review context, and the relevant MySQL Connector/J native conversion sources.

Not Reviewed Scope: no live MySQL/MariaDB/Doris server reproduction, no Proxy runtime validation, and the scoped Maven run was still in progress when the semantic blockers were already confirmed.

Need Expert Review: JDBC/domain review is sufficient after a fix.

Positive Feedback

The previous blocker is fixed in the right place: protocol selection now follows the executed database instead of the connection’s current database via ShardingSphereResultSet.java (line 409).
The non-MySQL unchanged path is still covered, and the review round did re-check latest head instead of reusing the older conclusion.
Major Issues

[P1] The new MySQL conversion is not actually MySQL-compatible. In ResultSetUtils.java (line 156), integer targets use BigDecimal#...ValueExact() and empty strings always throw. Native Connector/J instead accepts empty strings as zero/false by default and routes decimal/exponent strings through floating-point parsing before converting integral targets. That means cases like "" and "123.45" still do not match the MySQL behavior this PR says it is emulating, so the root cause is only partially repaired.
[P1] The added tests lock in the incompatible behavior. ResultSetUtilsTest.java (line 195) explicitly expects empty-string-to-integer conversion to fail under MySQL protocol, which is opposite to Connector/J’s default emptyStringsConvertToZero path. Under CODE_OF_CONDUCT.md:20 and CODE_OF_CONDUCT.md:90-99, this is not just a coverage gap; it is a regression oracle pointing at the wrong contract.
Newly Introduced Issues

The blast radius is broader than MySQL. ResultSetUtils.java (line 134) also enables this path for trunk-MySQL dialects such as MariaDB, Doris, and H2, but the PR only validates MySQL and PostgreSQL cases. I locally checked H2 as a counterexample path and its typed-conversion behavior is not simply “treat everything like MySQL,” so this shared-path expansion still needs explicit compatibility evidence.
Next Steps

Please align convertMySQLStringValue with Connector/J’s conversion matrix instead of hand-rolling an exact-value subset. At minimum, cover empty-string defaults and decimal/exponent input for integral targets.
Please add one-to-one regression tests for "", "123.45", and primitive targets such as int.class / boolean.class on the MySQL protocol path.
Please either limit the new compatibility branch to true MySQL protocol only, or add explicit compatibility rationale and regression tests for MariaDB/Doris/H2 before keeping trunk-database opt-in.
Multi-Round Comparison

Fixed from previous round: protocol detection now uses the executed database instead of the connection’s current database.
Still unresolved/new on latest head: the actual conversion matrix still does not match native MySQL for empty strings and non-integral numeric strings, and the trunk-dialect blast radius is still unvalidated.

terrymanu (changes requested):
The latest head is still not mergeable. The problem is no longer the earlier protocol selection issue, but that the claimed "MySQL-compatible typed conversion" in the current version still does not match the actual behavior of the official Connector/J.

Decision

Merge Verdict: Not Mergeable
Reviewed Scope: The 6 changed files in the latest PR head 490d169b5dbc (2026-03-22), the context of issue #37757 and its maintainer design/review comments, the execution path in ShardingSphereStatement / ShardingSpherePreparedStatement, and the corresponding typed getObject source code in the official MySQL Connector/J.
Not Reviewed Scope: No real MySQL 8.x JDBC reproduction was performed, no Proxy runtime validation was performed, and no local Maven run was executed on the PR worktree.
Need Expert Review: A JDBC / infra maintainer re-review is sufficient; no separate security review is needed at this point.
There are several things in this revision that are in the right direction: ShardingSphereResultSet has already been changed to use the protocol of the actual usedDatabase, and it caches protocolType inside the result set; at the same time, the latest head has also narrowed the blast radius back to type == "MySQL", instead of extending it to trunk dialects. GitHub checks on 490d169b5dbc are also all green.

Major Issues

[P1] The current implementation compresses MySQL typed conversion into a unified branch of "trim + string parsing", but the official Connector/J actually dispatches by target type to different ValueFactory implementations, so this revision is still not natively MySQL-compatible.
In ResultSetUtils.java:156-246, everything now goes through convertMySQLStringValue(...); however, the official ResultSetImpl#getObject(int, Class<T>) dispatches separately to booleanValueFactory, byteValueFactory, shortValueFactory, integerValueFactory, and so on.
This already has several concrete counterexamples:
Byte.class: the PR converts "127" to 127 in ResultSetUtils.java:166-168, and locks that behavior in ResultSetUtilsTest.java:151-153; but the official ByteValueFactory#createFromBytes does not numerically parse multi-character strings in that way.
Boolean.class: in ResultSetUtils.java:232-246, the PR only accepts 1/-1/0 and a small set of textual literals; the official BooleanValueFactory#createFromBytes also treats numeric strings as numeric booleans, so adjacent valid inputs such as "2", "0.1", and "1e3" would produce different results.
Whitespace handling: the PR first applies trim() in ResultSetUtils.java:157, and asserts in ResultSetUtilsTest.java:161-163 that " 123 " can be converted to 123; the official AbstractNumericValueFactory#createFromBytes does not trim, which is another explicit behavioral mismatch.
This shows that although the PR reaches the main symptom path from the issue, it still does not truly complete the "native MySQL compatibility" goal.
Newly Introduced Issues

The latest commit adds/adjusts tests that lock in the wrong contract, especially the assertions in ResultSetUtilsTest.java:151-163 for Byte and integer strings with surrounding whitespace. According to CODE_OF_CONDUCT.md:20,90-99, tests should precisely express the target contract; these assertions currently preserve behavior that is not native MySQL behavior.
Conversely, the counterexamples that actually determine mergeability are still not covered: at minimum, the MySQL-path behavior should be added for "2" / "0.1" / "1e3" under Boolean.class, and for multi-character strings under Byte.class.
Next-Step Suggestions

Please keep the parts that were fixed correctly in this round: usedDatabase protocol propagation, caching protocolType inside the result set, and MySQL-only gating.
Please change MySQL typed conversion to align with Connector/J's dispatch-by-target-type semantics, instead of continuing to expand a unified string parser.
If the intent is to keep the current more permissive behavior, please first make it explicit in issue #37757 / the PR that the target is changing from "native MySQL compatibility" to a "ShardingSphere-defined compatibility subset"; otherwise, this will continue to conflict with the PR title and test target.
Please also roll back or rewrite the incorrect oracle in ResultSetUtilsTest, and then add the counterexample tests above before proceeding to the next review round.
Multi-Round Comparison

Closed: the issue of mixing current database and used database, the repeated protocol lookup on the hot path, and the expansion to trunk dialects such as MariaDB/Doris/H2.
Still unresolved in the current head: the latest revision still does not implement MySQL-compatible typed conversion according to the official Connector/J's dispatch-by-type semantics; there are still explicit mismatches in Boolean, Byte, and whitespace string handling.

terrymanu:
Please fix spotless

terrymanu (changes requested):
### Decision

- **Reviewed Scope:** Latest PR head `0e86fbfb24c59ef176a17d24302c89492bc79133`, local merge-base `5f39c14e3f88a579d14a03e95fa53b45a6e60595`; GitHub `/pulls/38444/files` matched the local triple-dot file list. Reviewed 7 changed files: `.github/workflows/e2e-sql.yml`, `infra/executor/.../ResultSetUtils.java`, `infra/executor/.../ResultSetUtilsTest.java`, `jdbc/.../ShardingSphereResultSet.java`, `jdbc/.../ShardingSpherePreparedStatement.java`, `jdbc/.../ShardingSphereStatement.java`, and `jdbc/.../ShardingSphereResultSetTest.java`. Also reviewed linked issue `#37757`, prior GitHub-visible change requests, and official Connector/J docs/source for typed conversion behavior.
- **Not Reviewed Scope:** No GitHub Actions/check-run status was used. No live MySQL/MariaDB/Doris JDBC reproduction, no Proxy runtime validation, and no local Maven run on a PR checkout.
- **Need Expert Review:** JDBC/infra maintainer re-review is needed after the conversion and scope issues are fixed. CI/workflow owner review is needed only if the workflow change is intentionally split into a separate CI PR.

### Positive Feedback

- The latest version fixes several earlier blockers: protocol selection now uses the executed database path, protocol type is cached in `ShardingSphereResultSet`, non-MySQL behavior remains strict, and the tests now cover empty strings, decimal/exponent integer input, boolean numeric strings, whitespace rejection, and used-database protocol selection.
- The PR also narrows activation to `protocolType.getType() == "MySQL"`, which avoids the earlier trunk-dialect blast radius for MariaDB/Doris/H2.

### Major Issues

- **[P1] MySQL typed string conversion still differs from Connector/J for adjacent numeric inputs** (`infra/executor/src/main/java/org/apache/shardingsphere/infra/executor/sql/execute/result/query/impl/driver/jdbc/type/util/ResultSetUtils.java:202`)
  - **Symptom:** For integer-like targets, `convertMySQLIntegerValue` falls back to `new BigDecimal(value)` when the string has no dot or exponent. That accepts plain leading-plus inputs such as `"+123"`; for decimal strings such as `"+123.45"`, the PR also accepts them through `Double.parseDouble(...)`.
  - **Risk:** Connector/J's numeric byte parsing path only treats strings as numeric when they match its integer/floating patterns or exponent path; plain leading-plus integer/decimal strings do not match those patterns. Because this PR claims MySQL-compatible conversion, accepting values that Connector/J rejects changes the contract and can hide bad data instead of surfacing the same conversion failure.
  - **Action:** Please align the parser with Connector/J's `AbstractNumericValueFactory#createFromBytes` behavior and add direct MySQL-path tests for leading-plus integer and decimal strings, for example `"+123"` and `"+123.45"` with `Integer.class` / `BigDecimal.class`.

- **[P1] `Byte.class` conversion checks Java character length instead of encoded byte length** (`infra/executor/src/main/java/org/apache/shardingsphere/infra/executor/sql/execute/result/query/impl/driver/jdbc/type/util/ResultSetUtils.java:240`)
  - **Symptom:** `convertMySQLByteValue` allows any one-Java-character string and casts `value.charAt(0)` to `byte`.
  - **Risk:** Connector/J's `ByteValueFactory#createFromBytes` checks the encoded byte array length and only returns the first byte when the byte length is exactly one. A single multi-byte character would be converted by this PR but rejected by Connector/J, so the Byte branch still locks in a non-native oracle.
  - **Action:** Please base this branch on encoded byte length, not `String.length()`, and add a regression test for a single non-ASCII/multi-byte character converted to `Byte.class`.

- **[P2] MySQL-specific semantics still live inside a shared infra utility** (`infra/executor/src/main/java/org/apache/shardingsphere/infra/executor/sql/execute/result/query/impl/driver/jdbc/type/util/ResultSetUtils.java:140`)
  - **Symptom:** The shared `infra/executor` utility now contains the hard-coded `"MySQL"` check plus `convertMySQL...` helper methods.
  - **Risk:** The generic hook is useful, but the target-dialect conversion matrix is now owned by shared result-set utility code. That makes future dialect additions harder to review and conflicts with the issue's original direction of a dialect-aware converter boundary.
  - **Action:** Please move the MySQL conversion matrix behind a dialect-owned converter/SPI boundary, or document why this shared utility is the intentional owner and add generic shared-path tests proving the disabled/non-target path remains stable.

### Unrelated Changes

- **Please roll back the workflow change from this PR** (`.github/workflows/e2e-sql.yml:37`)
  - **Symptom:** The latest head adds `TESTCONTAINERS_RYUK_DISABLED: true` to the global E2E SQL workflow environment.
  - **Risk:** Disabling Testcontainers Ryuk changes resource-cleanup behavior for the whole E2E SQL workflow and is not related to JDBC `getObject(int, Class<T>)` typed string conversion. This is a substantive scope expansion, not import/format churn.
  - **Action:** Please remove this workflow change from the JDBC conversion PR. If it is needed, it should be submitted separately with CI-specific rationale and ownership review.

### Next Steps

- Please remove or split the `.github/workflows/e2e-sql.yml` change.
- Please align `ResultSetUtils` with Connector/J's numeric and byte conversion dispatch for the counterexamples above.
- Please add one-to-one regression tests for leading-plus numeric strings and single multi-byte-character `Byte.class` conversion.
- Please revisit the shared-layer ownership of MySQL-specific conversion logic.
- After the code changes, please run focused verification such as `./mvnw -pl infra/executor,jdbc -am -DskipITs -Dspotless.skip=true -Dtest=ResultSetUtilsTest,ShardingSphereResultSetTest -Dsurefire.failIfNoSpecifiedTests=false test`, then the repository Spotless/Checkstyle gates.

### Multi-Round Comparison

- **Fixed:** Previous current-database vs used-database protocol selection blocker.
- **Fixed:** Previous hot-path protocol lookup issue; protocol type is now cached in the result set.
- **Fixed:** Previous broad trunk-dialect activation risk; latest code gates on exact MySQL protocol type.
- **Partially fixed:** The MySQL conversion matrix now covers several earlier counterexamples, but still diverges from Connector/J for leading-plus numeric strings and multi-byte `Byte.class` inputs.
- **Newly introduced:** The E2E SQL workflow change is unrelated to the PR goal and should be rolled back or split.

terrymanu (approved):
### Summary

- **Reason:** The latest head fixes the linked JDBC/MySQL typed string conversion path, keeps non-MySQL behavior strict, and the scoped PR-head tests plus Spotless/Checkstyle gates passed.

### Evidence

- The linked issue `#37757` asks for ShardingSphere JDBC to handle legacy MySQL `CHAR`/`VARCHAR` numeric values when callers use `ResultSet#getObject(int, Class<T>)`, especially `Integer.class`.
- The root path now reaches a dialect converter from `ShardingSphereResultSet#getObject(int, Class<T>)` via the executed database protocol, not the connection's current database (`jdbc/src/main/java/org/apache/shardingsphere/driver/jdbc/core/resultset/ShardingSphereResultSet.java:385`, `jdbc/src/main/java/org/apache/shardingsphere/driver/jdbc/core/resultset/ShardingSphereResultSet.java:412`).
- `ShardingSphereStatement` and `ShardingSpherePreparedStatement` expose the actual used database selected during bind/prepare, which covers cross-database execution (`jdbc/src/main/java/org/apache/shardingsphere/driver/jdbc/core/statement/ShardingSphereStatement.java:270`, `jdbc/src/main/java/org/apache/shardingsphere/driver/jdbc/core/statement/ShardingSpherePreparedStatement.java:148`).
- The MySQL-specific behavior is isolated behind `DialectResultSetValueConverter` and the JDBC SPI service file, while shared `ResultSetUtils` keeps strict default conversion; `ResultSetUtilsTest` verifies `String -> Integer` still fails outside the dialect path (`infra/executor/src/test/java/org/apache/shardingsphere/infra/executor/sql/execute/result/query/impl/driver/jdbc/type/util/ResultSetUtilsTest.java:129`).
- The converter's target-type dispatch was checked against official Connector/J source: `ResultSetImpl#getObject(int, Class<T>)`, `AbstractNumericValueFactory#createFromBytes`, `ByteValueFactory#createFromBytes`, and `BooleanValueFactory#createFromBytes`.
  - https://github.com/mysql/mysql-connector-j/blob/release/9.x/src/main/user-impl/java/com/mysql/cj/jdbc/result/ResultSetImpl.java#L1273-L1328
  - https://github.com/mysql/mysql-connector-j/blob/release/9.x/src/main/core-impl/java/com/mysql/cj/result/AbstractNumericValueFactory.java#L47-L67
  - https://github.com/mysql/mysql-connector-j/blob/release/9.x/src/main/core-impl/java/com/mysql/cj/result/ByteValueFactory.java#L92-L103
  - https://github.com/mysql/mysql-connector-j/blob/release/9.x/src/main/core-impl/java/com/mysql/cj/result/BooleanValueFactory.java#L83-L107
- `MySQLResultSetValueConverterTest` covers string-to-boolean, byte, short, integer, long, float, double, BigDecimal, empty string defaults, decimal/exponent integral conversion, invalid input, leading-plus rejection, and multi-byte byte rejection (`jdbc/src/test/java/org/apache/shardingsphere/driver/jdbc/core/resultset/MySQLResultSetValueConverterTest.java:43`).
- `ShardingSphereResultSetTest` covers MySQL success, PostgreSQL disabled behavior, MariaDB trunk-database disabled behavior, Statement used-database protocol selection, and PreparedStatement used-database protocol selection (`jdbc/src/test/java/org/apache/shardingsphere/driver/jdbc/core/resultset/ShardingSphereResultSetTest.java:595`).
- Fresh risk scan found no dependency, parser, SQL syntax, CI workflow, config, or unrelated scope changes in the latest GitHub file list. No direct `ConcurrentHashMap#computeIfAbsent` was introduced on the JDBC DML/DQL path.

### Multi-Round Comparison

- **Fixed:** The previous current-database vs used-database protocol-selection blocker is covered by both Statement and PreparedStatement tests.
- **Fixed:** The previous hot-path repeated metadata lookup risk is avoided by selecting the converter once during result-set construction.
- **Fixed:** The previous broad trunk-dialect activation risk is closed; MariaDB with MySQL trunk remains strict unless its own converter is registered.
- **Fixed:** The previous Connector/J conversion counterexamples for empty strings, decimal/exponent strings, boolean numeric strings, leading-plus strings, and multi-byte byte inputs are now covered by the converter tests.
- **Fixed:** The previous unrelated E2E workflow change is no longer present in the latest PR file list.

### Review Details

- **Reviewed Scope:** Latest PR head `fbd60fe429791d9a1182faa968e48e54228b7a9c`; local merge-base `5f39c14e3f88a579d14a03e95fa53b45a6e60595`; local triple-dot file list matched GitHub `/pulls/38444/files`. Reviewed all 10 changed files in `infra/executor` and `jdbc`, linked issue `#37757`, GitHub-visible prior review rounds, relevant JDBC result-set execution context, and official Connector/J source for the conversion semantics.
- **Not Reviewed Scope:** No GitHub Actions/check-run status was used. No live MySQL server reproduction, no Proxy runtime validation, and no full `./mvnw clean install -B -T1C -Pcheck` run were performed.
- **Verification:** On a detached PR-head worktree: `./mvnw -pl infra/executor,jdbc -am -DskipITs -Dspotless.skip=true -Dtest=ResultSetUtilsTest,MySQLResultSetValueConverterTest,ShardingSphereResultSetTest -Dsurefire.failIfNoSpecifiedTests=false test` exited `0` (`ResultSetUtilsTest`: 40 tests, `MySQLResultSetValueConverterTest`: 26 tests, `ShardingSphereResultSetTest`: 87 tests). `./mvnw -pl infra/executor,jdbc -am -DskipTests spotless:check checkstyle:check -Pcheck -T1C` exited `0`.

PR #38566, opened 2026-03-27
Fix PostgreSQL proxy timestamp text output adding trailing .0

### LINKED ISSUE #37437: PostgreSQL proxy: timestamp output adds trailing ".0" compared to native PostgreSQL

shardingsphere-proxy 5.5.2
postgres 17.6 as backend

### Expected behavior

```sql
DROP TABLE IF EXISTS t1;
CREATE TABLE t1 (c8 TIMESTAMP);

INSERT INTO t1 VALUES ('1973-06-03 10:30:01'); -- no fractional seconds
SELECT c8 FROM t1;
```
should return (in native PG):
```
1973-06-03 10:30:01
```

### Actual behavior
```sql
testdb=> CREATE TABLE t100 (c8 TIMESTAMP);
CREATE TABLE
testdb=> INSERT INTO t100 VALUES ('1973-06-03 10:30:01');
INSERT 0 1
testdb=> SELECT c8 FROM t100 ;
          c8           
-----------------------
 1973-06-03 10:30:01.0
(1 row)
```

The semantic value seems identical, but the textual representation differs and may break clients/tests relying on stable output formatting.

We guess the bug is related to `Timestamp.toString()` in `java.sql.Timestamp`.

[ss_postgres.zip](https://github.com/user-attachments/files/24253131/ss_postgres.zip)

run `docker compose up -d` in this file.

SQL to execute are shown above

daguimu:
## Problem

When querying a `TIMESTAMP` column without fractional seconds through ShardingSphere-Proxy (PostgreSQL protocol), the result includes a trailing `.0` (e.g., `1973-06-03 10:30:01.0`), whereas native PostgreSQL returns `1973-06-03 10:30:01`. This formatting mismatch can break clients and tests relying on stable textual output.

## Root Cause

In `PostgreSQLDataRowPacket#writeTextValue`, non-binary column values fall through to the generic `each.toString()` branch. When the value is a `java.sql.Timestamp`, `Timestamp.toString()` always emits a fractional part (at minimum `.0`), which does not match PostgreSQL's text protocol behavior of omitting trailing zero fractional seconds.

## Fix

- Added a type-aware branch for `java.sql.Timestamp` in `writeTextValue` that formats the timestamp according to PostgreSQL text output rules:
  - Omits the fractional part entirely when there are no fractional seconds
  - Uses microsecond precision (matching PostgreSQL's internal precision) and strips trailing zeros from the fractional part
- The formatting logic is encapsulated in a private `formatTimestamp` method within `PostgreSQLDataRowPacket`

## Tests Added

| Change Point | Test |
|-------------|------|
| Timestamp without fractional seconds omits `.0` | `timestamp_no_fractional` — verifies `1973-06-03 10:30:01` (exact reproduction from issue) |
| Timestamp with fractional seconds preserved | `timestamp_with_fractional` — verifies `2024-01-15 14:30:45.123` |
| Trailing zeros in fractional part stripped | `timestamp_trailing_zeros_stripped` — verifies `.120` becomes `.12` |
| Full microsecond precision supported | `timestamp_microsecond_precision` — verifies `2024-01-15 14:30:45.123456` |

## Impact

Only affects PostgreSQL text protocol timestamp formatting in the proxy. No changes to binary protocol, other data types, or any other module. Existing behavior for all non-Timestamp types is unchanged.

Fixes #37437

### DIFF
--- a/database/protocol/dialect/postgresql/src/main/java/org/apache/shardingsphere/database/protocol/postgresql/packet/command/query/PostgreSQLDataRowPacket.java
+++ b/database/protocol/dialect/postgresql/src/main/java/org/apache/shardingsphere/database/protocol/postgresql/packet/command/query/PostgreSQLDataRowPacket.java
@@ -30,6 +30,8 @@ import org.apache.shardingsphere.database.protocol.postgresql.payload.PostgreSQL
 import java.nio.charset.StandardCharsets;
 import java.sql.SQLException;
 import java.sql.SQLXML;
+import java.sql.Timestamp;
+import java.time.format.DateTimeFormatter;
 import java.util.Collection;
 
 /**
@@ -40,6 +42,8 @@ import java.util.Collection;
 public final class PostgreSQLDataRowPacket extends PostgreSQLIdentifierPacket {
     
     private static final byte[] HEX_DIGITS = "0123456789abcdef".getBytes(StandardCharsets.US_ASCII);
+
+    private static final DateTimeFormatter TIMESTAMP_FORMATTER = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss");
     
     private final Collection<Object> data;
     
@@ -79,6 +83,10 @@ public final class PostgreSQLDataRowPacket extends PostgreSQLIdentifierPacket {
             byte[] columnData = ((Boolean) each ? "t" : "f").getBytes(payload.getCharset());
             payload.writeInt4(columnData.length);
             payload.writeBytes(columnData);
+        } else if (each instanceof Timestamp) {
+            byte[] columnData = formatTimestamp((Timestamp) each).getBytes(payload.getCharset());
+            payload.writeInt4(columnData.length);
+            payload.writeBytes(columnData);
         } else {
             byte[] columnData = each.toString().getBytes(payload.getCharset());
             payload.writeInt4(columnData.length);
@@ -86,6 +94,22 @@ public final class PostgreSQLDataRowPacket extends PostgreSQLIdentifierPacket {
         }
     }
     
+    private String formatTimestamp(final Timestamp timestamp) {
+        String baseTime = timestamp.toLocalDateTime().format(TIMESTAMP_FORMATTER);
+        int micros = timestamp.getNanos() / 1000;
+        if (0 == micros) {
+            return baseTime;
+        }
+        StringBuilder result = new StringBuilder(baseTime).append('.');
+        result.append(String.format("%06d", micros));
+        int length = result.length();
+        while (result.charAt(length - 1) == '0') {
+            length--;
+        }
+        result.setLength(length);
+        return result.toString();
+    }
+
     private byte[] encodeByteaText(final byte[] value) {
         byte[] result = new byte[value.length * 2 + 2];
         result[0] = '\\';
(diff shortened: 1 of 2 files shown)


daguimu:
Hi, I have just addressed the Spotless violations in a follow-up commit; CI should now pass. Would appreciate a review when someone has time. Thanks!

terrymanu (changes requested):
### Decision

- **Reviewed Scope:** PR #38566 latest head `9a703ce57a7d19168a1e30a8f1dd45ad1a277744`; local merge-base `605db75e72b290d5136c8f8658b91f6ff84c05a8`; GitHub `/pulls/38566/files` matched the local triple-dot diff exactly. Reviewed `database/protocol/dialect/postgresql/.../PostgreSQLDataRowPacket.java`, `database/protocol/dialect/postgresql/.../PostgreSQLDataRowPacketTest.java`, PostgreSQL simple/extended text call sites, and the openGauss simple-query call site that also uses `PostgreSQLDataRowPacket`.
- **Not Reviewed Scope:** Maven tests and Proxy integration smoke were not run locally; CI/check-runs were intentionally not used. No SQL parser changes are involved.
- **Need Expert Review:** PostgreSQL/openGauss protocol maintainer re-check is useful after the locale issue is fixed; no security or dependency review is required.

### Positive Feedback

- The change targets the right root-cause area: PostgreSQL text row encoding previously fell through to `Timestamp.toString()`, which explains the reported trailing `.0`.
- The added tests cover the reported no-fraction case, fractional seconds, trailing-zero trimming, and microsecond precision.
- The PR scope is small and the GitHub file list shows no unrelated files.

### Major Issues

- **[P1] Timestamp fractional output is locale-sensitive** (`database/protocol/dialect/postgresql/src/main/java/org/apache/shardingsphere/database/protocol/postgresql/packet/command/query/PostgreSQLDataRowPacket.java:104`)
  - **Symptom:** `formatTimestamp` pads microseconds with `String.format("%06d", micros)`, which uses the JVM default locale when no locale is supplied. Java `Formatter` documents that numeric conversions localize digits under the active locale: https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/Formatter.html
  - **Risk:** PostgreSQL wire text becomes locale-dependent for fractional timestamps. With `Locale.forLanguageTag("th-TH-u-nu-thai")`, the new logic emits fractional digits with code points `U+0E51..U+0E56` instead of ASCII `1..6`, while PostgreSQL ISO timestamp output is expected to be stable text. This affects PostgreSQL simple query, PostgreSQL extended text results, and openGauss simple query because they all route through `PostgreSQLDataRowPacket`.
  - **Action:** Please make the microsecond padding locale-independent, for example with `String.format(Locale.ROOT, "%06d", micros)` or manual ASCII zero-padding, and add a regression test that temporarily sets/restores a non-Latin default locale and verifies ASCII timestamp output.

### Next Steps

- Please fix the locale-sensitive padding in `formatTimestamp`.
- Please add one focused regression test for a non-Latin default locale and restore the original locale in `finally` or an equivalent cleanup path.
- After the fix, please run the scoped module test with dependencies, for example: `./mvnw -pl database/protocol/dialect/postgresql -am -DskipITs -Dspotless.skip=true -Dtest=PostgreSQLDataRowPacketTest -Dsurefire.failIfNoSpecifiedTests=false test`.

terrymanu:
Hi @daguimu, thanks again for working on this PostgreSQL protocol issue.

The fix direction looks reasonable because it targets the `Timestamp.toString()` fallback in `PostgreSQLDataRowPacket`, which matches the root cause of #37437. However, this PR is still not ready to merge because the timestamp fractional padding is locale-sensitive, as described in the previous review, and there has been no response or update for a long time after the change request.

Could you please update the PR to make the padding locale-independent, such as by using `Locale.ROOT` or manual ASCII zero-padding, and add the requested regression test for a non-Latin default locale? If there is still no response or update, we may close this PR later to keep the review queue clear.

terrymanu:
Thanks for your contribution. The locale-sensitive timestamp padding issue raised in the previous review remains unresolved, and there has been no response or update after the change request and the follow-up notice. We are therefore closing this PR to keep the review queue clear.
The overall fix direction still looks reasonable. Please feel free to reopen this PR or submit a new one after making the timestamp padding locale-independent and adding the requested regression test for a non-Latin default locale. The linked issue #37437 will remain open.

PR #38569, opened 2026-03-27
Fix test parameter parsing treating leading-zero strings as octal

### LINKED ISSUE #31001: SQL-rewrite-test phone number createInputParameter error
## Bug Report

master

ALL

### Expected behavior
SQL rewrite test ,insert phone number shoud regarded as a String;

### Actual behavior
SQL rewrite test ,insert phone number now regarded as a number;

I am working for issue 30227 insert SQL. 

- I want to insert data, a virtual phone number '15755557777';
- This data is encrypted by AES with the 123456abc key to '04844448888';
- When insert use mode 'with_columns_for_parameters',data '04844448888' is passed as an argument to method `SQLRewriteEngineTestParametersBuilder.createInputParameter`.
- In modthod,'04844448888' is treated as a number.
image

- Because it is too long and begin with '0','04844448888'is regarded as a base-8 number.
- But it incloud number 8,exception throw from this: `java.lang.NumberFormatException: For input string: "4844448888" under radix 8`
- 
image

-https://github.com/zzyReal666/shardingsphere.git 
-branch : issue-30227

daguimu:
## Problem

`SQLRewriteEngineTestParametersBuilder.createInputParameter()` uses `NumberUtils.createNumber()` to parse test parameter strings. When a parameter string consists entirely of digits but starts with '0' (e.g., encrypted phone number `04844448888`), `NumberUtils.createNumber()` interprets it as an octal literal per Java conventions, throwing `NumberFormatException` since digits 8 and 9 are invalid in octal.

Even for valid octal-looking strings like `04100001111` (digits 0-7 only), the value is silently parsed as octal instead of decimal, producing a wrong numeric value.

## Root Cause

`StringUtils.isNumeric("04844448888")` returns `true` (all characters are digits), but `NumberUtils.createNumber("04844448888")` follows Java number literal conventions where leading `0` means octal. SQL test parameters should always use decimal interpretation — SQL has no octal number syntax.

## Fix

Replaced `NumberUtils.createNumber()` with `Integer.valueOf()` / `Long.valueOf()` which always parse as decimal:
- Try `Integer.valueOf()` first (preserves existing Integer return type for small values)
- Fall back to `Long.valueOf()` for values exceeding Integer range
- Fall back to returning the original string if both overflow

Removed the unused `NumberUtils` import.

## Tests Added

| Change Point | Test |
|-------------|------|
| Regular integer parsed as decimal Integer | `assertCreateInputParameterWithRegularInteger` — `"1"` → `Integer(1)` |
| Large integer within int range | `assertCreateInputParameterWithLargeNumber` — `"1000"` → `Integer(1000)` |
| Leading-zero string that was crashing (octal with invalid digits) | `assertCreateInputParameterWithLeadingZeroDigits` — `"04844448888"` → `Long(4844448888)` |
| Leading-zero string that was silently mis-parsed as octal | `assertCreateInputParameterWithLeadingZeroValidOctalLikeDigits` — `"04100001111"` → `Long(4100001111)` |
| Single zero | `assertCreateInputParameterWithSingleZero` — `"0"` → `Integer(0)` |
| Non-numeric string | `assertCreateInputParameterWithString` — `"aaa"` → `String("aaa")` |
| NULL literal | `assertCreateInputParameterWithNull` — `"NULL"` → `null` |
| Long overflow fallback | `assertCreateInputParameterWithLongOverflow` — returns String |

## Impact

Only affects the SQL rewrite test parameter builder. No runtime code changes. All existing test parameters (which have no leading-zero numeric values) produce identical results.

Fixes #31001

### DIFF
--- a/test/it/rewriter/src/test/java/org/apache/shardingsphere/test/it/rewriter/engine/parameter/SQLRewriteEngineTestParametersBuilder.java
+++ b/test/it/rewriter/src/test/java/org/apache/shardingsphere/test/it/rewriter/engine/parameter/SQLRewriteEngineTestParametersBuilder.java
@@ -22,7 +22,6 @@ import com.google.common.base.Strings;
 import lombok.AccessLevel;
 import lombok.NoArgsConstructor;
 import org.apache.commons.lang3.StringUtils;
-import org.apache.commons.lang3.math.NumberUtils;
 import org.apache.shardingsphere.test.it.rewriter.engine.type.SQLExecuteType;
 import org.apache.shardingsphere.test.it.rewriter.entity.RewriteAssertionEntity;
 import org.apache.shardingsphere.test.it.rewriter.entity.RewriteAssertionsRootEntity;
@@ -124,7 +123,14 @@ public final class SQLRewriteEngineTestParametersBuilder {
     
     private static Object createInputParameter(final String inputParam) {
         if (StringUtils.isNumeric(inputParam)) {
-            return NumberUtils.createNumber(inputParam);
+            try {
+                return Integer.valueOf(inputParam);
+            } catch (final NumberFormatException ignored) {
+            }
+            try {
+                return Long.valueOf(inputParam);
+            } catch (final NumberFormatException ignored) {
+            }
         }
         return "NULL".equals(inputParam) ? null : inputParam;
     }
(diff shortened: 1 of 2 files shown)


daguimu:
Hi, I have just addressed both the Spotless and Checkstyle violations in a follow-up commit (switched to `org.hamcrest.Matchers` and replaced `assertInstanceOf` with the whitelisted `isA` matcher). CI should now pass. Would appreciate a review when someone has time. Thanks!

terrymanu (changes requested):
Please fix conflict first

terrymanu:
Hi @daguimu, thanks again for working on this issue.

I am going to close this PR because it has had no response or update for a long time after the request to fix conflicts, and the current approach also needs to be revisited.

For #31001, the expected behavior is not only to avoid octal parsing. We also need to distinguish string-like parameters, such as phone numbers, from numeric parameters based on the test input semantics. The current change parses leading-zero values such as `04844448888` as decimal `Long`, which may not match the issue expectation that phone-number-like values should be treated as strings.

The linked issue remains open, and you are very welcome to submit a new PR later with a rebased branch, a clarified parsing rule, and matching regression tests.

PR #38595, opened 2026-04-07
Fix: Handle MySQL NULL schema behavior

### LINKED ISSUE #28469: MySQL's behavior without current schema
## Question
- When using `shardingsphere-jdbc` on `MySQL`, we can give `jdbcUrl` without schema
  - ex) `jdbc:mysql://localhost:3306/`
  - Then `SELECT DATABASE()` returns null
- In this case, the masking feature is failed with `org.apache.shardingsphere.infra.exception.TableNotExistsException: Table or view 'actor' does not exist.`.
  - Reference: #27879
- Reason: metaDataContext queries to Database with below query. (Maybe generated with `ResultSet java.sql.DatabaseMetaData.getTables()`)
 
   ```sql
    SELECT TABLE_SCHEMA            AS TABLE_CAT,
    NULL                    AS TABLE_SCHEM,
    TABLE_NAME,
    CASE
        WHEN TABLE_TYPE = 'BASE TABLE' THEN CASE
                                                WHEN TABLE_SCHEMA = 'mysql' OR TABLE_SCHEMA = 'performance_schema'
                                                    THEN 'SYSTEM TABLE'
                                                ELSE 'TABLE' END
        WHEN TABLE_TYPE = 'TEMPORARY' THEN 'LOCAL_TEMPORARY'
        ELSE TABLE_TYPE END AS TABLE_TYPE,
    TABLE_COMMENT           AS REMARKS,
    NULL                    AS TYPE_CAT,
    NULL                    AS TYPE_SCHEM,
    NULL                    AS TYPE_NAME,
    NULL                    AS SELF_REFERENCING_COL_NAME,
    NULL                    AS REF_GENERATION
    FROM INFORMATION_SCHEMA.TABLES
    WHERE TABLE_SCHEMA = ''   -- <-- This part makes empty result
    HAVING TABLE_TYPE IN ('TABLE', 'VIEW', 'SYSTEM TABLE', 'SYSTEM VIEW', null)
    ORDER BY TABLE_TYPE, TABLE_SCHEMA, TABLE_NAME;
    ```

Is this behavior is intended? I think we need to fetch all schemas in this case.
Also, even though having current schema, quering to other schema's data is being failure, we need to fetch all schemas also.

AtharvaMa:
Fixes #28469 

  -Changed `MySQLMetaDataLoader.java` so that it works right when `Connection.`getCatalog() gives back null.
  -Added a specific unit test `assertLoadWithNullCatalog` in `MySQLMetaDataLoaderTest.java` to simulate the `null` catalog behavior and verify the metadata loads successfully.
  -Update the logic that incorrectly fell back to an empty string `""` when the catalog was null(null != ""), preventing the subsequent `TableNotExistsException` and NPEs.

---

### DIFF
--- a/database/connector/dialect/mysql/src/main/java/org/apache/shardingsphere/database/connector/mysql/metadata/data/loader/MySQLMetaDataLoader.java
+++ b/database/connector/dialect/mysql/src/main/java/org/apache/shardingsphere/database/connector/mysql/metadata/data/loader/MySQLMetaDataLoader.java
@@ -206,10 +206,10 @@ public final class MySQLMetaDataLoader implements DialectMetaDataLoader {
     }
     
     private String getDatabaseName(final Connection connection, final Collection<String> tableNames) throws SQLException {
-
         String catalog = connection.getCatalog();
-        String databaseName = (catalog==null || catalog.isEmpty()) ?  GlobalDataSourceRegistry.getInstance().getCachedDatabaseTables().get(tableNames.iterator().next())  : catalog;
-        return  databaseName;
+        return (null == catalog || catalog.isEmpty())
+                ? GlobalDataSourceRegistry.getInstance().getCachedDatabaseTables().get(tableNames.iterator().next())
+                : catalog;
     }
     
     @Override
--- a/database/connector/dialect/mysql/src/main/java/org/apache/shardingsphere/database/connector/mysql/metadata/data/loader/MySQLMetaDataLoader.java
+++ b/database/connector/dialect/mysql/src/main/java/org/apache/shardingsphere/database/connector/mysql/metadata/data/loader/MySQLMetaDataLoader.java
@@ -206,7 +206,10 @@ public final class MySQLMetaDataLoader implements DialectMetaDataLoader {
     }
     
     private String getDatabaseName(final Connection connection, final Collection<String> tableNames) throws SQLException {
-        return "".equals(connection.getCatalog()) ? GlobalDataSourceRegistry.getInstance().getCachedDatabaseTables().get(tableNames.iterator().next()) : connection.getCatalog();
+
+        String catalog = connection.getCatalog();
+        String databaseName = (catalog==null || catalog.isEmpty()) ?  GlobalDataSourceRegistry.getInstance().getCachedDatabaseTables().get(tableNames.iterator().next())  : catalog;
+        return  databaseName;
     }
     
     @Override
--- a/database/connector/dialect/mysql/src/test/java/org/apache/shardingsphere/database/connector/mysql/metadata/data/loader/MySQLMetaDataLoaderTest.java
+++ b/database/connector/dialect/mysql/src/test/java/org/apache/shardingsphere/database/connector/mysql/metadata/data/loader/MySQLMetaDataLoaderTest.java
@@ -82,14 +82,14 @@ class MySQLMetaDataLoaderTest {
         assertThat(actual.size(), is(1));
         assertTrue(actual.iterator().next().getTables().isEmpty());
     }
-
+    
     @SuppressWarnings("JDBCResourceOpenedButNotSafelyClosed")
     @Test
-    void assertLoadWithNullCatalog() throws SQLException{
+    void assertLoadWithNullCatalog() throws SQLException {
         DataSource dataSource = mockDataSource();
         ResultSet tableMetaDataResultSet = mockTableMetaDataResultSet();
         when(dataSource.getConnection().prepareStatement(TABLE_METADATA_SQL_WITH_TABLE).executeQuery()).thenReturn(tableMetaDataResultSet);
-
+        
         ResultSet simpleIndexMetaDataResultSet = mockSimpleIndexMetaDataResultSet();
         when(dataSource.getConnection().prepareStatement(INDEX_METADATA_SQL).executeQuery()).thenReturn(simpleIndexMetaDataResultSet);
         when(dataSource.getConnection().prepareStatement(VIEW_METADATA_SQL).executeQuery()).thenReturn(mock(ResultSet.class));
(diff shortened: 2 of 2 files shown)


AtharvaMa:
Hi maintainers, could you please review my PR and trigger the CI/CD checks?

AtharvaMa:
Hi maintainers, There is a styling error which I resolve could you please trigger the CI/CD checks..?
Also, as I'm new to open source, I’d appreciate any feedback or suggestions you might have on the PR

AtharvaMa:
Hi everyone, I’m just checking in on the status of this PR. I’m happy to make any requested changes or address any feedback to help move this forward. Please let me know if there is anything I can do to assist with the review process. Thanks

terrymanu (changes requested):
### Decision

- **Reviewed Scope:** Latest PR head `c0d23757d626c74e760a39f8ac9f3750ce8de0d1`; local merge-base `04cecfc6f980319ad3aa5c8fabce30f9b06ea61f`; GitHub `/pulls/38595/files` matched local `git diff --name-status` with 2 changed files: `database/connector/dialect/mysql/src/main/java/org/apache/shardingsphere/database/connector/mysql/metadata/data/loader/MySQLMetaDataLoader.java` and `database/connector/dialect/mysql/src/test/java/org/apache/shardingsphere/database/connector/mysql/metadata/data/loader/MySQLMetaDataLoaderTest.java`. Supporting paths checked: `infra/common/src/main/java/org/apache/shardingsphere/infra/metadata/database/schema/util/SchemaMetaDataUtils.java`, `infra/common/src/main/java/org/apache/shardingsphere/infra/datanode/DataNodes.java`, and `features/mask/core/src/main/java/org/apache/shardingsphere/mask/rule/MaskRule.java`.
- **Not Reviewed Scope:** No GitHub Actions / CI status, no live MySQL JDBC or Proxy smoke, and no Maven test run. SQL parser family scan is not applicable because no parser files changed.
- **Need Expert Review:** Yes, a MySQL metadata/schema-loading maintainer should re-check the revised production-path behavior. No separate security or dependency review is needed.

### Positive Feedback

- The change is small and targets the right MySQL metadata loader method.
- The new test covers one simple `null` catalog case where the fallback table-to-schema cache is already populated.

### Major Issues

- **[Blocker] Null-catalog root cause is not fixed for the reported masking-only topology** (`database/connector/dialect/mysql/src/main/java/org/apache/shardingsphere/database/connector/mysql/metadata/data/loader/MySQLMetaDataLoader.java:208`)
  - **Symptom:** The new branch falls back to `GlobalDataSourceRegistry.getInstance().getCachedDatabaseTables().get(tableNames.iterator().next())` when `Connection#getCatalog()` is `null` or empty (`MySQLMetaDataLoader.java:209-211`). However, the added test seeds that cache directly (`MySQLMetaDataLoaderTest.java:97-100`) instead of proving that the real schema-building path populates it. In the masking-only scenario from issue #28469, `MaskRule` contributes a `MaskTableMapperRuleAttribute` only (`features/mask/core/src/main/java/org/apache/shardingsphere/mask/rule/MaskRule.java:57`), so `DataNodes#getDataNodes(table)` has no data node schema to cache. `SchemaMetaDataUtils` only writes the fallback cache while iterating data nodes whose data source name contains a schema (`infra/common/src/main/java/org/apache/shardingsphere/infra/metadata/database/schema/util/SchemaMetaDataUtils.java:96-105`).
  - **Risk:** For `jdbc:mysql://host:3306/` with no selected database, which MySQL documents as a valid state where `DATABASE()` returns `NULL`, this can still bind a `null` schema into the metadata SQL and return empty metadata. That leaves the original `TableNotExistsException` / wrong metadata path unresolved for the reported ShardingSphere-JDBC + Masking use case.
  - **Action:** Please fix the production fallback source, not only the cached-subset symptom. Either carry the authoritative schema into the loader material, or explicitly load the relevant schemas when MySQL has no current catalog. Please add a regression test through `SchemaMetaDataUtils` or `GenericSchemaBuilder` for the masking-only no-current-schema path, without manually inserting into `GlobalDataSourceRegistry`.

### Newly Introduced Issues

- **[Blocker] `null` catalog with an empty table set can now fail before metadata loading** (`database/connector/dialect/mysql/src/main/java/org/apache/shardingsphere/database/connector/mysql/metadata/data/loader/MySQLMetaDataLoader.java:210`)
  - **Symptom:** The new `null == catalog || catalog.isEmpty()` branch calls `tableNames.iterator().next()` (`MySQLMetaDataLoader.java:211`). The existing empty-table test only covers a non-empty catalog (`MySQLMetaDataLoaderTest.java:268`), so the adjacent case `actualTableNames = empty` plus `catalog = null` is unvalidated.
  - **Risk:** The MySQL loader's existing "load without requested table names" path can regress to `NoSuchElementException` before it decides whether to load all metadata, return empty metadata, or use an explicit schema.
  - **Action:** Please define the expected behavior for empty requested table names when MySQL has no current catalog, guard the fallback accordingly, and add a dedicated regression test.

### Next Steps

- Rework schema fallback so it comes from an authoritative production source instead of a manually seeded `tableName -> database` global cache.
- Add production-path regression coverage for ShardingSphere-JDBC + Masking with a MySQL URL that has no schema selected.
- Add at least one counterexample for fallback binding, such as missing cache, empty requested table set, or same table name in two schemas.
- Keep the current non-empty catalog and empty-string catalog cases covered.
- Run the scoped verification after the fix, for example: `./mvnw -pl database/connector/dialect/mysql -am -DskipITs -Dspotless.skip=true -Dtest=MySQLMetaDataLoaderTest -Dsurefire.failIfNoSpecifiedTests=false test`, then run the project-required Spotless and Checkstyle gates.

terrymanu:
Hi @AtharvaMa, thanks again for working on this issue.

This PR has had no response or update for a long time after the change request. Also, the current approach is not ready to merge because it still relies on the table-to-schema fallback cache and does not prove that the reported masking-only no-current-schema path in #28469 is fixed.

To keep the review queue clear, we will close this PR soon due to inactivity. You are very welcome to reopen it or submit a new PR later with a revised root-cause fix and production-path regression tests.

terrymanu:
Hi @AtharvaMa, thanks again for the contribution.

I am going to close this PR now. The branch has had no update after the previous change request, and the narrow null-catalog handling has since been addressed on master by #38855 with additional coverage. This PR is now outdated against master, and the remaining broader no-current-schema / all-schema behavior from #28469 should be handled in a fresh PR if someone wants to continue that scope.

You are very welcome to open a new PR based on the latest master if you would like to keep working on the remaining issue. Thanks again.

PR #38602, opened 2026-04-09
Fix null-safe identifier matching for quoted normalized lookups

menghaoranss:
 This PR fixes a regression in StandardIdentifierCaseRule.matches(...) introduced while aligning quoted normalized lookup behavior.

The previous change correctly enabled quoted NORMALIZED matching for cases like MySQL with lower_case_table_names=1/2, but it accidentally removed the existing null guard in matches(...). As a result, StandardIdentifierCaseRuleTest could throw NullPointerException when either storedName or actualIdentifier was null.

- Root Cause

StandardIdentifierCaseRule.matches(...) now supports two quoted paths:

LookupMode.EXACT: compare stored and actual identifiers directly
LookupMode.NORMALIZED: compare normalized identifiers
During that refactor, the original null handling was dropped.
For unquoted matching, the method still evaluates:

unquotedStoredNamePredicate.test(storedName)
So when storedName or actualIdentifier is null, the predicate/normalizer path can fail before any equality check.

- Fix

Restore the null-safe guard at the start of matches(...):

if either storedName or actualIdentifier is null
return Objects.equals(storedName, actualIdentifier)
This preserves the quoted normalized matching fix and restores the original null-safe behavior.

- Tests

Updated StandardIdentifierCaseRuleTest to cover:

quoted normalized match
quoted null names
quoted null actual identifier
existing unquoted null scenarios

---

### DIFF
--- a/database/connector/core/src/main/java/org/apache/shardingsphere/database/connector/core/metadata/identifier/StandardIdentifierCaseRule.java
+++ b/database/connector/core/src/main/java/org/apache/shardingsphere/database/connector/core/metadata/identifier/StandardIdentifierCaseRule.java
@@ -50,11 +50,13 @@ public final class StandardIdentifierCaseRule implements IdentifierCaseRule {
     
     @Override
     public boolean matches(final String storedName, final String actualIdentifier, final QuoteCharacter quoteCharacter) {
-        if (QuoteCharacter.NONE != quoteCharacter) {
-            return Objects.equals(storedName, actualIdentifier);
-        }
         if (null == storedName || null == actualIdentifier) {
             return Objects.equals(storedName, actualIdentifier);
+        }
+        if (QuoteCharacter.NONE != quoteCharacter) {
+            return LookupMode.EXACT == quotedLookupMode
+                    ? Objects.equals(storedName, actualIdentifier)
+                    : Objects.equals(normalize(storedName), normalize(actualIdentifier));
         }
         return unquotedStoredNamePredicate.test(storedName) && Objects.equals(normalize(storedName), normalize(actualIdentifier));
     }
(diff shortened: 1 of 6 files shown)


menghaoranss:
```
Merge Verdict: Mergeable

Reviewed Scope

StandardIdentifierCaseRule.java
IdentifierCaseRuleSetsTest.java
StandardIdentifierCaseRuleTest.java
MySQLIdentifierCaseRuleProviderTest.java
IdentifierIndexTest.java
DatabaseIdentifierContextFactoryTest.java
Not Reviewed Scope

I did not run a full Proxy/JDBC end-to-end runtime reproduction.
I did not separately inspect GitHub PR comments or CI UI state.
Need Expert Review

No
Basis

The PR fixes the actual root cause in the shared matching path instead of adding fallback-only handling.
The previous inconsistency was:
MySQL insensitive rules declared quoted lookup as NORMALIZED
but StandardIdentifierCaseRule.matches(...) still treated all quoted identifiers as exact matches
The latest implementation now respects quotedLookupMode:
quoted EXACT remains exact
quoted NORMALIZED now compares normalized names
The original symptom path is no longer only partially covered by tests.
DatabaseIdentifierContextFactoryTest.java (line 161) now validates the metadata assembly path through:
DatabaseIdentifierContextFactory.create(...)
MySQL lower_case_table_names=1
ShardingSphereSchema.refreshIdentifierContext(...)
quoted table lookup with `T_MASK`
Shared-path regression risk was rechecked outside the target case:
generic insensitive quoted lookup still stays EXACT
Oracle still uses newUpperCaseRuleSet(), so quoted exact behavior is unchanged
Disabled-path behavior was also validated:
lower_case_table_names=0 still does not enable the MySQL insensitive rule
```

taojintianxia (approved):

Issue #38178, opened 2026-02-24
Consider adjusting the SHARDING API to bring keyGenerateStrategy to the same level as tables to support column-level auto-increment and global auto-increment


zhaojinchao95:
Hi all,

 To better solve the primary key generation problem in distributed systems，we will to change sharding api to to support column-level auto-increment and global auto-increment.

Before:

```
rules:
- !SHARDING
  tables:
    t_order:
      actualDataNodes: ds_${0..1}.t_order_${0..1}
      tableStrategy:
        standard:
          shardingColumn: order_id
          shardingAlgorithmName: t_order_inline
      keyGenerateStrategy:
        column: order_item_id
        keyGeneratorName: snowflake
  keyGenerators:
    snowflake:
      type: SNOWFLAKE
```

After：

```
rules:
- !SHARDING
  tables:
    t_order:
      actualDataNodes: ds_${0..1}.t_order_${0..1}
      tableStrategy:
        standard:
          shardingColumn: order_id
          shardingAlgorithmName: t_order_inline
  keyGenerateStrategies:
    t_order:
       keyGenerateType: column
       keyGeneratorName: snowflake
       logicTable: t_order
       keyGenerateColumn: id
  keyGenerators:
    snowflake:
      type: SNOWFLAKE
```

ym0506:
Thanks for the proposal. I’m not a maintainer, but this direction looks reasonable to me.
Could we clarify backward compatibility and precedence when old/new configs coexist?

zhaojinchao95:
> Thanks for the proposal. I’m not a maintainer, but this direction looks reasonable to me. Could we clarify backward compatibility and precedence when old/new configs coexist?

Hello, thank you for your feedback. this API adjustment is significant, and maintaining multiple versions in the future would complicate matters.

ym0506:
Thanks for the update and for merging #38294.

Would you like to keep #38178 open for follow-up items (for example DistSQL/docs migration guidance), or close this issue and track remaining work separately?

If helpful, I can help with docs/tests for migration behavior.

zhaojinchao95:
> Thanks for the update and for merging [#38294](https://github.com/apache/shardingsphere/pull/38294).
> 
> Would you like to keep [#38178](https://github.com/apache/shardingsphere/issues/38178) open for follow-up items (for example DistSQL/docs migration guidance), or close this issue and track remaining work separately?
> 
> If helpful, I can help with docs/tests for migration behavior.


That's great, welcome!

PR #38609, opened 2026-04-10
Remove keyGenerateStrategy for ShardingTableRuleConfiguration

zhaojinchao95:
For #38178


linghengqian (dismissed):
I don't believe the failure of `GraalVM - GraalVM CE for JDK 24.0.2 on ubuntu-latest` will prevent the current PR from being merged. It simply requires updating the GraalVM Reachability Metadata JSON according to the instructions at https://shardingsphere.apache.org/document/current/en/user-manual/shardingsphere-jdbc/graalvm-native-image/development/ . I will elaborate on the fix later.

zhaojinchao95:
• The current review rule does not allow skipping this check.

  This PR still includes an unrelated change in .github/workflows/graalvm.yml, so I cannot mark it as Mergeable in its current state. Even though the defaultKeyGenerateStrategy regression appears to be fixed, unrelated changes still need to be removed from
  this PR rather than deferred to a follow-up PR.

  If the goal is to confirm whether the core sharding change is otherwise acceptable, my current view is:

  - The defaultKeyGenerateStrategy semantic regression has been addressed.
  - The remaining blocker is that .github/workflows/graalvm.yml is still part of this PR and is unrelated to the stated goal.

  So the current conclusion is still:

  - Merge Verdict: Not Mergeable

  The minimum next step is:

  - Remove .github/workflows/graalvm.yml from this PR, and then I can do a final focused pass.

menghaoranss (approved):

PR #38629, opened 2026-04-16
Pipeline: Fix MySQL datetime(4) zero-value binlog decoding in migration

### LINKED ISSUE #35531: pipeline migration:  parse binlog error when mysql column type datetime(4) in increment task

5.5.3-SNAPSHOT

 ShardingSphere-Proxy

### Expected behavior

Binlog parser should correctly handle datetime(4) columns with special value '0000-00-00 00:00:00.0000'

### Actual behavior

in  aws rds 8.0.32, the following table is allowed

CREATE TABLE `reproduce_datetime_parse_error`
(
    `id`      bigint      NOT NULL AUTO_INCREMENT,
    `updated` datetime(4) NOT NULL DEFAULT '0000-00-00 00:00:00.0000',
    PRIMARY KEY (`id`)
) ENGINE = InnoDB;

insert into reproduce_datetime_parse_error values ();

**Binlog parser not read all bytes of column datetime(4),  which should 7 bytes, but only read 5 bytes ,lost fractional seconds** 

**When the parsing error occurs, the binlog client becomes corrupted**

sandynz:
Fixes #35531.

  - Pipeline: Fix MySQL datetime(4) zero-value binlog decoding in migration

---


taojintianxia (approved):

sandynz:
## Scope

This PR fixes the binlog decoding misalignment for zero-value `datetime(4)` during MySQL migration incremental stage.

The root cause is that `DATETIME2` returned early on the zero-value branch and skipped consuming the fractional bytes required by `datetime(4)`, which misaligned decoding for following columns and subsequent events.

## Main Changes

- Fixed `MySQLDatetime2BinlogProtocolValue` so the zero-value branch still consumes fractional bytes according to `columnMeta`.
- Added regression unit tests for zero-value `DATETIME2` with fractional precision.
- Updated `MySQLTimeTypesMigrationE2EIT` so the E2E reliably hits this scenario.

## E2E Validation Improvement

The E2E path was tightened so the same test can clearly distinguish behavior before and after the fix:

- It explicitly waits for `EXECUTE_INCREMENTAL_TASK` before inserting the second row, so the second row does not fall into inventory stage.
- A trailing `date` column is placed after `t_datetime_4`, so missing fractional-byte consumption immediately breaks decoding of the following column.
- The inventory row writes `t_datetime_4 = NULL`, while the incremental row hits the `datetime(4)` zero default, avoiding an earlier interruption caused by inventory JDBC converting zero datetime to `null`.

With this adjustment:

- On the reverted state, the same `MySQLTimeTypesMigrationE2EIT` reliably reproduces the incremental decode failure.
- After restoring the fix, the same `MySQLTimeTypesMigrationE2EIT` passes and completes `DATA_MATCH` verification.

## Verification

- `./mvnw -pl database/protocol/dialect/mysql -Dtest=MySQLDatetime2BinlogProtocolValueTest,MySQLFractionalSecondsTest test -Dsurefire.failIfNoSpecifiedTests=false -DskipITs -Dspotless.skip=true -Dcheckstyle.skip=true`
  - Exit code: 0
- `./mvnw -pl database/protocol/dialect/mysql -DskipTests -Pcheck -Dspotless.skip=true checkstyle:check`
  - Exit code: 0
- `./mvnw -pl test/e2e/operation/pipeline -DskipTests -Pcheck -Dspotless.skip=true checkstyle:check`
  - Exit code: 0
- `./mvnw -pl test/e2e/operation/pipeline -Dtest=MySQLTimeTypesMigrationE2EIT test -Dsurefire.failIfNoSpecifiedTests=false -DskipITs -Dspotless.skip=true -Dcheckstyle.skip=true`
  - Reverted state: Exit code 1
  - Fixed state: Exit code 0

## Notes

- Runtime validation was performed in NATIVE mode with an external `proxy/bootstrap/.../Bootstrap.java`.
- The fix scope stays limited to `DATETIME2`, matching the reported issue, and does not extend to `TIMESTAMP2` / `TIME2`.

PR #38655, opened 2026-04-27
Fix BOOLEAN type handling in Firebird Proxy  

makssent:
## Changes proposed in this pull request:

  - Introduce FirebirdBooleanBinaryProtocolValue as a dedicated handler for the Firebird BOOLEAN wire protocol type, reading/writing 4 bytes and correctly converting Boolean values to 1/0                                                                                                                                                                                                    
                                                                                                                                                                                                                                    
  ---
                                                                                                                                                                                                                                    
  - [x] My code follows the code of conduct of this project.                    
  - [x] I have passed maven check locally : ./mvnw clean install -B -T1C -Dmaven.javadoc.skip -Dmaven.jacoco.skip -e.
  - [ ] I have updated the Release Notes of the current development version.


terrymanu (changes requested):
### Decision

- **Reviewed Scope:** PR https://github.com/apache/shardingsphere/pull/38655 latest head `498b11efed129b4d0e2effc7968865e12110d5bd`; fetched base ref `apache/master` at `72f436c6ea6637fb9f8648265c62c745041b61a0`; local merge-base `44e922be4ea1562a68b1a57132b0cb56e511591e`. Local triple-dot file list matched GitHub `/pulls/38655/files`: 6 Firebird protocol/test files under `database/protocol/dialect/firebird`.
- **Not Reviewed Scope:** CI/check-runs, full Maven build, live Firebird client/server interoperability, and non-Firebird dialect runtime paths.
- **Need Expert Review:** Yes, Firebird wire-protocol review is needed after the byte-level encoding is corrected.

### Positive Feedback

- Introducing a dedicated `FirebirdBooleanBinaryProtocolValue` is the right direction; BOOLEAN should not keep sharing the old int1 handler.

### Major Issues

- **[P1] BOOLEAN row data is encoded with the wrong byte order and still does not prove the insert path** (`database/protocol/dialect/firebird/src/main/java/org/apache/shardingsphere/database/protocol/firebird/packet/command/query/statement/execute/protocol/FirebirdBooleanBinaryProtocolValue.java:29`)
  - **Symptom:** `read()` uses `payload.readInt4()` and `write()` uses `payload.writeInt4(...)`. In this codebase, those methods delegate to `ByteBuf.readInt()` / `ByteBuf.writeInt()`, while Firebird’s wire protocol documents `blr_bool` row data as a `Byte` padded to length 4, or equivalently a 4-byte value whose first byte is `1`/`0` / little-endian `Int32`: https://www.firebirdsql.org/file/documentation/html/en/firebirddocs/wireprotocol/firebird-wire-protocol.html#wireprotocol-row-data-column-value
  - **Risk:** For a documented TRUE value `01 00 00 00`, `readInt4()` reads `16777216`, not `1` or `true`. On output, `writeInt4(1)` writes `00 00 00 01`, not the documented `01 00 00 00`. This means both the original INSERT/EXECUTE symptom and SELECT/FETCH response path can remain incompatible with Firebird clients.
  - **Action:** Please implement boolean wire encoding explicitly: read the boolean byte and skip its 3 padding bytes, or add/use a little-endian 4-byte read; write `01 00 00 00` / `00 00 00 00` for true/false. Also convert inbound boolean values to Java `Boolean` before they are forwarded through `FirebirdExecuteStatementCommandExecutor` (`proxy/frontend/dialect/firebird/src/main/java/org/apache/shardingsphere/proxy/frontend/firebird/command/query/statement/execute/FirebirdExecuteStatementCommandExecutor.java:74`), because the current executor passes `packet.getParameterValues()` directly into binding and `QueryContext`.

- **[P1] Tests mock the helper API instead of validating real protocol bytes** (`database/protocol/dialect/firebird/src/test/java/org/apache/shardingsphere/database/protocol/firebird/packet/command/query/statement/execute/protocol/FirebirdBooleanBinaryProtocolValueTest.java:39`)
  - **Symptom:** The new tests stub `payload.readInt4()` and verify `payload.writeInt4(1)`, so they lock in the current helper call rather than verifying Firebird’s on-wire bytes. `FirebirdExecuteStatementPacketTest` also expects a BOOLEAN parameter value of `123` (`database/protocol/dialect/firebird/src/test/java/org/apache/shardingsphere/database/protocol/firebird/packet/command/query/statement/execute/FirebirdExecuteStatementPacketTest.java:153`), which is not a valid boolean counterexample.
  - **Risk:** These tests pass even when real Firebird-compatible bytes are read or written incorrectly, so the regression suite does not protect the root-cause fix.
  - **Action:** Please add ByteBuf-backed tests for the actual byte sequences `01 00 00 00` and `00 00 00 00` on both read and write paths, plus a production-path response test through `FirebirdFetchResponsePacket.writeRowData(...)` (`database/protocol/dialect/firebird/src/main/java/org/apache/shardingsphere/database/protocol/firebird/packet/generic/FirebirdFetchResponsePacket.java:63`). The execute-packet test should assert true/false/null behavior rather than an arbitrary integer.

### Unrelated Changes

- `database/protocol/dialect/firebird/src/test/java/org/apache/shardingsphere/database/protocol/firebird/packet/command/query/statement/execute/protocol/FirebirdInt1BinaryProtocolValueTest.java:26` only reorders static imports and is not tied to the BOOLEAN fix. Please roll it back unless it is strictly required by the final Spotless/Checkstyle run for this PR.

### Next Steps

- Fix BOOLEAN read/write against documented Firebird row-data bytes.
- Convert inbound BOOLEAN parameters to Java `Boolean` before backend binding.
- Replace mock-only tests with ByteBuf-backed protocol tests for true, false, and null/counterexample paths.
- Add response-path coverage for BOOLEAN fetch/stored-procedure row output.
- Roll back the unrelated import-only test change if it is not required.
- Re-run scoped verification with dependent modules, for example `./mvnw -pl database/protocol/dialect/firebird,proxy/frontend/dialect/firebird -am -DskipITs -Dspotless.skip=true -Dsurefire.failIfNoSpecifiedTests=false test`, then run the project formatting/style gates required by this repo.

### Evidence Supplement

- GitHub PR file scope and local triple-dot diff matched exactly.
- No GitHub-visible previous review comments or review threads were found, so no multi-round comparison is included.
- I did not use GitHub Actions or check-run status for this verdict.
- I did not run Maven locally because the byte-level protocol blocker is visible from the reviewed code and official Firebird wire-protocol documentation.

terrymanu (approved):
### Decision

- **Merge Verdict: Mergeable**
- **Reviewed Scope:** PR #38655 latest head `8b1dd539a2102e5d053033b3bc1b2db7fe599e0b`; the local triple-dot diff file list matches GitHub `/pulls/38655/files`, covering 7 Firebird protocol/test files under `database/protocol/dialect/firebird`.
- **Not Reviewed Scope:** GitHub Actions/check-runs, full Maven build, and live Firebird client/server interoperability.
- **Need Expert Review:** Not strictly required; an additional Firebird protocol maintainer review would still be helpful.

### Basis

- The latest revision fixes the previous BOOLEAN byte-order issue. `FirebirdBooleanBinaryProtocolValue` now reads the first byte and skips the 3 padding bytes, and writes `01 00 00 00` / `00 00 00 00`.
- This matches the Firebird official wire protocol definition of `blr_bool`: `Byte` padded to length 4, with `1` for true and `0` for false.
- Inbound execute parameters now parse `blr_bool` into Java `Boolean`, instead of passing BOOLEAN values forward as int1-style integer values.
- The fetch/SQL response path now uses the dedicated BOOLEAN handler to write the correct row-data bytes.
- The change is limited to the Firebird protocol path and does not affect non-Firebird dialects or shared SQL parser paths.

### Verification

- Reviewed the latest PR head and confirmed the local triple-dot diff boundary matches GitHub’s file list.
- Checked the Firebird official wire protocol documentation for BOOLEAN row-data encoding: https://www.firebirdsql.org/file/documentation/html/en/firebirddocs/wireprotocol/firebird-wire-protocol.html#wireprotocol-row-data-column-value
- The added/updated tests cover BOOLEAN read/write, factory mapping, execute parameter parsing, and fetch response output.
- GitHub Actions/check-run status was not used for this merge decision.

Issue #27879, opened 2023-08-03
MySQL's ResultSet of `SELECT *` is wrong when with Masking Feature


134130:
## Bug Report

**For English only**, other languages will not accept.

Before report a bug, make sure you have:

- Searched open and closed [GitHub issues](https://github.com/apache/shardingsphere/issues).
- Read documentation: [ShardingSphere Doc](https://shardingsphere.apache.org/document/current/en/overview).

Please pay attention on issues you submitted, because we maybe need more details. 
If no response anymore and we cannot reproduce it on current information, we will **close it**.

Please answer these questions before submitting your issue. Thanks!

### Which version of ShardingSphere did you use?
5.4.0

### Which project did you use? ShardingSphere-JDBC or ShardingSphere-Proxy?
ShardingSphere-JDBC with SharingSphere-Mask

### Expected behavior
```json
[
  {
    "actor_id": 77,
    "film_id": 8,
    "last_update": "2006-02-15 05:03:42",
    "first_name": "CARY",
    "last_name": "M********Y",
    "title": "AIRPORT POLLOCK",
    "description": "A Epic Tale of a Moose And a Girl who must Confront a Monkey in Ancient India",
    "release_year": 2006,
    "language_id": 1,
    "original_language_id": null,
    "rental_duration": 6,
    "rental_rate": 4.99,
    "length": 54,
    "replacement_cost": 15.99,
    "rating": "R",
    "special_features": "Trailers"
  }
]
```

### Actual behavior
```json
[
  {
    "actor_id": 1,
    "first_name": 1,
    "last_name": "2006-02-15 05:05:03",
    "last_update": 79
  }
]
```

### Reason analyze (If you can)
May MaskingContext pollute MetadataContext

### Steps to reproduce the behavior, such as: SQL to execute, sharding rule configuration, when exception occur etc.

1. build.gradle.kts
	```kotlin
	dependencies {
	    // ShardingSphere
	    implementation("org.apache.shardingsphere:shardingsphere:5.4.0")
	    implementation("org.apache.shardingsphere:shardingsphere-jdbc:5.4.0")
	    implementation("org.apache.shardingsphere:shardingsphere-jdbc-core:5.4.0")
	
	    // ShardingSphere Mask
	    implementation("org.apache.shardingsphere:shardingsphere-mask:5.4.0")
	    implementation("org.apache.shardingsphere:shardingsphere-mask-api:5.4.0")
[...]

PR #38855, opened 2026-06-16
Fix MySQLMetaDataLoader NPE risk when JDBC catalog is null

nassirsultan:
Resolves the null-vs-empty-string catalog handling bug described in
apache/shardingsphere#28469: getDatabaseName() checked catalog with
"".equals(...), which doesn't catch a null catalog (returned by
MySQL's getCatalog() when no schema is selected on the connection),
causing the cached-database-name fallback to be skipped and masking
to fail with TableNotExistsException for a specific, named table.

Replaced the check with Guava's Strings.isNullOrEmpty(), consistent
with existing usage elsewhere in the same file.

Also fixes a related crash this change could otherwise introduce:
when no catalog is selected and no specific table names are requested
(the "load every table" startup case), the null-catalog fallback
previously assumed a table name was always available to look up in
the cache. getDatabaseName() now checks tableNames.isEmpty() first and
returns null gracefully in that case, rather than throwing
NoSuchElementException.

Scope: this PR fixes the specific null-catalog handling bug and the
empty-tableNames crash above. It does not implement fetching/querying
across all schemas when no current schema is selected and no specific
table is requested - the original report also raises that as a
separate question ("I think we need to fetch all schemas in this
case"), which is a meaningfully larger feature (schema enumeration,
cross-schema querying, naming-collision handling) rather than a
regression test that fits this fix.

Partially addresses #28469

- Fix null-catalog handling in MySQLMetaDataLoader.getDatabaseName()
- Guard against empty tableNames when no catalog fallback is available
- Add regression tests for both the null-catalog fallback and the
  empty-tableNames edge case, each verified to fail without the
  corresponding fix and pass with it

- [x] I have passed maven check locally: `./mvnw clean install -B -T1C -Dmaven.javadoc.skip -Dmaven.jacoco.skip -e`

### DIFF
--- a/database/connector/dialect/mysql/src/main/java/org/apache/shardingsphere/database/connector/mysql/metadata/data/loader/MySQLMetaDataLoader.java
+++ b/database/connector/dialect/mysql/src/main/java/org/apache/shardingsphere/database/connector/mysql/metadata/data/loader/MySQLMetaDataLoader.java
@@ -206,7 +206,7 @@ public final class MySQLMetaDataLoader implements DialectMetaDataLoader {
     }
     
     private String getDatabaseName(final Connection connection, final Collection<String> tableNames) throws SQLException {
-        return "".equals(connection.getCatalog()) ? GlobalDataSourceRegistry.getInstance().getCachedDatabaseTables().get(tableNames.iterator().next()) : connection.getCatalog();
+        return Strings.isNullOrEmpty(connection.getCatalog()) ? GlobalDataSourceRegistry.getInstance().getCachedDatabaseTables().get(tableNames.iterator().next()) : connection.getCatalog();
     }
     
     @Override
--- a/database/connector/dialect/mysql/src/main/java/org/apache/shardingsphere/database/connector/mysql/metadata/data/loader/MySQLMetaDataLoader.java
+++ b/database/connector/dialect/mysql/src/main/java/org/apache/shardingsphere/database/connector/mysql/metadata/data/loader/MySQLMetaDataLoader.java
@@ -206,7 +206,11 @@ public final class MySQLMetaDataLoader implements DialectMetaDataLoader {
     }
     
     private String getDatabaseName(final Connection connection, final Collection<String> tableNames) throws SQLException {
-        return Strings.isNullOrEmpty(connection.getCatalog()) ? GlobalDataSourceRegistry.getInstance().getCachedDatabaseTables().get(tableNames.iterator().next()) : connection.getCatalog();
+        String catalog = connection.getCatalog();
+        if (!Strings.isNullOrEmpty(catalog)) {
+            return catalog;
+        }
+        return tableNames.isEmpty() ? null : GlobalDataSourceRegistry.getInstance().getCachedDatabaseTables().get(tableNames.iterator().next());
     }
     
     @Override
--- a/database/connector/dialect/mysql/src/test/java/org/apache/shardingsphere/database/connector/mysql/metadata/data/loader/MySQLMetaDataLoaderTest.java
+++ b/database/connector/dialect/mysql/src/test/java/org/apache/shardingsphere/database/connector/mysql/metadata/data/loader/MySQLMetaDataLoaderTest.java
@@ -50,6 +50,7 @@ import static org.hamcrest.Matchers.is;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.mockito.Mockito.RETURNS_DEEP_STUBS;
 import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
 class MySQLMetaDataLoaderTest {
@@ -87,7 +88,7 @@ class MySQLMetaDataLoaderTest {
     @ParameterizedTest(name = "{0}")
     @MethodSource("loadArguments")
     void assertLoad(final String name, final Collection<String> actualTableNames, final String tableMetaDataSQL,
-                    final boolean emptyCatalog, final boolean includeView, final boolean includeConstraints, final boolean includeCompositeIndex) throws SQLException {
+                    final String catalog, final String expectedDatabaseName, final boolean includeView, final boolean includeConstraints, final boolean includeCompositeIndex) throws SQLException {
         DataSource dataSource = mockDataSource();
         ResultSet tableMetaDataResultSet = mockTableMetaDataResultSet();
         when(dataSource.getConnection().prepareStatement(tableMetaDataSQL).executeQuery()).thenReturn(tableMetaDataResultSet);
@@ -105,10 +106,11 @@ class MySQLMetaDataLoaderTest {
         Map<String, String> cachedDatabaseTables = GlobalDataSourceRegistry.getInstance().getCachedDatabaseTables();
         String previous = cachedDatabaseTables.put("tbl", "fallback_db");
         try {
-            when(dataSource.getConnection().getCatalog()).thenReturn(emptyCatalog ? "" : "sharding_db");
+            when(dataSource.getConnection().getCatalog()).thenReturn(catalog);
             DataTypeRegistry.load(dataSource, "MySQL");
             Collection<SchemaMetaData> actual = dialectMetaDataLoader.load(new MetaDataLoaderMaterial(actualTableNames, "foo_ds", dataSource, databaseType, "sharding_db"));
             assertTableMetaData(actual, includeConstraints, includeCompositeIndex);
+            verify(dataSource.getConnection().prepareStatement(tableMetaDataSQL)).setString(1, expectedDatabaseName);
         } finally {
             if (null == previous) {
                 cachedDatabaseTables.remove("tbl");
@@ -237,8 +239,9 @@ class MySQLMetaDataLoaderTest {
     
     private static Stream<Arguments> loadArguments() {
         return Stream.of(
-                Arguments.of("load without requested table names", Collections.emptyList(), TABLE_METADATA_SQL, false, false, false, false),
-                Arguments.of("load with requested table names", Collections.singletonList("tbl"), TABLE_METADATA_SQL_WITH_TABLE, false, false, false, false),
-                Arguments.of("load with empty catalog fallback and view constraints", Collections.singletonList("tbl"), TABLE_METADATA_SQL_WITH_TABLE, true, true, true, true));
+                Arguments.of("load without requested table names", Collections.emptyList(), TABLE_METADATA_SQL, "sharding_db", "sharding_db", false, false, false),
+                Arguments.of("load with requested table names", Collections.singletonList("tbl"), TABLE_METADATA_SQL_WITH_TABLE, "sharding_db", "sharding_db", false, false, false),
+                Arguments.of("load with empty catalog fallback and view constraints", Collections.singletonList("tbl"), TABLE_METADATA_SQL_WITH_TABLE, "", "fallback_db", true, true, true),
+                Arguments.of("load with null catalog fallback", Collections.singletonList("tbl"), TABLE_METADATA_SQL_WITH_TABLE, null, "fallback_db", true, true, true));
     }
 }
--- a/database/connector/dialect/mysql/src/test/java/org/apache/shardingsphere/database/connector/mysql/metadata/data/loader/MySQLMetaDataLoaderTest.java
+++ b/database/connector/dialect/mysql/src/test/java/org/apache/shardingsphere/database/connector/mysql/metadata/data/loader/MySQLMetaDataLoaderTest.java
@@ -84,6 +84,17 @@ class MySQLMetaDataLoaderTest {
         assertTrue(actual.iterator().next().getTables().isEmpty());
     }
     
+    @SuppressWarnings("JDBCResourceOpenedButNotSafelyClosed")
+    @Test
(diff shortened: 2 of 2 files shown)


terrymanu (changes requested):
### Summary

- **Merge Decision: Not Mergeable**
- **Reason:** The null-catalog fallback still assumes a table-name fallback is always available, so the no-current-database path from the linked issue remains incomplete and can fail before metadata is loaded.

### Positive Feedback

- The table-specific null/empty catalog fallback is in the right direction, and the added assertion checks the bound database name for that narrow path.

### Issues

- **[P1] Null catalog fallback can fail for full metadata load** (`database/connector/dialect/mysql/src/main/java/org/apache/shardingsphere/database/connector/mysql/metadata/data/loader/MySQLMetaDataLoader.java:209`)
  - **Problem:** `Strings.isNullOrEmpty(connection.getCatalog())` now sends both empty and null catalogs to `GlobalDataSourceRegistry.getInstance().getCachedDatabaseTables().get(tableNames.iterator().next())`. The loader already has a supported “load without requested table names” case in `MySQLMetaDataLoaderTest.java:242`, but that branch has no first table name; when `getCatalog()` is null, it can fail before binding the metadata SQL. This is relevant to #28469 because MySQL documents that `DATABASE()` returns `NULL` when no default database is selected, and JDBC `Connection#getCatalog()` may return null if there is no current catalog.
  - **Impact:** A JDBC URL without a schema can still fail or miss metadata for the full/no-table-filter path, and `Fixes #28469` can close a public issue whose no-schema and other-schema requirements are not fully validated.
  - **Required Change:** Please make null/empty-catalog behavior explicit when `actualTableNames` is empty instead of unconditionally reading the first table, then add regression coverage for null catalog with no requested table names and map the remaining #28469 cross-schema requirement to code/tests. If this PR is intentionally only a narrow table-specific fallback, please change the PR text from `Fixes #28469` to a partial reference so the broader issue remains open.

### Review Details

- **Reviewed Scope:** PR #38855 latest head `e563e4dd6f4e30f4c4cd5a93f01ea552b9c512b4`, merge-base `c215ada3e5bf9dfbd02f25f9a710ca1fb68c8015`; reviewed `MySQLMetaDataLoader.java`, `MySQLMetaDataLoaderTest.java`, linked issue #28469, and related public issue context. GitHub `/pulls/38855/files` returned 2 files and matched the local triple-dot diff file list.
- **Not Reviewed Scope:** CI/check-run status, live MySQL/ShardingSphere-JDBC masking reproduction, and modules outside the two changed files except for focused metadata-material/cache path inspection.
- **Verification:** No Maven command was run; this review found a deterministic code/linked-issue blocker by static review. Public semantic references checked: MySQL information functions documentation and Java SE `Connection#getCatalog()` documentation.
- **Release Note / User Docs:** Not required for a narrow internal metadata bug fix that restores expected MySQL no-schema behavior; docs/release notes are not the current blocker.

nassirsultan:
Thanks for the detailed review — pushed a fix for the empty-tableNames crash, with a regression test verified to fail without it (NoSuchElementException) and pass with it.

On the remaining point: I went back through the original issue and the linked triage comment, and both describe a single named table failing under the masking feature with no schema selected - I don't see a cross-schema scenario mentioned anywhere in either. My read is that this PR fully resolves what #28469 actually describes, and "cross-schema" might be pointing at a separate, broader concern (e.g. schema handling across multiple sharded data sources) that wasn't part of the original report. Happy to be corrected if I'm missing something, but wanted to flag that distinction rather than assume.

terrymanu (changes requested):
### Summary

- **Merge Decision: Not Mergeable**
- **Reason:** The latest patch avoids the previous empty-table-name crash, but the no-current-schema metadata path required by #28469 still binds a null schema filter and now tests the empty-result behavior as success.

### Positive Feedback

- The table-specific null/empty catalog fallback is moving in the right direction, and the latest head adds coverage for the previously reported empty `tableNames` crash path.

### Issues

- **[P1] No-current-schema full metadata load still returns no table metadata** (`database/connector/dialect/mysql/src/main/java/org/apache/shardingsphere/database/connector/mysql/metadata/data/loader/MySQLMetaDataLoader.java:213`)
  - **Problem:** When `Connection#getCatalog()` is null and `actualTableNames` is empty, `getDatabaseName()` now returns `null`, while the full metadata SQL still uses `WHERE TABLE_SCHEMA=?` (`MySQLMetaDataLoader.java:54` and `MySQLMetaDataLoader.java:165`). MySQL documents that `DATABASE()` returns `NULL` when no default database is selected, and MySQL NULL comparison with `=` does not match rows. The new regression test also passes an empty table list with a null catalog and asserts that the returned schema has no tables (`MySQLMetaDataLoaderTest.java:89` and `MySQLMetaDataLoaderTest.java:95`), which codifies the missing metadata result instead of proving the all-schema behavior requested in #28469.
  - **Impact:** A ShardingSphere-JDBC MySQL connection without a schema can still fail to load the metadata needed by the masking/single-table scenario from #28469, and the PR can over-close the issue while leaving the no-schema and other-schema requirements unresolved.
  - **Required Change:** Please make the no-current-schema behavior explicit and load the intended schemas when no table-name fallback is available, then add regression coverage for the no-current-schema full metadata path and the other-schema/cross-schema requirement described in #28469. If this PR is intentionally limited to the table-specific cached fallback, please narrow the PR text from `Fixes #28469` to a partial reference so the broader issue remains open.

### Multi-Round Comparison

- The previous blocker about `tableNames.iterator().next()` on an empty collection is **partially fixed**: the latest head no longer crashes there, but it replaces the crash with a null schema filter and does not prove the linked issue's expected metadata-loading behavior.
- The previous request to map the remaining #28469 cross-schema requirement to code/tests is **not fixed**: no new test covers the issue's "fetch all schemas" or other-schema path.

### Review Details

- **Reviewed Scope:** PR #38855 latest head `bfe565cde8682d878e9a8ef56a94dd9fa49fceb6`, local merge-base `ac8881fa2d48cdd9549e89a09b391d868ced3cbc`; GitHub `/pulls/38855/files` returned 2 files and matched the local triple-dot diff: `MySQLMetaDataLoader.java` and `MySQLMetaDataLoaderTest.java`. Also reviewed linked issue #28469, referenced issue #27879, and focused caller/context paths in `SchemaMetaDataUtils` and `MetaDataLoader`.
- **Not Reviewed Scope:** CI/check-run status, live MySQL/ShardingSphere-JDBC masking reproduction, and a full repository build.
- **Verification:** Public GitHub API reads for PR/issues/file list: exit 0; local PR ref fetch and triple-dot file-list cross-check: exit 0. No Maven command was run because the merge blocker is deterministic from the latest code, tests, linked issue scope, and official semantics. Semantic references checked: [MySQL `DATABASE()`](https://dev.mysql.com/doc/refman/9.7/en/information-functions.html), [MySQL NULL comparisons](https://dev.mysql.com/doc/refman/9.7/en/working-with-null.html), and [Java `Connection#getCatalog()`](https://docs.oracle.com/javase/8/docs/api/java/sql/Connection.html#getCatalog--).
- **Release Note / User Docs:** Not required for this narrow internal metadata bug-fix attempt; the current blocker is root-cause completeness and regression coverage.

nassirsultan:
Thanks for clarifying — that's a fair distinction, and I want to make sure we're aligned on scope before going further.

What this PR does: fixes the null-vs-empty-string mismatch that breaks masking for a specific, named table when no schema is selected (the exact scenario in the original report and the linked triage comment), and closes the empty-tableNames crash that surfaced as a side effect of the first fix.

What this PR does not do: implement schema discovery across multiple/all schemas when no schema is selected and no specific table is requested. That's a meaningfully different capability - querying across every schema on the server rather than failing gracefully when none is selected - and I don't see it described in #28469's original report or the linked fix recommendation.

If that broader multi-schema discovery is something the project wants, I'd suggest it become its own tracked issue, since it's a real feature with its own design questions (which schemas to query, how to handle naming collisions across them, etc.) rather than a regression test we can bolt onto this fix.

For this PR, I'd like to change the description from "Fixes #28469" to "Addresses #28469" to reflect that scope honestly, and treat full multi-schema discovery as a separate follow-up issue. Does that line up with how you'd scope it, or is there something specific in the current code that still needs to change?

terrymanu:
Thanks for clarifying. I agree that full multi-schema discovery is a broader behavior than the narrow null/empty catalog fallback fixed in this PR.

One important point: #28469's original report does mention fetching all schemas and the other-schema case, so this PR should not use a closing keyword for that issue. Please update the PR description from `Fixes #28469` to something like `Partially addresses #28469` or `Addresses part of #28469`, and explicitly mention that full no-current-schema / cross-schema metadata discovery remains out of scope for this PR.

After that, we can re-review this PR as a narrow partial fix. If the intention is still to close #28469 in this PR, then the previous request still stands: the PR needs to implement and test the no-current-schema full metadata path and the other-schema/cross-schema behavior.

nassirsultan:
Updated the PR description to "Partially addresses #28469" with explicit scope notes, per your feedback. Thanks for catching that — I'd missed the "fetch all schemas" line at the end of the original report.

terrymanu:
Thanks for updating the PR description and clarifying the scope.

I agree with the narrowed scope: this PR can be reviewed as a partial fix for the null/empty catalog fallback and the empty `tableNames` crash, while full no-current-schema / cross-schema metadata discovery remains out of scope.

One last thing before this can be merged: the commit message still contains `Fixes #28469`. Since #28469 also mentions fetching all schemas and the other-schema case, please reword or squash the commit so the closing keyword becomes something like `Partially addresses #28469` or `Addresses part of #28469`.

After that, I think this PR can be re-reviewed as a narrow partial fix.

nassirsultan:
Reworded the commit message so it no longer uses a closing keyword for #28469. It now reads "Partially addresses #28469", matching the PR description.

linghengqian (commented):
Your branch seems to have been messed up by your merge operation, and I can't even approve CI to start.

nassirsultan:
Cleaned up the branch — reset onto current master and re-applied just the two real commits, so the history should be simple to review now. Let me know if CI can run.

linghengqian (commented):
> Let me know if CI can run.

LGTM😄

nassirsultan:
👍

terrymanu (approved):
### Summary

- **Reason:** The latest PR is now scoped as a partial fix for #28469, and the null/empty catalog fallback plus empty `tableNames` guard are covered by focused tests and local style/test verification.

### Evidence

- `database/connector/dialect/mysql/src/main/java/org/apache/shardingsphere/database/connector/mysql/metadata/data/loader/MySQLMetaDataLoader.java:208` now reads `Connection#getCatalog()` once, keeps the normal non-empty catalog path, falls back to `GlobalDataSourceRegistry` for null/empty catalog with requested table names, and avoids the previous empty-collection crash for the no-requested-table case.
- `database/connector/dialect/mysql/src/test/java/org/apache/shardingsphere/database/connector/mysql/metadata/data/loader/MySQLMetaDataLoaderTest.java:89` covers null catalog with no requested table names, and `MySQLMetaDataLoaderTest.java:251` covers both empty-catalog and null-catalog named-table fallback to the cached database.
- The previous public blockers are closed for the narrowed scope: the PR description and commits now say `Partially addresses #28469`, the broader all-schema/cross-schema discovery request remains explicitly out of scope, and the empty `tableNames` crash has a regression test.
- Scope/ownership is limited to the MySQL metadata loader and its tests; no shared API, dependency, distribution, protocol, parser, or public configuration surface is changed.
- MySQL and JDBC semantics support the trigger condition: MySQL documents that `DATABASE()` returns `NULL` when no default database exists, and Java `Connection#getCatalog()` may return `null` when there is no current catalog.

### Review Details

- **Reviewed Scope:** PR #38855 latest head `335a92c22d80741ca7cb8f4b475547065420d479`, base `master` at `c0c01668ff2f6921a504bfa86542a30ef0a8753a`, local merge-base `c0c01668ff2f6921a504bfa86542a30ef0a8753a`; GitHub `/pulls/38855/files` returned 2 files and matched the local triple-dot diff: `MySQLMetaDataLoader.java` and `MySQLMetaDataLoaderTest.java`. Also reviewed linked issue #28469, public PR comments/reviews, `GlobalDataSourceRegistry`, `SchemaMetaDataUtils`, `MetaDataLoader`, and official MySQL/JDBC docs.
- **Not Reviewed Scope:** CI/check-run status, live MySQL/ShardingSphere-JDBC masking reproduction, and the broader full no-current-schema / cross-schema metadata discovery behavior that the PR now explicitly leaves out of scope.
- **Verification:** Public GitHub REST metadata/file/comment/review reads: exit 0; local PR ref fetch and inventory cross-check: exit 0; `./mvnw -pl database/connector/dialect/mysql -am -DskipITs -Dspotless.skip=true -Dtest=MySQLMetaDataLoaderTest -Dsurefire.failIfNoSpecifiedTests=false test`: exit 0, `Tests run: 6, Failures: 0, Errors: 0, Skipped: 0`; `./mvnw -pl database/connector/dialect/mysql -DskipTests -DskipITs spotless:check -Pcheck -T1C`: exit 0; `./mvnw -pl database/connector/dialect/mysql -DskipTests -DskipITs checkstyle:check -Pcheck -T1C`: exit 0.
- **Release Note / User Docs:** Not required for this narrow internal metadata bug fix; it restores expected MySQL null-catalog handling for the named-table fallback path without adding user-facing configuration, migration steps, or new documented behavior.

nassirsultan:
Thanks for the review and the guidance throughout — glad to see this merged!

Issue #38857, opened 2026-06-17
Nightly CI failure in NativeTest CI on Ubuntu and Windows


RaigorJiang:
Action link: https://github.com/apache/shardingsphere/actions/runs/27625450071

<img width="1369" height="686" alt="Image" src="https://github.com/user-attachments/assets/0c2309a5-ddb6-4fc0-8881-8812d3de308e" />

---

### CI Job

  Nightly CI failed in the following jobs:

  - `NativeTest CI - GraalVM CE on ubuntu-latest`
  - `NativeTest CI - GraalVM CE on windows-latest`

  ### Failure Summary

  The failures are concentrated in the GraalVM native test workflow, but Ubuntu and Windows show different failure modes.

  #### 1. Windows: Docker/Testcontainers environment is unavailable

  In `NativeTest CI - GraalVM CE on windows-latest`, the workflow runs:

  ```bash
  ./mvnw -PgenerateMetadata -e -T 1C clean verify
  ```

  The job first installs and starts Rancher Desktop, then runs the generateMetadata profile. However, Testcontainers cannot find a valid Docker environment.

  Relevant log excerpt:

  DockerClientProviderStrategy - Could not find a valid Docker environment. Please check configuration.
  Attempted configurations were:
  NpipeSocketClientProviderStrategy: failed with exception RuntimeException
  Root cause MalformedChunkCodingException (Bad chunk header: )
  As no valid configuration was found, execution cannot continue.

  After that, many native tests fail with the same root cause:

  Previous attempts to find a Docker environment failed. Will not retry.

  The final Failsafe summary is:

  Tests run: 24, Failures: 0, Errors: 16, Skipped: 1

  Affected tests include:

  - ClickHouseTest
  - DorisFETest
  - FirebirdTest
  - MySQLTest
  - OpenGaussTest
  - PostgresTest
  - PrestoTest
  - SQLServerTest
  - Hive-related native tests
  - SeataTest
  - Proxy MySQL/PostgreSQL native tests

  This looks like a CI environment/readiness issue rather than a failure in a specific database test. Rancher Desktop reaches a backend state accepted by the script, but Docker
  is still not usable by Testcontainers through the expected Windows Docker API.

  #### 2. Ubuntu: Native image execution fails because of missing reachability metadata

  In NativeTest CI - GraalVM CE on ubuntu-latest, the workflow runs:

[...]

PR #38682, opened 2026-05-10
Support building Proxy Native via GraalVM CE for JDK 25

linghengqian:
For #29052 .

Changes proposed in this pull request:
  - The current PR has some preceding PRs, 
    - https://github.com/apache/shardingsphere/pull/38721
    - https://github.com/apache/shardingsphere/pull/38722
    - https://github.com/apache/shardingsphere/pull/38880
    - https://github.com/apache/shardingsphere/pull/38878
  - Support building Proxy Native via GraalVM CE for JDK 25. These are the key steps for using https://github.com/graalvm/native-build-tools/releases/tag/1.0.0 . 
  - Increase the timeout for detecting the ShardingSphere proxy in nativetest to 5 minutes. Also see https://github.com/netty/netty/wiki/Java-24-and-sun.misc.Unsafe . Netty no longer utilizes `Unsafe` on JDK 25; consequently, ShardingSphere Proxy has slowed down, exposing issues that had previously gone unnoticed.
```bash
[INFO] Running org.apache.shardingsphere.test.natived.proxy.databases.MySQLTest
[INFO] MultiKeyMap stripe configuration: 32 locks for 16 cores
WARNING: A restricted method in java.lang.System has been called
WARNING: java.lang.System::loadLibrary has been called by io.netty.util.internal.NativeLibraryUtil in an unnamed module (file:/home/lingh/.m2/repository/io/netty/netty-common/4.2.9.Final/netty-common-4.2.9.Final.jar)
WARNING: Use --enable-native-access=ALL-UNNAMED to avoid a warning for callers in this module
WARNING: Restricted methods will be blocked in a future release unless native access is enabled

WARNING: A terminally deprecated method in sun.misc.Unsafe has been called
WARNING: sun.misc.Unsafe::objectFieldOffset has been called by com.github.benmanes.caffeine.cache.UnsafeAccess (file:/home/lingh/.m2/repository/com/github/ben-manes/caffeine/caffeine/2.9.3/caffeine-2.9.3.jar)
WARNING: Please consider reporting this to the maintainers of class com.github.benmanes.caffeine.cache.UnsafeAccess
WARNING: sun.misc.Unsafe::objectFieldOffset will be removed in a future release
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 356.9 s -- in org.apache.shardingsphere.test.natived.proxy.databases.MySQLTest
[...]

PR #35905, opened 2025-07-06
Enable nativeTest on Windows Server 2025 CI via Docker CLI using remote Podman connection

linghengqian:
For #35052 .

Changes proposed in this pull request:
  - Enable nativeTest on Windows Server 2025 CI via Docker CLI using remote Podman connection.
  - Due to the controversy at https://github.com/rancher-sandbox/rancher-desktop/issues/8888 , I do not plan to introduce the steps to install rancher desktop in the github actions pipeline of shardingsphere immediately.
  - Due to unresolved https://github.com/testcontainers/testcontainers-java/issues/10381 and , we cannot configure Docker CLI to use a remote Docker Daemon.

---

Before committing this PR, I'm sure that I have checked the following options:
- [x] My code follows the [code of conduct](https://shardingsphere.apache.org/community/en/involved/conduct/code/) of this project.
- [x] I have self-reviewed the commit code.
- [x] I have (or in comment I request) added corresponding labels for the pull request.
- [x] I have passed maven check locally : `./mvnw clean install -B -T1C -Dmaven.javadoc.skip -Dmaven.jacoco.skip -e`.
- [x] I have made corresponding changes to the documentation.
- [ ] I have added corresponding unit tests for my changes.
- [ ] I have updated the Release Notes of the current development version. For more details, see [Update Release Note](https://shardingsphere.apache.org/community/en/involved/contribute/contributor/)

PR #38859, opened 2026-06-17
Stabilize GraalVM native test CI

RaigorJiang:
For #38857

  ## Summary

  This PR stabilizes the GraalVM native test CI by fixing three failure paths found in PR CI:

  - Add native reachability metadata for `NodePath` field access used by native tests.
  - Strengthen Windows Docker readiness verification before running Maven tests that depend on Testcontainers.
  - Close the native Zookeeper test datasource before the embedded Zookeeper server shuts down.

  ## Background

  The GraalVM native test workflow failed on both Ubuntu and Windows, with different causes across CI attempts.

  On Ubuntu, the native image test binary failed with `MissingReflectionRegistrationError` when `NodePathSegment` reflected fields from `TableMetaDataNodePath`:

  ```text
  org.graalvm.nativeimage.MissingReflectionRegistrationError:
  The program tried to reflectively read or write field:
  private final SchemaMetaDataNodePath TableMetaDataNodePath.schema
  without it being registered for runtime reflection.
  ```
  NodePathGenerator and NodePathSegment build node paths by reading @NodePathEntity variables through reflection. In native image tests, these fields must be registered in
  reachability metadata.

  On Windows, the CI setup accepted Rancher Desktop readiness, but Testcontainers later failed during Maven execution because Docker was still not usable:

  DockerClientProviderStrategy - Could not find a valid Docker environment.
  NpipeSocketClientProviderStrategy ... MalformedChunkCodingException: Bad chunk header

  This made the failure appear as many downstream native test errors instead of a clear Docker readiness failure.

  After those two issues were addressed, both Ubuntu and Windows native test jobs exposed the same remaining failure in ZookeeperTest cleanup:

  ZookeeperTest.afterEach - ClusterRepositoryPersistException
  Caused by: KeeperException$ConnectionLossException

  The test closed ShardingSphereDataSource from @AfterEach, after the TestingServer created by try-with-resources had already shut down. Closing the datasource closes
  ContextManager, which tries to take the compute node offline from Zookeeper. At that point the embedded Zookeeper server was already unavailable.

  ## Changes

  - Add native reachability metadata for NodePathSegment field access:
      - DatabaseMetaDataNodePath.databaseName
      - SchemaMetaDataNodePath.database
      - SchemaMetaDataNodePath.schemaName
      - TableMetaDataNodePath.schema
      - TableMetaDataNodePath.tableName

  - Update Windows Docker readiness checks:
      - Keep checking Docker server API readiness with:
          - docker version --format "{{.Server.Version}}"
          - docker info --format "{{.ServerVersion}}"

      - Add a short container lifecycle smoke check:
          - remove any stale shardingsphere-docker-readiness container
          - run hello-world with --rm

      - Keep the existing retry path that restarts Rancher Desktop before the final attempt.

  - Update ZookeeperTest cleanup order:
      - close logicDataSource while TestingServer is still alive
      - keep @AfterEach as a fallback cleanup path
      - clear the Zookeeper system property in finally
      - avoid closing the same datasource twice

  ## Motivation

  The native test CI should fail at the real setup, metadata, or resource lifecycle boundary instead of surfacing misleading downstream test errors.

  The reachability metadata change fixes the native image runtime reflection failure for node path generation.

  The Windows Docker smoke check makes the setup script verify that Docker can actually run containers before Maven starts, which better matches what Testcontainers needs
  during -PgenerateMetadata.

  The Zookeeper cleanup change ensures ContextManager.close() can take the compute node offline before the embedded Zookeeper server is closed.

  ## Verification

  Ran locally:

  jq empty test/native/src/test/resources/META-INF/native-image/shardingsphere-test-native/reachability-metadata.json

  Result: success

  Ran locally:

  ./mvnw -pl test/native -am -Dspotless.apply.skip=true -Dcheckstyle.skip=true -Djacoco.skip=true -DskipITs
  -Dtest=org.apache.shardingsphere.test.natived.jdbc.modes.cluster.ZookeeperTest -Dsurefire.failIfNoSpecifiedTests=false test

  Result: BUILD SUCCESS

  Note: ZookeeperTest is annotated with @EnabledInNativeImage, so the scoped JVM test verifies compilation and test discovery while the actual execution path is covered by the
  GraalVM native test CI.

  Ran locally:

  ./mvnw spotless:check -Pcheck -T1C

  Result: BUILD SUCCESS

  Ran locally:

  ./mvnw checkstyle:check -Pcheck -T1C

  Result: BUILD SUCCESS

  The PR GraalVM native test CI has passed after the Zookeeper cleanup adjustment.

### DIFF
--- a/test/native/src/test/resources/META-INF/native-image/shardingsphere-test-native/reachability-metadata.json
+++ b/test/native/src/test/resources/META-INF/native-image/shardingsphere-test-native/reachability-metadata.json
@@ -26,6 +26,45 @@
     },
     {
       "condition": {
+        "typeReached": "org.apache.shardingsphere.mode.node.path.engine.generator.NodePathSegment"
+      },
+      "type": "org.apache.shardingsphere.mode.node.path.type.database.metadata.DatabaseMetaDataNodePath",
+      "fields": [
+        {
+          "name": "databaseName"
+        }
+      ]
+    },
+    {
+      "condition": {
+        "typeReached": "org.apache.shardingsphere.mode.node.path.engine.generator.NodePathSegment"
+      },
+      "type": "org.apache.shardingsphere.mode.node.path.type.database.metadata.schema.SchemaMetaDataNodePath",
+      "fields": [
+        {
+          "name": "database"
+        },
+        {
+          "name": "schemaName"
+        }
+      ]
+    },
+    {
+      "condition": {
+        "typeReached": "org.apache.shardingsphere.mode.node.path.engine.generator.NodePathSegment"
+      },
+      "type": "org.apache.shardingsphere.mode.node.path.type.database.metadata.schema.TableMetaDataNodePath",
+      "fields": [
+        {
+          "name": "schema"
+        },
+        {
+          "name": "tableName"
+        }
+      ]
+    },
+    {
+      "condition": {
         "typeReached": "org.apache.shardingsphere.test.natived.jdbc.transactions.xa.NarayanaTest"
       },
       "type": "org.apache.shardingsphere.test.natived.jdbc.transactions.xa.NarayanaTest",
--- a/test/native/src/test/resources/META-INF/native-image/shardingsphere-test-native/reachability-metadata.json
+++ b/test/native/src/test/resources/META-INF/native-image/shardingsphere-test-native/reachability-metadata.json
@@ -26,45 +26,6 @@
     },
     {
       "condition": {
-        "typeReached": "org.apache.shardingsphere.mode.node.path.engine.generator.NodePathSegment"
-      },
-      "type": "org.apache.shardingsphere.mode.node.path.type.database.metadata.DatabaseMetaDataNodePath",
-      "fields": [
-        {
-          "name": "databaseName"
-        }
-      ]
-    },
-    {
-      "condition": {
-        "typeReached": "org.apache.shardingsphere.mode.node.path.engine.generator.NodePathSegment"
-      },
-      "type": "org.apache.shardingsphere.mode.node.path.type.database.metadata.schema.SchemaMetaDataNodePath",
-      "fields": [
-        {
-          "name": "database"
-        },
-        {
-          "name": "schemaName"
-        }
-      ]
-    },
-    {
-      "condition": {
-        "typeReached": "org.apache.shardingsphere.mode.node.path.engine.generator.NodePathSegment"
-      },
-      "type": "org.apache.shardingsphere.mode.node.path.type.database.metadata.schema.TableMetaDataNodePath",
-      "fields": [
-        {
-          "name": "schema"
-        },
-        {
-          "name": "tableName"
-        }
-      ]
-    },
-    {
-      "condition": {
         "typeReached": "org.apache.shardingsphere.test.natived.jdbc.transactions.xa.NarayanaTest"
       },
       "type": "org.apache.shardingsphere.test.natived.jdbc.transactions.xa.NarayanaTest",
--- a/test/native/src/test/resources/test-native/ps1/wait-for-docker-daemon.ps1
+++ b/test/native/src/test/resources/test-native/ps1/wait-for-docker-daemon.ps1
@@ -18,7 +18,8 @@
 #
 
 # This file is only used in the PowerShell 7 of ShardingSphere in GitHub Actions environment and should not be executed manually in a development environment.
-# Background information can be found at https://github.com/apache/shardingsphere/pull/35905 .
+# Background information for the Windows Server 2025 Rancher Desktop setup can be found at https://github.com/apache/shardingsphere/pull/35905 .
+# Docker daemon readiness instability on affected WSL2 versions may also relate to https://github.com/microsoft/WSL/pull/40519 .
 param (
     [int] $MaxAttempts = 2,
     [int] $TimeoutMinutes = 5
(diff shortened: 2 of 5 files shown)


taojintianxia (dismissed):

test/native/src/test/resources/test-native/ps1/wait-for-docker-daemon.ps1
```diff
@@ -0,0 +1,106 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# This file is only used in the PowerShell 7 of ShardingSphere in GitHub Actions environment and should not be executed manually in a development environment.
+# Background information can be found at https://github.com/apache/shardingsphere/pull/35905 .
```
linghengqian: Huh? You misspelled the URL? I think the instability of Rancher Desktop on Windows Server 2025 is related to a cgroup bug in WSL2. See https://github.com/microsoft/WSL/pull/40519 .

By the way, WSL 2.5.x does not have this bug. Microsoft's infrastructure efforts in recent months have been quite alarming.

test/native/src/test/resources/META-INF/native-image/shardingsphere-test-native/reachability-metadata.json
```diff
@@ -24,6 +24,45 @@
...
+      "type": "org.apache.shardingsphere.mode.node.path.type.database.metadata.schema.TableMetaDataNodePath",
+      "fields": [
+        {
+          "name": "schema"
+        },
+        {
+          "name": "tableName"
+        }
+      ]
+    },
```
linghengqian: Why not include it in https://github.com/apache/shardingsphere/blob/master/infra/reachability-metadata/src/main/resources/META-INF/native-image/org.apache.shardingsphere/shardingsphere-infra-reachability-metadata/reachability-metadata.json ? Also, this JSON should be captured by the GraalVM tracing agent, see the changes in https://github.com/apache/shardingsphere/pull/38682 .

test/native/src/test/resources/test-native/ps1/wait-for-docker-daemon.ps1
RaigorJiang: Thanks for the context. The `#35905` link was intentional because it is the PR that introduced the Windows Server 2025 Rancher Desktop setup for native tests. I agree the WSL2 cgroup issue is useful background for the Docker readiness instability, so I will clarify the comment and add the WSL reference as additional context.

test/native/src/test/resources/META-INF/native-image/shardingsphere-test-native/reachability-metadata.json
RaigorJiang: Thanks, this makes sense. The new entries describe reflection used by `NodePathSegment` against ShardingSphere node path classes, so keeping them only in the test-native metadata is not ideal.

I will re-check this with the `generateMetadata` profile first and move the necessary metadata to the shared reachability metadata output if the tracing agent captures it. If the agent does not produce a stable entry for this path, I will move the minimal manual metadata to the infra reachability metadata file and explain why it is maintained there instead of under `test/native`.

test/native/src/test/resources/test-native/ps1/wait-for-docker-daemon.ps1
linghengqian: Given that this is a bug in WSL, I don't think the contents of this new PowerShell file will be useful unless you intend to downgrade to WSL2 in CI. The failure of CI has already proven that Rancher Desktop is usable, but the Docker daemon fails to start due to a cgroup bug.

RaigorJiang: @linghengqian Yes, so the failure of Nightly CI has to be put on hold for now. Do you have any other suggestions?

linghengqian: For Windows CI to succeed, you can only repeatedly restart CI. I don't think the humans at Microsoft could do anything before they realized they'd messed up WSL. The maintenance of the entire WSL repository was taken over by agents.

RaigorJiang: This is a sad story. Thank you!

linghengqian (commented):
So you just want to update the contents of `ZookeeperTest.java`? You can simply delete all changes except for `ZookeeperTest.java`, merge the PR, and then I can pull the changes from https://github.com/apache/shardingsphere/pull/38682.

Or you can tell me what's wrong with https://github.com/apache/shardingsphere/pull/38682 preventing it from being merged.

RaigorJiang:
> So you just want to update the contents of `ZookeeperTest.java`? You can simply delete all changes except for `ZookeeperTest.java`, merge the PR, and then I can pull the changes from #38682.
> 
> Or you can tell me what's wrong with #38682 preventing it from being merged.

Sorry, I didn't actually pay attention to #38682. I just wanted to try and fix the Nightly CI failure issue.

#38682 may require confirmation from the previous reviewer.

RaigorJiang:
See #38880

PR #38863, opened 2026-06-18
Ignore sort direction when comparing group by and order by items (#23014)

### LINKED ISSUE #23014: SelectStatementContext#isSameGroupByAndOrderByItems has bug
https://github.com/apache/shardingsphere/blob/764c13f2f098d3f13911e31aa93a61f12d49a332/infra/binder/src/main/java/org/apache/shardingsphere/infra/binder/statement/dml/SelectStatementContext.java#L297

### Bug Report
the method SelectStatementContext#isSameGroupByAndOrderByItems is not right;  OrderByItem in SelectStatementContext#groupByContext is default ASC，But OrderByItem in SelectStatementContext#orderByContext  couild be ASC or DESC;So when my sql is desc, it is always return false;

5.1.2

ShardingSphere-JDBC

### Expected behavior
my sql is :
select column1,column2,count(0) as statisticsnum from my_table a  group by column1,column2 order by column1 desc,column2 desc  limit 0,1000
It should be rewrite to :
select column1,column2,count(0) as statisticsnum from my_table_one group by column1,column2 order by column1 desc,column2 desc  limit 0,1000；
select column1,column2,count(0) as statisticsnum from my_table_two group by column1,column2 order by column1 desc,column2 desc  limit 0,1000;

### Actual behavior
It was  rewrited to :
select column1,column2,count(0) as statisticsnum from my_table_one group by column1,column2 order by column1 desc,column2 desc  limit 0,2147483647；
select column1,column2,count(0) as statisticsnum from my_table_two group by column1,column2 order by column1 desc,column2 desc  limit 0,2147483647;

when  PaginationContext#getRevisedRowCount  rewrite the limit part of sql，It use SelectStatementContext#isSameGroupByAndOrderByItems to judge if need to get max row count;but SelectStatementContext#isSameGroupByAndOrderByItems is always return false when orderByContext is desc;

I think SelectStatementContext#isSameGroupByAndOrderByItems should be return true when groupByContext.getItems().size() == groupByContext.getItems() where orderByContext.getItems() is List<ColumnOrderByItemSegment>;

Ayasaz:
Fixes #23014.

  - `SelectStatementBaseContext#isSameGroupByAndOrderByItems` compared the order by item lists via `equals()`, which includes the sort direction (`OrderByItem#equals` factors in `OrderDirection`). As a result a query like `GROUP BY col1, col2 ORDER BY col1 DESC, col2 DESC` was treated as having different group by / order by items.
  - That false negative caused two regressions on the same column sequence with a different direction: pagination was expanded to `LIMIT 0, 2147483647` (full fetch, see `PaginationContext#isMaxRowCount`), and the merge engine fell back from streaming to memory group merge (`ShardingDQLResultMerger#getGroupByMergedResult`).
  - Group by carries no direction by SQL semantics. When order by references the same column sequence the grouping keys stay contiguous, so streaming group merge remains valid and the original pagination can be preserved regardless of asc/desc; the direction is already honored by the merge comparator (`GroupByRowComparator`). The check now compares the item index sequence only and ignores direction.
  - Updated the unit tests that previously relied on same-column / different-direction to force the memory-merge or LIMIT-max path so they use genuinely different items, and added cases covering the fixed behavior.

This follows the root-cause analysis and fix proposal from @terrymanu on the issue.

---

- [ ] I have passed maven check locally : `./mvnw clean install ...`. (Ran `test` for the affected modules `infra/binder/core`, `infra/rewrite/core`, `features/sharding/core` — all green; have not run the full multi-module `clean install`.)
- [ ] I have made corresponding changes to the documentation. (Behavior fix only, no doc change needed.)
- [ ] I have updated the Release Notes of the current development version.


linghengqian (commented):
Please remove claude from the commit submitter list; I don't think a non-human needs to be here.

PR #38864, opened 2026-06-18
Fix SHOW UNLOADED SINGLE TABLES for uppercase loaded table names

### LINKED ISSUE #38862: `show unloaded single tables` will produce inaccurate results when table names contain uppercase letters
5.5.4-SNAPSHOT 96b8227550c0c350785596a8186feccc3bfcc9b5

ShardingSphere-Proxy

### Expected behavior
A single table that has already been loaded should not appear in the result set of the `show unloaded single tables` query.

### Actual behavior
When table names contain uppercase letters, `show unloaded single tables` cannot handle them correctly.

The ShowUnloadedSingleTablesExecutor class contains a toLowerCase conversion.

1. Before load any single tables

Image

2. Show single tables
Image

3. After load, show unloaded single tables

Image

RaigorJiang:
Fixes #38862.

### Type of Change

  - Bug fix

  ### What Changed

  This PR fixes `SHOW UNLOADED SINGLE TABLES` when loaded single table names contain uppercase letters.

  Previously, `ShowUnloadedSingleTablesExecutor` checked whether `actualDataNodes` contained the original loaded table key, but removed the lower-case form of that key. For an
  uppercase table such as `T_ORDER`, the executor matched `T_ORDER` but attempted to remove `t_order`, so the already-loaded table could still appear in the unloaded single
  table result.

  This PR removes the exact matched key instead.

  ### Why This Change Is Needed

  `SHOW UNLOADED SINGLE TABLES` should only return single tables that have not been loaded. A table that has already been loaded should not remain in the result set just
  because its name contains uppercase letters.

  ### Testing

  - `./mvnw -pl kernel/single/distsql/handler -am -DskipITs -Dspotless.skip=true -Dtest=ShowUnloadedSingleTablesExecutorTest -DfailIfNoTests=false
  -Dsurefire.failIfNoSpecifiedTests=false test`
  - `./mvnw -pl kernel/single/distsql/handler checkstyle:check -Pcheck -T1C`
  - `./mvnw -pl kernel/single/distsql/handler spotless:check -Pcheck -T1C`

  ### Coverage

  A regression test was added for the fully-loaded uppercase table case:

  - actual data node key: `T_ORDER`
  - loaded data node key: `T_ORDER`
  - expected unloaded result: empty

  Target class coverage after the change:

  - `ShowUnloadedSingleTablesExecutor`
    - CLASS: `100.00%`
    - LINE: `100.00%`
    - BRANCH: `100.00%`

### DIFF
--- a/kernel/single/distsql/handler/src/main/java/org/apache/shardingsphere/single/distsql/handler/query/ShowUnloadedSingleTablesExecutor.java
+++ b/kernel/single/distsql/handler/src/main/java/org/apache/shardingsphere/single/distsql/handler/query/ShowUnloadedSingleTablesExecutor.java
@@ -71,7 +71,7 @@ public final class ShowUnloadedSingleTablesExecutor implements DistSQLQueryExecu
         for (Entry<String, Collection<DataNode>> entry : rule.getSingleTableDataNodes().entrySet()) {
             if (actualDataNodes.containsKey(entry.getKey())) {
                 if (entry.getValue().containsAll(actualDataNodes.get(entry.getKey()))) {
-                    actualDataNodes.remove(entry.getKey().toLowerCase());
+                    actualDataNodes.remove(entry.getKey());
                     continue;
                 }
                 Collection<DataNode> tableNodes = actualDataNodes.get(entry.getKey());
--- a/kernel/single/distsql/handler/src/test/java/org/apache/shardingsphere/single/distsql/handler/query/ShowUnloadedSingleTablesExecutorTest.java
+++ b/kernel/single/distsql/handler/src/test/java/org/apache/shardingsphere/single/distsql/handler/query/ShowUnloadedSingleTablesExecutorTest.java
@@ -245,6 +245,10 @@ class ShowUnloadedSingleTablesExecutorTest {
                 Collections.singletonMap("t_order", Collections.singleton(new DataNode("ds_0", (String) null, "t_order"))));
         Map<String, Collection<DataNode>> allLoadedRuleDataNodes = new HashMap<>(
                 Collections.singletonMap("t_order", Collections.singleton(new DataNode("ds_0", (String) null, "t_order"))));
+        Map<String, Collection<DataNode>> allLoadedUpperCaseActualDataNodes = new HashMap<>(
+                Collections.singletonMap("T_ORDER", Collections.singleton(new DataNode("ds_0", (String) null, "T_ORDER"))));
+        Map<String, Collection<DataNode>> allLoadedUpperCaseRuleDataNodes = new HashMap<>(
+                Collections.singletonMap("T_ORDER", Collections.singleton(new DataNode("ds_0", (String) null, "T_ORDER"))));
         Map<String, Collection<DataNode>> partiallyLoadedActualDataNodes = new HashMap<>(
                 Collections.singletonMap("t_order", new LinkedList<>(Arrays.asList(new DataNode("ds_0", "public", "t_order"), new DataNode("ds_1", "public", "t_order")))));
         Map<String, Collection<DataNode>> partiallyLoadedRuleDataNodes = new HashMap<>(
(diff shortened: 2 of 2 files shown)


RaigorJiang:
### Summary

  - **Reason:** The PR fixes the exact key-normalization bug reported in #38862 and includes a focused regression test for the uppercase loaded-table path.

  ### Evidence

  - Root-cause fix evidence:
    - Issue #38862 reports that `SHOW UNLOADED SINGLE TABLES` can still return an already-loaded single table when the table name contains uppercase letters.
    - The documented behavior of `SHOW UNLOADED SINGLE TABLES` is to query unloaded single tables, so an already-loaded table should be excluded from the result.
    - In `kernel/single/distsql/handler/src/main/java/org/apache/shardingsphere/single/distsql/handler/query/ShowUnloadedSingleTablesExecutor.java`, the existing logic first
    checks `actualDataNodes.containsKey(entry.getKey())` and then handles fully-loaded tables.
    - The PR changes the fully-loaded removal from `actualDataNodes.remove(entry.getKey().toLowerCase())` to `actualDataNodes.remove(entry.getKey())`, so the key used for
    removal now matches the key used for lookup.
    - This directly fixes the reported trigger: an uppercase key such as `T_ORDER` is no longer matched and then removed as `t_order`.
  - Linked issue completeness:
    - PR #38864 claims `Fixes #38862`.
    - The issue scope is Proxy / Single DistSQL behavior for uppercase table names in `SHOW UNLOADED SINGLE TABLES`.
    - The PR maps the issue’s required behavior to one production change and one regression validation point: a fully-loaded uppercase table is removed from the unloaded
    result.
    - There are no issue comments adding extra topology, database-version, configuration, or follow-up requirements beyond the reported uppercase loaded-table behavior.
  - Test adequacy:
    - `kernel/single/distsql/handler/src/test/java/org/apache/shardingsphere/single/distsql/handler/query/ShowUnloadedSingleTablesExecutorTest.java` adds the `all loaded
    uppercase tables are excluded` scenario.
    - The new case uses uppercase actual and loaded data-node keys, both `T_ORDER`, and expects an empty unloaded result.
    - Existing adjacent cases continue covering lowercase fully-loaded tables, partially-loaded table nodes, unmatched rule keys, and result ordering by table, storage unit,
    and schema.
  - Scope and ownership:
    - The PR changes only the Single DistSQL handler production class and its existing unit test.
    - No parser grammar, SQL syntax, public API/SPI contract, metadata storage format, configuration key, dependency, distribution artifact, or diagnostic output is changed.
    - The fix stays in the owning feature module and does not move Single-specific behavior into shared infrastructure.
  - Regression and side-effect review:
    - Lowercase behavior is preserved because the same original key is used for both lookup and removal.
    - Partial-load behavior is unchanged because the `removeIf` branch is untouched.
    - Unmatched loaded keys remain unaffected because the existing `containsKey` guard is unchanged.
    - The patch does not add loops, remote calls, blocking I/O, fallback behavior, hidden state, or high-frequency DML/DQL `ConcurrentHashMap#computeIfAbsent` usage.
  - Release and documentation impact:
    - No release note is required because this is a narrow bug fix restoring the already documented meaning of `SHOW UNLOADED SINGLE TABLES`.
    - No user documentation update is required because the SQL syntax, return columns, configuration, and user workflow do not change.

  ### Review Details

  - **Reviewed Scope:** PR #38864 latest head `ce1eaabbf2bcdea4f32e772ed03b0c1d7ae8a082`, base `master` at `96b8227550c0c350785596a8186feccc3bfcc9b5`, local merge-base
  `96b8227550c0c350785596a8186feccc3bfcc9b5`. The local triple-dot diff file list matched GitHub `/pulls/38864/files`. Reviewed files:
    - `kernel/single/distsql/handler/src/main/java/org/apache/shardingsphere/single/distsql/handler/query/ShowUnloadedSingleTablesExecutor.java`
    - `kernel/single/distsql/handler/src/test/java/org/apache/shardingsphere/single/distsql/handler/query/ShowUnloadedSingleTablesExecutorTest.java`
  - **Not Reviewed Scope:** GitHub Actions / CI status, full repository build, live Proxy manual reproduction, E2E tests, cluster governance behavior, and unrelated modules
  outside the two PR files.
  - **Verification:** Reviewer-run local verification:
    - `./mvnw -pl kernel/single/distsql/handler -am -DskipITs -Dspotless.skip=true -Dtest=ShowUnloadedSingleTablesExecutorTest -DfailIfNoTests=false
    -Dsurefire.failIfNoSpecifiedTests=false test` exited `0`; `Tests run: 11, Failures: 0, Errors: 0, Skipped: 0`.
    - `./mvnw -pl kernel/single/distsql/handler checkstyle:check -Pcheck -T1C` exited `0`.
    - `./mvnw -pl kernel/single/distsql/handler spotless:check -Pcheck -T1C` exited `0`.
  - **Release Note / User Docs:** Not required. The PR restores existing documented behavior for `SHOW UNLOADED SINGLE TABLES` and does not introduce new syntax,
  configuration, migration steps, diagnostics, or operator-facing behavior.

taojintianxia (approved):

PR #38897, opened 2026-06-23
Fix temporal literal rendering in insert values

iamhucong:
### Motivation

`TemporalLiteralExpressionSegment` keeps the raw literal value for consumers while preserving the original SQL literal text, such as `DATE '2017-08-08'`, through `getText()`.

`InsertValue` renders insert value expressions back to SQL. Since `TemporalLiteralExpressionSegment` extends `LiteralExpressionSegment`, it could otherwise fall through the generic literal formatter and render a date literal as a plain string literal.

### Changes

- Return `TemporalLiteralExpressionSegment#getText()` before generic literal formatting in `InsertValue`.
- Add a focused unit test for preserving `DATE '2017-08-08'` in insert values.

### Verification

- `./mvnw -pl infra/rewrite/core -am -DskipITs -Dspotless.skip=true -Dtest=InsertValueTest -Dsurefire.failIfNoSpecifiedTests=false test`


iamhucong:
### Summary

   **Review Result: Mergeable**

   Reason: The PR fixes the shared SQL rewrite rendering point directly. `TemporalLiteralExpressionSegment` is already a `LiteralExpressionSegment`, but it also preserves the original SQL text through `getText()`. Returning
 that text before generic literal formatting in `InsertValue` prevents temporal literals such as `DATE '2017-08-08'` from being rendered as ordinary string literals, without changing parameter markers, non-temporal literals,
 or other expression rendering.

Verification:

- `./mvnw -pl infra/rewrite/core -am -DskipITs -Dspotless.skip=true -Dtest=InsertValueTest -Dsurefire.failIfNoSpecifiedTests=false test`

strongduanmu (approved):

PR #38927, opened 2026-06-26
Improve error and exception handling in Firebird Proxy

makssent:
  - Add Firebird dialect protocol exceptions and map them to native gdscodes / SQL states:                                                                                                                                          
    `InvalidStatementHandleException`, `InvalidTransactionHandleException`,                                                                                                                                                         
    `InvalidBatchHandleException`, `BatchAlreadyOpenedException`, `BatchTooBigException`,                                                                                                                                           
    `InvalidBatchMessageFormatException`, `InvalidBatchParameterVersionException`,                                                                                                                                                  
    `InvalidBlobHandleException`, `InvalidBlobIdException`, plus `TABLE_EXISTS`,                                                                                                                                                    
    `COLUMN_UNKNOWN`, `DYNAMIC_SQL_ERROR` and `CHARSET_NOT_FOUND` errors.                                                                                                                                                           
  - Replace generic protocol errors in command executors (batch create/send/execute,                                                                                                                                                
    blob open/close/cancel/put-segment, execute statement, commit/rollback transaction)                                                                                                                                             
    with the proper Firebird dialect exceptions.                                                                                                                                                                                    
  - Handle codec-level decode/encode errors by emitting a Firebird error response packet                                                                                                                                            
    (and closing the channel on framing corruption) instead of letting raw exceptions propagate.                                                                                                                                    
  - Add batch error handling: translate `BatchUpdateException` into a batch completion                                                                                                                                              
    state response carrying per-element detailed errors (halt-at-first-error semantics).                                                                                                                                            
  - Add `FirebirdErrorSegmentResolver` and improve `FirebirdStatusVector` rendering of                                                                                                                                              
    error messages and SQL state from native error codes.                                                                                                                                                                           
  - Map invalid attach `lc_ctype` to `CHARSET_NOT_FOUND`.                                                                                                                                                                           
  - Parser: add `UNION` (select-combine) support and `CREATE DATABASE` / `DROP DATABASE`.                                                                                                                                           
  - Add unit tests for all new exceptions and the updated command executors.
                                                                                                                                                                                                                        
  ---                                                                                                                                                                                                                               
                                                                                                                                                                                                                                    

terrymanu (changes requested):
### Summary

**Reason:** The PR improves Firebird error/protocol handling, but it still leaves protocol-visible lifecycle bugs and Firebird parser semantics that conflict with official Firebird behavior. These are user-visible correctness issues and should be fixed before merge.

### Issues

#### P1 Blob handles remain usable after close

- Problem: `FirebirdCloseBlobCommandExecutor` closes the upload, but `FirebirdBlobUploadCache.closeUpload` only marks it closed and keeps the handle mapping. `getBlobId` still returns the blob id, so `FirebirdPutBlobSegmentCommandExecutor` and `FirebirdCancelBlobCommandExecutor` continue to accept the same handle after close.
  - `proxy/frontend/dialect/firebird/src/main/java/org/apache/shardingsphere/proxy/frontend/firebird/command/query/blob/executors/FirebirdCloseBlobCommandExecutor.java:44`
  - `proxy/frontend/dialect/firebird/src/main/java/org/apache/shardingsphere/proxy/frontend/firebird/command/query/blob/upload/FirebirdBlobUploadCache.java:102`
  - `proxy/frontend/dialect/firebird/src/main/java/org/apache/shardingsphere/proxy/frontend/firebird/command/query/blob/executors/FirebirdPutBlobSegmentCommandExecutor.java:44`
- Impact: A closed blob handle can still accept more segments or be cancelled successfully. Firebird wire protocol documents blob close as invalidating the blob handle, so this returns success for operations that should fail with an invalid blob handle.
- Required Change: Invalidate/remove the blob handle on close while preserving any data needed for later parameter binding, and add tests for close followed by put/close/cancel on the same handle.

#### P1 Transaction validation accepts stale or already-ended handles

- Problem: commit and rollback validation only rejects handles less than or equal to zero or greater than the maximum generated id. `FirebirdTransactionIdGenerator` tracks a monotonic counter, not active transaction state.
  - `proxy/frontend/dialect/firebird/src/main/java/org/apache/shardingsphere/proxy/frontend/firebird/command/query/transaction/FirebirdCommitTransactionCommandExecutor.java:44`
  - `proxy/frontend/dialect/firebird/src/main/java/org/apache/shardingsphere/proxy/frontend/firebird/command/query/transaction/FirebirdRollbackTransactionCommandExecutor.java:44`
  - `proxy/frontend/dialect/firebird/src/main/java/org/apache/shardingsphere/proxy/frontend/firebird/command/query/transaction/FirebirdTransactionIdGenerator.java:61`
- Impact: After transaction 1 is committed or rolled back, or after transaction 2 is created, transaction 1 can still pass validation. Double commit, double rollback, or operations on an old handle can be incorrectly accepted instead of returning Firebird invalid transaction handle semantics.
- Required Change: Track active transaction handles and invalidate them on commit/rollback. Add tests for double commit, double rollback, and reuse of an older handle after a newer handle has been issued.

#### P1 Firebird CREATE/DROP DATABASE parser semantics are incorrect

- Problem: The Firebird grammar accepts `CREATE DATABASE foo`, `DROP DATABASE foo`, and `DROP SCHEMA foo`, and the visitor derives a database name from a path string. Official Firebird syntax uses a quoted filespec for `CREATE DATABASE` / `CREATE SCHEMA`, while `DROP DATABASE` drops the current database and takes no name. `DROP SCHEMA` is not a valid Firebird alias for `DROP DATABASE`.
  - `parser/sql/engine/dialect/firebird/src/main/antlr4/imports/firebird/DDLStatement.g4:192`
  - `parser/sql/engine/dialect/firebird/src/main/java/org/apache/shardingsphere/sql/parser/engine/firebird/visitor/statement/type/FirebirdDDLStatementVisitor.java:270`
- Impact: The parser accepts invalid Firebird SQL and rejects valid `DROP DATABASE`. This also exposes an AST contract that does not match Firebird behavior.
- Required Change: Align grammar and visitor behavior with Firebird syntax: `CREATE DATABASE` / `CREATE SCHEMA` should take the quoted filespec, and `DROP DATABASE` should take no name and should not accept `SCHEMA`. Add Firebird parser IT cases for `CREATE DATABASE '...'`, `CREATE SCHEMA '...'`, and `DROP DATABASE`.

#### P2 Firebird UNION ALL behavior is changed without a Firebird parser fixture

- Problem: `FirebirdDMLStatementVisitor` now maps combine statements containing `ALL` to `UNION_ALL`, but the added Firebird fixture only covers plain `UNION`.
  - `parser/sql/engine/dialect/firebird/src/main/java/org/apache/shardingsphere/sql/parser/engine/firebird/visitor/statement/type/FirebirdDMLStatementVisitor.java:266`
  - `test/it/parser/src/main/resources/sql/supported/dml/select-combine.xml:22`
  - `test/it/parser/src/main/resources/case/dml/select-combine.xml:74`
- Impact: Firebird-specific `UNION ALL` parsing can regress without being caught by parser ITs.
- Required Change: Add a Firebird `UNION ALL` parser fixture with expected `combine-type="UNION_ALL"`.

### Review Details

- Reviewed Scope: PR #38927 at head `c89bb19af14f9d1fe2db2194ea6ab5d7787d9d63`, base `master` at `0023d933d18fddbfdfd9c8e17eb60437817cb253`, all 67 changed files. Reviewed Firebird exception mapping, protocol error/status vector handling, blob lifecycle, transaction lifecycle, batch handling, statement execution, charset/authentication paths, and parser changes.
- SQL Parser Scope: Target dialect is Firebird. Related repository references checked were SQL92 DDL grammar inheritance and Oracle `DROP DATABASE` no-name AST handling. Official documentation checked: [Firebird 5.0 DDL](https://firebirdsql.org/file/documentation/chunk/en/refdocs/fblangref50/fblangref50-ddl.html), [Firebird 5.0 DML](https://www.firebirdsql.org/file/documentation/chunk/en/refdocs/fblangref50/fblangref50-dml.html), and [Firebird Wire Protocol](https://firebirdsql.org/file/documentation/html/en/firebirddocs/wireprotocol/firebird-wire-protocol.html).
- Not Reviewed Scope: I did not run a live Firebird server, Jaybird native-client smoke test, packet capture, or local Maven test suite. I also did not review code outside the 67-file PR diff except adjacent lifecycle/parser references needed to validate the changed behavior.
- Verification: Confirmed GitHub PR metadata, head SHA, and 67/67 changed-file parity against the local triple-dot diff. Rebuilt the review inventory. Inspected changed protocol/parser code, related tests, official Firebird docs, and current GitHub check-run status. GitHub checks for the reviewed head were 29 successful and 4 skipped.
- Release Note / User Docs: No release note is required for the exception-mapping/internal protocol fixes themselves. If this PR intentionally changes Firebird parser support for database DDL, the parser fixtures should be corrected first; user-facing documentation only needs updates if ShardingSphere already documents that SQL surface.

terrymanu (changes requested):
### Summary

**Reason:** The direction is useful, but the current head still has six protocol- or query-visible correctness boundaries that can lose accepted batch state, misidentify failures, corrupt transaction ownership, or change valid Firebird SQL semantics.

### Issues

#### P1 Logical transaction handles do not own backend transactions

- **Problem:** [`FirebirdTransactionIdGenerator`](https://github.com/apache/shardingsphere/blob/ea40870c2f6d40bc7f83282220991752f877b5be/proxy/frontend/dialect/firebird/src/main/java/org/apache/shardingsphere/proxy/frontend/firebird/command/query/transaction/FirebirdTransactionIdGenerator.java#L38-L89) tracks multiple active IDs, but [`FirebirdStartTransactionCommandExecutor`](https://github.com/apache/shardingsphere/blob/ea40870c2f6d40bc7f83282220991752f877b5be/proxy/frontend/dialect/firebird/src/main/java/org/apache/shardingsphere/proxy/frontend/firebird/command/query/transaction/FirebirdStartTransactionCommandExecutor.java#L41-L46) and commit/rollback still operate on one connection-wide backend transaction. Starting handles 1 and 2, then committing 1, leaves 2 marked active after its shared backend transaction has ended.
- **Impact:** Handle 2 can subsequently “commit” successfully without owning a transaction. Firebird supports multiple transactions per attachment, so Set membership alone does not implement handle ownership.
- **Required Change:** Either implement backend transaction state keyed by Firebird handle and route statements/commit/rollback through it, or reject a second concurrent start. Add overlapping-handle tests.

#### P1 Routed batch failures are serialized in the wrong domain

- **Problem:** [`FirebirdBatchedStatementsExecutor`](https://github.com/apache/shardingsphere/blob/ea40870c2f6d40bc7f83282220991752f877b5be/proxy/frontend/dialect/firebird/src/main/java/org/apache/shardingsphere/proxy/frontend/firebird/command/query/batch/FirebirdBatchedStatementsExecutor.java#L134-L147) records original message indexes for successful routed batches, but a route-local `BatchUpdateException` escapes without that mapping. [`FirebirdBatchExecuteCommandExecutor`](https://github.com/apache/shardingsphere/blob/ea40870c2f6d40bc7f83282220991752f877b5be/proxy/frontend/dialect/firebird/src/main/java/org/apache/shardingsphere/proxy/frontend/firebird/command/query/batch/FirebirdBatchExecuteCommandExecutor.java#L68-L98) then treats local offsets as global client indexes, reports the total queued count instead of the processed count, and can serialize JDBC `EXECUTE_FAILED` (`-3`) instead of Firebird’s `-1`.
- **Impact:** A routed failure can be attributed to the wrong client row and carry incompatible completion counts/status values.
- **Required Change:** Preserve execution-unit failure context, translate it back to original client indexes, and construct a Firebird-domain completion result before serialization. Cover a multi-route failure.

#### P1 Oversized `BATCH_MSG` clears previously accepted rows

- **Problem:** [`FirebirdBatchSendMessageCommandExecutor`](https://github.com/apache/shardingsphere/blob/ea40870c2f6d40bc7f83282220991752f877b5be/proxy/frontend/dialect/firebird/src/main/java/org/apache/shardingsphere/proxy/frontend/firebird/command/query/batch/FirebirdBatchSendMessageCommandExecutor.java#L47-L62) calls `batchStatement.reset()` before returning `BatchTooBigException`. The batch remains registered and the connection remains open, so a later execute succeeds with an empty batch. Native Firebird checks the limit before mutating stored batch data ([Firebird 5.0.4 source](https://github.com/FirebirdSQL/firebird/blob/v5.0.4/src/dsql/DsqlBatch.cpp#L850-L854)).
- **Impact:** A recoverable oversized addition silently deletes earlier accepted rows; the following execute can report success without executing them.
- **Required Change:** Reject only the oversized addition and preserve existing batch state. Add an accepted-send → rejected-send → execute regression.

#### P1 Batch-create errors lose native Firebird identities

- **Problem:** [`FirebirdDialectExceptionMapper`](https://github.com/apache/shardingsphere/blob/ea40870c2f6d40bc7f83282220991752f877b5be/database/exception/dialect/firebird/src/main/java/org/apache/shardingsphere/database/exception/firebird/mapper/FirebirdDialectExceptionMapper.java#L61-L75) passes only the received version to the two-argument `isc_batch_param_version` template. [`FirebirdErrorSegmentResolver`](https://github.com/apache/shardingsphere/blob/ea40870c2f6d40bc7f83282220991752f877b5be/database/protocol/dialect/firebird/src/main/java/org/apache/shardingsphere/database/protocol/firebird/err/FirebirdErrorSegmentResolver.java#L55-L68) therefore downgrades it to `isc_random`. The batch-create path also maps both missing/mismatched metadata and invalid message length to `SQLDA_ERROR`, although Firebird distinguishes `isc_batch_param` from `isc_dsql_sqlda_err` ([official error catalog](https://firebirdsql.org/file/documentation/chunk/en/refdocs/fblangref50/fblangref50-appx02-sqlcodes.html), [Firebird source](https://github.com/FirebirdSQL/firebird/blob/v4.0.7/src/dsql/DsqlBatch.cpp#L236-L241)).
- **Impact:** Clients receive incorrect GDS codes and cannot reliably classify batch-creation failures.
- **Required Change:** Preserve structured template arguments and distinct failure identities through the final status vector. Test all conditions at the packet boundary.

#### P1 Newly parsed database DDL cannot execute through the generic handlers

- **Problem:** [`FirebirdDDLStatementVisitor`](https://github.com/apache/shardingsphere/blob/ea40870c2f6d40bc7f83282220991752f877b5be/parser/sql/engine/dialect/firebird/src/main/java/org/apache/shardingsphere/sql/parser/engine/firebird/visitor/statement/type/FirebirdDDLStatementVisitor.java#L269-L284) correctly emits a name-less `DROP DATABASE`, but [`DropDatabaseProxyBackendHandler`](https://github.com/apache/shardingsphere/blob/ea40870c2f6d40bc7f83282220991752f877b5be/proxy/backend/core/src/main/java/org/apache/shardingsphere/proxy/backend/handler/database/type/DropDatabaseProxyBackendHandler.java#L52-L71) immediately calls `getDatabaseName().toLowerCase()`. Valid Firebird `DROP DATABASE` therefore throws `NullPointerException`. Documented remote CREATE syntax such as `CREATE DATABASE 'baseserver:test'` is also accepted by the parser but rejected later by the generic `:` name validation ([Firebird DDL reference](https://www.firebirdsql.org/file/documentation/chunk/en/refdocs/fblangref50/fblangref50-ddl.html)).
- **Impact:** The new syntax support parses valid SQL that cannot execute.
- **Required Change:** Resolve current-database and filespec semantics before entering generic handlers, or add a Firebird-specific execution boundary. Add execution-level tests for name-less DROP and supported/rejected remote CREATE forms.

#### P1 UNION AST construction changes valid query semantics

- **Problem:** [`FirebirdDMLStatementVisitor`](https://github.com/apache/shardingsphere/blob/ea40870c2f6d40bc7f83282220991752f877b5be/parser/sql/engine/dialect/firebird/src/main/java/org/apache/shardingsphere/sql/parser/engine/firebird/visitor/statement/type/FirebirdDMLStatementVisitor.java#L265-L284) searches raw source text for `ALL`. Consequently, `UNION /* ALL */ SELECT` becomes `UNION_ALL`, even though Firebird ignores comment text. The grammar also attaches a trailing `ORDER BY` to the final branch, while the visitor does not propagate it to the combined statement. Firebird requires that order to apply to the entire union ([comments](https://www.firebirdsql.org/file/documentation/chunk/en/refdocs/fblangref50/fblangref50-structure-comments.html), [UNION ordering](https://www.firebirdsql.org/file/documentation/chunk/en/refdocs/fblangref50/fblangref50-dml.html)).
- **Impact:** SQL federation can retain duplicates for plain UNION or order only the right operand.
- **Required Change:** Derive operators from parser tokens/context and model trailing clauses at the combined-query level. Add commented UNION and global `ORDER BY` fixtures.

### Multi-Round Comparison

- The previous blob-handle finding is withdrawn: blob paths were removed from the current PR scope.
- Sole-handle double commit/rollback validation is fixed, but overlapping handles remain unsupported while still being accepted.
- Firebird CREATE/DROP grammar syntax is corrected; the parser-to-execution integration remains broken.
- The missing `UNION ALL` fixture is fixed; comment and global-order semantics remain uncovered.

### Review Details

- **Review Focus:** Code Correctness Review.
- **Reviewed Scope:** Current head `ea40870c2f6d40bc7f83282220991752f877b5be`, base/merge-base `2e4bc504ed792cee62ae56fbb511c11c5bbbf470`, all 60 files: 30 exception/protocol, 20 proxy, and 10 parser/fixture files.
- **Scope Integrity:** GitHub file list and local triple-dot diff match 60/60; current head was rechecked before finalizing.
- **Verification:** Seven changed parser XML files passed `xmllint`; focused exact-head parser probes reproduced both UNION defects; the full-coverage ledger validated with 29 production files and 31 test/fixture files reviewed.
- **Not Reviewed:** Live Firebird/native-client smoke, packet capture, and Maven suites were not run because the shared worktree is on an unrelated dirty commit. **CI not reviewed by request.**
- **PR Description:** Please remove the blob-change claims that no longer exist in the current diff.

terrymanu (changes requested):
## Review Result

**Not Mergeable — Change Request**

### P1: Batch execution does not actually stop at the first error

The code catches `BatchUpdateException` and continues collecting results from other routed executions. A later write may succeed, while the response reports that execution stopped at an earlier failure.

- [Error handling](https://github.com/apache/shardingsphere/blob/da58d88bfdf7af5228c382da11ac854f223e4659/proxy/frontend/dialect/firebird/src/main/java/org/apache/shardingsphere/proxy/frontend/firebird/command/query/batch/FirebirdBatchedStatementsExecutor.java#L302-L317)
- [Completion result](https://github.com/apache/shardingsphere/blob/da58d88bfdf7af5228c382da11ac854f223e4659/proxy/frontend/dialect/firebird/src/main/java/org/apache/shardingsphere/proxy/frontend/firebird/command/query/batch/FirebirdBatchedStatementsExecutor.java#L212-L239)

**Required:** Ensure later batch messages are not executed after the first failure, or report every message that was actually executed. Add a test where the first route fails and a later route would otherwise succeed.

### P2: The new Firebird UNION test uses invalid SQL

The PR marks `UNION ... ORDER BY COL1` as supported Firebird SQL:

- [Parser fixture](https://github.com/apache/shardingsphere/blob/da58d88bfdf7af5228c382da11ac854f223e4659/test/it/parser/src/main/resources/sql/supported/dml/select-combine.xml#L20-L27)

Firebird requires an ordinal for a UNION-level `ORDER BY`, such as `ORDER BY 1`.

**Required:** Replace `ORDER BY COL1` with `ORDER BY 1` and update the expected parser result.

CI not reviewed by request.

terrymanu (changes requested):
### Correction

- **Previous Finding:** Firebird requires an ordinal for a UNION-level `ORDER BY`.
- **Reason:** The [Firebird 5.0 Language Reference](https://www.firebirdsql.org/file/documentation/html/en/refdocs/fblangref50/firebird-50-language-reference.html) explicitly allows column names or aliases from the first contributing query. The current `ORDER BY 1` fixture is valid, but the ordinal-only requirement was incorrect.

### Summary

**Reason:** The latest head improves batch failure mapping and transaction lifecycle tracking, but two protocol-visible correctness gaps remain: batch execution ignores Firebird’s requested error mode, and statement/batch execution ignores the transaction handle supplied by the client.

### Issues

#### P1 Firebird batch error mode is not honored

- **Problem:** [`FirebirdBatchCreateCommandExecutor`](https://github.com/apache/shardingsphere/blob/a4c795274a7aaea04aaa4eadc0d88a9d73c00cbb/proxy/frontend/dialect/firebird/src/main/java/org/apache/shardingsphere/proxy/frontend/firebird/command/query/batch/FirebirdBatchCreateCommandExecutor.java#L52-L58) does not recognize `TAG_MULTIERROR`; unknown tags are simply [skipped](https://github.com/apache/shardingsphere/blob/a4c795274a7aaea04aaa4eadc0d88a9d73c00cbb/proxy/frontend/dialect/firebird/src/main/java/org/apache/shardingsphere/proxy/frontend/firebird/command/query/batch/FirebirdBatchCreateCommandExecutor.java#L122-L138). The execution callback then converts `BatchUpdateException` into a normal result and permits other routed units to continue. Completion construction also marks every same-route offset beyond the JDBC update-count array as failed, although those trailing messages were not necessarily executed ([completion mapping](https://github.com/apache/shardingsphere/blob/a4c795274a7aaea04aaa4eadc0d88a9d73c00cbb/proxy/frontend/dialect/firebird/src/main/java/org/apache/shardingsphere/proxy/frontend/firebird/command/query/batch/FirebirdBatchedStatementsExecutor.java#L214-L255), [exception path](https://github.com/apache/shardingsphere/blob/a4c795274a7aaea04aaa4eadc0d88a9d73c00cbb/proxy/frontend/dialect/firebird/src/main/java/org/apache/shardingsphere/proxy/frontend/firebird/command/query/batch/FirebirdBatchedStatementsExecutor.java#L313-L317)). The latest test explicitly expects a later route to execute after the first route fails ([test](https://github.com/apache/shardingsphere/blob/a4c795274a7aaea04aaa4eadc0d88a9d73c00cbb/proxy/frontend/dialect/firebird/src/test/java/org/apache/shardingsphere/proxy/frontend/firebird/command/query/batch/FirebirdBatchedStatementsExecutorTest.java#L203-L237)).
- **Impact:** Clients using the default mode can observe writes that native Firebird would have stopped before, while same-route messages that JDBC never executed can be reported as failures. This makes the completion vector unreliable.
- **Required Change:** Parse and retain `TAG_MULTIERROR`. When it is absent or false, halt at the first failed client message and report counts only through that failure. When true, continue and report every actually executed message. Add same-route and multi-route tests for both modes. This matches Jaybird 5.0.10’s [halt/continue contract](https://github.com/FirebirdSQL/jaybird/blob/v5.0.10/src/main/org/firebirdsql/gds/ng/FbBatchConfig.java#L47-L53) and [completion semantics](https://github.com/FirebirdSQL/jaybird/blob/v5.0.10/src/main/org/firebirdsql/gds/ng/BatchCompletion.java#L64-L89).

#### P1 Statement and batch execution ignore the supplied transaction handle

- **Problem:** The wire packets retain a transaction handle for [statement execution](https://github.com/apache/shardingsphere/blob/a4c795274a7aaea04aaa4eadc0d88a9d73c00cbb/database/protocol/dialect/firebird/src/main/java/org/apache/shardingsphere/database/protocol/firebird/packet/command/query/statement/execute/FirebirdExecuteStatementPacket.java#L41-L65) and [batch execution](https://github.com/apache/shardingsphere/blob/a4c795274a7aaea04aaa4eadc0d88a9d73c00cbb/database/protocol/dialect/firebird/src/main/java/org/apache/shardingsphere/database/protocol/firebird/packet/command/query/batch/FirebirdBatchExecuteCommandPacket.java#L27-L35), as required by the [Firebird wire protocol](https://firebirdsql.org/file/documentation/html/en/firebirddocs/wireprotocol/firebird-wire-protocol.html). However, neither [`FirebirdExecuteStatementCommandExecutor`](https://github.com/apache/shardingsphere/blob/a4c795274a7aaea04aaa4eadc0d88a9d73c00cbb/proxy/frontend/dialect/firebird/src/main/java/org/apache/shardingsphere/proxy/frontend/firebird/command/query/statement/execute/FirebirdExecuteStatementCommandExecutor.java#L73-L99) nor [`FirebirdBatchExecuteCommandExecutor`](https://github.com/apache/shardingsphere/blob/a4c795274a7aaea04aaa4eadc0d88a9d73c00cbb/proxy/frontend/dialect/firebird/src/main/java/org/apache/shardingsphere/proxy/frontend/firebird/command/query/batch/FirebirdBatchExecuteCommandExecutor.java#L43-L61) reads or validates it.
- **Impact:** After transaction 1 is closed and transaction 2 is started, an execute request carrying stale handle 1 runs against the connection-wide backend transaction for handle 2 instead of returning `InvalidTransactionHandleException`. A stale client handle can therefore read or modify data in a different logical transaction.
- **Required Change:** Before backend execution, validate or bind the supplied handle according to the active Firebird transaction state. Add statement and batch regressions for start 1 → commit/rollback 1 → start 2 → execute using stale handle 1, while preserving any valid no-transaction/auto-commit behavior defined by the protocol.

### Multi-Round Comparison

- Batch failure-to-client-index mapping is improved, but the requested/default error mode remains unimplemented.
- Oversized batch sends now preserve previously accepted messages.
- Concurrent logical transactions are now rejected, resolving the connection-wide ownership conflict, but execution still accepts stale handles.
- Database DDL execution and UNION comment/global-order handling are fixed.
- The prior ordinal-only UNION `ORDER BY` finding is withdrawn as documented above.

### Review Details

- **Reviewed Scope:** Current head `a4c795274a7aaea04aaa4eadc0d88a9d73c00cbb`, base and local merge-base `2e4bc504ed792cee62ae56fbb511c11c5bbbf470`, all 71 changed files: 36 production/grammar files and 35 tests/parser fixtures. GitHub’s file list matched the local triple-dot diff 71/71.
- **SQL Parser Scope:** Target dialect Firebird; related SQL92 combine grammar and Oracle name-less `DROP DATABASE` AST handling were checked. Syntax was compared with the official [Firebird 5.0 DDL reference](https://www.firebirdsql.org/file/documentation/chunk/en/refdocs/fblangref50/fblangref50-ddl.html), [language reference](https://www.firebirdsql.org/file/documentation/html/en/refdocs/fblangref50/firebird-50-language-reference.html), and [wire protocol](https://firebirdsql.org/file/documentation/html/en/firebirddocs/wireprotocol/firebird-wire-protocol.html).
- **Verification:** Exact-head source and call-site scans completed; all seven changed parser XML files passed `xmllint` with exit code 0. The full-coverage ledger validated after two adversarial passes, with zero new findings in the final pass. Read-only GitHub REST metadata, file, commit, and review queries completed without an evidence gap.
- **Not Reviewed Scope:** Live Firebird/native-client smoke, packet capture, and local Maven suites were not run because the shared checkout contains unrelated uncommitted changes.
- **Release Note / User Docs:** No release note or user documentation is required because these are internal compatibility/error-handling changes without configuration or migration requirements.

terrymanu (changes requested):
### Summary

**Reason:** One merge blocker remains. The latest commit attempts to implement halt-on-first-error semantics for `multiError=false`, but only handles the case with multiple execution units. Within a single execution unit, statements after the first failure may still execute while the response reports them as unprocessed. This can make the client-visible completion state inconsistent with the actual data state.

### Issues

#### P1 A single execution unit can execute and hide writes after the first failure

- **Problem:** [`isHaltedOnFirstFailure()`](https://github.com/apache/shardingsphere/blob/e2dcd969533463cf8319873df9678817ba81e434/proxy/frontend/dialect/firebird/src/main/java/org/apache/shardingsphere/proxy/frontend/firebird/command/query/batch/FirebirdBatchedStatementsExecutor.java#L393-L395) returns `true` only when `multiError=false` and more than one execution unit exists. When all client messages belong to one execution unit, the execution plan combines them into one JDBC batch ([execution plan](https://github.com/apache/shardingsphere/blob/e2dcd969533463cf8319873df9678817ba81e434/proxy/frontend/dialect/firebird/src/main/java/org/apache/shardingsphere/proxy/frontend/firebird/command/query/batch/FirebirdBatchedStatementsExecutor.java#L343-L390)) and submits them through one `PreparedStatement.executeBatch()` call ([execution path](https://github.com/apache/shardingsphere/blob/e2dcd969533463cf8319873df9678817ba81e434/proxy/frontend/dialect/firebird/src/main/java/org/apache/shardingsphere/proxy/frontend/firebird/command/query/batch/FirebirdBatchedStatementsExecutor.java#L431-L446)). After a failure, the completion response is then truncated at the first failed message ([completion calculation](https://github.com/apache/shardingsphere/blob/e2dcd969533463cf8319873df9678817ba81e434/proxy/frontend/dialect/firebird/src/main/java/org/apache/shardingsphere/proxy/frontend/firebird/command/query/batch/FirebirdBatchedStatementsExecutor.java#L305-L320)).

- **Impact:** ShardingSphere uses MySQL Connector/J 8.4.0 ([dependency version](https://github.com/apache/shardingsphere/blob/e2dcd969533463cf8319873df9678817ba81e434/pom.xml#L129)), where `continueBatchOnError` defaults to `true` ([official documentation](https://dev.mysql.com/doc/connector-j/en/connector-j-connp-props-statements.html)). For example, a single execution unit may contain:

  1. `INSERT A`, which succeeds;
  2. `INSERT B`, which fails;
  3. `INSERT C`, which succeeds.

  The JDBC driver may continue and execute the third statement after the second one fails, while the Firebird completion reports only the first two messages. If the client subsequently commits the transaction, `INSERT C` is persisted even though the response represented it as unprocessed.

- **Regression Origin:** Submitting same-route parameters in one JDBC batch predates this change. Before the latest commit, however, later successful update counts returned by the driver were still included in the completion. Commit [`e2dcd969533`](https://github.com/apache/shardingsphere/commit/e2dcd969533463cf8319873df9678817ba81e434) introduced first-failure truncation without preventing continued execution within one execution unit. Therefore, the specific inconsistency where an executed statement is hidden from the response is introduced by the latest change and should not be deferred as a pre-existing limitation.

- **Required Change:** When `multiError=false`, execution must follow client-message order regardless of the number of execution units:

  1. Do not combine consecutive client messages from the same execution unit into one JDBC batch.
  2. Submit the next client message only after the preceding message succeeds.
  3. After the first failure, do not submit any later message to any execution unit.
  4. Preserve the current continue-on-error behavior for `multiError=true`.

  Please add a regression test with three messages in one execution unit, make the second message fail, and verify that the third message is never added or executed. The current test contains only two messages with the second one failing ([existing test](https://github.com/apache/shardingsphere/blob/e2dcd969533463cf8319873df9678817ba81e434/proxy/frontend/dialect/firebird/src/test/java/org/apache/shardingsphere/proxy/frontend/firebird/command/query/batch/FirebirdBatchedStatementsExecutorTest.java#L284-L294)), so it cannot detect execution after the failure.

### Review Details

- **Review Focus:** Mergeability Review.
- **Reviewed Scope:** PR #38927 at head `e2dcd969533463cf8319873df9678817ba81e434`; base and merge-base `2e4bc504ed792cee62ae56fbb511c11c5bbbf470`. GitHub and local changed-file scopes matched at 73/73 files.
- **CI:** All 33 checks completed, with no failed or pending checks.
- **Verification:** The current head passed 26 Firebird exception tests, 66 Firebird protocol tests, and 102 Firebird Proxy tests. The effective Connector/J 8.4.0 default was also verified. Existing tests do not cover the three-message counterexample described above.
- **Not Reviewed Scope:** A live three-message reproduction using a Firebird client with MySQL storage was not run. The production execution path and the exact driver default provide sufficient deterministic evidence.
- **Release Note / User Docs:** No update is required.
- **Other Findings:** Protocol-completeness, documentation, and code-quality observations are intentionally omitted because they are not merge blockers.

terrymanu (approved):
### Result

**Review Result: Mergeable**

No major code-correctness issue requiring a merge block was found in the current revision. I recommend merging this PR and addressing global pagination for Firebird `UNION` queries as a follow-up improvement.

### Evidence

- The previously reported halt-on-first batch issue has been fixed. When `multiError=false`, messages are executed in client order and execution stops after the first failure ([FirebirdBatchedStatementsExecutor.java](https://github.com/apache/shardingsphere/blob/9bcacd46593750f381d67f1be2be7421552fcba2/proxy/frontend/dialect/firebird/src/main/java/org/apache/shardingsphere/proxy/frontend/firebird/command/query/batch/FirebirdBatchedStatementsExecutor.java#L252-L263)).
- The batch-completion response field order matches the project’s [Jaybird 5.0.10](https://github.com/FirebirdSQL/jaybird/blob/v5.0.10/src/main/org/firebirdsql/gds/ng/wire/version16/V16WireOperations.java#L43-L68) dependency.
- No independent major defects were found in exception status vectors, batch lifecycle handling, transaction-handle validation, or character-set error conversion.
- Trailing pagination for `UNION` is still not assigned to the outer `SelectStatement` ([FirebirdDMLStatementVisitor.java](https://github.com/apache/shardingsphere/blob/9bcacd46593750f381d67f1be2be7421552fcba2/parser/sql/engine/dialect/firebird/src/main/java/org/apache/shardingsphere/sql/parser/engine/firebird/visitor/statement/type/FirebirdDMLStatementVisitor.java#L269-L337)), but this is classified as a non-blocking follow-up.

### Follow-up Recommendation

Add parser cases for Firebird `UNION ... ROWS`, `UNION ... OFFSET/FETCH`, and `UNION ... ORDER BY ... ROWS`, and preserve trailing pagination in the combined query’s outer `LimitSegment`.

### Coverage

- Reviewed head: `9bcacd46593750f381d67f1be2be7421552fcba2`
- Base: `959db03e1b23d36daca2fe4e03a874dee50a45f9`
- Accounted for all 73 files returned by GitHub.
- Covered exception mapping, protocol encoding and decoding, status vectors, batch execution and lifecycle, transaction validation, character-set handling, and Firebird SQL parsing.
- All seven changed parser XML files passed well-formedness validation.
- This conclusion covers code correctness only; CI was not reviewed.

PR #38932, opened 2026-06-26
Compute the Snowflake epoch in UTC instead of the system timezone

vasiliy-mikhailov:
The static `EPOCH` for `2016-11-01 00:00:00` is converted to epoch millis using the JVM's **default timezone** offset (`ZoneId.systemDefault()...getOffset(Instant.now())`). The Snowflake key timestamps therefore shift with the server's timezone, which can cause duplicate or mis-ordered keys across deployments in different zones. Use `ZoneOffset.UTC` so the epoch is timezone-independent.

Added a test (run under `-Duser.timezone=Asia/Shanghai`) asserting the epoch equals the UTC-based value.

## Verifying this change

The added `SnowflakeKeyGenerateAlgorithmTest#assertEpochShouldBeTimezoneIndependent` fails before this change and passes after. The bug only manifests when the system timezone is not UTC, so the test is run in a non-UTC zone:

Before the fix:

```
$ TZ=America/New_York mvn test -Dtest=SnowflakeKeyGenerateAlgorithmTest -pl infra/algorithm/type/key-generator/type/snowflake
[INFO] Running org.apache.shardingsphere.infra.algorithm.keygen.snowflake.SnowflakeKeyGenerateAlgorithmTest
[ERROR] Tests run: 12, Failures: 1, Errors: 0, Skipped: 0 <<< FAILURE!
[ERROR]   SnowflakeKeyGenerateAlgorithmTest.assertEpochShouldBeTimezoneIndependent
[INFO] BUILD FAILURE
```

After the fix:

```
$ TZ=America/New_York mvn test -Dtest=SnowflakeKeyGenerateAlgorithmTest -pl infra/algorithm/type/key-generator/type/snowflake
[INFO] Running org.apache.shardingsphere.infra.algorithm.keygen.snowflake.SnowflakeKeyGenerateAlgorithmTest
[INFO] Tests run: 12, Failures: 0, Errors: 0, Skipped: 0
[INFO] BUILD SUCCESS
```

## AI assistance disclosure

This contribution was produced with the help of an AI pipeline. The pipeline processed a large amount of source code to surface suspected bugs, reproduced a subset of them with failing unit tests and generated candidate fixes, and prepared pull requests from the ones that held up. Each PR was then reviewed and verified by a human before being opened: the fix and test were checked by hand and the test was confirmed to fail before the change and pass after.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -31,6 +31,7 @@
 1. Sharding: Fix generated actual index names exceeding database identifier length limits while preserving legacy generated index name compatibility - [#38449](https://github.com/apache/shardingsphere/pull/38449)
 1. Sharding: Fix AUTO_INTERVAL sharding failure under JVM default locales that use comma decimal separators - [#38806](https://github.com/apache/shardingsphere/pull/38806)
 1. DistSQL: Fix case-sensitive storage unit matching in `SHOW RULES USED STORAGE UNIT` - [#38848](https://github.com/apache/shardingsphere/pull/38848)
+1. Sharding: Compute the Snowflake key generator epoch in UTC instead of the JVM default timezone - [#38932](https://github.com/apache/shardingsphere/pull/38932)
 
 ### Enhancements
 
--- a/infra/algorithm/type/key-generator/type/snowflake/pom.xml
+++ b/infra/algorithm/type/key-generator/type/snowflake/pom.xml
@@ -33,7 +33,7 @@
             <version>${project.version}</version>
         </dependency>
     </dependencies>
-
+    
     <build>
         <plugins>
             <plugin>
--- a/infra/algorithm/type/key-generator/type/snowflake/pom.xml
+++ b/infra/algorithm/type/key-generator/type/snowflake/pom.xml
@@ -33,4 +33,15 @@
             <version>${project.version}</version>
         </dependency>
     </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <configuration>
+                    <argLine>${argLine} -Duser.timezone=Asia/Shanghai</argLine>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
 </project>
--- a/infra/algorithm/type/key-generator/type/snowflake/pom.xml
+++ b/infra/algorithm/type/key-generator/type/snowflake/pom.xml
@@ -26,6 +26,10 @@
     <artifactId>shardingsphere-infra-algorithm-key-generator-snowflake</artifactId>
     <name>${project.artifactId}</name>
     
+    <properties>
+        <argLine>-Duser.timezone=Asia/Shanghai</argLine>
+    </properties>
+    
     <dependencies>
         <dependency>
             <groupId>org.apache.shardingsphere</groupId>
@@ -33,15 +37,4 @@
             <version>${project.version}</version>
         </dependency>
     </dependencies>
-    
-    <build>
-        <plugins>
-            <plugin>
-                <artifactId>maven-surefire-plugin</artifactId>
-                <configuration>
-                    <argLine>${argLine} -Duser.timezone=Asia/Shanghai</argLine>
-                </configuration>
-            </plugin>
-        </plugins>
-    </build>
 </project>
--- a/infra/algorithm/type/key-generator/type/snowflake/src/main/java/org/apache/shardingsphere/infra/algorithm/keygen/snowflake/SnowflakeKeyGenerateAlgorithm.java
+++ b/infra/algorithm/type/key-generator/type/snowflake/src/main/java/org/apache/shardingsphere/infra/algorithm/keygen/snowflake/SnowflakeKeyGenerateAlgorithm.java
@@ -28,9 +28,8 @@ import org.apache.shardingsphere.infra.exception.ShardingSpherePreconditions;
 import org.apache.shardingsphere.infra.instance.ComputeNodeInstanceContext;
 import org.apache.shardingsphere.infra.instance.ComputeNodeInstanceContextAware;
 
-import java.time.Instant;
 import java.time.LocalDateTime;
-import java.time.ZoneId;
+import java.time.ZoneOffset;
 import java.util.Collection;
 import java.util.LinkedList;
 import java.util.Properties;
@@ -91,7 +90,7 @@ public final class SnowflakeKeyGenerateAlgorithm implements KeyGenerateAlgorithm
     private int maxTolerateTimeDifferenceMillis;
     
     static {
-        EPOCH = LocalDateTime.of(2016, 11, 1, 0, 0, 0).toInstant(ZoneId.systemDefault().getRules().getOffset(Instant.now())).toEpochMilli();
+        EPOCH = LocalDateTime.of(2016, 11, 1, 0, 0, 0).toInstant(ZoneOffset.UTC).toEpochMilli();
     }
     
     @Override
--- a/infra/algorithm/type/key-generator/type/snowflake/src/test/java/org/apache/shardingsphere/infra/algorithm/keygen/snowflake/SnowflakeKeyGenerateAlgorithmTest.java
+++ b/infra/algorithm/type/key-generator/type/snowflake/src/test/java/org/apache/shardingsphere/infra/algorithm/keygen/snowflake/SnowflakeKeyGenerateAlgorithmTest.java
@@ -52,6 +52,9 @@ import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
+import java.time.LocalDateTime;
+import java.time.ZoneOffset;
+
 class SnowflakeKeyGenerateAlgorithmTest {
     
     private static final long DEFAULT_SEQUENCE_BITS = 12L;
@@ -59,6 +62,14 @@ class SnowflakeKeyGenerateAlgorithmTest {
     private static final int DEFAULT_KEY_AMOUNT = 10;
     
     private static final ComputeNodeInstanceContext INSTANCE;
+    
+    @Test
+    void assertEpochShouldBeTimezoneIndependent() {
+        // The EPOCH is supposed to represent 2016-11-01 00:00:00 UTC.
+        // It MUST equal the UTC-based value regardless of system timezone.
+        long expectedUtcEpoch = LocalDateTime.of(2016, 11, 1, 0, 0, 0).toInstant(ZoneOffset.UTC).toEpochMilli();
+        assertThat(SnowflakeKeyGenerateAlgorithm.EPOCH, is(expectedUtcEpoch));
+    }
     
     static {
         ComputeNodeInstanceContext computeNodeInstanceContext = mock(ComputeNodeInstanceContext.class);


linghengqian (commented):
Have you ever opened an issue to showcase the minimum reproducible unit tests? I recall encountering a similar issue in places like https://github.com/apache/shardingsphere/blob/master/test/native/src/test/resources/test-native/yaml/jdbc/databases/hive/zsd.yaml , but I couldn't consistently reproduce it and had to use `max-tolerate-time-difference-milliseconds`.

Of course, you should at least fix the CI of the current PR.

terrymanu:
Please fix checkstyle first

vasiliy-mikhailov:
Done, thanks. Removed the two unused imports (`java.time.Instant`, `java.time.ZoneId`); `./mvnw checkstyle:check -Pcheck` is clean now.

vasiliy-mikhailov:
Thanks @linghengqian. The added `SnowflakeKeyGenerateAlgorithmTest#assertEpochShouldBeTimezoneIndependent` is the minimal reproducer, and it's deterministic rather than timing-based: it asserts the configured EPOCH equals `LocalDateTime.of(2016, 11, 1, 0, 0, 0).toInstant(ZoneOffset.UTC).toEpochMilli()` (the fixed UTC epoch), regardless of the JVM default timezone. On master the EPOCH is parsed in the system timezone, so the test fails whenever the default zone isn't UTC (e.g. `-Duser.timezone=America/New_York`); with the fix it passes in any zone. Since it compares the computed epoch constant, not clock skew, it doesn't need a `max-tolerate-time-difference-milliseconds` tolerance. CI (checkstyle) is fixed now too.

terrymanu (changes requested):
### Summary

**Reason:** The code direction is correct, but the regression guard is still conditional on external JVM timezone setup, and the user-visible Snowflake key-generation behavior change needs release-note coverage.

### Issues

#### P2 - Make the timezone regression test self-contained

**Problem:** `SnowflakeKeyGenerateAlgorithmTest#assertEpochShouldBeTimezoneIndependent` asserts the UTC epoch value, but it does not set or otherwise guarantee a non-UTC JVM default timezone. With the old implementation, the same assertion would still pass when the test JVM runs in UTC. The module POM has no timezone override, and the root Surefire configuration only passes `${argLine}` (`pom.xml:821-826`).

**Impact:** The PR fixes the production code, but the added test does not deterministically guard the actual trigger condition in normal test execution. A future regression could still pass CI if the JVM default timezone is UTC.

**Required Change:** Please make the regression test or its scoped test configuration force a non-UTC timezone before `SnowflakeKeyGenerateAlgorithm` is initialized, so the test fails on the previous implementation without relying on the developer or CI environment.

#### P2 - Add release-note coverage for the Snowflake epoch behavior change

**Problem:** The PR changes `SnowflakeKeyGenerateAlgorithm.EPOCH` from a JVM-default-timezone calculation to UTC (`SnowflakeKeyGenerateAlgorithm.java:92-94`). Generated keys use `(currentMillis - EPOCH)` in the timestamp component (`SnowflakeKeyGenerateAlgorithm.java:149`), so non-UTC deployments will observe different generated key values after upgrade. `RELEASE-NOTES.md` has a 5.5.4 bug-fix section (`RELEASE-NOTES.md:9`) but this PR does not add an entry.

**Impact:** This is user-visible behavior for the default Snowflake key generator, especially for users operating in non-UTC JVM timezones who need to understand the key-ordering/compatibility effect during upgrade.

**Required Change:** Please add a concise 5.5.4 bug-fix release-note entry describing that Snowflake epoch calculation is now UTC-based and timezone-independent.

### Review Details

**Reviewed Scope:** Latest PR head `7fd350bdac48d2b6d1c2629e98779b21a3077f30`; base ref `master` fetched at `ea4cacb2e1350d42a66bf62d4a725477ce37615b`; local merge-base `6349cd08e9660fbf2ad1d912e7181117a772c460`. GitHub `/pulls/38932/files` matched the local triple-dot file list. Reviewed:
- `infra/algorithm/type/key-generator/type/snowflake/src/main/java/org/apache/shardingsphere/infra/algorithm/keygen/snowflake/SnowflakeKeyGenerateAlgorithm.java`
- `infra/algorithm/type/key-generator/type/snowflake/src/test/java/org/apache/shardingsphere/infra/algorithm/keygen/snowflake/SnowflakeKeyGenerateAlgorithmTest.java`
- Release-note and relevant Maven test-timezone configuration surfaces.

**Not Reviewed Scope:** No Proxy/JDBC E2E, native-image, or distributed runtime smoke was reviewed because the PR is limited to the Snowflake key generator unit path.

**Verification:** Local verification was run on a clean detached worktree at the PR head:
- `./mvnw -pl infra/algorithm/type/key-generator/type/snowflake -Pcheck -T1C checkstyle:check` exited `0`.
- `./mvnw -pl infra/algorithm/type/key-generator/type/snowflake -Pcheck -T1C spotless:check` exited `0`.
- `./mvnw -pl infra/algorithm/type/key-generator/type/snowflake -DskipITs -Dspotless.skip=true -Duser.timezone=Asia/Shanghai -Dtest=SnowflakeKeyGenerateAlgorithmTest test` exited `0`; `Tests run: 12, Failures: 0, Errors: 0`.
- GitHub check-runs for the head commit were accessible: `79` check-runs, all `success`.

**Release Note / User Docs:** Release note is required and missing. User docs were not required because the Snowflake epoch is not documented as a configurable user-facing option in the checked user-manual paths.

vasiliy-mikhailov:
Thanks for the thorough review, both points addressed.

1. Timezone test: since EPOCH is a static final computed at class load, a test-level TimeZone.setDefault cannot reliably control it (the class may already be loaded by an earlier test in the JVM). I added a scoped Surefire user.timezone=Asia/Shanghai to the snowflake module pom, which forces a non-UTC default for the whole test JVM before the class loads. Verified locally: with this override the test passes on the new UTC-based code, and fails on the previous implementation (Expected: is <1477958400000L>, the old offset-based value is 8 hours off), so it is now a real regression regardless of the runner timezone.

2. Release notes: added a 5.5.4 Bug Fixes entry for the UTC epoch change. Note for users: non-UTC deployments will see different generated key values after upgrading, since the epoch baseline shifts to UTC.

Pushed both.

terrymanu (changes requested):
### Summary

**Reason:** The production UTC epoch fix and release-note coverage are aligned, but the new module Surefire configuration replaces inherited JVM test flags while adding the timezone guard. That is a PR-caused build-configuration regression and should be fixed before merge.

### Issues

#### P2 - Preserve inherited Surefire JVM flags while forcing the test timezone

**Problem:** `infra/algorithm/type/key-generator/type/snowflake/pom.xml:42` adds a module-level Surefire `<argLine>${argLine} -Duser.timezone=Asia/Shanghai</argLine>`. In the effective POM for this module, that becomes the module’s active Surefire `argLine`, so it drops the root `jdk11+` Surefire flags from `pom.xml:1141`, including the shared `--add-opens ...` entries and `-XX:+EnableDynamicAgentLoading`.

The module test path still uses Mockito/Byte Buddy and reflective member access, for example `SnowflakeKeyGenerateAlgorithmTest.java:75` and `SnowflakeKeyGenerateAlgorithmTest.java:198`. In a focused local test run on the latest PR head, the tests passed, but the JVM emitted the dynamic-agent warning that asks for `-XX:+EnableDynamicAgentLoading` and notes that dynamic agent loading will be disallowed by default in a future release.

**Impact:** The timezone regression guard is now deterministic, but this module no longer runs tests with the same JDK compatibility flags as the rest of the reactor. That weakens the shared test-runtime setup for current newer-JDK diagnostics and future JDK behavior.

**Required Change:** Please keep the non-UTC timezone guard while preserving the inherited root Surefire `argLine`. For example, set the timezone through configuration that merges with the parent Surefire configuration, or otherwise retain the root `jdk11+` JVM flags alongside `-Duser.timezone=Asia/Shanghai`. Please re-check the effective POM and focused Snowflake test after the change.

### Multi-Round Comparison

- Previous timezone-test request: addressed functionally by forcing a non-UTC timezone before `SnowflakeKeyGenerateAlgorithm` is initialized; the remaining issue is the inherited `argLine` regression above.
- Previous release-note request: addressed by `RELEASE-NOTES.md:34`.

### Review Details

**Review Focus:** Code Correctness Review. CI not reviewed by request.

**Reviewed Scope:** Latest PR head `f345f89dcdf201e71d1b36646fe23c163384fb8b`; base ref `master` with GitHub metadata base SHA `505424233e2c320db166895f2db4d4f1faaa73d1`; current fetched `apache/master` was `b6ff7ce98d22f2afc3691a766c01f01f8717310d`; local merge-base `6349cd08e9660fbf2ad1d912e7181117a772c460`. GitHub `/pulls/38932/files` matched the local triple-dot file list. Reviewed:
- `RELEASE-NOTES.md`
- `infra/algorithm/type/key-generator/type/snowflake/pom.xml`
- `infra/algorithm/type/key-generator/type/snowflake/src/main/java/org/apache/shardingsphere/infra/algorithm/keygen/snowflake/SnowflakeKeyGenerateAlgorithm.java`
- `infra/algorithm/type/key-generator/type/snowflake/src/test/java/org/apache/shardingsphere/infra/algorithm/keygen/snowflake/SnowflakeKeyGenerateAlgorithmTest.java`

**Not Reviewed Scope:** GitHub Actions/check-runs/logs, Proxy/JDBC E2E, native-image, and distributed runtime smoke were not reviewed because this was a code-correctness review of the Snowflake key-generator module path.

**Verification:** Commands were run on a detached local checkout at PR head `f345f89dcdf201e71d1b36646fe23c163384fb8b`.
- `./mvnw -pl infra/algorithm/type/key-generator/type/snowflake -Pcheck -T1C checkstyle:check` exited `0`; Checkstyle reported `0` violations.
- `./mvnw -pl infra/algorithm/type/key-generator/type/snowflake -Pcheck -T1C -DskipITs -Dspotless.skip=true -Dtest=SnowflakeKeyGenerateAlgorithmTest test` exited `0`; `Tests run: 12, Failures: 0, Errors: 0, Skipped: 0`.
- `./mvnw -pl infra/algorithm/type/key-generator/type/snowflake -Pcheck -T1C spotless:check` exited `0`.
- `./mvnw -pl infra/algorithm/type/key-generator/type/snowflake -Pcheck -T1C help:effective-pom` exited `0`; the effective module Surefire `argLine` was `${argLine} -Duser.timezone=Asia/Shanghai`, without the root `jdk11+` flags.

**Release Note / User Docs:** Release note is present at `RELEASE-NOTES.md:34`. Additional user docs are not required for this PR because no configuration key, SQL syntax, or documented user-facing API is changed.

vasiliy-mikhailov:
Good catch, thanks. You are right that the module argLine was replacing the inherited jdk11+ flags instead of extending them. I moved the timezone into the argLine property so it composes with the parent Surefire config: the effective argLine now keeps the four --add-opens entries and -XX:+EnableDynamicAgentLoading and appends -Duser.timezone=Asia/Shanghai. Verified locally: the focused Snowflake test passes with no dynamic-agent warning (flags present), and it still fails on the pre-fix implementation (the timezone guard is intact). Pushed.

terrymanu (approved):
### Summary

**Review Result: Mergeable**

**Reason:** For this Code Correctness Review, the latest PR head is code-scope ready: the Snowflake epoch root cause is fixed, the deterministic non-UTC regression guard and release note are present, and CI was not reviewed by request.

### Evidence

- `infra/algorithm/type/key-generator/type/snowflake/src/main/java/org/apache/shardingsphere/infra/algorithm/keygen/snowflake/SnowflakeKeyGenerateAlgorithm.java:92-94` now computes `EPOCH` with `ZoneOffset.UTC`. Generated IDs use `(currentMillis - EPOCH)` at line 149, so the actual key timestamp baseline is no longer derived from the JVM default timezone.
- `infra/algorithm/type/key-generator/type/snowflake/pom.xml:29-31` forces the Snowflake module tests to run under `Asia/Shanghai`, and the effective Surefire `argLine` keeps the inherited `--add-opens` flags plus `-XX:+EnableDynamicAgentLoading`.
- `infra/algorithm/type/key-generator/type/snowflake/src/test/java/org/apache/shardingsphere/infra/algorithm/keygen/snowflake/SnowflakeKeyGenerateAlgorithmTest.java:66-71` covers the UTC epoch value. Under the previous implementation and the module’s non-UTC timezone, this assertion would be offset by 8 hours, so it guards the reported trigger condition.
- `RELEASE-NOTES.md:34` documents the user-visible Snowflake epoch behavior change for 5.5.4.
- The previous public change requests were rechecked: the release-note gap is addressed, and the latest POM change resolves the inherited `argLine` regression.

### Review Details

**Review Focus:** Code Correctness Review. CI not reviewed by request.

**Reviewed Scope:** Latest PR head `3986158483c587629685247fa98e56f8c70638b0`; base ref `master` with GitHub metadata base SHA `505424233e2c320db166895f2db4d4f1faaa73d1`; local merge-base `6349cd08e9660fbf2ad1d912e7181117a772c460`. GitHub `/pulls/38932/files` matched the local triple-dot file list. Reviewed:
- `RELEASE-NOTES.md`
- `infra/algorithm/type/key-generator/type/snowflake/pom.xml`
- `infra/algorithm/type/key-generator/type/snowflake/src/main/java/org/apache/shardingsphere/infra/algorithm/keygen/snowflake/SnowflakeKeyGenerateAlgorithm.java`
- `infra/algorithm/type/key-generator/type/snowflake/src/test/java/org/apache/shardingsphere/infra/algorithm/keygen/snowflake/SnowflakeKeyGenerateAlgorithmTest.java`

**Not Reviewed Scope:** GitHub Actions/check-runs/logs, Proxy/JDBC E2E, native-image, and distributed runtime smoke were not reviewed because this was a code-correctness review of the Snowflake key-generator module path.

**Verification:**
- `git fetch apache master pull/38932/head:refs/remotes/apache/pr/38932` exited `0`.
- Local triple-dot file list matched GitHub `/pulls/38932/files`.
- `./mvnw -pl infra/algorithm/type/key-generator/type/snowflake -Pcheck -T1C checkstyle:check` exited `0`; Checkstyle reported `0` violations.
- `./mvnw -pl infra/algorithm/type/key-generator/type/snowflake -Pcheck -T1C spotless:check` exited `0`.
- `./mvnw -pl infra/algorithm/type/key-generator/type/snowflake -Pcheck -T1C help:effective-pom -Doutput=<temp file>` exited `0`; effective Surefire `argLine` contained `-Duser.timezone=Asia/Shanghai`, the inherited `--add-opens` flags, and `-XX:+EnableDynamicAgentLoading`.
- `./mvnw -pl infra/algorithm/type/key-generator/type/snowflake -Pcheck -T1C -DskipITs -Dspotless.skip=true -Dtest=SnowflakeKeyGenerateAlgorithmTest test` exited `0`; `Tests run: 12, Failures: 0, Errors: 0, Skipped: 0`.
- No GitHub endpoint required for this selected review focus was inaccessible; CI endpoints were intentionally not queried.

**Release Note / User Docs:** Release note is required and verified at `RELEASE-NOTES.md:34`. Additional user docs are not required because this PR does not change a configuration key, SQL syntax, or documented user-facing API.

PR #39077, opened 2026-07-12
Fix wrong target class name in StaticMethodAdviceExecutor error logs

### LINKED ISSUE #39076: StaticMethodAdviceExecutor error logs report java.lang.Class instead of the target class name
master @ 82d63200491

N/A (ShardingSphere-Agent)

### Expected behavior
When a static-method plugin advice fails, the SEVERE log should name the target class (e.g. `com.foo.Bar`) so the failing instrumentation can be diagnosed.

### Actual behavior
The log always prints `java.lang.Class` as the target class name, regardless of which class was instrumented.

In `StaticMethodAdviceExecutor` the three error handlers `adviceBefore` (line 92), `adviceThrow` (line 108) and `adviceAfter` (line 124) format the `{1}` placeholder with `klass.getClass().getName()`. `klass` is the `@Origin final Class<?>` target class itself, so `klass.getClass()` yields `java.lang.Class` and its name is a constant. The correct call is `klass.getName()`. The sibling `InstanceMethodAdviceExecutor` uses `target.getClass().getName()` because `target` is an instance, confirming the intent is the real target class name.

Register a static-method advice whose `beforeMethod`/`afterMethod`/`onThrowing` throws; the resulting SEVERE log shows `in class \`java.lang.Class\`` instead of the instrumented class.

N/A

thswlsqls:
Fixes #39076.

  - Replace `klass.getClass().getName()` with `klass.getName()` in the three error handlers of `StaticMethodAdviceExecutor` (`adviceBefore`, `adviceThrow`, `adviceAfter`).
  - `klass` is the `@Origin final Class<?>` target class, so `klass.getClass()` always resolved to `java.lang.Class`, making every failure log report `java.lang.Class` instead of the instrumented class.
  - Aligns with the sibling `InstanceMethodAdviceExecutor`, which correctly uses `target.getClass().getName()` because `target` is an instance.
  - Add focused log assertions for the before, throwing, and after failure paths to verify each SEVERE `LogRecord` reports the real target class name.

---

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -39,7 +39,7 @@
 1. DistSQL: Fix case-sensitive storage unit matching in `SHOW RULES USED STORAGE UNIT` - [#38848](https://github.com/apache/shardingsphere/pull/38848)
 1. Sharding: Compute the Snowflake key generator epoch in UTC instead of the JVM default timezone - [#38932](https://github.com/apache/shardingsphere/pull/38932)
 1. Proxy: Fix MySQL BLOB data corruption when string-like prepared statement parameters target BLOB columns - [#39072](https://github.com/apache/shardingsphere/pull/39072)
-1. Agent: Fix wrong target class name in StaticMethodAdviceExecutor error logs - [#PR](https://github.com/apache/shardingsphere/pull/PR)
+1. Agent: Fix wrong target class name in StaticMethodAdviceExecutor error logs - [#39077](https://github.com/apache/shardingsphere/pull/39077)
 
 ### Enhancements
 
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -39,6 +39,7 @@
 1. DistSQL: Fix case-sensitive storage unit matching in `SHOW RULES USED STORAGE UNIT` - [#38848](https://github.com/apache/shardingsphere/pull/38848)
 1. Sharding: Compute the Snowflake key generator epoch in UTC instead of the JVM default timezone - [#38932](https://github.com/apache/shardingsphere/pull/38932)
 1. Proxy: Fix MySQL BLOB data corruption when string-like prepared statement parameters target BLOB columns - [#39072](https://github.com/apache/shardingsphere/pull/39072)
+1. Agent: Fix wrong target class name in StaticMethodAdviceExecutor error logs - [#PR](https://github.com/apache/shardingsphere/pull/PR)
 
 ### Enhancements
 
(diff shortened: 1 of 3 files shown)


terrymanu (approved):
### Summary

**Review Result: Mergeable**

**Reason:** This is a code-scope-only result. The PR fixes the diagnostic root cause across all reported paths with focused regression coverage; CI was not reviewed by request.

### Evidence

- [`StaticMethodAdviceExecutor.java`](https://github.com/apache/shardingsphere/blob/f14a10e4b63dc558745af1ad0d8e0b1ddda45505/agent/core/src/main/java/org/apache/shardingsphere/agent/core/advisor/executor/type/StaticMethodAdviceExecutor.java#L92-L124) now uses `klass.getName()` in the before, throwing-handler, and after failure logs. Since `klass` is already the instrumented `Class<?>`, this directly replaces the constant `java.lang.Class` value produced by `klass.getClass().getName()`.
- [`StaticMethodAdviceExecutorTest.java`](https://github.com/apache/shardingsphere/blob/f14a10e4b63dc558745af1ad0d8e0b1ddda45505/agent/core/src/test/java/org/apache/shardingsphere/agent/core/advisor/executor/type/StaticMethodAdviceExecutorTest.java#L135-L195) independently exercises all three failure paths through the public `advice(...)` entry and verifies the emitted `LogRecord` contains the target class name. Each test would fail against the base implementation.
- Successful execution, disabled advice, original-exception propagation, and `finally` behavior remain unchanged. No API, SPI, configuration, dependency, lifecycle, packaging, security, or meaningful performance impact was introduced.
- [`RELEASE-NOTES.md:42`](https://github.com/apache/shardingsphere/blob/f14a10e4b63dc558745af1ad0d8e0b1ddda45505/RELEASE-NOTES.md#L42) accurately records the user-visible diagnostic fix.

### Review Details

- **Reviewed Scope:** All three GitHub-listed files in PR #39077; latest head `f14a10e4b63dc558745af1ad0d8e0b1ddda45505`; base and local merge-base `620e40a33c3b10a0091dd148e48b21375e156d3e`. The local triple-dot file list exactly matched GitHub `/pulls/39077/files`.
- **Not Reviewed Scope:** CI, GitHub Actions, and runtime agent integration outside the changed diagnostic path. No authoritative changed file was omitted.
- **Verification:** GitHub PR/issue metadata, files, commits, comments, and reviews were accessible. Immutable `git diff`, `git show`, `git grep`, inventory, related executor-path inspection, and Byte Buddy 1.18.8 binding inspection completed successfully. Maven was not run because the active checkout was not the PR head and contained unrelated user changes, so it would not provide fresh-head evidence.
- **Release Note / User Docs:** Release-note entry verified. Separate user documentation, migration, or rollback guidance is not required for this diagnostic-only correction.

PR #38904, opened 2026-06-24
Support SQLServer table variable declaration parse

ClaireLytt:
for https://github.com/apache/shardingsphere/issues/30227
official document:https://learn.microsoft.com/en-us/sql/t-sql/queries/update-transact-sql?view=sql-server-ver17#m-specifying-a-table-variable-as-the-target-object

https://learn.microsoft.com/en-us/sql/t-sql/statements/create-procedure-transact-sql?view=sql-server-ver17
https://learn.microsoft.com/en-us/sql/t-sql/statements/create-function-transact-sql?view=sql-server-ver17
https://learn.microsoft.com/en-us/sql/t-sql/statements/create-trigger-transact-sql?view=sql-server-ver17

Changes proposed in this pull request:
  -

---

Before committing this PR, I'm sure that I have checked the following options:
- [x] My code follows the [code of conduct](https://shardingsphere.apache.org/community/en/involved/conduct/code/) of this project.
- [x] I have self-reviewed the commit code.
- [ ] I have (or in comment I request) added corresponding labels for the pull request.
- [x] I have passed maven check locally : `./mvnw clean install -B -T1C -Dmaven.javadoc.skip -Dmaven.jacoco.skip -e`.
- [ ] I have made corresponding changes to the documentation.
- [x] I have added corresponding unit tests for my changes.
- [x] I have updated the Release Notes of the current development version. For more details, see [Update Release Note](https://shardingsphere.apache.org/community/en/involved/contribute/contributor/)

Issue #30227, opened 2024-02-21
Improve SQL support for SQLServer database when use encrypt feature


strongduanmu:
### Is your feature request related to a problem?

https://github.com/apache/shardingsphere/issues?q=is%3Aopen+is%3Aissue+label%3A%22in%3A+SQL+parse%22+label%3A%22db%3A+SQLServer%22

### Describe the feature you would like.

Through the efforts of many developers in the community, ShardingSphere's support for SQLServer DML syntax is constantly improving. I think it's time to go a step further, and on top of SQL parsing, we can use encrypt feature and test SQLServer DML statements. 

When we test SQLServer DML encryption, there will be some problems, it may be SQL parsing problems, it may be encryption rewriting problems, and it may even be SQL Binder problems. What we have to do is find these problems and fix them.

The following shows the steps required to complete the task:

1. Install SQLServer 2017 on your machine, please refer to https://hub.docker.com/_/microsoft-mssql-server;
2. Then config encrypt rule like following:

```yaml
databaseName: encrypt_db

dataSources:
  db0:
    dataSourceClassName: com.zaxxer.hikari.HikariDataSource
    driverClassName: com.microsoft.sqlserver
    jdbcUrl: jdbc:sqlserver://db1.database.windows.net;databaseName=db1;loginTimeout=300;encrypt=false;
    username: adm

rules:
- !SINGLE
  tables:
    - "*.*"
- !ENCRYPT
  encryptors:
    aes_encryptor:
      type: AES
      props:
        aes-key-value: 123456abc
    like_encryptor:
      type: IT.ENCRYPT.LIKE.FIXTURE
      props:
        mask: 4093
  tables:
    t_user:
      columns:
        user_name:
          cipher:
            name: user_name_cipher
            encryptorName: aes_encryptor
          likeQuery:
            name: user_name_like
            encryptorName: like_encryptor
        password:
          cipher:
            name: password_cipher
            encryptorName: aes_encryptor
        email:
          cipher:
            name: email_cipher
            encryptorName: aes_encryptor
        telephone:
          cipher:
            name: user_telephone_cipher
            encryptorName: aes_encryptor
          likeQuery:
            name: user_telephone_like
            encryptorName: like_encryptor
    t_merchant:
      columns:
        business_code:
          cipher:
            name: business_code_cipher
            encryptorName: aes_encryptor
          likeQuery:
            name: business_code_like
            encryptorName: like_encryptor
        telephone:
          cipher:
[...]

PR #39093, opened 2026-07-14
Support internal binding and extraction for SQL Server UPDATE table variable targets

ClaireLytt:
for https://github.com/apache/shardingsphere/issues/30227  https://github.com/apache/shardingsphere/pull/38904
official document:https://learn.microsoft.com/en-us/sql/t-sql/queries/update-transact-sql?view=sql-server-ver17#m-specifying-a-table-variable-as-the-target-object

  -

---


ClaireLytt:
This PR fixes the SQL Server error “org.apache.shardingsphere.infra.exception.kernel.metadata.TableNotFoundException: Table or view '@MyTableVar' does not exist.” when running UPDATE test

terrymanu (changes requested):
### Summary

**Reason:** The unquoted SQL Server table-variable path is implemented and tested, but the prefix-only detection also classifies valid delimited physical table names beginning with `@` as variables. This can bypass Encrypt rewriting.

### Issues

- **P1: Delimited table names beginning with `@` are misclassified as table variables**

  **Problem:** Both `TableExtractor.isVariableTableTarget` and `SimpleTableSegmentBinder.isVariableTable` inspect only the unwrapped identifier value with `startsWith("@")`; neither checks its `QuoteCharacter`. SQL Server distinguishes regular identifiers beginning with `@` from delimited identifiers such as `[@MyTableVar]`, which remain object names ([Microsoft identifier documentation](https://learn.microsoft.com/en-us/sql/relational-databases/databases/database-identifiers?view=sql-server-ver17)). A latest-head production parser probe confirmed that `UPDATE [@MyTableVar] ...` produces `IdentifierValue("@MyTableVar", BRACKETS)`, but `TableExtractor` now emits no target table.

  `UpdateStatementBaseContext` builds `TablesContext` from that extractor result. When the target is the only table, `EncryptSQLRewriteContextDecorator` consequently sees no Encrypt table and returns without rewriting. The binder also suppresses physical-table existence validation for the same quoted name.

  **Impact:** A valid update against a physical SQL Server Encrypt table whose delimited name begins with `@` can skip Encrypt rewriting or receive incorrect metadata validation. This is a regression from the merge-base behavior, which extracted every non-alias update target.

  **Required Change:** Make table-variable recognition quote-aware at both detection sites: the prefix must identify a variable only when the identifier is unquoted (`QuoteCharacter.NONE`). Add regression coverage proving that bracket-delimited and double-quoted physical names beginning with `@` remain in `TablesContext`, retain normal binder validation, and activate Encrypt rewriting, while unquoted `@MyTableVar` continues through the table-variable path.

### Review Details

- **Reviewed Scope:** All 20 files at head `f395ec505f78`, covering SQL Server metadata, parser statement extraction, binder behavior, SQL federation conversion tests, Encrypt rewriter integration fixtures, build configuration, release notes, and repository ignore configuration. The local triple-dot inventory matched GitHub’s latest PR file list.
- **Not Reviewed Scope:** Live execution against a native SQL Server instance and unrelated DML/MERGE work tracked by the broader umbrella issue.
- **Verification:** 58 focused connector/parser/binder/converter tests passed; all 186 `EncryptSQLRewriterIT` cases passed; scoped Spotless and Checkstyle checks passed with no violations. A production parser-to-extractor probe reproduced the quoted-identifier regression while confirming the unquoted variable path.
- **Release Note / User Docs:** The release note is present and appropriately scoped. No additional user documentation is required for this change.

terrymanu (changes requested):
### Summary

- **Feedback Mode:** Needs Discussion
- **Reason:** The tested rewrite is not reachable through a valid ShardingSphere execution topology for SQL Server table variables, and binder precedence can still confuse an unquoted variable with a same-named delimited physical table.

### Issues

#### P1 — The claimed Encrypt path is not runnable in the current execution topology

- **Problem:** SQL Server table variables exist only within the batch, stored procedure, or function where they are declared ([Microsoft documentation](https://learn.microsoft.com/en-us/sql/t-sql/data-types/table-transact-sql?view=sql-server-ver17)). However, the new test supplies only a standalone [`UPDATE @MyTableVar`](https://github.com/apache/shardingsphere/blob/481389309471509462978368c495c3853df18521/test/it/rewriter/src/test/resources/scenario/encrypt/case/query-with-cipher/dml/update/update.xml#L128-L130). ShardingSphere’s SQL Server grammar accepts one statement followed by EOF ([`SQLServerStatement.g4`](https://github.com/apache/shardingsphere/blob/481389309471509462978368c495c3853df18521/parser/sql/engine/dialect/sqlserver/src/main/antlr4/org/apache/shardingsphere/sql/parser/autogen/SQLServerStatement.g4#L22-L85)); the JDBC path parses each submitted SQL independently ([`ShardingSphereStatement.java`](https://github.com/apache/shardingsphere/blob/481389309471509462978368c495c3853df18521/jdbc/src/main/java/org/apache/shardingsphere/driver/jdbc/core/statement/ShardingSphereStatement.java#L255-L263)), and statement batching executes each SQL separately ([`BatchStatementExecutor.java`](https://github.com/apache/shardingsphere/blob/481389309471509462978368c495c3853df18521/jdbc/src/main/java/org/apache/shardingsphere/driver/executor/engine/batch/statement/BatchStatementExecutor.java#L52-L57)). A procedure does not close the gap because the SQL Server visitor currently discards its body ([`SQLServerDDLStatementVisitor.java`](https://github.com/apache/shardingsphere/blob/481389309471509462978368c495c3853df18521/parser/sql/engine/dialect/sqlserver/src/main/java/org/apache/shardingsphere/sql/parser/engine/sqlserver/visitor/statement/type/SQLServerDDLStatementVisitor.java#L392-L394)).
- **Impact:** The rewrite fixture proves an isolated internal stage, but no current public execution path can both keep `@MyTableVar` in scope and apply the Encrypt rewrite. The [`RELEASE-NOTES.md` claim](https://github.com/apache/shardingsphere/blob/481389309471509462978368c495c3853df18521/RELEASE-NOTES.md#L80) therefore advertises support that applications cannot use.
- **Discussion Needed:** Please pause this implementation direction and align with maintainers on the supported topology: multi-statement batch rewriting, routine-body rewriting, or explicitly scoped internal groundwork without a user-facing support claim. Recommend `type: discussion`.

#### P1 — Physical metadata can shadow an unquoted table variable

- **Problem:** SQL Server distinguishes unquoted `@MyTableVar`, which is necessarily a variable, from the delimited physical identifier `[@MyTableVar]` ([Microsoft identifier rules](https://learn.microsoft.com/en-us/sql/relational-databases/databases/database-identifiers?view=sql-server-ver17)). In [`SimpleTableSegmentBinder.java`](https://github.com/apache/shardingsphere/blob/481389309471509462978368c495c3853df18521/infra/binder/core/src/main/java/org/apache/shardingsphere/infra/binder/engine/segment/dml/from/type/SimpleTableSegmentBinder.java#L394-L413), physical metadata lookup occurs before `isVariableTable`. Because metadata retains the name `@MyTableVar` without delimiter identity, [`IdentifierIndex.find`](https://github.com/apache/shardingsphere/blob/481389309471509462978368c495c3853df18521/infra/common/src/main/java/org/apache/shardingsphere/infra/metadata/identifier/IdentifierIndex.java#L177-L185) matches it and returns a physical-table binder context. The table-variable flag is never set, so [`ColumnSegmentBinder`](https://github.com/apache/shardingsphere/blob/481389309471509462978368c495c3853df18521/infra/binder/core/src/main/java/org/apache/shardingsphere/infra/binder/engine/segment/dml/expression/type/ColumnSegmentBinder.java#L320-L341) validates the variable’s columns against the unrelated physical table.
- **Impact:** A valid variable update can raise a false `ColumnNotFoundException` or receive incorrect physical-table column ownership whenever a delimited physical table with the same text exists.
- **Discussion Needed:** Please include identifier-kind ownership and precedence in the reopened design. An unquoted `@` target must remain a variable even when identical text exists in physical metadata, while a delimited target must remain a physical object.

### Multi-Round Comparison

The previous quoted-identifier blocker is fixed at head `481389309471`: both classification sites now require `QuoteCharacter.NONE`, and regression tests cover bracket- and quote-delimited physical names. The blockers above are newly discovered: the same-name metadata precedence case remains uncovered, and the standalone rewrite still lacks a runnable production topology.

### Review Details

- **Reviewed Scope:** All 20 changed files at head `481389309471509462978368c495c3853df18521`; local merge-base `299926f1bae87123f75cdf31e27245b06b758f8f`; local changed-file list matched GitHub `/pulls/39093/files` 20/20. Reviewed database connector metadata, SQL Server activation, parser table extraction, binder contexts and column binding, federation conversion coverage, Encrypt rewrite fixtures, test-scope dependencies, release notes, and the latest delta from the previous public review. Target dialect: SQL Server; no branch dialect applies. Syntax validation used Microsoft’s [`UPDATE`](https://learn.microsoft.com/en-us/sql/t-sql/queries/update-transact-sql?view=sql-server-ver17), [`table`](https://learn.microsoft.com/en-us/sql/t-sql/data-types/table-transact-sql?view=sql-server-ver17), and [identifier](https://learn.microsoft.com/en-us/sql/relational-databases/databases/database-identifiers?view=sql-server-ver17) documentation, plus `docs/document/content/features/encrypt/limitations.en.md` and `docs/document/content/dev-manual/sql-parser.en.md`.
- **Not Reviewed Scope:** GitHub Actions/check runs and live execution against a SQL Server backend.
- **Verification:** Scoped `spotless:check` and `checkstyle:check` completed with exit code 0. Focused SQL Server metadata, extractor, binder, context, and federation tests completed with 62/62 tests passing and exit code 0. `EncryptSQLRewriterIT` completed with 186/186 tests passing and exit code 0. The GitHub app connector was unavailable because no session was logged in; read-only GitHub REST provided the PR metadata, files, commits, comments, and reviews, so this did not affect the findings.
- **Release Note / User Docs:** A release-note entry is present, but its support claim is not accurate until the execution-topology discussion is resolved. Existing Encrypt documentation explicitly states that simultaneous semicolon-separated statements are unsupported.

ClaireLytt:
Could maintainers please advise which direction I should take?

* multi-statement batch rewriting
* routine-body rewriting
* internal groundwork only, without a user-facing support claim

For this PR, I lean toward option 3: keep the binder/table-extractor groundwork and tests, but avoid a user-facing Encrypt support claim until a runnable path exists. I’m open to maintainer guidance if another direction is preferred.

Also, should I remove or soften the RELEASE-NOTES / Encrypt “supported” wording until a runnable path exists?

btw, @terrymanu Could you please add a tag `type: discussion`? Thanks!

terrymanu:
Thanks for clarifying. Please proceed with **option 3** and keep this PR limited to internal support for binding, extracting, and rewriting SQL Server `UPDATE` targets. Multi-statement batch and routine-body rewriting should be handled separately in #30227.

Please address the remaining items:

1. Remove the Encrypt release-note entry and revise the PR title/description to avoid claiming user-facing support. The rewriter IT can remain as component-level coverage.
2. Restrict table-variable handling in `SimpleTableSegmentBinder` to SQL Server `UPDATE` targets. It currently also affects `SELECT`, `INSERT`, and `DELETE`, while `TableExtractor` only handles `UPDATE`.
3. Detect unquoted `@` table variables before external-context and physical-metadata lookup. Add regression tests for same-named delimited physical tables or CTEs. Bracketed and double-quoted names must remain physical identifiers.
4. Remove the unrelated `.cursor/` change.

The quoted-identifier issue is already fixed. With the scope clarified, a `type: discussion` label is not needed.

ClaireLytt:
@terrymanu 
Could you please rerun failed E2E test? It seemed to failed because of network problem.

terrymanu (changes requested):
### Summary

**Reason:** The overall direction and SPI placement are acceptable, but the latest head still misses two valid SQL Server `UPDATE` table-variable paths and contains a release-note typo. The linked PostgreSQL failure is repeatable, but the evidence attributes it to base commit [#39241](https://github.com/apache/shardingsphere/pull/39241), not to this PR. Therefore, this needs concrete changes rather than reopening the design discussion.

### Issues

#### P1: Aliased table-variable targets are still extracted as physical tables

**Problem:** [`TableExtractor.extractTablesFromUpdate`](https://github.com/apache/shardingsphere/blob/9c4fd3b086998ac4d734a6c2f8aa5b4fab2740d4/parser/sql/statement/core/src/main/java/org/apache/shardingsphere/sql/parser/statement/core/extractor/TableExtractor.java#L319-L343) excludes a variable only when the direct `UPDATE` target itself starts with `@`. Consider this valid alias form:

    UPDATE target
    SET target.Value = 1
    FROM @MyTableVar AS target;

The direct target is the alias `target`, while line 323 unconditionally extracts the `FROM` segment. Consequently, `@MyTableVar` is retained as a physical/rewrite table. A current-head parser-to-extractor probe reproduced this result.

This syntax is supported by SQL Server: the `UPDATE` target may be a table alias and a table variable may be used as a table source ([Microsoft `UPDATE` documentation](https://learn.microsoft.com/en-us/sql/t-sql/queries/update-transact-sql?view=sql-server-ver17)). The current test covers only the direct `UPDATE @MyTableVar ... FROM Employee` form and therefore does not protect this branch.

**Impact:** The aliased form violates the PR’s internal extraction contract and leaks the variable into `TablesContext` and downstream rewrite-table processing as though it were a physical table.

**Required Change:** Resolve an aliased `UPDATE` target against its `FROM` source and exclude that source when it is an unquoted SQL Server table variable, while preserving other physical `FROM` tables. Please add focused extractor/context coverage for the alias form above and for a join containing both physical tables and an aliased table-variable target.

#### P1: Column binding depends on the order of `FROM` sources

**Problem:** [`ColumnSegmentBinder.isSkipColumnBind`](https://github.com/apache/shardingsphere/blob/9c4fd3b086998ac4d734a6c2f8aa5b4fab2740d4/infra/binder/core/src/main/java/org/apache/shardingsphere/infra/binder/engine/segment/dml/expression/type/ColumnSegmentBinder.java#L320-L342) returns immediately for the first `SimpleTableSegmentBinderContext`, even when `isMetadataUnavailable(...)` is false. It therefore never checks a later context containing the table-variable flag.

This is observable when a physical table precedes the variable target in a join. [`JoinTableSegmentBinder`](https://github.com/apache/shardingsphere/blob/9c4fd3b086998ac4d734a6c2f8aa5b4fab2740d4/infra/binder/core/src/main/java/org/apache/shardingsphere/infra/binder/engine/segment/dml/from/type/JoinTableSegmentBinder.java#L69-L83) binds the left side before the right side, and [`UpdateStatementBinder`](https://github.com/apache/shardingsphere/blob/9c4fd3b086998ac4d734a6c2f8aa5b4fab2740d4/infra/binder/core/src/main/java/org/apache/shardingsphere/infra/binder/engine/statement/dml/UpdateStatementBinder.java#L45-L60) binds `FROM` before an alias target. A current-head binder probe with the physical context first and the variable context later failed with `ColumnNotFoundException`.

The new unit test supplies only one table-variable context, so it does not exercise the ordering boundary.

**Impact:** A valid multi-source SQL Server `UPDATE` can still fail binding depending solely on `FROM`/join order.

**Required Change:** Evaluate all applicable contexts before deciding that metadata is available—for example, aggregate whether any function, DB-link, or table-variable context requires skipping column validation. Please add regression coverage with a physical context before the table-variable context and retain coverage for the reverse order and existing DB-link behavior.

#### P2: Fix the release-note typo introduced by the latest rerun commit

**Problem:** [`RELEASE-NOTES.md`](https://github.com/apache/shardingsphere/blob/9c4fd3b086998ac4d734a6c2f8aa5b4fab2740d4/RELEASE-NOTES.md#L80) currently says `SQL Binder: upport internal binding and extraction...`.

**Impact:** This is user-facing release output.

**Required Change:** Restore `Support`.

### Multi-Round Comparison

- The earlier quoted-identifier and physical-metadata-precedence concerns have been addressed.
- The earlier topology discussion is resolved: the maintainer explicitly selected internal groundwork, deferred multi-statement/routine-body support to #30227, and stated that `type: discussion` is unnecessary.
- The public SPI was already present when that direction was accepted. Its default `Optional.empty()` contract preserves existing implementations, the SQL Server override contains the dialect-specific value, and dialect metadata is the existing owner for this capability. The SPI change is therefore not itself a blocker.
- The two P1 findings above are newly confirmed boundary gaps in the current implementation.

### Review Details

- **Review Focus:** Mergeability Review, including the linked GitHub Actions failure.
- **Reviewed Scope:** All 19 files at head `9c4fd3b086998ac4d734a6c2f8aa5b4fab2740d4`; base `master` at `a14fa59423cd16b75e081e680417e199377eac3d`; merge-base `299926f1bae87123f75cdf31e27245b06b758f8f`. The authoritative file list matched GitHub `/pulls/39093/files` 19/19. Reviewed SQL Server metadata/SPI activation, shared statement extraction, binder context and ordering, federation coverage, Encrypt rewrite fixtures, test-scope dependencies, release notes, and the latest rerun delta. Target dialect: SQL Server; shared default behavior for other dialects was checked. Syntax evidence used Microsoft’s [`UPDATE`](https://learn.microsoft.com/en-us/sql/t-sql/queries/update-transact-sql?view=sql-server-ver17) and [`table`](https://learn.microsoft.com/en-us/sql/t-sql/data-types/table-transact-sql?view=sql-server-ver17) documentation.
- **Not Reviewed Scope:** Live execution against a native SQL Server instance; multi-statement and routine-body support tracked by #30227; implementation of the unrelated PostgreSQL base regression.
- **Verification:** Focused current-head connector, extractor, binder, statement-context, and federation tests passed with exit code 0, 65 tests total. The alias extractor probe exited 1 after reproducing `@MyTableVar` in the extracted rewrite tables. The ordered multi-context binder probe exited 1 with the expected `ColumnNotFoundException`.
- **CI Attribution:** The exact PostgreSQL passthrough job and all four `RETURNING` variants passed on [run 30194375859](https://github.com/apache/shardingsphere/actions/runs/30194375859/job/89773973182), where head `1020724e` was merged into base `b084c233`. The next run merged this PR into base `c11a89b`; the only base delta was [`Fix PostgreSQL composite column type OID resolution (#39241)`](https://github.com/apache/shardingsphere/commit/c11a89b038ac984f79c28280bcd52280f6857587), after which the same four cases failed in [run 30245404560](https://github.com/apache/shardingsphere/actions/runs/30245404560/job/89914540716), [run 30252395522](https://github.com/apache/shardingsphere/actions/runs/30252395522/job/89941614236), and the [linked current job](https://github.com/apache/shardingsphere/actions/runs/30325097107/job/90192243745?pr=39093). PostgreSQL JDBC 42.7.8 resolves the newly exposed composite OID through a metadata query containing `s.r` ([`TypeInfoCache#getSQLTypeQuery`](https://github.com/pgjdbc/pgjdbc/blob/REL42.7.8/pgjdbc/src/main/java/org/postgresql/jdbc/TypeInfoCache.java#L242-L267)), which matches the failure exactly. This is a base regression exposed by this matrix, not a regression introduced by #39093.
- **GitHub Evidence Access:** The GitHub app session was unavailable; authenticated read-only REST supplied PR metadata, commits, files, reviews, workflow attempts, jobs, and logs, so there is no evidence gap affecting this result.
- **Release Note / User Docs:** The release note is correctly scoped to internal binder/extractor support, apart from the P2 typo. Additional user documentation is not required because this PR does not claim runnable public table-variable support.

### Next Steps

1. Address the two SQL Server boundary cases and the release-note typo in this PR.
2. Do not add PostgreSQL protocol/OID changes to #39093. The linked CI job should be rerun after the #39241 base regression is fixed or the PR is updated onto a repaired base.

ClaireLytt:
@terrymanu 
Thanks for the review. The two P1s are addressed as follows:
**P1 — Aliased table-variable targets were still extracted as physical tables**

**Problem:** For `UPDATE target ... FROM @MyTableVar AS target`, `@MyTableVar` was still extracted from `FROM` as a physical/rewrite table because only a direct `@` UPDATE target was excluded.

**Code changes:**
- `parser/sql/statement/core/.../extractor/TableExtractor.java`
  - `extractTablesFromUpdate()`
  - `findAliasedVariableTableSource()`
  - `findFromTableSegmentByAliasTarget()`
  - `extractTablesFromTableSegment(..., skipSimpleTableSegment)`
- Tests: `TableExtractorTest`, `UpdateStatementContextTest`

**P1 — Column binding depended on the order of FROM sources**

**Problem:** `isSkipColumnBind()` returned on the first `SimpleTableSegmentBinderContext`, so a physical table before a table-variable context could cause a false `ColumnNotFoundException`.

**Code changes:**
- `infra/binder/core/.../expression/type/ColumnSegmentBinder.java`
  - `isSkipColumnBind()`
  - `containsMetadataUnavailableContext()` (new)
- Tests: `ColumnSegmentBinderTest` (physical-before-variable, variable-before-physical, owner-qualified wrong column, DB-link)

ClaireLytt:
I don't add PostgreSQL protocol/OID changes to this pr and merge lastest master code in this branch, but still failed in postgresql e2e test.
Could you please help me to check this problem whenever you have time? Thanks a lot! @terrymanu

terrymanu (changes requested):
### Result

Feedback Mode: Change Request

**Blocking Issues: 1**

The maintainer has confirmed that this PR should provide internal binding, extraction, and component-level rewriting support. However, the current head still does not correctly bind an aliased SQL Server table-variable target and requires a concrete code change.

### Blocking Issues

#### P1 — Preserve the original table-variable name when binding an aliased UPDATE target

**Evidence**

SQL Server allows an alias from the `FROM` table source to be used as the UPDATE target, for example:

    UPDATE target
    SET Value = 1
    FROM @MyTableVar AS target;

This behavior is supported by Microsoft's [`UPDATE`](https://learn.microsoft.com/en-us/sql/t-sql/queries/update-transact-sql?view=sql-server-ver17) and [table variable](https://learn.microsoft.com/en-us/sql/t-sql/data-types/table-transact-sql?view=sql-server-ver17) documentation.

At the current head:

1. [`UpdateStatementBinder`](https://github.com/apache/shardingsphere/blob/f2ca43a193566d8fb86f56a31d5c1d0222dd7d78/infra/binder/core/src/main/java/org/apache/shardingsphere/infra/binder/engine/statement/dml/UpdateStatementBinder.java#L47-L50) binds `FROM @MyTableVar AS target` before binding the UPDATE target `target`.
2. [`createSimpleTableBinderContext`](https://github.com/apache/shardingsphere/blob/f2ca43a193566d8fb86f56a31d5c1d0222dd7d78/infra/binder/core/src/main/java/org/apache/shardingsphere/infra/binder/engine/segment/dml/from/type/SimpleTableSegmentBinder.java#L415-L420) marks this source with `containsTableVariable`, but constructs its context without an `originalTableName`.
3. [`bindUpdateTargetTableAlias`](https://github.com/apache/shardingsphere/blob/f2ca43a193566d8fb86f56a31d5c1d0222dd7d78/infra/binder/core/src/main/java/org/apache/shardingsphere/infra/binder/engine/segment/dml/from/type/SimpleTableSegmentBinder.java#L217-L237) therefore cannot recover `@MyTableVar` from the context and falls back to the UPDATE target text, `target`.
4. The resulting bound statement has both its table name and alias set to `target`. The correct result should have table name `@MyTableVar` and alias `target`.
5. [`UpdateStatementConverter`](https://github.com/apache/shardingsphere/blob/f2ca43a193566d8fb86f56a31d5c1d0222dd7d78/kernel/sql-federation/compiler/src/main/java/org/apache/shardingsphere/sqlfederation/compiler/sql/ast/converter/statement/type/UpdateStatementConverter.java#L54-L76) checks only that an alias is present, so it accepts this incorrectly bound statement and converts `target` as the target table.

The existing alias tests only verify that a manually constructed statement excludes the table variable from extraction:

- [`TableExtractorTest`](https://github.com/apache/shardingsphere/blob/f2ca43a193566d8fb86f56a31d5c1d0222dd7d78/parser/sql/statement/core/src/test/java/org/apache/shardingsphere/sql/parser/statement/core/extractor/TableExtractorTest.java#L186-L225)

The existing binder and converter tests cover only the direct `UPDATE @MyTableVar` form:

- [`UpdateStatementBinderTest`](https://github.com/apache/shardingsphere/blob/f2ca43a193566d8fb86f56a31d5c1d0222dd7d78/infra/binder/core/src/test/java/org/apache/shardingsphere/infra/binder/engine/statement/dml/UpdateStatementBinderTest.java#L175-L201)
- [`UpdateStatementConverterTest`](https://github.com/apache/shardingsphere/blob/f2ca43a193566d8fb86f56a31d5c1d0222dd7d78/kernel/sql-federation/compiler/src/test/java/org/apache/shardingsphere/sqlfederation/compiler/sql/ast/converter/statement/type/UpdateStatementConverterTest.java#L55-L65)

**Impact**

The valid aliased form loses the table variable's actual identity. Every downstream component consuming the bound `UpdateStatement` sees `target` as the target table instead of `@MyTableVar`, violating the internal binding contract introduced by this PR.

**Required Change**

Preserve the table variable's original `IdentifierValue` when creating the `SimpleTableSegmentBinderContext` for a SQL Server UPDATE `FROM` source so that alias resolution produces `@MyTableVar`.

Add a focused `UpdateStatementBinderTest` that verifies at least the following:

- the bound table name is `@MyTableVar`;
- the bound alias is `target`;
- `targetTableIsFromAlias` remains `true`;
- the table variable is not included as a physical table in `TablesContext`.

### Coverage

- Review Focus: Code Correctness Review. This is a code-scope-only result; CI was not queried or reviewed.
- Reviewed Head: `f2ca43a193566d8fb86f56a31d5c1d0222dd7d78`.
- Base: `master` at `e24b3761dd5e72a2b24d2bd05818af3942cf3564`.
- Scope: All 19 files in GitHub's current changed-file list were reviewed.
- Behavior clusters: SQL Server dialect metadata, table-variable recognition and binder context, column binding, UPDATE table extraction, SQL Federation conversion, rewriter component coverage, dependencies, and release notes.
- Requirements: The maintainer explicitly selected the internal binding/extraction/rewriting direction. Multi-statement and routine-body support remain tracked by #30227, so this is a concrete implementation defect rather than a reason to reopen the design discussion.
- Multi-round status: The previous quoted-identifier, metadata-precedence, column-context-order, alias-extraction, and release-note findings are fixed. This finding was already present in the previously reviewed version but was missed during that review; it was not introduced by the latest extractor fix.
- Discovery: The root-cause/behavior, blast-radius/contracts, and tests/runtime/operations review lenses were completed. No unresolved evidence gap remains that could change the blocker set.

ClaireLytt:
@terrymanu
Thanks for the review. The latest review comment is addressed as follows:
- `SimpleTableSegmentBinder#createSimpleTableBinderContext` — when `isVariableTable(...)`, also `setOriginalTableName(tableName)` so the FROM context keeps `@MyTableVar`.
- `UpdateStatementBinderTest#assertBindAliasedTableVariableTarget` — assert bound name `@MyTableVar`, alias `target`, `targetTableIsFromAlias=true`, and `@MyTableVar` is not in `TablesContext` as a physical table.

ClaireLytt:
@terrymanu 
I suspect the e2e tests might be unstable — my last commit failed on the PostgreSQL test, but this commit failed on the MySQL test. Could you help me check this issue when you get a chance? Thanks a lot!

terrymanu (changes requested):
### Result

Feedback Mode: Change Request
**Blocking Issues: 2**

The current head still mishandles two valid SQL Server table-variable `UPDATE` forms. Both defects can expose table variables as physical tables, violating the PR’s binding and extraction contract.

### Blocking Issues

1. **P1 — Aliased table-variable sources lose variable precedence when metadata contains the same name**
   - **Evidence:** [`createSimpleTableBinderContext`](https://github.com/apache/shardingsphere/blob/61e6b60e2d49f20652cbc9ca01df2fb2a3cae4cc/infra/binder/core/src/main/java/org/apache/shardingsphere/infra/binder/engine/segment/dml/from/type/SimpleTableSegmentBinder.java#L406-L431) prioritizes variable handling only when the segment is the direct target object. An aliased source such as `UPDATE target SET Value = 1 FROM @MyTableVar AS target` reaches external and physical-table lookup before `containsTableVariable` is set. Microsoft documents unquoted `@` identifiers as variables rather than ordinary database objects. [Database identifiers](https://learn.microsoft.com/en-us/sql/relational-databases/databases/database-identifiers?view=sql-server-ver17)
   - **Reproduction:** Against current-head classes, adding same-named physical metadata for `@MyTableVar` causes the statement above to bind against that physical table and fail with `ColumnNotFoundException` when `Value` is absent.
   - **Impact:** A valid table-variable target can be mistaken for a physical table or external context, causing incorrect metadata validation and binding. Existing collision tests cover only direct targets, while the aliased-target test uses no conflicting metadata and no assignment columns.
   - **Required Change:** Classify every applicable unquoted SQL Server `@` table segment participating in `UPDATE` as a table variable before external-context and physical-metadata lookup, including the `FROM` source selected through the target alias. Add aliased-target collision coverage for both physical metadata and external/CTE context.

2. **P1 — A direct table-variable target repeated in `FROM` is extracted as a rewrite table**
   - **Evidence:** [`extractTablesFromUpdate`](https://github.com/apache/shardingsphere/blob/61e6b60e2d49f20652cbc9ca01df2fb2a3cae4cc/parser/sql/statement/core/src/main/java/org/apache/shardingsphere/sql/parser/statement/core/extractor/TableExtractor.java#L327-L345) skips a variable source only when `targetTableIsFromAlias` is true. Therefore `UPDATE @MyTableVar SET Value = 1 FROM @MyTableVar` extracts the `FROM` occurrence. Microsoft permits the updated object’s single `FROM` reference to appear with or without an alias, and table variables are valid in `UPDATE`. [UPDATE syntax](https://learn.microsoft.com/en-us/sql/t-sql/queries/update-transact-sql?view=sql-server-ver17) [Table variables](https://learn.microsoft.com/en-us/sql/t-sql/data-types/table-transact-sql?view=sql-server-ver17)
   - **Reproduction:** The current-head statement model sets `targetTableIsFromAlias=false`, and `UpdateStatementContext.getTablesContext().getTableNames()` returns `[@MyTableVar]`.
   - **Impact:** Downstream rewrite and rule logic sees the internal table variable as a physical logical table, defeating the extraction boundary introduced by this PR.
   - **Required Change:** Exclude the matching unquoted table-variable source from `FROM` extraction even when no alias is used, while continuing to extract unrelated physical and joined tables. Add focused extraction/context coverage for this form.

### Coverage

- Reviewed current public head `61e6b60e2d49f20652cbc9ca01df2fb2a3cae4cc` against `master`.
- Accounted for all 19 files in the final public diff across metadata SPI, SQL Server metadata, binder behavior, table extraction, converter coverage, rewriter fixtures, POMs, and release notes.
- Reviewed all 13 public patch commits and the available discussion/review history. The latest commit fixes the earlier original-table-name loss for aliased variables, but does not affect either issue above.
- Verified the additive metadata SPI default preserves existing dialect behavior; only SQL Server supplies the `@` prefix, and quoted identifiers remain physical identifiers.
- Focused metadata, extractor, binder, and statement-context test suites passed.
- Scoped Spotless and Checkstyle checks passed with zero violations.
- `UpdateStatementConverterTest` passed. `EncryptSQLRewriterIT` was not valid local evidence: all 189 cases failed during setup with the same missing `DialectSQLOption` class caused by an inconsistent local dependency classpath; the class is absent from the reviewed head source. This was classified as a verification-environment failure, not a third PR defect.
- This is a code-correctness review; CI status was not queried or included.

ClaireLytt:
I will add more tests following all your review comments and post a summary comment later.
Really appreciate your thorough review and valuable feedback. Thank you for your patience with this PR!@terrymanu

terrymanu (changes requested):
### Result

Feedback Mode: Change Request

**Blocking Issues: 1**

A valid SQL Server `UPDATE` can still expose a secondary table variable as a physical rewrite table.

### Blocking Issues

#### P1 — Exclude every table-variable source from UPDATE table extraction

- **Evidence:** [`extractTablesFromUpdate`](https://github.com/apache/shardingsphere/blob/d16dbc19467e758f79ec44676ed47c0a24f0292d/parser/sql/statement/core/src/main/java/org/apache/shardingsphere/sql/parser/statement/core/extractor/TableExtractor.java#L327-L364) selects at most one variable source to skip. The generic extraction path then adds every other `SimpleTableSegment` to `rewriteTables`. Microsoft documents that table variables may be used anywhere a table expression is accepted, including `UPDATE ... FROM` joins ([table variables](https://learn.microsoft.com/en-us/sql/t-sql/data-types/table-transact-sql?view=sql-server-ver17#remarks), [UPDATE syntax](https://learn.microsoft.com/en-us/sql/t-sql/queries/update-transact-sql?view=sql-server-ver17#from-table_source)). A current-head regression probe using `@TargetVar` as the target and `@SourceVar` as an additional joined source failed because `@SourceVar` remained in `rewriteTables`.
- **Impact:** [`UpdateStatementBaseContext`](https://github.com/apache/shardingsphere/blob/d16dbc19467e758f79ec44676ed47c0a24f0292d/infra/binder/core/src/main/java/org/apache/shardingsphere/infra/binder/context/statement/type/dml/UpdateStatementBaseContext.java#L67-L70) exposes that variable through `TablesContext`. Rule and rewrite consumers such as [`EncryptSQLRewriteContextDecorator`](https://github.com/apache/shardingsphere/blob/d16dbc19467e758f79ec44676ed47c0a24f0292d/features/encrypt/core/src/main/java/org/apache/shardingsphere/encrypt/rewrite/context/EncryptSQLRewriteContextDecorator.java#L71-L75) can consequently treat it as a logical physical table, including matching a same-named delimited physical-table rule.
- **Required Change:** Exclude all unquoted SQL Server table-variable segments encountered during `UPDATE` extraction, not only the occurrence selected as the target source. Preserve bracket- and double-quoted physical identifiers and unrelated physical/joined tables. Add focused extraction/context coverage for direct and aliased variable targets with an additional table-variable source.

### Coverage

- Reviewed head `d16dbc19467e758f79ec44676ed47c0a24f0292d`; base and merge-base `e24b3761dd5e72a2b24d2bd05818af3942cf3564`.
- Accounted for all 19 authoritative GitHub files; the GitHub file list and public PR diff matched 19/19.
- Reviewed clusters: metadata SPI and SQL Server activation; binder precedence and column binding; UPDATE extraction/context; SQL Federation conversion; Encrypt component coverage; test dependencies; release notes.
- The maintainer-approved scope is internal binding, extraction, and component-level rewriting; public batch/routine execution remains outside this PR.
- Both blockers reported at head `61e6b60e…` are fixed. The additional-source gap predates the latest commit and was missed in the previous review.
- Existing focused connector, parser, binder, context, and converter tests passed: 80 tests, zero failures. The isolated additional-source regression probe failed with `expected false but was true`.
- Target dialect: SQL Server; no branch dialect applies. Other dialects retain the SPI’s empty default. Official Microsoft SQL Server 17 documentation was checked.
- Root-cause/behavior, blast-radius/contracts, tests/runtime/operations, latest-delta, and final convergence passes completed with no unresolved evidence gap.
- Code-correctness scope only; CI was not queried or reviewed.

terrymanu (changes requested):
### Result

**Feedback Mode:** Change Request  
**Blocking Issues:** 1

The overall direction has already been accepted: this PR is limited to internal binding, extraction, and component-level rewriting for SQL Server `UPDATE` table-variable targets. Multi-statement and routine-body support remain separate work.

However, the current implementation still has a correctness defect inside that accepted scope.

### Blocking Issue

#### P1 — Preserve target-column ownership when a table-variable column collides with a physical source column

**Evidence**

Microsoft documents that SQL Server permits `@table_variable` as an `UPDATE` target and that `SET column_name` refers to a column of the object being updated, not an arbitrary same-named column from the `FROM` sources. See the [Microsoft `UPDATE` documentation](https://learn.microsoft.com/en-us/sql/t-sql/queries/update-transact-sql?view=sql-server-ver17).

The current binding flow violates that rule:

1. [`UpdateStatementBinder`](https://github.com/apache/shardingsphere/blob/15d59dc0fb4c21c8222d81a723ead37145a951d2/infra/binder/core/src/main/java/org/apache/shardingsphere/infra/binder/engine/statement/dml/UpdateStatementBinder.java#L45-L60) binds the table-variable target and physical `FROM` sources into the same table-context collection before binding the `SET` assignments.
2. For an unqualified assignment column, [`ColumnSegmentBinder`](https://github.com/apache/shardingsphere/blob/15d59dc0fb4c21c8222d81a723ead37145a951d2/infra/binder/core/src/main/java/org/apache/shardingsphere/infra/binder/engine/segment/dml/expression/type/ColumnSegmentBinder.java#L155-L167) searches all available table contexts.
3. The table-variable context has no column projections because its metadata is unavailable.
4. If a joined physical table contains the same column name, [`getInputInfoFromTableBinderContexts`](https://github.com/apache/shardingsphere/blob/15d59dc0fb4c21c8222d81a723ead37145a951d2/infra/binder/core/src/main/java/org/apache/shardingsphere/infra/binder/engine/segment/dml/expression/type/ColumnSegmentBinder.java#L233-L252) selects that physical column and assigns its table ownership to the target column.
5. `isSkipColumnBind` only suppresses the not-found validation. It does not correct a column that has already been matched to the wrong physical table.
6. Encrypt then uses the incorrect `ColumnSegmentBoundInfo.originalTable` and applies the physical table’s Encrypt rule to the table-variable assignment.

A current-head regression probe used:

    UPDATE @MyTableVar
    SET VacationNote = 'annual leave'
    FROM HumanResources.Employee AS e
    WHERE e.BusinessEntityID = EmpID

`VacationNote` is a column of the table-variable target in this statement. The joined `Employee` table also has a `VacationNote` column configured for Encrypt.

Expected rewritten SQL:

    UPDATE @MyTableVar
    SET VacationNote = 'annual leave'
    FROM HumanResources.Employee AS e
    WHERE e.BusinessEntityID = EmpID

Actual rewritten SQL:

    UPDATE @MyTableVar
    SET [vacation_note_cipher] = 'encrypt_annual leave'
    FROM HumanResources.Employee AS e
    WHERE e.BusinessEntityID = EmpID

The isolated regression run completed 198 `EncryptSQLRewriterIT` cases with exactly one failure: the collision case above.

**Why the existing tests miss it**

The current table-variable rewrite cases use `NewVacationNote` as the target column, while the physical `Employee` Encrypt rule applies to `VacationNote`. Therefore, they verify the no-collision path only:

- [Existing table-variable cases](https://github.com/apache/shardingsphere/blob/15d59dc0fb4c21c8222d81a723ead37145a951d2/test/it/rewriter/src/test/resources/scenario/encrypt/case/query-with-cipher/dml/update/update.xml#L138-L150)
- [Employee Encrypt configuration](https://github.com/apache/shardingsphere/blob/15d59dc0fb4c21c8222d81a723ead37145a951d2/test/it/rewriter/src/test/resources/scenario/encrypt/config/query-with-cipher.yaml#L183-L188)

The binder tests similarly use assignment names that do not exist in the physical source metadata, so they prove that unknown columns are accepted but do not verify target ownership when a physical match exists.

**Impact**

- Valid SQL can be rewritten to update a cipher column belonging to a different table.
- Literal assignment values are incorrectly encrypted.
- Parameter assignments can enter the same incorrect parameter-rewrite path.
- Any downstream component relying on the assignment column’s bound table can observe incorrect ownership.

**Required Change**

Preserve the ownership of `SET` target columns when the SQL Server `UPDATE` target is a table variable whose metadata is unavailable.

The fix should occur at the binding/ownership boundary. Adding an Encrypt-only guard would leave the incorrect `ColumnSegmentBoundInfo` visible to other consumers.

Please add focused regression coverage for:

- A direct table-variable target whose assignment column has the same name as a physical `FROM` column.
- An aliased table-variable target with the same collision.
- Both literal and parameter assignments.
- Confirmation that the target assignment is not rewritten or encrypted.
- Confirmation that qualified physical-source predicates continue to use their normal Encrypt rewriting.

### Coverage

- Reviewed head: `15d59dc0fb4c21c8222d81a723ead37145a951d2`
- Base and merge-base: `e24b3761dd5e72a2b24d2bd05818af3942cf3564`
- All 19 changed files were accounted for.
- The metadata SPI, SQL Server activation, table-variable recognition, binder context, column binding, table extraction, statement context, SQL Federation conversion, Encrypt rewrite fixtures, test dependencies, and release note were reviewed.
- The previously reported quoted-identifier, metadata-precedence, alias-resolution, source-order, and multiple-table-variable extraction defects are fixed.
- The new default SPI method preserves compatibility for existing dialect implementations; only SQL Server provides the `@` prefix.
- Existing focused connector, parser, binder, context, and converter tests passed: 117 tests, zero failures.
- The original `EncryptSQLRewriterIT` suite passed: 197 tests, zero failures.
- The collision regression probe failed deterministically: 198 tests, one failure.
- CI was not queried or included in this code-correctness result.
- Live SQL Server batch and routine-body execution were not reviewed because they are outside this PR’s confirmed scope.

terrymanu (approved):

terrymanu (changes requested):
### Result

**Feedback Mode:** Change Request

**Blocking Issues: 2**

The current head contains a reproducible incorrect SQL rewrite and introduces an under-specified public dialect contract. Both issues are within the stated SQL Server `UPDATE` table-variable scope and should be resolved before merge.

### Blocking Issues

#### 1. [P1] Preserve `SET` column ownership for table-variable update targets

**Affected path**

- [`ColumnSegmentBinder`](https://github.com/apache/shardingsphere/blob/15d59dc0fb4c21c8222d81a723ead37145a951d2/infra/binder/core/src/main/java/org/apache/shardingsphere/infra/binder/engine/segment/dml/expression/type/ColumnSegmentBinder.java#L233-L252)
- [`SimpleTableSegmentBinder`](https://github.com/apache/shardingsphere/blob/15d59dc0fb4c21c8222d81a723ead37145a951d2/infra/binder/core/src/main/java/org/apache/shardingsphere/infra/binder/engine/segment/dml/from/type/SimpleTableSegmentBinder.java#L399-L405)

**Problem**

`UpdateStatementBinder` puts the target table and every `FROM` table into the same binder-context collection before binding the assignment. The table-variable context contains no projection metadata, so an unqualified `SET` column cannot match it. If a physical table in `FROM` contains a column with the same name, `ColumnSegmentBinder` binds the assignment column to that physical table instead.

The new `containsTableVariable` handling only suppresses the column-not-found validation. It does not preserve ownership of assignment columns or prevent another table context from winning the lookup.

That contradicts the [SQL Server `UPDATE` contract](https://learn.microsoft.com/en-us/sql/t-sql/queries/update-transact-sql?view=sql-server-ver17): `@table_variable` is the update target, and an unqualified `SET column_name` belongs to that target.

**Reproduction**

    Input:
    UPDATE @MyTableVar
    SET VacationNote = 'annual leave'
    FROM HumanResources.Employee AS e
    WHERE e.BusinessEntityID = EmpID

    Expected:
    UPDATE @MyTableVar
    SET VacationNote = 'annual leave'
    FROM HumanResources.Employee AS e
    WHERE e.BusinessEntityID = EmpID

    Actual:
    UPDATE @MyTableVar
    SET [vacation_note_cipher] = 'encrypt_annual leave'
    FROM HumanResources.Employee AS e
    WHERE e.BusinessEntityID = EmpID

`VacationNote` is incorrectly bound to `HumanResources.Employee`. [`EncryptAssignmentTokenGenerator`](https://github.com/apache/shardingsphere/blob/15d59dc0fb4c21c8222d81a723ead37145a951d2/features/encrypt/core/src/main/java/org/apache/shardingsphere/encrypt/rewrite/token/generator/assignment/EncryptAssignmentTokenGenerator.java#L147-L160) then trusts that bound table and rewrites the table-variable assignment as an encrypted `Employee` column.

**Impact**

A valid SQL Server statement can be converted into invalid or semantically different SQL. The current tests do not expose this because their table-variable assignment columns (`NewVacationNote`, `NewVacationHours`) do not collide with columns in the configured physical table.

**Required change**

- Bind `SET` assignment columns to the effective `UPDATE` target before performing general expression-column lookup across `FROM` sources.
- When the target is a table variable without metadata, preserve its syntactic ownership without assigning a physical table’s `ColumnBoundInfo`.
- Do not solve this only inside Encrypt; the incorrect ownership originates in the binder and can affect every downstream consumer.
- Add regression coverage for:
  - Direct table-variable targets with a colliding `FROM` column.
  - Aliased table-variable targets with a colliding `FROM` column.
  - Literal and parameter rewrite paths.
  - A binder-level assertion proving that the assignment column is not bound to the physical source table.

#### 2. [P2] Replace `getVariableTableNamePrefix()` with a semantic table-variable contract

**Affected path**

- [`DialectDatabaseMetaData#getVariableTableNamePrefix`](https://github.com/apache/shardingsphere/blob/15d59dc0fb4c21c8222d81a723ead37145a951d2/database/connector/core/src/main/java/org/apache/shardingsphere/database/connector/core/metadata/database/metadata/DialectDatabaseMetaData.java#L212-L219)
- [`SQLServerDatabaseMetaData`](https://github.com/apache/shardingsphere/blob/15d59dc0fb4c21c8222d81a723ead37145a951d2/database/connector/dialect/sqlserver/src/main/java/org/apache/shardingsphere/database/connector/sql92/sqlserver/metadata/database/SQLServerDatabaseMetaData.java#L91-L94)
- [`SimpleTableSegmentBinder`](https://github.com/apache/shardingsphere/blob/15d59dc0fb4c21c8222d81a723ead37145a951d2/infra/binder/core/src/main/java/org/apache/shardingsphere/infra/binder/engine/segment/dml/from/type/SimpleTableSegmentBinder.java#L335-L347)
- [`TableExtractor`](https://github.com/apache/shardingsphere/blob/15d59dc0fb4c21c8222d81a723ead37145a951d2/parser/sql/statement/core/src/main/java/org/apache/shardingsphere/sql/parser/statement/core/extractor/TableExtractor.java#L350-L359)

**Problem**

`VariableTableNamePrefix` is not the database concept being modeled. The SQL Server term is **table variable**, so even a prefix-oriented name would be `TableVariableNamePrefix`.

More importantly, exposing `Optional<String>` publishes the lexical representation instead of the dialect behavior. Both consumers must independently reconstruct the real rule:

- Statement must be an `UPDATE`.
- The segment must represent a table source.
- The identifier must be unqualified.
- The identifier must be unquoted.
- The identifier must have dialect-specific table-variable syntax.

This duplicates semantic decisions across modules and makes the public contract easy to misuse. The same `@` spelling cannot safely be treated as a general cross-dialect table prefix: for example, [MySQL defines `@name` as a scalar user variable and explicitly does not allow it where a table identifier is required](https://dev.mysql.com/doc/refman/8.0/en/user-variables.html).

**Required change**

Replace the raw prefix getter with a semantic predicate owned by the dialect contract, for example:

    default boolean isTableVariableIdentifier(
            final String value, final QuoteCharacter quoteCharacter) {
        return false;
    }

The SQL Server implementation should return `true` only for an unquoted identifier with the supported `@` table-variable form. The binder and extractor may retain generic structural checks such as statement type and owner qualification, but they should not interpret a raw dialect prefix themselves.

If a prefix getter is retained, `getTableVariableNamePrefix()` is the minimum correct name, but it still leaves the semantic duplication unresolved.

Add focused tests for:

- Default dialect behavior returning `false`.
- SQL Server unquoted `@tableVariable`.
- Quoted `[@name]` and `"@name"` identifiers returning `false`.
- Ordinary SQL Server identifiers.
- Both production consumers using the semantic contract consistently.

### Coverage

- Reviewed PR: [#39093](https://github.com/apache/shardingsphere/pull/39093)
- Reviewed head: [`15d59dc0fb4c21c8222d81a723ead37145a951d2`](https://github.com/apache/shardingsphere/commit/15d59dc0fb4c21c8222d81a723ead37145a951d2)
- Base: `e24b3761dd5e72a2b24d2bd05818af3942cf3564`
- Authoritative GitHub file list and extracted base/head diff matched: 19 of 19 files.
- Reviewed behavior clusters:
  - Public dialect metadata contract.
  - SQL Server table-variable recognition.
  - Table extraction and rewrite-table exclusion.
  - Binder context creation and metadata-unavailable behavior.
  - `UPDATE` target and assignment binding.
  - SQL Federation conversion.
  - Encrypt token and parameter rewrite paths.
  - Unit and integration-test adequacy.
  - Maven dependency changes and release notes.
- Focused connector/parser/binder/converter verification: 117 tests, 0 failures.
- Existing `EncryptSQLRewriterIT`: 197 tests, 0 failures.
- Isolated colliding-column regression probe: 198 tests, 1 failure, reproducing the incorrect encrypted assignment shown above.
- No live SQL Server execution was performed; the deterministic binder-to-rewriter regression already establishes the merge-blocking correctness failure.
- CI status was not evaluated as part of this code-correctness review.

ClaireLytt:
@terrymanu 
Thank you for your guidence!
for p1, I have added test like
```
<rewrite-assertion id="update_physical_target_with_table_variable_source_colliding_set_column_for_literals" db-types="SQLServer">
        <input sql="UPDATE HumanResources.Employee SET VacationNote = 'x' FROM HumanResources.Employee AS e JOIN @MyTableVar ON e.BusinessEntityID = EmpID" />
        <output sql="UPDATE HumanResources.Employee SET [vacation_note_cipher] = 'encrypt_x' FROM HumanResources.Employee AS e JOIN @MyTableVar ON e.BusinessEntityID = EmpID" />
</rewrite-assertion>
    
<rewrite-assertion id="update_table_variable_colliding_set_column_from_employee_for_literals" db-types="SQLServer">
        <input sql="UPDATE @MyTableVar SET VacationNote = 'annual leave' FROM HumanResources.Employee AS e WHERE e.BusinessEntityID = EmpID AND e.VacationNote = 'annual leave'" />
        <output sql="UPDATE @MyTableVar SET VacationNote = 'annual leave' FROM HumanResources.Employee AS e WHERE e.BusinessEntityID = EmpID AND e.[vacation_note_cipher] = 'encrypt_annual leave'" />
</rewrite-assertion>
```




for p2, I have refactored the code

Please review again. I'm apreciate the time you spent on this pr!

terrymanu (changes requested):
### Result

Feedback Mode: Change Request

**Blocking Issues: 1**

The PR is not ready to merge. One confirmed correctness issue remains in the core binder-to-rewriter path.

### Blocking Issue

#### P1 — A quoted physical target can be mistaken for a table-variable target

In `UpdateStatementBinder.java:76-94`, target-variable matching compares only the identifier text:

```java
each.getValue().equalsIgnoreCase(targetTableName.getValue())
```

It ignores the quote character. Therefore, this valid SQL Server case is mis-associated:

```sql
DECLARE @MyTable TABLE (Remark varchar(100));

UPDATE [@MyTable]
SET Remark = 'x'
FROM @MyTable;
```

`[@MyTable]` is a physical table, while `@MyTable` is a table variable. SQL Server distinguishes regular and delimited identifiers.

Current path:

1. `[@MyTable]` is bound as a physical table.
2. `@MyTable` in the `FROM` clause is bound as a table-variable context.
3. `UpdateStatementBinder` matches the variable context to the physical target because both values are `@MyTable`.
4. `ColumnSegmentBinder.java:226-269` binds the `SET` column through the temporary-table path.
5. `EncryptAssignmentTokenGenerator.java:107-110` may derive a suffix-based column name instead of using the configured cipher column.

This can produce incorrect rewritten SQL.

The current tests cover quoted physical targets with ordinary or quoted physical sources, but not a quoted physical target combined with a same-text unquoted table-variable source:

- `UpdateStatementBinderTest.java:419-579`
- `update.xml:193-221`

Required changes:

- Match target and variable contexts using complete identifier semantics, including the quote character and owner/schema identity.
- Add binder regression coverage for `UPDATE [@MyTable] ... FROM @MyTable`.
- Add Encrypt rewrite coverage for both literal and parameter assignments.

### Review Scope and Conclusion

Reviewed the current PR head `eb7df42`, all 22 changed files, production paths, tests, parser and table-extraction behavior, Encrypt rewrite consumers, dependencies, release notes, public review history, and CI state.

All 80 current-head checks completed successfully. GitHub reports no merge conflict, but the latest public review state remains `CHANGES_REQUESTED`.

The two additional observations from the previous report are withdrawn; they are not merge blockers.

**Final recommendation: Do not merge until the P1 issue is fixed and covered by regression tests.**

ClaireLytt:
@terrymanu 
Thank you for your guidence!
I have added test like
```
UPDATE [@MyTable] SET ... FROM @MyTable

UPDATE "@MyTable" SET ... FROM @MyTable

UPDATE dbo.[@MyTable] SET ... FROM @MyTable

UPDATE dbo."@MyTable" SET ... FROM @MyTable
```
Please review again. I'm apreciate the time you spent on this pr!

terrymanu (changes requested):
### Result

Feedback Mode: Change Request

**Blocking Issues: 1**

Reason: The 2026-08-13 blocker (quoted physical target mistaken for a table-variable target) is fixed for the tested no-alias forms, and all earlier blockers remain fixed at the current head. However, the same contract is still violated when the FROM table variable carries an alias: value-only target-alias matching conflates a delimited physical table with a same-text unquoted table variable, so the bound UPDATE target loses its quote identity and SET-column ownership, extraction, and encrypt rewriting follow the wrong object. This finding was missed in the previous review round; it predates the latest two commits and is the complete remaining blocker set after a full-scope re-review.

### Blocking Issues

#### P1 — A quoted physical target is still mistaken for a table-variable target when the same-text variable appears in FROM with an alias

**Evidence**

- `parser/sql/engine/dialect/sqlserver/src/main/java/org/apache/shardingsphere/sql/parser/engine/sqlserver/visitor/statement/SQLServerStatementVisitor.java:1786-1792` — `isTableNameWithAliasInFromClause` sets `targetTableIsFromAlias` by comparing identifier values with `equalsIgnoreCase` only; `[@MyTable]` and `@MyTable` compare equal, so `UPDATE [@MyTable] SET Remark = 'x' FROM @MyTable AS x` is flagged as an alias-target update (the same check at `:1772-1783` for the alias-name form).
- `infra/binder/core/src/main/java/org/apache/shardingsphere/infra/binder/engine/segment/dml/from/type/SimpleTableSegmentBinder.java:257` and `:290-294` — with the flag set, FROM binds first and creates a variable context (`containsTableVariable=true`, `originalTableName=@MyTable`, unquoted); the target `[@MyTable]` then matches it through `isUpdateTargetTableAlias`/`isSameUpdateTargetTableContext`, again by value only, and `bindUpdateTargetTableAlias` (`:274-279`) overwrites the bound target identifier with the context's unquoted `@MyTable`, erasing the `BRACKETS` quote.
- `infra/binder/core/src/main/java/org/apache/shardingsphere/infra/binder/engine/statement/dml/UpdateStatementBinder.java:93-105` — `isTableVariableTarget` is quote-aware, but it now inspects the already-rewritten unquoted target, so it returns true and SET columns are bound to the variable (`TEMPORARY_TABLE`) instead of the physical table.
- `parser/sql/statement/core/src/main/java/org/apache/shardingsphere/sql/parser/statement/core/extractor/TableExtractor.java:326-336` — extraction keeps `targetTableIsFromAlias` and `excludeUnquotedTableVariables` removes the FROM variable, so the physical target never enters `rewriteTables` and no encrypt rewrite is generated for it.
- Counter-evidence checked: `assertBindBracketDelimitedAtSignTargetWithUnquotedSameTextTableVariableSourceRetainsPhysicalTableValidation` in `infra/binder/core/src/test/java/org/apache/shardingsphere/infra/binder/engine/statement/dml/UpdateStatementBinderTest.java` and the `update_bracket_delimited_at_sign_physical_target_with_unquoted_table_variable_source_*` IT cases in `test/it/rewriter/src/test/resources/scenario/encrypt/case/query-with-cipher/dml/update/update.xml` all pass because their FROM variable has no alias; no binder, extractor, or IT case covers the aliased form.

**Impact**

For the valid SQL Server statement `UPDATE [@MyTable] SET Remark = 'x' FROM @MyTable AS x` (physical table `[@MyTable]`, variable `@MyTable` aliased `x`), the SET column is bound to the table variable as `TEMPORARY_TABLE`, so encrypt rewriting configured on the physical `@MyTable` is skipped and plaintext is sent where a cipher column is expected. The symmetric form `UPDATE @MyTable SET Remark = 'x' FROM [@MyTable] AS src` fails the opposite way: the variable target is rebound to the quoted physical identifier, SET columns are resolved against the physical table (or throw `ColumnNotFoundException` when the column sets differ), and the physical table is extracted as the rewrite target, producing cipher columns inside a table-variable update. Before this PR both statements failed fast with `TableNotFoundException`; the PR makes them silently misbind, which is the exact error class the previous review round blocked on.

**Required Change**

Make the UPDATE target-alias association quote-aware end to end so a delimited `@`-prefixed physical table and an unquoted table variable with the same text are never conflated: `isTargetTableAliasInFromClause` in the SQLServer visitor must not treat a delimited target as the alias/table-name of an unquoted same-text variable source, and `isUpdateTargetTableAlias`/`isSameUpdateTargetTableContext` in `SimpleTableSegmentBinder` must not match a `containsTableVariable` context when the target segment is not itself the same unquoted variable (compare `IdentifierValue.getQuoteCharacter()`, mirroring `DialectDatabaseMetaData.isTableVariableIdentifier`). Add regressions for both directions — `UPDATE [@MyTable] … FROM @MyTable AS x` and `UPDATE @MyTable … FROM [@MyTable] AS src` — at the binder level (quote character and `TableSourceType` of the bound SET column), the extractor level, and as rewriter IT cases analogous to the existing unaliased ones.

### Coverage

- Reviewed head `0c7539a` (branch `declare`, fork `ClaireLytt/shardingsphere`), base `master@d8720e1`, 24 commits; latest delta after the 2026-08-13 review is commits `347afc1` and `0c7539a`, both re-verified in a full-scope convergence pass with no further independent candidates.
- Authoritative scope is the GitHub changed-file list (22 files, +2249/−24), all accounted for and mapped to six behavior clusters: dialect contract (`DialectDatabaseMetaData.isTableVariableIdentifier` + SQLServer override + tests), binder variable detection/context creation with precedence over external/physical/CTE metadata, UPDATE SET-column ownership (`UpdateStatementBinder`/`AssignmentSegmentBinder`/`ColumnSegmentBinder`), table extraction exclusions, component-level rewriter IT coverage (mock schemas, `update.xml`, `query-with-cipher.yaml`), and metadata (poms, release note, federation converter test). No churn-only files.
- All three discovery lenses completed; every prior public blocker re-verified at the head as fixed: variable detection before external-context and physical lookup, restriction to SQLServer UPDATE targets (`SELECT` still throws `TableNotFoundException`), default dialect unchanged, all FROM variable sources excluded from extraction, aliased-variable-target original-name preservation, SET-column ownership under collisions, and the quote-aware dialect contract replacing the under-specified one. The single remaining blocker above is classified as missed in the previous review.
- Dialect family: SQLServer is the only affected dialect (trunk visitor and shared binder/extractor paths gated by the dialect contract; no SQLServer branch dialect exists; default `isTableVariableIdentifier=false` leaves all other dialects unchanged, confirmed by the default-dialect tests). Official semantics per Microsoft UPDATE Transact-SQL (target object forms, FROM table sources with aliases, table variables) and the identifier documentation cited in prior review rounds.
- Local verification used read-only GitHub API access (anonymous route; no token configured, `gh` unavailable); the PR head ref is not present in the local repository and fetching was out of authority, so local triple-dot comparison and the inventory/ledger scripts were not applicable — the GitHub file list served as the authoritative scope.
- Non-blocking note: the `skipSimpleTableSegment` parameter threaded through `extractTablesFromTableSegment`/`extractTablesFromJoinTableSegment` in `TableExtractor.java:124-161` is always `Optional.empty()` after the 2026-08-11 rework and can be removed with the blocker fix or separately.
- This is a Code Correctness Review of code scope only; GitHub Actions and CI were not queried or reviewed.

ClaireLytt:
@terrymanu 
Thank you for your guidence!
I have added test in `TablesContext`, retain normal binder validation, and activate Encrypt rewriting
```
UPDATE [@MyTable]

SET Remark = 'x'

FROM @MyTable AS x



UPDATE @MyTable

SET Remark = 'x'

FROM [@MyTable] AS src



UPDATE "@MyTable"

SET Remark = 'x'

FROM @MyTable AS x



UPDATE @MyTable

SET Remark = 'x'

FROM “@MyTable” AS x



UPDATE dbo.[@MyTable]

SET Remark = 'x'

FROM @MyTable AS x



UPDATE @MyTable

SET Remark = 'x'

FROM dbo.[@MyTable] AS src



UPDATE dbo."@MyTable"

SET Remark = 'x'

FROM @MyTable AS x



UPDATE @MyTable

SET Remark = 'x'

FROM dbo."@MyTable" AS src
```
Please review again. I'm apreciate the time you spent on this pr!

ClaireLytt:
@terrymanu 
Could you please rerun failed e2e tests?

terrymanu (changes requested):
### Result

Feedback Mode: Change Request

**Blocking Issues: 3**

The current head still loses SQL Server identifier identity in one alias path, weakens physical-target column validation, and does not propagate the new metadata capability through the delegating contract.

### Blocking Issues

#### 1. Preserve quote identity when matching an UPDATE target to a FROM alias

- **Evidence:** [`isAliasInFromClause`](https://github.com/apache/shardingsphere/blob/ad1eb8d15de1e93aaaa76107782f72e565649d92/parser/sql/engine/dialect/sqlserver/src/main/java/org/apache/shardingsphere/sql/parser/engine/sqlserver/visitor/statement/SQLServerStatementVisitor.java#L1765-L1807) compares the target with `getAliasName()` as plain text, discarding the alias quote character. Consequently, `UPDATE @MyTable SET Remark = 'x' FROM [@MyTable] AS [@MyTable]` is classified as an alias-target update. The [binder alias-key fast path](https://github.com/apache/shardingsphere/blob/ad1eb8d15de1e93aaaa76107782f72e565649d92/infra/binder/core/src/main/java/org/apache/shardingsphere/infra/binder/engine/segment/dml/from/type/SimpleTableSegmentBinder.java#L242-L300) then replaces the unquoted variable target with the quoted physical-table context. Microsoft documents that unquoted `@name` denotes a variable while delimited identifiers have different semantics. [SQL Server identifiers](https://learn.microsoft.com/en-us/sql/relational-databases/databases/database-identifiers?view=sql-server-ver17)
- **Impact:** The target is bound and extracted as a physical table, so the Encrypt token and parameter rewriters can incorrectly rewrite the table-variable assignment to `remark_cipher`. The previous quote-identity blocker is therefore only partially fixed; the new tests use aliases `x` or `src` and miss this collision.
- **Required Change:** Preserve the alias `IdentifierValue`, including its quote character, through visitor detection, binder alias resolution, and extractor matching. Add parser, binder, extractor, literal-rewriter, and parameter-rewriter regressions for the same-text delimited alias case.

#### 2. Do not let a table-variable source disable validation of a physical target’s SET column

- **Evidence:** For `UPDATE e SET MissingColumn = 1 FROM HumanResources.Employee AS e CROSS JOIN @MyTableVar AS v`, the effective target is the known physical `Employee` table. However, [`UpdateStatementBinder`](https://github.com/apache/shardingsphere/blob/ad1eb8d15de1e93aaaa76107782f72e565649d92/infra/binder/core/src/main/java/org/apache/shardingsphere/infra/binder/engine/statement/dml/UpdateStatementBinder.java#L65-L104) passes every target/FROM context to SET-column binding. [`ColumnSegmentBinder`](https://github.com/apache/shardingsphere/blob/ad1eb8d15de1e93aaaa76107782f72e565649d92/infra/binder/core/src/main/java/org/apache/shardingsphere/infra/binder/engine/segment/dml/expression/type/ColumnSegmentBinder.java#L388-L405) then suppresses `ColumnNotFoundException` merely because the unrelated source context contains a table variable. SQL Server requires the assigned column to exist in the target table or view. [SQL Server UPDATE](https://learn.microsoft.com/en-us/sql/t-sql/queries/update-transact-sql?view=sql-server-ver17)
- **Impact:** An invalid physical-target assignment passes binding with temporary/unowned column metadata, deferring the error to the database and misleading ownership-based consumers.
- **Required Change:** For `SET_ASSIGNMENT_COLUMNS`, allow metadata-unavailable binding only when the effective UPDATE target itself is a table variable. Retain the relaxed behavior for source expressions, and add an `UpdateStatementBinder` regression asserting `ColumnNotFoundException` for the physical-target case above.

#### 3. Forward the new metadata capability through the delegating base class

- **Evidence:** [`DialectDatabaseMetaData`](https://github.com/apache/shardingsphere/blob/ad1eb8d15de1e93aaaa76107782f72e565649d92/database/connector/core/src/main/java/org/apache/shardingsphere/database/connector/core/metadata/database/metadata/DialectDatabaseMetaData.java#L213-L222) adds `isTableVariableIdentifier` with a default result of `false`, while SQL Server overrides it. [`AbstractDelegatingDialectDatabaseMetaData`](https://github.com/apache/shardingsphere/blob/ad1eb8d15de1e93aaaa76107782f72e565649d92/database/connector/core/src/main/java/org/apache/shardingsphere/database/connector/core/metadata/database/metadata/AbstractDelegatingDialectDatabaseMetaData.java#L42-L152) forwards every older dialect capability but not this method; its [delegation test](https://github.com/apache/shardingsphere/blob/ad1eb8d15de1e93aaaa76107782f72e565649d92/database/connector/core/src/test/java/org/apache/shardingsphere/database/connector/core/metadata/database/metadata/AbstractDelegatingDialectDatabaseMetaDataTest.java#L55-L93) also omits it.
- **Impact:** A branch or plugin metadata implementation delegating to SQL Server returns `false` for unquoted `@t`, causing parser, binder, extractor, and Encrypt consumers to take the physical-table path. MariaDB-over-MySQL does not expose the defect because both currently return `false`.
- **Required Change:** Delegate `isTableVariableIdentifier` in `AbstractDelegatingDialectDatabaseMetaData` and extend its forwarding contract test with a delegate returning `true`.

### Coverage

- Reviewed current head `ad1eb8d15de1e93aaaa76107782f72e565649d92`; the final GitHub check confirmed it remains the open PR head.
- Accounted for all 25 authoritative changed files across dialect metadata, parser/extraction, binder ownership, Encrypt rewriting, SQL Federation, and release metadata.
- Revalidated the linked issue, maintainer scope decision, and previous formal review; completed root-cause, blast-radius, tests/runtime, and zero-new-finding convergence passes with no unresolved coverage gaps.
- Focused connector/parser/binder/federation Maven suite: exit 0, 93 tests passed. The initial sandbox run exited before assertions because Mockito could not attach; the identical host-permission rerun passed.
- `EncryptSQLRewriterIT`: exit 0, 221 scenarios passed.
- Code-scope review only. Live SQL Server execution and CI were not queried or reviewed.

terrymanu:
Please resolve the conflicts first

terrymanu:
This issue still exists. The latest change fixes one case but breaks the mirrored case:

`UPDATE [@MyTable] SET Remark = 'x' FROM @MyTable AS [@MyTable];`

Here, the UPDATE target `[@MyTable]` exactly matches the FROM alias `[@MyTable]`, so the target should resolve to that alias, which points to the table variable `@MyTable`.

Please adjust `isAliasInFromClause` to determine alias matching by comparing the UPDATE target with the FROM alias, including their quoting. It should not require the UPDATE target and the original table in the FROM clause to have the same table-variable classification.

These two cases should produce different results:

- `UPDATE [@MyTable] ... FROM @MyTable AS [@MyTable]`: the target is the FROM alias.
- `UPDATE @MyTable ... FROM [@MyTable] AS [@MyTable]`: the target is not the quoted alias.

Please also update the related test cases:

- `target-table-is-from-alias` should be `true` for the first case.
- In the Encrypt rewrite result, columns belonging to the table variable should not be rewritten to cipher columns.

terrymanu:
After rechecking the cross-module changes, I’m keeping the scope concern. The common binding, extraction, and rewriting paths do need changes, but the current patch has grown beyond what is necessary.

Please simplify the PR as follows:

1. Keep `SQLServerStatementVisitor` as the single owner of `targetTableIsFromAlias`.

   `SimpleTableSegmentBinder` and `TableExtractor` should trust this value. They should not walk the FROM clause and re-evaluate whether the target, alias, and original source have the same table-variable classification.

   In particular, please simplify or remove:

   - `SimpleTableSegmentBinder#isTargetTableMatchedFromAlias`
   - The table-variable identity filtering in `findUpdateTargetFromTableContexts`
   - `TableExtractor#isFromAliasTarget`
   - `TableExtractor#isTargetMatchedInFrom`
   - `TableExtractor#matchesFromAlias`
   - `TableExtractor#matchesFromTableNameWithAlias`
   - The related recursive owner comparison

   Fixing only `SQLServerStatementVisitor#isAliasInFromClause` is not enough. The current Binder and Extractor checks would still reject:

   `UPDATE [@MyTable] SET Remark = 'x' FROM @MyTable AS [@MyTable]`

2. Remove the new `UpdateStatementConverterTest#assertConvertWithTableVariableTarget` test.

   SQL Federation production code is unchanged, and this test only verifies that the identifier string `@MyTableVar` is copied unchanged. It would already pass without this PR and does not test the new binding, extraction, or rewriting behavior.

3. Reduce the duplicated test matrix.

   The PR currently adds:

   - 34 tests to `UpdateStatementBinderTest`
   - 27 tests to `TableExtractorTest`
   - 20 tests to `UpdateStatementContextTest`
   - 47 Encrypt rewrite assertions
   - 8 parser cases

   Please keep one representative test for each distinct production branch instead of testing the full combination of bracket/double-quote, schema/no-schema, alias/no-alias, JOIN order, and literal/parameter at every layer.

   Most of the new `UpdateStatementContextTest` cases duplicate `TableExtractorTest`, because `UpdateStatementBaseContext` simply delegates table collection to `TableExtractor`. Only a small number of integration cases are needed there.

   For Encrypt rewriting, literal and parameter cases must both remain because they use different rewriters, but they do not both need every quoting, schema, alias, and JOIN combination.

4. Avoid adding unrelated test fixtures where existing ones can be reused.

   The same-named encrypted `@MyTable` fixture is useful because it verifies that an unquoted table variable takes precedence over physical metadata. However, the additional `HumanResources.Employee` metadata and Encrypt configuration are not essential if an existing encrypted table such as `ScrapReason` can cover the same physical-source and column-collision behavior.

Please keep the necessary shared changes:

- `DialectDatabaseMetaData#isTableVariableIdentifier`, its SQL Server implementation, delegation, and contract tests
- SQL Server UPDATE-specific table-variable handling in the Binder
- Binding SET columns to the table-variable target while preserving normal validation for physical targets
- Excluding table variables from `TableExtractor`
- Skipping table-variable target columns in both Encrypt token and parameter rewriting
- The test-scope SQL Server dependencies required to load the dialect metadata SPI

The goal is not to move everything back into the SQL Server modules. It is to keep the necessary shared pipeline changes while removing duplicated SQL semantics, unrelated coverage, and repeated test combinations.

ClaireLytt:
Thanks for your review, the scale of this pr is too big. So I think it's a better way to split it up into 3-4 prs! I will close this pr.

PR #39102, opened 2026-07-14
Add SYSTEM_USER to SQL92 unparenthesized function names

### LINKED ISSUE #39099: SQL92 dialect omits SYSTEM_USER from unparenthesized function names
master @ 620e40a33c3

Both (kernel-level binder)

### Expected behavior
Under the SQL92 baseline dialect, the niladic function `SYSTEM_USER` should bind as a function, like the sibling SQL-92 §6.8 user-value functions `USER`, `CURRENT_USER`, and `SESSION_USER`.

### Actual behavior
`SYSTEM_USER` is missing from `SQL92FunctionOption.UNPARENTHESIZED_FUNCTION_NAMES`, so the binder misclassifies it as a column reference instead of a function.

`SQL92FunctionOption` (database/connector/dialect/sql92/.../metadata/database/option/SQL92FunctionOption.java line 31-32) lists only 3 of the 4 SQL-92 user-value niladic functions. `ColumnSegmentBinder.isUnparenthesizedFunction()` (infra/binder/core/.../expression/type/ColumnSegmentBinder.java line 143) reads this set via `.contains()` to decide function vs column. The MySQL and SQLServer sibling options already include `SYSTEM_USER`.

Bind `SELECT SYSTEM_USER` against a SQL92-typed schema; the identifier is treated as a column rather than a niladic function.

N/A

thswlsqls:
Fixes #39099.

  - Add `SYSTEM_USER` to `SQL92FunctionOption.UNPARENTHESIZED_FUNCTION_NAMES` so it binds as a niladic function instead of a column.
  - `SYSTEM_USER` is one of the four SQL-92 §6.8 user-value niladic functions (`USER`/`CURRENT_USER`/`SESSION_USER`/`SYSTEM_USER`); the SQL92 baseline dialect previously listed only three.
  - Aligns with the MySQL and SQLServer sibling `DialectFunctionOption` implementations, which already include `SYSTEM_USER`.
  - Consumer `ColumnSegmentBinder.isUnparenthesizedFunction()` reads this set via `.contains()` to classify a bare identifier as function vs column.

---


terrymanu (changes requested):
### Summary

**Reason:** Adding `SYSTEM_USER` to the function-name set is correct, but it does not fix the linked issue. On the latest PR head, `SELECT SYSTEM_USER` is rejected by the SQL92 parser before the binder reads this set.

### Issues

#### P1 — The reported SQL does not reach this change

**Problem:** The new test verifies only that `SYSTEM_USER` exists in `SQL92FunctionOption`. An exact-head parser/visitor check still raises `SQLParsingException` for `SELECT SYSTEM_USER`.

**Impact:** The behavior described in #39099 remains unfixed, and the release-note claim is not yet accurate.

**Discussion Needed:** Please first confirm how bare `SYSTEM_USER` should be represented by the SQL92 parser and where its niladic-function classification belongs, then validate the complete parser-to-binder path.

### Review Details

- **Review Focus:** Code Correctness Review; CI not reviewed by request.
- **Reviewed Scope:** All three changed files at head `3c464e35c8ec49ab41fb2fee024235f80416323c`, plus the relevant SQL92 parser and binder path. The local file list matched GitHub.
- **Not Reviewed Scope:** GitHub Actions, the full Maven suite, and JDBC/Proxy E2E scenarios.
- **Verification:** `SQL92FunctionOptionTest` passed with 2 tests and 0 failures. A reviewer-run exact-head parser check confirmed that `SELECT SYSTEM_USER` still raises `SQLParsingException`.
- **Release Note / User Docs:** The release-note entry is present, but its claimed behavior is not yet implemented end to end.

thswlsqls:
Confirmed. In the SQL92 grammar the `SYSTEM_USER` token is declared in `SQL92Keyword.g4` but is not referenced by any parser rule — it is unreachable through `columnName → identifier → (IDENTIFIER_ | unreservedWord)` in `BaseRule.g4` and is not in `regularFunctionName`. So `SELECT SYSTEM_USER` never produces a `ColumnSegment` and fails in the parser before `ColumnSegmentBinder#isUnparenthesizedFunction` reads the set; the binder-set entry cannot fix #39099. The existing niladic names (`CURRENT_USER`, `SESSION_USER`, `USER`) have the same gap. I'll move the fix into the grammar — make `SYSTEM_USER` reachable as an unparenthesized/niladic function via `BaseRule.g4` — and add a parser-to-binder test that actually parses `SELECT SYSTEM_USER`.

thswlsqls:
Implemented as promised: `SYSTEM_USER` is now reachable in the SQL92 grammar via `unreservedWord`, so `SELECT SYSTEM_USER` parses to a column segment that `ColumnSegmentBinder` classifies as a niladic function through the unparenthesized-function set. Added the parser IT case `select_system_user_sql92` covering `SELECT SYSTEM_USER`.

terrymanu (changes requested):
### Summary

**Reason:** The latest commit fixes the previous parser-reachability issue, but the chosen representation conflates the bare SQL92 value specification with regular and quoted identifiers. This introduces a valid quoted-column regression and broadens `SYSTEM_USER` into unsupported identifier contexts.

### Issues

#### P1 — `SYSTEM_USER` is indistinguishable from quoted identifiers during binding

**Problem:** SQL-92 defines `SYSTEM_USER` as a reserved word and a bare `<general value specification>`, not an unreserved identifier ([SQL-92 dpANS X3H2-93-004, §§5.2 and 6.2](https://web.cecs.pdx.edu/~len/sql-92.pdf)). The patch instead adds it to [`unreservedWord`](https://github.com/apache/shardingsphere/blob/2ea43e5bbde1c798349c134c1a9ba1c971b71034/parser/sql/engine/dialect/sql92/src/main/antlr4/imports/sql92/BaseRule.g4#L65-L85), which also feeds table names, column names, owners, and [`regularFunctionName`](https://github.com/apache/shardingsphere/blob/2ea43e5bbde1c798349c134c1a9ba1c971b71034/parser/sql/engine/dialect/sql92/src/main/antlr4/imports/sql92/BaseRule.g4#L249-L255).

The metadata change then adds `SYSTEM_USER` to the [value-only function-name set](https://github.com/apache/shardingsphere/blob/2ea43e5bbde1c798349c134c1a9ba1c971b71034/database/connector/dialect/sql92/src/main/java/org/apache/shardingsphere/database/connector/sql92/metadata/database/option/SQL92FunctionOption.java#L31-L35). Double-quoted identifiers are parsed by [`IDENTIFIER_`](https://github.com/apache/shardingsphere/blob/2ea43e5bbde1c798349c134c1a9ba1c971b71034/parser/sql/engine/dialect/sql92/src/main/antlr4/imports/sql92/Literals.g4#L22-L25); [`IdentifierValue`](https://github.com/apache/shardingsphere/blob/2ea43e5bbde1c798349c134c1a9ba1c971b71034/parser/sql/statement/core/src/main/java/org/apache/shardingsphere/sql/parser/statement/core/value/identifier/IdentifierValue.java#L41-L47) unwraps their text while preserving the quote kind, but [`ColumnSegmentBinder`](https://github.com/apache/shardingsphere/blob/2ea43e5bbde1c798349c134c1a9ba1c971b71034/infra/binder/core/src/main/java/org/apache/shardingsphere/infra/binder/engine/segment/dml/expression/type/ColumnSegmentBinder.java#L81-L89) checks only the unwrapped value through [the set membership test](https://github.com/apache/shardingsphere/blob/2ea43e5bbde1c798349c134c1a9ba1c971b71034/infra/binder/core/src/main/java/org/apache/shardingsphere/infra/binder/engine/segment/dml/expression/type/ColumnSegmentBinder.java#L141-L143).

Consequently, a valid reference such as `SELECT t."SYSTEM_USER" FROM t` now returns before physical-column binding, although the quoted name is an identifier rather than the bare value specification.

**Impact:** Valid quoted columns named `SYSTEM_USER` lose their bound-column metadata. The grammar also admits the reserved value keyword through generic identifier productions, including forms such as `SYSTEM_USER()` that SQL-92 does not define. The new [parser case](https://github.com/apache/shardingsphere/blob/2ea43e5bbde1c798349c134c1a9ba1c971b71034/test/it/parser/src/main/resources/sql/supported/dml/select.xml#L465) and function-option unit assertion cover the bare input and set membership, but the [parser IT ends after AST assertion](https://github.com/apache/shardingsphere/blob/2ea43e5bbde1c798349c134c1a9ba1c971b71034/test/it/parser/src/main/java/org/apache/shardingsphere/test/it/sql/parser/internal/InternalSQLParserIT.java#L53-L65) and does not exercise this binder distinction.

**Discussion Needed:** Please pause the current implementation direction and agree on how SQL92 `SYSTEM_USER` should be represented so the bare value specification remains distinct from quoted, qualified, and generic identifiers. The discussion should establish the accepted bare/parenthesized/quoted boundaries and the parser-to-binder regression contract before implementation continues. `type: discussion` is appropriate.

### Multi-Round Comparison

- **Previous P1 — Fixed:** The prior finding that `SELECT SYSTEM_USER` never reached the binder is resolved by the latest grammar change and parser IT case.
- **Newly discovered blocker — Current:** The function-name entry introduced in the first PR revision strips the semantic distinction from quoted identifiers, while the latest grammar commit additionally exposes the keyword through generic identifier productions.

### Review Details

- **Review Focus:** Code Correctness Review; CI not reviewed by request.
- **Reviewed Scope:** All six GitHub-listed files at head `2ea43e5bbde1c798349c134c1a9ba1c971b71034`: `RELEASE-NOTES.md`, SQL92 function option and test, SQL92 `BaseRule.g4`, and both parser IT resources. Fetched `master` was `6e2462907e5880b2ac0c09fef24c1b2f182609d0`; the merge-base and PR base SHA were `c9b7a8f41f379844333b9285044d71ebeb191dbe`. The local triple-dot file list matched GitHub `/pulls/39102/files` 6/6. Target dialect: SQL92; repository conventions list [no SQL92 branch dialect](https://github.com/apache/shardingsphere/blob/2ea43e5bbde1c798349c134c1a9ba1c971b71034/CODE_OF_CONDUCT.md#L130-L140). The MySQL and SQLServer sibling implementations cited by the PR were checked against the [MySQL information-function syntax](https://dev.mysql.com/doc/refman/8.0/en/information-functions.html) and [SQLServer `SYSTEM_USER` syntax](https://learn.microsoft.com/en-us/sql/t-sql/functions/system-user-transact-sql?view=sql-server-ver17); their differing contracts do not establish SQL92 identifier semantics.
- **Not Reviewed Scope:** GitHub Actions, the full Maven suite, JDBC/Proxy E2E scenarios, and unrelated dialect behavior.
- **Verification:** Latest PR metadata, files, comments, and reviews were re-fetched successfully; the head remained `2ea43e5bbde1` and the six-file scope was unchanged. `build_review_inventory.py` completed with exit code 0 and confirmed the GitHub/local scope match. No reviewer-run Maven result is used because the active checkout was not the PR head; the semantic review used immutable exact-head git objects and public specifications. No GitHub evidence endpoint gap affected the result.
- **Release Note / User Docs:** The [release-note entry](https://github.com/apache/shardingsphere/blob/2ea43e5bbde1c798349c134c1a9ba1c971b71034/RELEASE-NOTES.md#L89) is present. Separate user documentation is not required for this parser compatibility fix.

thswlsqls:
Thanks for the detailed review. Agreed on the grammar point — routing it through `unreservedWord` also admits `SYSTEM_USER` as a table or owner name and as `SYSTEM_USER()`, which SQL-92 does not define, so I'll narrow it to the bare value specification. On the quoted case, `ColumnSegmentBinder` matches on the unwrapped identifier value and never inspects the quote kind, so `SELECT t."USER"` already skips binding on master today across every dialect that defines this set — that distinction looks binder-level rather than SQL92-specific. Happy to split it into a separate issue, or to handle it here if you prefer.

thswlsqls:
@terrymanu Gentle ping on my last comment — I'm waiting on your call for the quoted-identifier case (separate issue vs. handled here) before continuing.

terrymanu:
Thanks for following up. The finding is retained, with the scope clarified.

Please handle the `SYSTEM_USER`-specific boundary in this PR. Adding `SYSTEM_USER` to the [unparenthesized-function set](https://github.com/apache/shardingsphere/blob/2ea43e5bbde1c798349c134c1a9ba1c971b71034/database/connector/dialect/sql92/src/main/java/org/apache/shardingsphere/database/connector/sql92/metadata/database/option/SQL92FunctionOption.java#L31-L35) newly causes `t."SYSTEM_USER"` to take the [early-return path](https://github.com/apache/shardingsphere/blob/2ea43e5bbde1c798349c134c1a9ba1c971b71034/infra/binder/core/src/main/java/org/apache/shardingsphere/infra/binder/engine/segment/dml/expression/type/ColumnSegmentBinder.java#L141-L143) instead of being bound as a quoted column. Adding it to [`unreservedWord`](https://github.com/apache/shardingsphere/blob/2ea43e5bbde1c798349c134c1a9ba1c971b71034/parser/sql/engine/dialect/sql92/src/main/antlr4/imports/sql92/BaseRule.g4#L65-L85) also exposes it through generic identifier and regular-function contexts.

This PR does not need to solve the broader pre-existing quote-insensitive behavior for names already in the set or for other dialects; that can be tracked in a separate issue.

For this PR, please preserve these boundaries:

- Bare `SYSTEM_USER` is parsed and bound as the SQL-92 niladic value.
- `t."SYSTEM_USER"` remains a quoted column and receives normal column-bound metadata.
- Generic identifier uses and `SYSTEM_USER()` are not admitted unless supported by the SQL-92 contract.
- The regression test exercises the parser-to-binder path, not only the parser AST.

Once these `SYSTEM_USER` cases are covered, the broader binder behavior can be followed up separately.

thswlsqls:
@terrymanu Thanks for the clarification. Resuming on that scope — I'll also keep `t."SYSTEM_USER"` bound as a quoted column here, and the regression case will cover both through the binder rather than the parser AST.

thswlsqls:
@terrymanu Done, with one change of approach worth flagging. Keeping `SYSTEM_USER` in `UNPARENTHESIZED_FUNCTION_NAMES` cannot satisfy the quoted-column boundary, because `isUnparenthesizedFunction` matches the unwrapped identifier before any owner is read. The token is now admitted through `specialFunction` as a niladic value, dropped from `unreservedWord`, and the `SQL92FunctionOption` entry is reverted; `visitSpecialFunction` already yields the function segment, so the visitor is unchanged. A new `SQL92BinderIT` covers both boundaries through the binder: bare `SELECT SYSTEM_USER` binds as the niladic function, and `SELECT t."SYSTEM_USER" FROM t_system_user t` keeps full column bound info, which fails if the option entry comes back. That second case needed a `t_system_user` table in the shared binder fixture; the whole binder IT is green (103 cases), as is the SQL92 parser IT (210 cases).

terrymanu (approved):
### Result

**Review Result: Mergeable**

The latest head fixes the earlier parser-reachability and quoted-identifier regressions without broadening SQL92 identifier syntax.

### Evidence

- SQL-92 defines `SYSTEM_USER` as a reserved, bare general value specification—not a parenthesized function or generic identifier ([SQL-92 §§5.2 and 6.2](https://web.cecs.pdx.edu/~len/sql-92.pdf)).
- The grammar now admits it only through [`specialFunction → systemUserFunction`](https://github.com/apache/shardingsphere/blob/bad23543513ab89c7860aeecc3c899d9df8439f9/parser/sql/engine/dialect/sql92/src/main/antlr4/imports/sql92/BaseRule.g4#L221-L227), while generic identifiers and regular parenthesized functions remain separate.
- The visitor produces a zero-argument [`FunctionSegment`](https://github.com/apache/shardingsphere/blob/bad23543513ab89c7860aeecc3c899d9df8439f9/parser/sql/engine/dialect/sql92/src/main/java/org/apache/shardingsphere/sql/parser/engine/sql92/visitor/statement/SQL92StatementVisitor.java#L428-L487).
- `SYSTEM_USER` is absent from [`SQL92FunctionOption`](https://github.com/apache/shardingsphere/blob/bad23543513ab89c7860aeecc3c899d9df8439f9/database/connector/dialect/sql92/src/main/java/org/apache/shardingsphere/database/connector/sql92/metadata/database/option/SQL92FunctionOption.java#L31-L36), so quoted `t."SYSTEM_USER"` follows normal column binding.
- The new binder cases verify both the bare function and complete quoted-column metadata through the parser-to-binder path.
- Exact-head verification passed:
  - SQL92 parser: 208 tests, 0 failures/errors.
  - SQL92 binder: 2 tests, 0 failures/errors.
  - Boundary probes rejected `SYSTEM_USER()` and unquoted generic-identifier use with `SQLParsingException`.
  - All four changed XML resources validated successfully.

### Coverage

- Reviewed head: `bad23543513ab89c7860aeecc3c899d9df8439f9`.
- Base and merge-base: `c9b7a8f41f379844333b9285044d71ebeb191dbe`.
- Accounted for all 8 GitHub-listed files across grammar/AST behavior, parser expectations, parser-to-binder regression coverage, quoted-column binding, shared test metadata, and release notes.
- Rechecked linked issue #39099, the complete public discussion, both earlier reviews, and the latest delta. Both previous blocking findings are fixed.
- Completed root-cause, blast-radius/contracts, tests/runtime, and final convergence passes. No unresolved evidence gaps remain.
- SQL92 has no branch dialect; relevant MySQL and SQLServer implementations use independent grammars and contracts.
- Code-scope result only. CI and GitHub Actions were not queried or reviewed.

PR #38639, opened 2026-04-20
Revert "Fix incorrect routing when irrelevant sharding conditions are…

FlyingZC:
… present (#38527)"

This reverts commit dcbbaf9d

Fixes #ISSUSE_ID.

Changes proposed in this pull request:
  -

---

Before committing this PR, I'm sure that I have checked the following options:
- [ ] My code follows the [code of conduct](https://shardingsphere.apache.org/community/en/involved/conduct/code/) of this project.
- [ ] I have self-reviewed the commit code.
- [ ] I have (or in comment I request) added corresponding labels for the pull request.
- [ ] I have passed maven check locally : `./mvnw clean install -B -T1C -Dmaven.javadoc.skip -Dmaven.jacoco.skip -e`.
- [ ] I have made corresponding changes to the documentation.
- [ ] I have added corresponding unit tests for my changes.
- [ ] I have updated the Release Notes of the current development version. For more details, see [Update Release Note](https://shardingsphere.apache.org/community/en/involved/contribute/contributor/)

PR #39112, opened 2026-07-14
Skip sharding conditions irrelevant to the routed table

Develop-KIM:
Fixes #38456.

  - Skip sharding conditions that carry no value for the routed table or its binding table group, so a subquery over two non-binding sharding tables no longer broadcasts to every data node.
  - Record a placeholder entry in `originalDataNodes` for the skipped conditions, keeping it positionally aligned with the condition list.
  - Treat a condition that has no sharding values at all as relevant, preserving its existing full-route behaviour.

### Why the route explodes

For a subquery over two non-binding sharding tables, `ShardingConditionEngine` emits one `ShardingCondition` per table instead of merging them into a single condition the way it does for the equivalent JOIN. While routing one of the tables, the condition that belongs to the *other* table yields no sharding values, so `route0` reads it as "no condition given" and broadcasts. The reporter saw 512 actual SQLs (4 datasources x 128 shards) for a query that should reach one shard.

### Why this does not repeat #38527

[#38527](https://github.com/apache/shardingsphere/pull/38527) filtered the same conditions but dropped them with `continue`, and fell back to a bare `route0` when every condition was dropped. Two things went wrong, and [#38639](https://github.com/apache/shardingsphere/pull/38639) reverted it:

1. `RouteSQLRewriteEngine#buildRouteParameters` walks `originalDataNodes` **by index** and feeds that index straight to `paramBuilder.getParameters(count)`. Dropping an entry shifts every later index, so parameters can land on the wrong row. This PR appends an empty collection instead of skipping, which leaves the indices intact and reads as "matches every route unit" in `isInSameDataNode`, exactly as before.
2. A condition with no sharding values (for example an INSERT into a table with no sharding strategy) was classed as irrelevant and sent down the fallback path, which never populated `originalDataNodes`. `RouteSQLRewriteEngine#getParameters` then took the short path instead of `buildRouteParameters`. Here such a condition is treated as relevant, so its full-route behaviour is unchanged.

`assertRouteByShardingConditionWithoutShardingValues` covers the second point: dropping the empty-values guard turns `originalDataNodes` from `[4 nodes]` into `[]` and the test fails, so the regression cannot come back unnoticed.

### Scope

`routeByMixedConditionsWithCondition` is left alone on purpose, even though the issue mentions the same shape there. That path mixes hint strategies, where a condition that looks irrelevant can still contribute a valid hint value, so the same filter is not obviously safe. Happy to extend it in this PR or a follow-up if you would prefer.

### Verification

- `assertRouteByShardingConditionsWithIrrelevantTableCondition` fails on master (`expected: <1> but: was <4>`) and passes with the fix.
- `./mvnw -pl features/sharding/core -Drat.skip=true test` — 853 tests, checkstyle included.
- `./mvnw -pl infra/rewrite/core -Drat.skip=true test` — 87 tests.

This was written with AI assistance (disclosed in the commit trailer); I have reviewed the reasoning and the diff and can speak to every line.

---

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -41,6 +41,7 @@
 1. Proxy: Fix MySQL BLOB data corruption when string-like prepared statement parameters target BLOB columns - [#39072](https://github.com/apache/shardingsphere/pull/39072)
 1. Agent: Fix wrong target class name in StaticMethodAdviceExecutor error logs - [#39077](https://github.com/apache/shardingsphere/pull/39077)
 1. Metadata: Fix Oracle metadata version comparison skipping identity and collation columns on 18c and later - [#39104](https://github.com/apache/shardingsphere/pull/39104)
+1. Sharding: Skip sharding conditions irrelevant to the routed table to avoid full-route broadcast for subqueries over non-binding sharding tables - [#39112](https://github.com/apache/shardingsphere/pull/39112)
 
 ### Enhancements
 
--- a/features/sharding/core/src/main/java/org/apache/shardingsphere/sharding/route/engine/type/standard/ShardingStandardRouteEngine.java
+++ b/features/sharding/core/src/main/java/org/apache/shardingsphere/sharding/route/engine/type/standard/ShardingStandardRouteEngine.java
@@ -131,8 +131,11 @@ public final class ShardingStandardRouteEngine implements ShardingRouteEngine {
     private Collection<DataNode> routeByShardingConditionsWithCondition(final ShardingRule shardingRule, final ShardingTable shardingTable,
                                                                         final ShardingStrategy databaseShardingStrategy, final ShardingStrategy tableShardingStrategy) {
         Collection<DataNode> result = new LinkedList<>();
+        boolean anyConditionRelevant = false;
         for (ShardingCondition each : shardingConditions.getConditions()) {
-            Collection<DataNode> dataNodes = isConditionRelevantToTable(shardingRule, each)
+            boolean relevant = isConditionRelevantToTable(shardingRule, each);
+            anyConditionRelevant |= relevant;
+            Collection<DataNode> dataNodes = relevant
                     ? route0(shardingTable,
                             databaseShardingStrategy, getShardingValuesFromShardingConditions(shardingRule, databaseShardingStrategy.getShardingColumns(), each),
                             tableShardingStrategy, getShardingValuesFromShardingConditions(shardingRule, tableShardingStrategy.getShardingColumns(), each))
@@ -140,7 +143,10 @@ public final class ShardingStandardRouteEngine implements ShardingRouteEngine {
             result.addAll(dataNodes);
             originalDataNodes.add(dataNodes);
         }
-        return result.isEmpty()
+        // Full-route only when every condition was irrelevant; a relevant condition that legitimately
+        // routes to no target (e.g. an out-of-range interval value) must keep its empty result rather
+        // than broadcast to every shard.
+        return result.isEmpty() && !anyConditionRelevant
                 ? route0(shardingTable, databaseShardingStrategy, Collections.emptyList(), tableShardingStrategy, Collections.emptyList())
                 : result;
     }
--- a/features/sharding/core/src/main/java/org/apache/shardingsphere/sharding/route/engine/type/standard/ShardingStandardRouteEngine.java
+++ b/features/sharding/core/src/main/java/org/apache/shardingsphere/sharding/route/engine/type/standard/ShardingStandardRouteEngine.java
@@ -132,13 +132,33 @@ public final class ShardingStandardRouteEngine implements ShardingRouteEngine {
                                                                         final ShardingStrategy databaseShardingStrategy, final ShardingStrategy tableShardingStrategy) {
         Collection<DataNode> result = new LinkedList<>();
         for (ShardingCondition each : shardingConditions.getConditions()) {
-            Collection<DataNode> dataNodes = route0(shardingTable,
-                    databaseShardingStrategy, getShardingValuesFromShardingConditions(shardingRule, databaseShardingStrategy.getShardingColumns(), each),
-                    tableShardingStrategy, getShardingValuesFromShardingConditions(shardingRule, tableShardingStrategy.getShardingColumns(), each));
+            Collection<DataNode> dataNodes = isConditionRelevantToTable(shardingRule, each)
+                    ? route0(shardingTable,
+                            databaseShardingStrategy, getShardingValuesFromShardingConditions(shardingRule, databaseShardingStrategy.getShardingColumns(), each),
+                            tableShardingStrategy, getShardingValuesFromShardingConditions(shardingRule, tableShardingStrategy.getShardingColumns(), each))
+                    : Collections.<DataNode>emptyList();
             result.addAll(dataNodes);
             originalDataNodes.add(dataNodes);
         }
-        return result;
+        return result.isEmpty()
+                ? route0(shardingTable, databaseShardingStrategy, Collections.emptyList(), tableShardingStrategy, Collections.emptyList())
+                : result;
+    }
+    
+    private boolean isConditionRelevantToTable(final ShardingRule shardingRule, final ShardingCondition shardingCondition) {
+        if (shardingCondition.getValues().isEmpty()) {
+            return true;
+        }
+        for (ShardingConditionValue each : shardingCondition.getValues()) {
+            if (logicTableName.equalsIgnoreCase(each.getTableName())) {
+                return true;
+            }
+            Optional<BindingTableRule> bindingTableRule = shardingRule.findBindingTableRule(each.getTableName());
+            if (bindingTableRule.isPresent() && bindingTableRule.get().hasLogicTable(logicTableName)) {
+                return true;
+            }
+        }
+        return false;
     }
     
     private Collection<DataNode> routeByMixedConditions(final ShardingRule shardingRule, final ShardingTable shardingTable,
--- a/features/sharding/core/src/test/java/org/apache/shardingsphere/sharding/route/engine/type/standard/ShardingStandardRouteEngineTest.java
+++ b/features/sharding/core/src/test/java/org/apache/shardingsphere/sharding/route/engine/type/standard/ShardingStandardRouteEngineTest.java
@@ -24,7 +24,9 @@ import org.apache.shardingsphere.infra.hint.HintValueContext;
 import org.apache.shardingsphere.infra.route.context.RouteContext;
 import org.apache.shardingsphere.infra.route.context.RouteUnit;
 import org.apache.shardingsphere.sharding.exception.algorithm.ShardingRouteAlgorithmException;
+import org.apache.shardingsphere.sharding.route.engine.condition.ShardingCondition;
 import org.apache.shardingsphere.sharding.route.engine.condition.ShardingConditions;
+import org.apache.shardingsphere.sharding.route.engine.condition.value.ListShardingConditionValue;
 import org.apache.shardingsphere.sharding.route.engine.fixture.ShardingRouteEngineFixtureBuilder;
 import org.apache.shardingsphere.sharding.rule.ShardingRule;
 import org.junit.jupiter.api.AfterEach;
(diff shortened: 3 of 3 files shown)


terrymanu (changes requested):
### Summary

**Reason:** The irrelevant-condition filtering direction is sound, but the new fallback conflates “no relevant condition” with “a relevant condition routed to no table,” introducing a full-route regression.

### Issues

#### P1: Preserve legitimate zero-target routing results

**Problem:** [`routeByShardingConditionsWithCondition`](https://github.com/apache/shardingsphere/blob/451277b53152338bd492d9c9aae34b1ea6f0880a/features/sharding/core/src/main/java/org/apache/shardingsphere/sharding/route/engine/type/standard/ShardingStandardRouteEngine.java#L131-L145) now performs full routing whenever the accumulated result is empty. However, a relevant condition can legitimately produce no table targets: [`routeTables`](https://github.com/apache/shardingsphere/blob/451277b53152338bd492d9c9aae34b1ea6f0880a/features/sharding/core/src/main/java/org/apache/shardingsphere/sharding/route/engine/type/standard/ShardingStandardRouteEngine.java#L281-L289) permits an empty table-sharding result, and [`IntervalShardingAlgorithm`](https://github.com/apache/shardingsphere/blob/451277b53152338bd492d9c9aae34b1ea6f0880a/features/sharding/core/src/main/java/org/apache/shardingsphere/sharding/algorithm/sharding/datetime/IntervalShardingAlgorithm.java#L127-L136) returns no target for a precise value outside its configured interval. [`StandardShardingStrategy`](https://github.com/apache/shardingsphere/blob/451277b53152338bd492d9c9aae34b1ea6f0880a/features/sharding/core/src/main/java/org/apache/shardingsphere/sharding/route/strategy/type/standard/StandardShardingStrategy.java#L66-L77) preserves that as an empty result. Before this PR, the route engine returned the empty result; this PR reruns `route0` without sharding values and selects every configured table.

**Impact:** A query known to match no configured shard can be broadcast to every shard, changing both result semantics and routing cost.

**Required Change:** Track whether at least one condition was relevant independently from whether routing produced nodes. Use full-route fallback only when every condition was irrelevant; preserve an empty result from relevant conditions. Please add a focused regression test where a relevant table condition—such as an out-of-range `INTERVAL` value—produces no target and assert that no route units are created.

### Review Details

- **Review Focus:** Code Correctness Review; CI not reviewed by request.
- **Reviewed Scope:** All three changed files at head `451277b53152338bd492d9c9aae34b1ea6f0880a`: `RELEASE-NOTES.md`, `ShardingStandardRouteEngine.java`, and `ShardingStandardRouteEngineTest.java`. Supporting route, strategy, interval-algorithm, Cartesian-route, and rewrite paths were traced. Local merge-base: `7b8178371a5be5cebb11bdbf5b96fbb437953f48`. The local triple-dot file list matched GitHub `/pulls/39112/files`.
- **Not Reviewed Scope:** GitHub Actions/check runs and runtime E2E execution. No changed file was omitted.
- **Verification:** Public PR metadata, changed files, linked issue #38456, prior fixes/revert #38527/#38639, comments, and reviews were accessible. The bounded review inventory and triple-dot scope checks exited successfully. Maven was not run because the shared checkout is not at the PR head; non-head test results were not used as evidence.
- **Release Note / User Docs:** The release-note entry is present. Additional user documentation is not required because this is an internal routing correction without configuration or migration changes.

Develop-KIM:
Thanks for the careful review — you're right, the empty-result fallback conflated the two cases.

Fixed it: I now track whether any condition was relevant to the routed table (`anyConditionRelevant`) separately from whether routing produced nodes. The full-route fallback only fires when every condition was irrelevant; a relevant condition that legitimately routes to no target now keeps its empty result instead of broadcasting.

Added the regression test you asked for: `assertRouteByRelevantConditionRoutingToNoTargetKeepsEmptyResult` uses an out-of-range INTERVAL value (`2025-06-01` against a `2021-01-01`/`2021-01-02` range) so the condition is relevant to `t_interval_test` but the algorithm returns no target, and asserts zero route units. I confirmed it fails on the previous code (routed to 2 shards) and passes now.

terrymanu (changes requested):
### Summary

**Reason:** #38456 exposes a real routing problem, but the current evidence does not establish that it should be fixed inside the stable `ShardingStandardRouteEngine`. This implementation uses a boolean flag and empty collections to encode multiple routing states without carrying the relational and positive/negative semantics of the subquery. It already produces order-dependent and incomplete `RouteUnit`s in the Complex/Cartesian path. This is not a local implementation problem that can be solved by adding branches, extracting methods, or removing comments. The current direction should be paused and the ownership boundary reconsidered.

**Expert Review Needed:** Maintainers familiar with the Sharding Kernel, Complex/Cartesian routing, and Federation subquery semantics should define the end-to-end contract first.

### Issues

#### P1: This layer cannot determine the correct routing semantics of a subquery condition

**Problem:** [`ShardingStandardRouteEngine`](https://github.com/apache/shardingsphere/blob/64e1ed1a33e5cfd30fbc101494b860f4cc6b864b/features/sharding/core/src/main/java/org/apache/shardingsphere/sharding/route/engine/type/standard/ShardingStandardRouteEngine.java#L131-L167) introduces `anyConditionRelevant` and uses empty collections to represent all of the following:

- The condition belongs to another table and exists only to preserve the position in `originalDataNodes`.
- The condition is relevant to the current table, but the sharding algorithm produced no target.
- The routing result has not been initialized yet.

These distinctions exist only in a local boolean inside this method. Once the result enters `RouteContext`, they all collapse into an empty collection. At the same time, [`ShardingCartesianRouteEngine#getIntersectionDataSources`](https://github.com/apache/shardingsphere/blob/64e1ed1a33e5cfd30fbc101494b860f4cc6b864b/features/sharding/core/src/main/java/org/apache/shardingsphere/sharding/route/engine/type/complex/ShardingCartesianRouteEngine.java#L75-L84) also uses an empty set for both “not initialized” and “the intersection is empty”:

- If the empty child route is processed first, the following non-empty route repopulates the datasource set and produces a `RouteUnit` containing only the later table.
- If the empty child route is processed last, the intersection remains empty and no `RouteUnit` is produced.

The public reproduction against the current head has already shown that merely reversing the table order in the same positive `EXISTS` query produces either one route missing a participating table mapper or no route at all. [`ShardingTableToken`](https://github.com/apache/shardingsphere/blob/64e1ed1a33e5cfd30fbc101494b860f4cc6b864b/features/sharding/core/src/main/java/org/apache/shardingsphere/sharding/rewrite/token/pojo/ShardingTableToken.java#L56-L59) falls back to the logical table name when the mapper is missing, so the resulting physical SQL can reference a table that does not exist.

This layer also lacks enough information to determine the meaning of an empty child relation. An empty child under positive `EXISTS` or `IN` should eliminate the outer result, while the same empty child under `NOT EXISTS` must preserve the outer route. [`ShardingCondition`](https://github.com/apache/shardingsphere/blob/64e1ed1a33e5cfd30fbc101494b860f4cc6b864b/features/sharding/core/src/main/java/org/apache/shardingsphere/sharding/route/engine/condition/ShardingCondition.java#L33-L38) carries only values and a position; it does not carry this relational context. Filtering by table name cannot derive the correct result.

The newly added ordinary comment explains only one of these branches. The project rules state that code requiring explanatory comments should be expressed through extracted methods, and that only Javadoc, TODO, and FIXME comments are allowed ([`CODE_OF_CONDUCT.md`](https://github.com/apache/shardingsphere/blob/64e1ed1a33e5cfd30fbc101494b860f4cc6b864b/CODE_OF_CONDUCT.md#L50-L52), [L80-L83](https://github.com/apache/shardingsphere/blob/64e1ed1a33e5cfd30fbc101494b860f4cc6b864b/CODE_OF_CONDUCT.md#L80-L83)). However, removing the comment or extracting a method would not restore the missing cross-layer state contract. The comment is evidence of an implicit-state design problem, not merely a formatting issue.

The added tests construct `ShardingConditions` directly and invoke the Standard route engine ([`ShardingStandardRouteEngineTest`](https://github.com/apache/shardingsphere/blob/64e1ed1a33e5cfd30fbc101494b860f4cc6b864b/features/sharding/core/src/test/java/org/apache/shardingsphere/sharding/route/engine/type/standard/ShardingStandardRouteEngineTest.java#L208-L254)). They do not pass through Parser, Binder, Federation decision, Complex, Cartesian, or Rewrite. Therefore, they prove only that the local branch follows the assumptions of the current implementation; they do not prove the end-to-end semantics required by #38456.

**Impact:** This change affects a stable core routing path and its blast radius is not limited to #38456. The current head can turn an inefficient but fully mapped query into an order-dependent result and can even produce physical SQL with a missing participating-table mapping.

**Discussion Needed:** Please do not continue adding branches or state variables to `routeByShardingConditionsWithCondition`. I suggest pausing and withdrawing the current Standard-route implementation until the following points are agreed:

- Whether “one actual SQL” for a subquery over non-binding tables is part of the Kernel routing contract, or should instead be owned by the existing [`ShardingSQLFederationDecider`](https://github.com/apache/shardingsphere/blob/64e1ed1a33e5cfd30fbc101494b860f4cc6b864b/features/sharding/core/src/main/java/org/apache/shardingsphere/sharding/decider/ShardingSQLFederationDecider.java#L70-L95) and Federation path.
- Whether query-block identity, positive/negative relation semantics, and participating-table information should be owned by the Condition, Federation, or Complex routing layer.
- How `FULL_ROUTE`, `NO_TARGET`, and `NOT_PARTICIPATING` should be represented explicitly across Standard, Complex, Cartesian, and Rewrite instead of continuing to overload empty collections.

I do not recommend further iteration on the current implementation until this contract has been agreed.

### Multi-Round Comparison

- The first-round request to avoid broadcasting again when a relevant condition produces no target has been addressed at the standalone Standard-route level.
- The latest end-to-end analysis shows that this change passes an unmodeled empty result into Complex/Cartesian routing, exposing a more fundamental ownership and state-contract problem.
- Therefore, this round does not request additional local tests, comment cleanup, or further changes to the current method. It asks that the current implementation direction be paused until the design is resolved.
- No second independent core blocker was found.

### Review Details

- **Review Focus:** Code Correctness Review; CI not reviewed by request.
- **Reviewed Scope:** All three files listed by GitHub `/pulls/39112/files`: `RELEASE-NOTES.md`, `ShardingStandardRouteEngine.java`, and `ShardingStandardRouteEngineTest.java`. The latest head is `64e1ed1a33e5cfd30fbc101494b860f4cc6b864b`; the base and local merge-base are both `7b8178371a5be5cebb11bdbf5b96fbb437953f48`. The local triple-dot file list matched GitHub exactly. The adjacent Condition, Federation, Complex, Cartesian, RouteContext, SQL Rewrite, and table-token paths were also traced.
- **Not Reviewed Scope:** GitHub Actions/check runs, execution against a real database, and a new Federation runtime reproduction. No changed file was omitted.
- **Verification:** PR metadata, the complete changed-file list, issue #38456, PR #38527, revert PR #38639, public comments, and both review rounds were accessible. `git merge-base`, triple-dot scope verification, current-head static path tracing, and the full-coverage ledger validation all completed with exit `0`. Maven was not run because the shared checkout was not at the PR head, and results from non-current source were not used. The public positive/negative subquery and table-order reproductions from the same unchanged head were inspected.
- **Release Note / User Docs:** The release-note entry is present. The blocker is the implementation direction and routing contract; additional user documentation would not resolve it.
- **Dependency / Distribution:** No API, SPI, configuration, dependency, generated-resource, packaging, or distribution changes were introduced.

Develop-KIM:
Thanks for the thorough writeup, @terrymanu — this is convincing.

You're right that I was treating this as a local fix inside `ShardingStandardRouteEngine`. But encoding "belongs to another table" vs. "relevant but no target" vs. "not yet initialized" as a boolean plus empty collections can't carry the positive/negative (`EXISTS` vs `NOT EXISTS`) or query-block semantics the subquery actually needs, and — as you point out — the Cartesian intersection already makes the outcome order-dependent, so the same `EXISTS` query can lose a participating table or produce no route at all depending on child ordering. That's a cross-layer contract question — whether this belongs to the Federation path or the Kernel, and where query-block identity and relation semantics should live — not something to settle by adding branches from the outside.

So I'll withdraw this and close it, rather than leave it as a partial answer to #38456. If the ownership boundary gets defined and there's a well-scoped piece to pick up, I'd be glad to help. Thanks again for taking the time on the detailed review.

PR #39131, opened 2026-07-15
Fix select * is not supported for associated query

### LINKED ISSUE #28841: select * is not supported for associated query
## Bug Report

5.4.1

ShardingSphere-JDBC

### Expected behavior
select * is supported
### Actual behavior
The statements supported by version 5.0 are not supported by version 5.4.1
SELECT
	A.FLAG,
	B.* 
FROM
	TABLE_A A
	LEFT JOIN TABLE_B B ON A.ID = B.ID 
WHERE
	A.CD = '123'

Caused by: java.sql.SQLFeatureNotSupportedException: Can not get index from column label `ID`.
	at org.apache.shardingsphere.driver.jdbc.core.resultset.ShardingSphereResultSet.lambda$getIndexFromColumnLabelAndIndexMap$0(ShardingSphereResultSet.java:395)
	at org.apache.shardingsphere.infra.util.exception.ShardingSpherePreconditions.checkState(ShardingSpherePreconditions.java:41)
	at org.apache.shardingsphere.driver.jdbc.core.resultset.ShardingSphereResultSet.getIndexFromColumnLabelAndIndexMap(ShardingSphereResultSet.java:395)
	at org.apache.shardingsphere.driver.jdbc.core.resultset.ShardingSphereResultSet.getString(ShardingSphereResultSet.java:166)
	at com.zaxxer.hikari.pool.HikariProxyResultSet.getString(HikariProxyResultSet.java)
	at org.apache.ibatis.type.StringTypeHandler.getNullableResult(StringTypeHandler.java:37)
	at org.apache.ibatis.type.StringTypeHandler.getNullableResult(StringTypeHandler.java:26)
	at org.apache.ibatis.type.BaseTypeHandler.getResult(BaseTypeHandler.java:85)
	at org.apache.ibatis.executor.resultset.DefaultResultSetHandler.applyAutomaticMappings(DefaultResultSetHandler.java:560)
	at org.apache.ibatis.executor.resultset.DefaultResultSetHandler.getRowValue(DefaultResultSetHandler.java:402)
	at org.apache.ibatis.executor.resultset.DefaultResultSetHandler.handleRowValuesForSimpleResultMap(DefaultResultSetHandler.java:354)
	at org.apache.ibatis.executor.resultset.DefaultResultSetHandler.handleRowValues(DefaultResultSetHandler.java:328)
	at org.apache.ibatis.executor.resultset.DefaultResultSetHandler.handleResultSet(DefaultResultSetHandler.java:301)
	at org.apache.ibatis.executor.resultset.DefaultResultSetHandler.handleResultSets(DefaultResultSetHandler.java:194)
	at org.apache.ibatis.executor.statement.PreparedStatementHandler.query(PreparedStatementHandler.java:65)
	at org.apache.ibatis.executor.statement.RoutingStatementHandler.query(RoutingStatementHandler.java:79)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)

SELECT
	A.FLAG,
	B.* 
FROM
	TABLE_A A
	LEFT JOIN TABLE_B B ON A.ID = B.ID 
WHERE
	A.CD = '123'

ClaireLytt:
Fixes https://github.com/apache/shardingsphere/issues/28841

  -

---

### DIFF
--- a/jdbc/src/main/java/org/apache/shardingsphere/driver/jdbc/core/resultset/ShardingSphereResultSetUtils.java
+++ b/jdbc/src/main/java/org/apache/shardingsphere/driver/jdbc/core/resultset/ShardingSphereResultSetUtils.java
@@ -26,6 +26,7 @@ import org.apache.shardingsphere.infra.binder.context.statement.type.dml.SelectS
 import java.sql.ResultSetMetaData;
 import java.sql.SQLException;
 import java.util.Map;
+import java.util.Map.Entry;
 
 /**
  * Utility class for {@link ShardingSphereResultSet}.
@@ -43,8 +44,28 @@ public final class ShardingSphereResultSetUtils {
      */
     public static Map<String, Integer> createColumnLabelAndIndexMap(final SQLStatementContext sqlStatementContext, final ResultSetMetaData resultSetMetaData) throws SQLException {
         if (sqlStatementContext instanceof SelectStatementContext && ((SelectStatementContext) sqlStatementContext).containsDerivedProjections()) {
-            return ((SelectStatementContext) sqlStatementContext).getProjectionsContext().getColumnLabelAndIndexMap();
+            return createColumnLabelAndIndexMapWithExpandProjections(
+                    ((SelectStatementContext) sqlStatementContext).getProjectionsContext().getColumnLabelAndIndexMap(), resultSetMetaData);
         }
+        return createColumnLabelAndIndexMapFromMetaData(resultSetMetaData);
+    }
+    
+    private static Map<String, Integer> createColumnLabelAndIndexMapWithExpandProjections(final Map<String, Integer> expandedMap,
+                                                                                          final ResultSetMetaData resultSetMetaData) throws SQLException {
+        if (null == resultSetMetaData || expandedMap.size() == resultSetMetaData.getColumnCount()) {
+            return expandedMap;
+        }
+        Map<String, Integer> result = new CaseInsensitiveMap<>(resultSetMetaData.getColumnCount(), 1F);
+        for (int columnIndex = resultSetMetaData.getColumnCount(); columnIndex > 0; columnIndex--) {
+            result.put(resultSetMetaData.getColumnLabel(columnIndex), columnIndex);
+        }
+        for (Entry<String, Integer> entry : expandedMap.entrySet()) {
+            result.putIfAbsent(entry.getKey(), entry.getValue());
+        }
+        return result;
+    }
+    
+    private static Map<String, Integer> createColumnLabelAndIndexMapFromMetaData(final ResultSetMetaData resultSetMetaData) throws SQLException {
         Map<String, Integer> result = new CaseInsensitiveMap<>(resultSetMetaData.getColumnCount(), 1F);
         for (int columnIndex = resultSetMetaData.getColumnCount(); columnIndex > 0; columnIndex--) {
             result.put(resultSetMetaData.getColumnLabel(columnIndex), columnIndex);
--- a/jdbc/src/test/java/org/apache/shardingsphere/driver/jdbc/core/resultset/ShardingSphereResultSetUtilsTest.java
+++ b/jdbc/src/test/java/org/apache/shardingsphere/driver/jdbc/core/resultset/ShardingSphereResultSetUtilsTest.java
@@ -65,4 +65,21 @@ class ShardingSphereResultSetUtilsTest {
         Map<String, Integer> actual = ShardingSphereResultSetUtils.createColumnLabelAndIndexMap(selectStatementContext, null);
         assertThat(actual, is(expected));
     }
+    
+    @Test
+    void assertCreateColumnLabelAndIndexMapWhenExpandMapSizeMismatchMetaData() throws SQLException {
+        SelectStatementContext selectStatementContext = mock(SelectStatementContext.class);
+        when(selectStatementContext.containsDerivedProjections()).thenReturn(true);
+        List<Projection> projections = new ArrayList<>(1);
+        projections.add(new ColumnProjection(null, "status_new", null, mock(DatabaseType.class)));
+        ProjectionsContext projectionsContext = new ProjectionsContext(0, 0, false, projections);
+        when(selectStatementContext.getProjectionsContext()).thenReturn(projectionsContext);
+        ResultSetMetaData resultSetMetaData = mock(ResultSetMetaData.class);
+        when(resultSetMetaData.getColumnCount()).thenReturn(2);
+        when(resultSetMetaData.getColumnLabel(1)).thenReturn("order_id");
+        when(resultSetMetaData.getColumnLabel(2)).thenReturn("status_new");
+        Map<String, Integer> actual = ShardingSphereResultSetUtils.createColumnLabelAndIndexMap(selectStatementContext, resultSetMetaData);
+        assertThat(actual.get("order_id"), is(1));
+        assertThat(actual.get("status_new"), is(2));
+    }
 }
--- a/test/e2e/sql/src/test/java/org/apache/shardingsphere/test/e2e/sql/it/sql/dql/GeneralDQLE2EIT.java
+++ b/test/e2e/sql/src/test/java/org/apache/shardingsphere/test/e2e/sql/it/sql/dql/GeneralDQLE2EIT.java
@@ -17,6 +17,7 @@
 
 package org.apache.shardingsphere.test.e2e.sql.it.sql.dql;
 
+import org.apache.shardingsphere.infra.hint.HintManager;
 import org.apache.shardingsphere.test.e2e.env.runtime.E2ETestEnvironment;
 import org.apache.shardingsphere.test.e2e.sql.cases.value.SQLValue;
 import org.apache.shardingsphere.test.e2e.sql.framework.SQLE2EITArgumentsProvider;
@@ -39,10 +40,16 @@ import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
 
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 @SQLE2EITSettings(SQLCommandType.DQL)
 class GeneralDQLE2EIT extends BaseDQLE2EIT {
+    
+    // HintManager DATA_SOURCE_NAME keeps logic table name (no rewrite);
+    // bare labels from T.* must remain accessible via getObject(columnLabel).
+    private static final String SHORTHAND_ALIAS_SQL =
+            "SELECT T.*, T.status status_new FROM t_order T WHERE T.order_id = 1000";
     
     @ParameterizedTest(name = "{0}")
     @Execution(ExecutionMode.CONCURRENT)
@@ -256,6 +263,55 @@ class GeneralDQLE2EIT extends BaseDQLE2EIT {
                 assertResultSet(actualResultSet, expectedResultSet, testParam);
             }
         }
+    }
+    
+    @ParameterizedTest(name = "{0}")
+    @Execution(ExecutionMode.SAME_THREAD)
+    @EnabledIf("isEnabled")
+    @ArgumentsSource(HintManagerColumnLabelArgumentsProvider.class)
+    void assertColumnLabelWithHintManager(final AssertionTestParameter testParam) throws SQLException, IOException, JAXBException {
+        SQLE2EITContext context = new SQLE2EITContext(testParam);
+        executeDQL(context, () -> {
+            init(testParam, context);
+            assertDoesNotThrow(() -> executeShorthandAliasQueryWithoutDataSourceHint(),
+                    "Without data-source hint + T.*, T.status status_new should succeed");
+            assertDoesNotThrow(() -> executeShorthandAliasQueryWithDataSourceHint(),
+                    "HintManager.setDataSourceName + SELECT T.*, T.status status_new "
+                            + "should allow getObject by bare column label from T.*");
+        });
+    }
+    
+    private void executeShorthandAliasQueryWithoutDataSourceHint() throws SQLException {
+        try (
+                Connection connection = getEnvironmentEngine().getTargetDataSource().getConnection();
+                Statement statement = connection.createStatement();
+                ResultSet resultSet = statement.executeQuery(SHORTHAND_ALIAS_SQL)) {
+            assertTrue(resultSet.next(), "Expected row for order_id=1000");
+            assertBareColumnLabelsFromShorthandAccessible(resultSet);
+        }
+    }
+    
+    private void executeShorthandAliasQueryWithDataSourceHint() throws SQLException {
+        try (
+                HintManager hintManager = HintManager.getInstance();
+                Connection connection = getEnvironmentEngine().getTargetDataSource().getConnection();
+                Statement statement = connection.createStatement()) {
+            hintManager.setDataSourceName("tbl");
+            try (ResultSet resultSet = statement.executeQuery(SHORTHAND_ALIAS_SQL)) {
+                assertTrue(resultSet.next(), "Expected row for order_id=1000 on physical tbl.t_order");
+                assertBareColumnLabelsFromShorthandAccessible(resultSet);
+            }
+        }
+    }
+    
+    private void assertBareColumnLabelsFromShorthandAccessible(final ResultSet resultSet) throws SQLException {
+        // Look up bare labels by name (same pattern as MyBatis automatic mapping).
+        // Do not iterate ResultSetMetaData labels: when containsDerivedProjections()=true,
+        // ShardingSphereResultSetMetaData.getColumnLabel() also comes from expandProjections.
+        resultSet.getObject("order_id");
+        resultSet.getObject("user_id");
+        resultSet.getObject("status");
+        resultSet.getObject("status_new");
     }
     
     private static boolean isEnabled() {
--- a/test/e2e/sql/src/test/java/org/apache/shardingsphere/test/e2e/sql/it/sql/dql/HintManagerColumnLabelArgumentsProvider.java
+++ b/test/e2e/sql/src/test/java/org/apache/shardingsphere/test/e2e/sql/it/sql/dql/HintManagerColumnLabelArgumentsProvider.java
@@ -1,0 +1,64 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.shardingsphere.test.e2e.sql.it.sql.dql;
+
+import com.google.common.base.Preconditions;
+import org.apache.shardingsphere.test.e2e.sql.framework.param.array.E2ETestParameterFactory;
+import org.apache.shardingsphere.test.e2e.sql.framework.param.model.AssertionTestParameter;
+import org.apache.shardingsphere.test.e2e.sql.framework.type.SQLCommandType;
+import org.apache.shardingsphere.test.e2e.sql.framework.type.SQLExecuteType;
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.ArgumentsProvider;
+import org.junit.jupiter.params.support.ParameterDeclarations;
+
+import java.util.Collection;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/**
+ * Arguments provider for HintManager column label DQL cases.
+ */
+public final class HintManagerColumnLabelArgumentsProvider implements ArgumentsProvider {
+    
+    private static final String JOIN_SHORTHAND_SQL_PREFIX = "SELECT o.user_id, i.* FROM t_order o JOIN t_order_item i";
+    
+    @Override
+    public Stream<? extends Arguments> provideArguments(final ParameterDeclarations parameters, final ExtensionContext context) {
+        Collection<AssertionTestParameter> matched = E2ETestParameterFactory.getAssertionTestParameters(SQLCommandType.DQL).stream()
+                .filter(HintManagerColumnLabelArgumentsProvider::isTargetCase)
+                .collect(Collectors.toList());
+        Preconditions.checkState(!matched.isEmpty(),
+                "HintManager column label case not found. Check e2e-dql-select-join.xml contains "
+                        + "`SELECT o.user_id, i.* FROM t_order o JOIN t_order_item i ...` "
+                        + "and e2e-env.properties is tbl + MySQL + jdbc.");
+        return matched.stream().map(Arguments::of);
+    }
+    
+    private static boolean isTargetCase(final AssertionTestParameter testParam) {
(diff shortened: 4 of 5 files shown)


ClaireLytt:
Add e2e test cases to reproduce the issue

After implementing the fix

terrymanu (changes requested):
### Summary

**Reason:** The implementation direction aligns with the confirmed column-label mapping root cause, but the committed regression test does not exercise the reproduced failing path.

### Issues

#### P1: Commit the actual #28841 reproduction path

- **Problem:** The added case at [`e2e-dql-select-join.xml:136`](https://github.com/apache/shardingsphere/blob/938b331dc40e7f21ca6612133cc81b8e9a2fa9ca/test/e2e/sql/src/test/resources/cases/dql/e2e-dql-select-join.xml#L136-L138) is a plain single-table query. It covers neither the original mixed `A.column, B.*` join from [#28841](https://github.com/apache/shardingsphere/issues/28841) nor the later HintManager trigger. The committed runner simply executes this SQL normally at [`GeneralDQLE2EIT.java:121`](https://github.com/apache/shardingsphere/blob/938b331dc40e7f21ca6612133cc81b8e9a2fa9ca/test/e2e/sql/src/test/java/org/apache/shardingsphere/test/e2e/sql/it/sql/dql/GeneralDQLE2EIT.java#L121-L146). Public issue evidence reports that the single-table pattern succeeds without HintManager and fails with it ([discussion](https://github.com/apache/shardingsphere/issues/28841#issuecomment-3557403150)). The PR’s [failing screenshot](https://github.com/apache/shardingsphere/pull/39131#issuecomment-4979306775) invokes `assertIssue28841WithHintManager` and `assertColumnLabelAccessibleAfterNext`, but neither method nor equivalent setup exists at the latest head.
- **Impact:** The committed test does not reproduce the demonstrated failure, so the new metadata-map merge behavior can regress while the added case remains green.
- **Required Change:** Please commit a deterministic regression test that fails at the merge-base and passes with this fix—either the original mixed-projection join and topology from #28841 or the actual HintManager path shown in the public reproduction. Preserve label-based `ResultSet` access so the assertion verifies the missing-column-label behavior.

### Review Details

- **Reviewed Scope:** `RELEASE-NOTES.md`, JDBC result-set mapping, and the SQL E2E case at latest head `938b331dc40e7f21ca6612133cc81b8e9a2fa9ca`; PR base `master` at `1a7d3e67b3bb628db3f43f6b1f82c1274f5b4983`; local merge-base `7568693e8ee02637a47d3cca8acbf449797932e5`. The local triple-dot file list matched all three GitHub `/pulls/39131/files` entries.
- **Not Reviewed Scope:** GitHub Actions and CI; no local database E2E or Maven execution was performed.
- **Verification:** Latest-head XML validation passed (`xmllint`, exit 0). Latest-head source inspection confirmed that no committed HintManager reproduction harness exists. PR metadata, files, commits, issue comments, PR comments, and reviews were fetched through authenticated REST. The GitHub app connector was unavailable, but the fallback supplied every required endpoint, so there is no evidence-access gap.
- **Release Note / User Docs:** The release-note entry is present. Additional user documentation is not required because this fix introduces no configuration, migration, or public API change.

ClaireLytt:
@terrymanu 
Thanks for the review. Please check the HintManager / column-label path in `GeneralDQLE2EIT`.

* I tried multiple times, but could not reproduce the failure with:

```sql
SELECT o.user_id, i.* FROM t_order o JOIN t_order_item i ON o.order_id = i.order_id WHERE o.order_id = ?
```

* The only pattern I can reproduce reliably is:

```sql
SELECT T.*, T.status status_new FROM t_order T WHERE T.order_id = 1000
```

with `HintManager.setDataSourceName(...)` and label-based `ResultSet.getObject(...)`. 
* This is also the SQL used in the latest public reproduction comment on [#28841](https://github.com/apache/shardingsphere/issues/28841), so the current test follows that path.

* For the next commit, would you prefer that I remove the HintManager-related code from `GeneralDQLE2EIT`, and keep only the core fix plus dedicated unit/regression tests?

ClaireLytt:
Hi @terrymanu sorry to bother you again — whenever you have a moment, I'd really appreciate your review to help resolve a few open questions in this PR. Thanks so much for your time and patience!

terrymanu:
### Summary

**Reason:** Yes, this should be redesigned. The intended outcome is valid, but the current implementation infers the result shape by comparing the projection-map size with the JDBC column count. That does not represent the actual trigger and conflates two scenarios with opposite semantics: missing user-visible columns and additional internal derived columns.

### Issues

#### P1: Redefine the result-column mapping state and ownership boundary

- **Problem:** When a data-source hint is present, [`SQLBindEngine`](https://github.com/apache/shardingsphere/blob/d912eb7944f20d3dafd3275d60347b5241b5a5d6/infra/binder/core/src/main/java/org/apache/shardingsphere/infra/binder/engine/SQLBindEngine.java#L56-L83) skips SQL binding. Therefore, `T.*` has no actual projections to expand. [`ProjectionEngine`](https://github.com/apache/shardingsphere/blob/d912eb7944f20d3dafd3275d60347b5241b5a5d6/infra/binder/core/src/main/java/org/apache/shardingsphere/infra/binder/context/segment/select/projection/engine/ProjectionEngine.java#L101-L105) and [`ProjectionsContext`](https://github.com/apache/shardingsphere/blob/d912eb7944f20d3dafd3275d60347b5241b5a5d6/infra/binder/core/src/main/java/org/apache/shardingsphere/infra/binder/context/segment/select/projection/ProjectionsContext.java#L84-L93) consequently retain only the explicit `status_new` projection. This is the direct cause of the current HintManager reproduction.
- **Problem:** The PR identifies this state through a column-count mismatch in [`ShardingSphereResultSetUtils.java:53-65`](https://github.com/apache/shardingsphere/blob/d912eb7944f20d3dafd3275d60347b5241b5a5d6/jdbc/src/main/java/org/apache/shardingsphere/driver/jdbc/core/resultset/ShardingSphereResultSetUtils.java#L53-L65). However, normal multi-route rewriting also adds internal columns such as `ORDER_BY_DERIVED_0`, as demonstrated by the [existing rewrite case](https://github.com/apache/shardingsphere/blob/d912eb7944f20d3dafd3275d60347b5241b5a5d6/test/it/rewriter/src/test/resources/scenario/sharding/case/dml/select.xml#L199-L202). Both scenarios produce `expandedMap.size() != columnCount`, but the former should trust physical metadata while the latter must hide internal derived columns.
- **Impact:** The current logic adds physical internal labels to the mapping used by `ResultSet#getObject(String)`, while [`ShardingSphereResultSetMetaData`](https://github.com/apache/shardingsphere/blob/d912eb7944f20d3dafd3275d60347b5241b5a5d6/jdbc/src/main/java/org/apache/shardingsphere/driver/jdbc/core/resultset/ShardingSphereResultSetMetaData.java#L48-L104) still reports column counts and labels from logical projections. The two JDBC-facing contracts therefore no longer share one authoritative result shape. The equal-size-but-different-label case also continues to return an incomplete mapping, as noted in the [latest review comment](https://github.com/apache/shardingsphere/pull/39131#issuecomment-5099857414).
- **Discussion Needed:** Please pause the current implementation direction and first define one consistent result-shape contract: how to explicitly represent “bound with complete projections” versus “unbound with an unexpanded shorthand projection”; when physical metadata is authoritative; when internal merge columns must be filtered; and how JDBC `ResultSet`, `ResultSetMetaData`, and existing Proxy header consumers should share the same decision. The implementation and regression tests should be reconsidered after the owner and compatibility boundary are agreed upon.

### Multi-Round Comparison

- The previous request to commit the actual HintManager path is **partially fixed**: the latest head now calls `HintManager.setDataSourceName("tbl")`.
- However, the test at [`GeneralDQLE2EIT.java:307-314`](https://github.com/apache/shardingsphere/blob/d912eb7944f20d3dafd3275d60347b5241b5a5d6/test/e2e/sql/src/test/java/org/apache/shardingsphere/test/e2e/sql/it/sql/dql/GeneralDQLE2EIT.java#L307-L314) deliberately bypasses `ResultSetMetaData` and only verifies that label-based access does not throw. It cannot protect the unified result-shape contract or verify that labels resolve to the correct indexes. The test structure should be rebuilt after the design discussion instead of extending the current provider.

### Review Details

- **Reviewed Scope:** All five changed files at latest head `d912eb7944f20d3dafd3275d60347b5241b5a5d6`; base `master` at `80e6541874d47b5814e3d1400c925d6ff3afa0ed`; GitHub compare merge-base `7568693e8ee02637a47d3cca8acbf449797932e5`. The GitHub `/pulls/39131/files` list exactly matched the compare file list. The binder, projection context, JDBC metadata, sharding rewrite, and Proxy header consumer paths were also traced.
- **Not Reviewed Scope:** GitHub Actions, CI, and database runtime E2E.
- **Verification:** PR metadata, files, commits, comments, reviews, the linked issue, and compare data were retrieved successfully through REST, exit 0. Scoped `rg` and `git show` checks of the relevant call paths and history also succeeded, exit 0. Maven and E2E were not run because the latest-head static state-propagation path is sufficient to establish this design-level result.
- **Release Note / User Docs:** The current five-file scope no longer includes `RELEASE-NOTES.md`. This user-visible JDBC bug fix still requires its release-note entry to be restored after the design is agreed upon. No additional user documentation is required.

ClaireLytt:
@terrymanu 
Thank you for the detailed review. I now understand the core issues:

State ambiguity: The current fix uses column-count/label mismatch as a proxy to detect the "unbound shorthand" state, but this conflates two opposite scenarios — HintManager skipping bind (missing user-visible columns) vs. normal rewrite adding internal derived columns (e.g. ORDER_BY_DERIVED_0). The former should trust physical metadata, while the latter must hide internal columns.

Contract split: The PR only changes columnLabelAndIndexMap (used by ResultSet#getObject(String)) but does not update ShardingSphereResultSetMetaData, which still reports column count and labels from expandProjections. After the merge, the two JDBC-facing APIs describe different result shapes.

Root cause: When HintManager.setDataSourceName(...) is present, SQLBindEngine.isNeedBind() returns false, so ShorthandProjectionSegmentBinder never expands T.*. The ProjectionsContext then only contains the explicit projection (status_new), while the physical result set has all columns. This is the actual state that needs to be explicitly represented — not inferred from mismatch.

I'll pause the current implementation direction. Before re-implementing, I'd like to align on the design:

**Proposed approach:**

- **Step 1**: Add an explicit flag (e.g. `boundWithCompleteProjections`) on `SelectStatementContext` or `SelectStatementBaseContext`, set by `SQLBindEngine` — `true` when bind was executed, `false` when skipped due to data-source hint. **Purpose**: make the bind state explicitly queryable downstream, instead of inferring it from column-count or label mismatch.

- **Step 2**: Introduce a unified "visible result columns" resolution point that checks this flag:
  - `true` → use `expandProjections` (filters out `DerivedProjection` as today)
  - `false` → use physical `ResultSetMetaData`, filtering out any `DerivedColumn.isDerivedColumnName(...)` labels
  
  **Purpose**: produce one authoritative "which columns the user should see" answer, regardless of which path was taken, so that HintManager's unbound shorthand and normal rewrite's internal derived columns are handled with opposite — and correct — logic.

- **Step 3**: Have `ResultSet#getObject(String)`, `ResultSetMetaData`, and Proxy `QueryHeaderBuilderEngine` all consume from this single resolution point. **Purpose**: ensure all JDBC-facing contracts share the same result shape, eliminating the current split where `getObject` sees a merged map while `getMetaData` still reports from `expandProjections`.

Hopefully, the ideas in my proposed approach address the four questions raised in your latest review comments. Does this direction make sense, or do you have a different ownership boundary in mind? Happy to discuss before coding.

terrymanu:
The finding is retained, and the proposal needs one more design revision before implementation.

Your understanding of the root cause and the split result-shape contract is correct. I also agree that `ResultSet#getObject(String)`, `ResultSetMetaData`, and Proxy headers should consume one authoritative result-shape decision.

However, `boundWithCompleteProjections` should not be derived directly from whether [`SQLBindEngine.isNeedBind()`](https://github.com/apache/shardingsphere/blob/d912eb7944f20d3dafd3275d60347b5241b5a5d6/infra/binder/core/src/main/java/org/apache/shardingsphere/infra/binder/engine/SQLBindEngine.java#L56-L84) returned true. A hinted query with only explicit projections skips binding but still has a complete user-visible projection list. `SelectStatementContext` is also constructed outside `SQLBindEngine`, so a flag owned only by that engine would not reliably describe the context state.

Likewise, filtering physical metadata with [`DerivedColumn.isDerivedColumnName(...)`](https://github.com/apache/shardingsphere/blob/d912eb7944f20d3dafd3275d60347b5241b5a5d6/infra/binder/core/src/main/java/org/apache/shardingsphere/infra/binder/context/segment/select/projection/DerivedColumn.java#L55-L67) is still name-based inference. It matches reserved prefixes and cannot by itself prove that a physical column was internally generated. Any resolver that hides columns must also preserve the mapping from each visible column index to its physical result-set index.

Please revise the proposal around the semantic state that actually matters: whether a shorthand projection remains unexpanded. `ProjectionsContext` is the closest existing owner because it already contains the shorthand projection’s actual columns and the expanded projections. Then define a runtime result-shape mapping that combines this semantic state with physical metadata and explicit knowledge of internally generated projections.

Before implementation, please cover at least these cases in the design and tests:

1. Data-source hint with an unexpanded shorthand projection.
2. Data-source hint with only explicit, already-complete projections.
3. Normal rewriting with internal derived columns.
4. Equal column counts but different labels.
5. Consistent labels, counts, and indexes across `ResultSet`, `ResultSetMetaData`, and Proxy headers.

Please update the design proposal first; no code change is needed until this ownership and mapping contract is agreed.

ClaireLytt:
@terrymanu Thank you for the clarification. I will revise the design before implementation.

## Root Cause

The root cause is not simply a column-count or label mismatch. The real state that matters is whether a shorthand projection remains unexpanded.

For the HintManager case:

```sql
SELECT T.*, T.status status_new FROM t_order T WHERE T.order_id = 1000
```

`T.*` may remain unexpanded, so `ProjectionsContext` only contains the explicit projection (`status_new`), while the physical result set contains all user-visible columns.

The current fix also creates a split contract: `ResultSet#getObject(String)` uses the merged `columnLabelAndIndexMap`, while `ResultSetMetaData` still reports from `expandProjections`.

## Proposed Changes

### 1. Problem: bind state does not equal projection completeness

A data-source hint may skip SQL binding, but that does not always mean the projection list is incomplete.

For example, this hinted query has only explicit projections:

```sql
SELECT T.order_id, T.status status_new FROM t_order T WHERE T.order_id = 1000
```

Even if binding is skipped, the user-visible projection list is already complete:

```text
order_id, status_new
```

The real problematic case is a hinted query with shorthand projection:

```sql
SELECT T.*, T.status status_new FROM t_order T WHERE T.order_id = 1000
```

Here `T.*` needs binding to expand into actual columns. If it remains unexpanded, `ProjectionsContext` may only contain the explicit projection `status_new`, while the physical result set contains `order_id`, `user_id`, `status`, and `status_new`.

**Solution**: Do not derive the state from `SQLBindEngine.isNeedBind()`. Move the state ownership to `ProjectionsContext`, and let it explicitly expose whether it contains an unexpanded shorthand projection.

### 2. Problem: internal columns cannot be identified only by name prefix

Normal SQL rewrite may add internal columns for merge, for example:

```sql
SELECT t_account.amount FROM t_account ORDER BY t_account.account_id ASC
```

may be rewritten to:

```sql
SELECT t_account_0.amount, t_account_0.account_id AS ORDER_BY_DERIVED_0
FROM t_account_0
ORDER BY t_account_0.account_id ASC
```

`ORDER_BY_DERIVED_0` is an internal column and must not be exposed to users.

However, filtering by name alone is unsafe:

```java
DerivedColumn.isDerivedColumnName("ORDER_BY_DERIVED_0")
```

because a real physical table could also have a user-defined column with the same name.

**Solution**: I will not hide physical metadata columns only because their labels match reserved derived-column prefixes such as `ORDER_BY_DERIVED_`.

Instead, I will make the resolver use explicit internal-projection information from the current statement. For example, when ShardingSphere generates an internal projection such as `ORDER_BY_DERIVED_0`, that projection should be represented as an internally generated projection in the statement/projection context. The resolver will build an internal-column set from those generated projections first, and then scan the physical `ResultSetMetaData`.

A physical metadata column will be hidden only when it matches an internally generated projection that ShardingSphere created for this statement. If a real physical table happens to have a user-defined column named `ORDER_BY_DERIVED_0`, but there is no corresponding internally generated projection for the current statement, I will keep it visible.

This means the filtering rule becomes:

```text
hide the column only if:
  it is present in physical metadata
  and it corresponds to an explicitly generated internal projection of this statement
```

### 3. Problem: hiding physical columns changes index mapping

If the physical result set is:

```text
physical metadata:
1 amount
2 ORDER_BY_DERIVED_0
3 status
```

and `ORDER_BY_DERIVED_0` is an internal column, the user-visible result should be:

```text
visible columns:
1 amount -> physical index 1
2 status -> physical index 3
```

So the visible column index is not always the same as the physical result-set index.

**Solution**: Introduce a shared runtime visible result-shape mapping. Each visible column item should contain:

```text
visibleColumnIndex
physicalColumnIndex
columnLabel
columnName
```

`ResultSet#getObject(String)`, `ResultSetMetaData`, and Proxy query headers should all consume this same mapping, so labels, counts, and indexes are resolved from one authoritative result shape.

## Test Cases

I will cover these cases before implementation:

1. Data-source hint with unexpanded shorthand projection:

```sql
SELECT T.*, T.status status_new FROM t_order T WHERE T.order_id = 1000
```

2. Data-source hint with only explicit projections:

```sql
SELECT T.order_id, T.status status_new FROM t_order T WHERE T.order_id = 1000
```

3. Normal rewrite with internal derived columns:

```sql
SELECT t_account.amount FROM t_account ORDER BY t_account.account_id ASC
```

where the rewritten SQL may include `ORDER_BY_DERIVED_0`, which must not be exposed.

```sql
SELECT t_account.ORDER_BY_DERIVED_0 FROM t_account
```
A real user column named ORDER_BY_DERIVED_0 must remain visible.

```sql
SELECT t_account.amount AS ORDER_BY_DERIVED_0 FROM t_account
ORDER BY t_account.account_id ASC
```
A user-defined alias named ORDER_BY_DERIVED_0 must remain visible, while the internally generated merge column must be hidden.

4. Equal column counts but different labels.

```sql
SELECT COUNT(DISTINCT user_id) FROM t_order
```
ProjectionEngine assigns the internal alias AGGREGATION_DISTINCT_DERIVED_0 when the user does not provide one, so the physical result set reports one column labeled AGGREGATION_DISTINCT_DERIVED_0. Meanwhile ProjectionsContext.createColumnLabelAndIndexMap keys the same column by its expression COUNT(DISTINCT user_id), because the label matches a reserved derived-column prefix.

Both sides have exactly one column, but the labels differ. A resolver must not treat this as an unexpanded-shorthand case, and must not expose AGGREGATION_DISTINCT_DERIVED_0 to users.

5. Consistent labels, counts, and indexes across `ResultSet#getObject(String)`, `ResultSetMetaData`, and Proxy query headers.
For each case above, I will verify that `ResultSet#getObject(String)`, `ResultSetMetaData`, and Proxy query headers consume the same visible result-shape mapping.
Specifically:
- `ResultSet#getObject(String)` should resolve only user-visible labels, and each label should map to the correct physical result-set index.
- `ResultSetMetaData#getColumnCount()` should report the visible column count from the same mapping.
- `ResultSetMetaData#getColumnLabel(int)` should report visible labels from the same mapping and in the same order.
- Proxy query headers should expose the same visible labels and order as `ResultSetMetaData`.
- Internal generated columns should not be visible through any of the three consumers.

Does this revised direction match your expected ownership and mapping boundary?

terrymanu:
The revised design direction matches the expected ownership and mapping boundary, and the design discussion is resolved at the proposal level. To be explicit about where this leaves the PR: the previous review result still stands. Nothing in the current head is approved by this reply — it only unblocks the re-implementation, and the PR remains not mergeable until the new implementation lands with the agreed tests and a green CI run.

**Current head state**

- The single commit of 2026-07-15 now conflicts with master (`mergeable_state: dirty`).
- 24 of the 88 checks on that head are red, all of them in the `E2E - SQL (Smoke)` matrix across every adapter, mode, database, and scenario combination.
- Whether those failures come from this PR's own delta or from the stale base is not worth establishing, because the current implementation is superseded by the redesign below. Please do not spend time fixing the current head or its CI; rebase onto the latest master and build the new implementation there, so the conflicts and the red smoke matrix are discarded together.

**Design pillars confirmed against the current code**

1. `ProjectionsContext` as the owner of the unexpanded-shorthand state is correct and requires no new data model. `ProjectionEngine` builds `ShorthandProjection` from `ShorthandProjectionSegment.getActualProjectionSegments()`, which only `ShorthandProjectionSegmentBinder` populates during binding (infra/binder/core/src/main/java/org/apache/shardingsphere/infra/binder/engine/segment/dml/projection/type/ShorthandProjectionSegmentBinder.java#L83-L89). When `SQLBindEngine.isNeedBind()` returns false, that collection stays empty, and `createExpandProjections()` silently drops the shorthand today (infra/binder/core/src/main/java/org/apache/shardingsphere/infra/binder/context/segment/select/projection/ProjectionsContext.java#L84-L94). A query method such as `containsUnexpandedShorthandProjection()` can therefore expose the state without size or label inference.

2. For internal columns, consume the existing explicit representation instead of introducing a parallel one. Internally generated projections are already tracked: `ORDER_BY_DERIVED_n` / `GROUP_BY_DERIVED_n` are created as `DerivedProjection` by `ProjectionsContextEngine.getDerivedOrderColumns()` (infra/binder/core/src/main/java/org/apache/shardingsphere/infra/binder/context/segment/select/projection/engine/ProjectionsContextEngine.java#L86-L96), AVG-derived `COUNT`/`SUM` projections live on `AggregationProjection.getDerivedAggregationProjections()`, and the rewriter injects exactly those objects (features/sharding/core/src/main/java/org/apache/shardingsphere/sharding/rewrite/token/generator/impl/ShardingProjectionsTokenGenerator.java). The current branch on `containsDerivedProjections()` in `ShardingSphereResultSetUtils.createColumnLabelAndIndexMap()` (jdbc/src/main/java/org/apache/shardingsphere/driver/jdbc/core/resultset/ShardingSphereResultSetUtils.java#L44-L53) is the maintained precedent for consuming that explicit knowledge.

3. The shared visible result-shape mapping carrying `visibleColumnIndex`, `physicalColumnIndex`, `columnLabel`, and `columnName`, consumed by `ResultSet#getObject(String)`, `ResultSetMetaData`, and the Proxy `QueryHeaderBuilderEngine` (proxy/backend/core/src/main/java/org/apache/shardingsphere/proxy/backend/response/header/query/QueryHeaderBuilderEngine.java#L98-L104 still resolves from `expandProjections`), closes the split result-shape contract.

**Two clarifications to fold into the design**

- Treat relabeling and hiding as two explicit resolver outcomes. `AGGREGATION_DISTINCT_DERIVED_n` (infra/binder/core/src/main/java/org/apache/shardingsphere/infra/binder/context/segment/select/projection/engine/ProjectionEngine.java#L123-L124) is not a hidden column: for `SELECT COUNT(DISTINCT user_id) FROM t_order` the column is user-visible and only its physical label is internal. The mapping must express "physical label → user-visible label (the original expression)", which is what `ProjectionsContext.createColumnLabelAndIndexMap()` already keys by expression today (ProjectionsContext.java#L182).

- Add a sixth test case where the two states coexist: a data-source hint with an unexpanded shorthand plus an order-by item absent from the projections, for example `SELECT T.*, T.status status_new FROM t_order T ORDER BY T.user_id`. The skipped bind does not prevent `ProjectionsContextEngine` from appending derived order-by columns, because an unexpanded shorthand cannot satisfy `containsProjection()`, so the context can hold an unexpanded shorthand and internally generated projections at the same time. The resolver must then trust physical metadata for the shorthand columns and still hide the internally generated column in one pass.

**Minimum next actions**

1. Rebase onto the latest master; do not fix the current head.
2. Implement per the three pillars and two clarifications above.
3. Rebuild the test structure rather than extending the current `GeneralDQLE2EIT` provider, and cover the six scenarios, including consistent labels, counts, and indexes across `ResultSet#getObject(String)`, `ResultSetMetaData`, and Proxy query headers.
4. Restore the `RELEASE-NOTES.md` entry.
5. A green CI run on the new head is required before the next review round.

PR #39156, opened 2026-07-16
Support encrypt rewrite for SQL Server OPENQUERY UPDATE with narrow SELECT ... FROM shape

ClaireLytt:
for https://github.com/apache/shardingsphere/issues/30227
official document: https://learn.microsoft.com/en-us/sql/t-sql/queries/update-transact-sql?view=sql-server-ver17#p-updating-data-in-a-remote-table-by-using-the-openquery-function

  -

---


ClaireLytt:
**Problem**  
`UPDATE OPENQUERY (...)` targets a `FunctionTableSegment`, not a `SimpleTableSegment`. Encrypt rewrite only checked `TablesContext.getSimpleTables()`, so OPENQUERY updates were skipped and Logic SQL equaled Actual SQL (logical columns/values were not rewritten).

**Approach**  
1. **Decorator gate** — For OPENQUERY updates, detect encrypt tables by parsing the inner `FROM schema.table` in the OPENQUERY SQL string.  
2. **Generator gate** — Let `EncryptUpdateAssignmentTokenGenerator` run when the update target is OPENQUERY.  
3. **Rewrite** — In `EncryptAssignmentTokenGenerator`, rewrite both:
   - `SET` clause: logic column → cipher column + encrypted value  
   - OPENQUERY string: logic column → cipher column (`EncryptOpenQuerySQLToken`)

**Shared logic** — `EncryptOpenQueryUtils` centralizes OPENQUERY detection, table matching, and schema extraction

terrymanu (changes requested):
### Summary

**Reason:** The PR identifies the correct root cause—`OPENQUERY` is represented as a `FunctionTableSegment`—but the latest head still has three functional blockers and one repository-standard violation.

### Issues

#### P1 — Parameterized assignments are not encrypted

**Problem:** The new token path supports `SET GroupName = ?`, but the parameter rewriter still resolves the table exclusively from `ColumnSegmentBoundInfo#getOriginalTable()` and skips the assignment when that value is empty. Function-table columns use the temporary-table binding path, which the new test also models with an empty original table. See [EncryptAssignmentTokenGeneratorTest.java#L146-L156](https://github.com/apache/shardingsphere/blob/d04c4f8d45adf970516b1f4dc3c3fdc9e2a5e39f/features/encrypt/core/src/test/java/org/apache/shardingsphere/encrypt/rewrite/token/generator/assignment/EncryptAssignmentTokenGeneratorTest.java#L146-L156) and [EncryptAssignmentParameterRewriter.java#L68-L84](https://github.com/apache/shardingsphere/blob/d04c4f8d45adf970516b1f4dc3c3fdc9e2a5e39f/features/encrypt/core/src/main/java/org/apache/shardingsphere/encrypt/rewrite/parameter/rewriter/EncryptAssignmentParameterRewriter.java#L68-L84).

**Impact:** A prepared `UPDATE OPENQUERY(...) SET GroupName = ?` rewrites the target column to the cipher column but sends the original plaintext parameter.

**Required Change:** Please make assignment-parameter rewriting resolve the OPENQUERY encrypt table and schema through the same target context as literal assignments. Add a rewrite test using an outer parameter marker, including derived cipher/assisted/like parameters where configured.

#### P1 — Multiple assignments produce overlapping OPENQUERY tokens

**Problem:** The generator creates one full-range `EncryptOpenQuerySQLToken` for every encrypted assignment ([EncryptAssignmentTokenGenerator.java#L90-L135](https://github.com/apache/shardingsphere/blob/d04c4f8d45adf970516b1f4dc3c3fdc9e2a5e39f/features/encrypt/core/src/main/java/org/apache/shardingsphere/encrypt/rewrite/token/generator/assignment/EncryptAssignmentTokenGenerator.java#L90-L135)). Tokens for the same string literal have identical ranges, and the rewrite engine retains only the first overlapping token ([AbstractSQLBuilder.java#L45-L55](https://github.com/apache/shardingsphere/blob/d04c4f8d45adf970516b1f4dc3c3fdc9e2a5e39f/infra/rewrite/core/src/main/java/org/apache/shardingsphere/infra/rewrite/sql/impl/AbstractSQLBuilder.java#L45-L55)).

**Impact:** For two encrypted columns, only one inner projection is rewritten while both outer assignments use physical column names. Configurations with assisted or like-query columns have the same mismatch because those physical columns are added to the outer assignment but not the OPENQUERY rowset. SQL Server permits multiple SET assignments and requires updated columns to exist in the target rowset ([Microsoft UPDATE documentation](https://learn.microsoft.com/en-us/sql/t-sql/queries/update-transact-sql?view=sql-server-2017)).

**Required Change:** Please generate one composed inner-query token per OPENQUERY target from the complete physical-column mapping. Cover two encrypted assignments and a column with assisted/like-query derived columns.

#### P1 — Regex rewriting can change pass-through query semantics

**Problem:** Table discovery uses `FROM\s+([^\s,;]+)`, and column rewriting replaces every matching word in the raw pass-through string ([EncryptOpenQueryUtils.java#L41-L116](https://github.com/apache/shardingsphere/blob/d04c4f8d45adf970516b1f4dc3c3fdc9e2a5e39f/features/encrypt/core/src/main/java/org/apache/shardingsphere/encrypt/rewrite/token/generator/assignment/EncryptOpenQueryUtils.java#L41-L116), [EncryptAssignmentTokenGenerator.java#L131-L135](https://github.com/apache/shardingsphere/blob/d04c4f8d45adf970516b1f4dc3c3fdc9e2a5e39f/features/encrypt/core/src/main/java/org/apache/shardingsphere/encrypt/rewrite/token/generator/assignment/EncryptAssignmentTokenGenerator.java#L131-L135)). OPENQUERY’s second argument is an opaque query executed by the linked server, and Microsoft’s examples include quoted predicates ([Microsoft OPENQUERY documentation](https://learn.microsoft.com/en-us/sql/t-sql/functions/openquery-transact-sql?view=sql-server-2017)).

For example, rewriting `GroupName` also changes the literal in `WHERE Note = ''GroupName''`. The table matcher also fails on a valid delimited identifier such as `[Human Resources].[Department]`.

**Impact:** Valid SQL can select a different remote rowset or silently bypass encryption-table detection.

**Required Change:** Please replace raw regex rewriting with a structure-aware, quote-aware mechanism appropriate to the supported linked-server dialect. If support is intentionally limited to a narrower query shape, enforce that boundary and document it. Add cases for string literals, delimited identifiers, multipart names, and predicates on encrypted columns.

#### P2 — New methods use `Optional` parameters

**Problem:** The new internal flow passes `Optional<TableSegment>` and `Optional<OpenQueryContext>` as method parameters ([EncryptAssignmentTokenGenerator.java#L90-L128](https://github.com/apache/shardingsphere/blob/d04c4f8d45adf970516b1f4dc3c3fdc9e2a5e39f/features/encrypt/core/src/main/java/org/apache/shardingsphere/encrypt/rewrite/token/generator/assignment/EncryptAssignmentTokenGenerator.java#L90-L128)). The repository coding standard explicitly forbids `Optional` method parameters and requires minimal access control ([CODE_OF_CONDUCT.md#L61-L65](https://github.com/apache/shardingsphere/blob/d04c4f8d45adf970516b1f4dc3c3fdc9e2a5e39f/CODE_OF_CONDUCT.md#L61-L65)).

**Impact:** The patch does not meet the project’s production-code quality gate and makes the normal/OpenQuery paths less explicit.

**Required Change:** Please keep `Optional` as a return/local representation and express the two execution paths without optional parameters. Also use the minimum required visibility for the new overload.

### Review Details

- **Reviewed Scope:** All 13 files in `features/encrypt/core`, `test/it/rewriter`, and `RELEASE-NOTES.md` at head `d04c4f8d45adf970516b1f4dc3c3fdc9e2a5e39f`; base `master` SHA `6ecd9c15b0ea089c6036fa4d1776098a19b03cb1`; merge-base `cb50781bc2a08ba20d7b9a66ec5e017fc2bf17c5`. The local triple-dot file list matched GitHub `/pulls/39156/files`.
- **Not Reviewed Scope:** GitHub Actions/check runs and live SQL Server/OLE DB execution. These omissions do not affect the static blockers above.
- **Verification:** Public PR metadata, files, issue comments, and reviews were refreshed successfully; the head and file list remained unchanged. Local source-path and token-overlap analysis completed successfully. Maven was not run because an isolated latest-head checkout was unavailable; the findings do not depend on build or CI results.
- **Release Note / User Docs:** The release-note entry is present. User documentation is missing for the implementation’s current query-shape restrictions; it is required if support remains intentionally limited.

ClaireLytt:
**P1 — Parameterized assignments are not encrypted**  
→ **Code:** `EncryptAssignmentParameterRewriter.java` 
→ **Unit tests:** `EncryptAssignmentTokenGeneratorTest`
→ **`update.xml`:**  add new tests for `SET GroupName = ?` and expands to cipher / assisted / like parameters  

**P1 — Multiple assignments produce overlapping OPENQUERY tokens**  
→ **Code:** `EncryptAssignmentTokenGenerator.java` 
→ **Unit tests:** `EncryptAssignmentTokenGeneratorTest`
→ **`update.xml`:**  
-  two encrypted columns rewritten together  
-  derived cipher/assisted/like columns 

**P1 — Regex rewriting can change pass-through query semantics**  
→ **Code:** `EncryptOpenQueryPassThroughSQL.java` — structure/quote-aware `parse` + SELECT-list-only `rewrite`; unsupported shapes throw `UnsupportedEncryptSQLException`. `EncryptOpenQueryUtils.java` / `EncryptAssignmentTokenGenerator.java` delegate to it (no raw whole-string regex replace).  
→ **Unit tests:** `EncryptOpenQueryPassThroughSQLTest` — multipart / delimited names; WHERE on encrypted columns preserved; string literal `Note = 'GroupName'` not rewritten; unsupported SELECT literal/expression, space-delimited id, three-part name, JOIN rejected; derived-column rewrite. `EncryptAssignmentTokenGeneratorTest#assertGenerateSQLTokenWithOpenQueryPreservesWhereClauseColumnRef` / `#assertGenerateSQLTokenWithUnsupportedOpenQuerySelectLiteralExpectsException`.  
→ **`update.xml`:**  
- `*_predicate_on_encrypted_column_for_literals` — WHERE column name left as-is  
- `*_string_literal_predicate_for_literals` — literal containing column name not rewritten  
- `*_delimited_multipart_table_name_for_literals` / `*_for_parameters` — `[HumanResources].[Department]` supported  

**P2 — New methods use Optional parameters**  
Fixed. Split into two overloads without optional/`null` parameters: `generateSQLTokens(tables, set)` for normal UPDATE and `generateSQLTokens(tables, set, openQueryTable)` for OPENQUERY. The caller keeps the table segment as a local and routes by path. The OPENQUERY overload is package-private.

terrymanu (changes requested):
### Summary

**Reason:** The latest head fixes the earlier parameter-rewrite and overlapping-token defects, but the new pass-through parser still permits encryption bypasses and incorrect physical SQL. The result is code-scope only; CI was not reviewed by request.

### Issues

#### P1 — Encrypted predicates remain on logical columns

**Problem:** `rewrite` transforms only the `SELECT` list and appends the entire remainder unchanged ([EncryptOpenQueryPassThroughSQL.java#L116-L123](https://github.com/apache/shardingsphere/blob/f115808b981a886bb39f593200871c1b30099a31/features/encrypt/core/src/main/java/org/apache/shardingsphere/encrypt/rewrite/token/generator/assignment/EncryptOpenQueryPassThroughSQL.java#L116-L123)). The new test explicitly expects `WHERE GroupName IS NOT NULL` to remain logical while the projection becomes `group_name_cipher` ([EncryptOpenQueryPassThroughSQLTest.java#L67-L75](https://github.com/apache/shardingsphere/blob/f115808b981a886bb39f593200871c1b30099a31/features/encrypt/core/src/test/java/org/apache/shardingsphere/encrypt/rewrite/token/generator/assignment/EncryptOpenQueryPassThroughSQLTest.java#L67-L75)).

The changed configuration defines `GroupName` with only `group_name_cipher`, not a plaintext query column ([query-with-cipher.yaml#L161-L180](https://github.com/apache/shardingsphere/blob/f115808b981a886bb39f593200871c1b30099a31/test/it/rewriter/src/test/resources/scenario/encrypt/config/query-with-cipher.yaml#L161-L180)). Normal encrypt rewriting replaces predicate columns with their cipher or assisted-query columns ([EncryptPredicateColumnTokenGenerator.java#L112-L122](https://github.com/apache/shardingsphere/blob/f115808b981a886bb39f593200871c1b30099a31/features/encrypt/core/src/main/java/org/apache/shardingsphere/encrypt/rewrite/token/generator/predicate/EncryptPredicateColumnTokenGenerator.java#L112-L122)). OPENQUERY executes its pass-through query on the linked server, so the outer rewrite cannot repair that unchanged predicate later ([Microsoft OPENQUERY documentation](https://learn.microsoft.com/en-us/sql/t-sql/functions/openquery-transact-sql?view=sql-server-2017)).

**Impact:** Predicates on encrypted columns can fail against the physical table, query an optional stale plaintext column, or select a different rowset from normal encrypt semantics before the update is applied.

**Required Change:** Please either rewrite predicates using the full `EncryptTable` query semantics—including encrypted literal values—or reject pass-through predicates that reference encrypted logical columns. Replace the current “preserves encrypted predicate” expectations with physical-predicate or explicit-unsupported assertions.

#### P1 — Identifier handling can silently bypass encryption or emit invalid physical SQL

**Problem:** Target discovery accepts only bracketed or alphanumeric/underscore identifier parts ([EncryptOpenQueryPassThroughSQL.java#L258-L282](https://github.com/apache/shardingsphere/blob/f115808b981a886bb39f593200871c1b30099a31/features/encrypt/core/src/main/java/org/apache/shardingsphere/encrypt/rewrite/token/generator/assignment/EncryptOpenQueryPassThroughSQL.java#L258-L282)). For a valid SQL Server target such as `"HumanResources"."Department"`, discovery returns empty, `findEncryptTable` returns empty ([EncryptOpenQueryUtils.java#L76-L82](https://github.com/apache/shardingsphere/blob/f115808b981a886bb39f593200871c1b30099a31/features/encrypt/core/src/main/java/org/apache/shardingsphere/encrypt/rewrite/token/generator/assignment/EncryptOpenQueryUtils.java#L76-L82)), and the decorator returns before validation or rewriting ([EncryptSQLRewriteContextDecorator.java#L54-L57](https://github.com/apache/shardingsphere/blob/f115808b981a886bb39f593200871c1b30099a31/features/encrypt/core/src/main/java/org/apache/shardingsphere/encrypt/rewrite/context/EncryptSQLRewriteContextDecorator.java#L54-L57)). SQL Server supports both bracketed and double-quoted delimited identifiers ([Microsoft identifier documentation](https://learn.microsoft.com/en-us/sql/relational-databases/databases/database-identifiers?view=sql-server-ver17)).

The emitter also writes cipher, assisted, and like-query names without identifier quoting ([EncryptOpenQueryPassThroughSQL.java#L359-L363](https://github.com/apache/shardingsphere/blob/f115808b981a886bb39f593200871c1b30099a31/features/encrypt/core/src/main/java/org/apache/shardingsphere/encrypt/rewrite/token/generator/assignment/EncryptOpenQueryPassThroughSQL.java#L359-L363)). Encrypt configuration validation requires these names to be non-empty but does not restrict them to regular identifiers ([EncryptRuleConfigurationChecker.java#L64-L77](https://github.com/apache/shardingsphere/blob/f115808b981a886bb39f593200871c1b30099a31/features/encrypt/core/src/main/java/org/apache/shardingsphere/encrypt/checker/config/EncryptRuleConfigurationChecker.java#L64-L77)).

**Impact:** A valid delimited target can be forwarded without encryption, while configured physical names containing spaces, reserved words, or special characters produce invalid pass-through SQL.

**Required Change:** Please use one authoritative identifier model for target discovery, validation, and emission. It should support or deliberately reject valid delimiter forms only after identifying the encrypt target, handle escaped delimiters, and quote physical column names for the supported remote dialect—or reject configurations that cannot be emitted safely. Add end-to-end cases for double-quoted targets, escaped bracket identifiers, and physical names requiring delimiters.

#### P1 — The narrow query-shape gate accepts unsupported multi-source and set queries

**Problem:** `validateRemainder` rejects only text containing `JOIN` ([EncryptOpenQueryPassThroughSQL.java#L158-L163](https://github.com/apache/shardingsphere/blob/f115808b981a886bb39f593200871c1b30099a31/features/encrypt/core/src/main/java/org/apache/shardingsphere/encrypt/rewrite/token/generator/assignment/EncryptOpenQueryPassThroughSQL.java#L158-L163)). Therefore, examples such as these pass validation:

- `SELECT GroupName FROM dbo.Department, dbo.Other`
- `SELECT GroupName FROM dbo.Department UNION SELECT GroupName FROM dbo.Other`
- `SELECT GroupName FROM dbo.Department CROSS APPLY ...`

The implementation rewrites only the first `SELECT` list and appends the remainder unchanged. SQL Server officially supports comma-separated table sources and APPLY in `FROM`, as well as UNION query expressions ([Microsoft FROM documentation](https://learn.microsoft.com/en-us/sql/t-sql/queries/from-transact-sql?view=sql-server-ver17), [Microsoft UNION documentation](https://learn.microsoft.com/en-us/sql/t-sql/language-elements/set-operators-union-transact-sql?view=sql-server-ver17)).

**Impact:** Shapes outside the claimed narrow boundary can receive a partial logical-to-physical rewrite instead of the promised explicit unsupported error, producing mixed rowset semantics or provider-dependent failures.

**Required Change:** Please validate the complete pass-through structure before changing parameters or generating tokens. Explicitly reject every table-source, set-operation, and trailing-clause form outside the supported grammar, and add counterexample tests for comma sources, APPLY, UNION, EXCEPT, and INTERSECT.

#### P2 — The user-facing support boundary is undocumented

**Problem:** The release note claims general OPENQUERY update support ([RELEASE-NOTES.md#L79](https://github.com/apache/shardingsphere/blob/f115808b981a886bb39f593200871c1b30099a31/RELEASE-NOTES.md#L79)), while the implementation rejects SELECT expressions, SELECT literals, space-containing identifiers, three-part table names, and JOINs ([EncryptOpenQueryPassThroughSQL.java#L36-L46](https://github.com/apache/shardingsphere/blob/f115808b981a886bb39f593200871c1b30099a31/features/encrypt/core/src/main/java/org/apache/shardingsphere/encrypt/rewrite/token/generator/assignment/EncryptOpenQueryPassThroughSQL.java#L36-L46)). Neither encrypt limitations document was updated.

**Impact:** Users cannot determine the accepted query shape, linked-server/provider boundary, or migration limitations without encountering runtime exceptions or incorrect rewrites.

**Required Change:** Please narrow the release-note wording and document the exact supported OPENQUERY shape and rejected predicates, identifiers, table sources, set operations, and remote-dialect assumptions in both `limitations.en.md` and `limitations.cn.md`.

### Multi-Round Comparison

- **Fixed:** Outer parameter markers now resolve the OPENQUERY encrypt table, including assisted/like derived parameters, with integration coverage.
- **Fixed:** Multiple assignments now produce one composed OPENQUERY token.
- **Partially fixed:** Raw whole-string regex replacement was removed and string literals are preserved, but encrypted predicates and unsupported identifier/query shapes remain incorrect.
- **Fixed:** The assignment generator no longer passes `Optional` parameters, and the OPENQUERY overload now has package-private visibility.
- **Newly introduced:** The custom pass-through parser’s detector, validator, and emitter do not share a complete fail-closed identifier/query model.

### Review Details

- **Reviewed Scope:** All 16 changed files in `features/encrypt/core`, `test/it/rewriter`, and `RELEASE-NOTES.md` at head `f115808b981a886bb39f593200871c1b30099a31`; base `master` SHA `6ecd9c15b0ea089c6036fa4d1776098a19b03cb1`; local merge-base `cb50781bc2a08ba20d7b9a66ec5e017fc2bf17c5`. The final local triple-dot file list matched GitHub `/pulls/39156/files`.
- **Not Reviewed Scope:** GitHub Actions/check runs and live SQL Server/OLE DB provider execution. These omissions do not affect the source-path and contract blockers above.
- **Verification:** GitHub PR metadata, files, commits, comments, and reviews were refreshed after the head changed and again before publication. The exact-head review inventory completed successfully with all 16 files matched. Maven was not run because the active checkout was not the PR head and differed in every changed file; the findings are proven directly by the latest-head source paths and official SQL Server contracts.
- **Release Note / User Docs:** Release-note entry present but broader than the implementation. Required English and Chinese limitations documentation is missing.

ClaireLytt:
This PR adds encryption rewrite support for SQL Server `UPDATE OPENQUERY(...)` statements.

It detects encrypted tables inside `FunctionTableSegment`, rewrites logical columns and assignment values to physical encrypted columns, and supports literal and parameterized assignments, multiple columns, assisted-query columns, and like-query columns.

It also introduces structure-aware parsing for the inner `SELECT ... FROM` query, safely handles quoted identifiers, preserves string literals, and explicitly rejects unsupported query shapes. Comprehensive unit and SQL rewrite integration tests cover the new behavior and edge cases.

I may open a follow-up issue to address the OPENQUERY scenarios that aren't supported in this PR.

terrymanu (changes requested):
## Summary

- **Review Result:** Not Mergeable
- Feedback Mode: Change Request
- Reason: The current revision follows the correct root-cause direction, but four P1 correctness gaps can still reject valid `OPENQUERY` statements, bypass encryption rewrite, or emit malformed/unsupported pass-through SQL.

## Issues

### P1 — OPENQUERY text is parsed and emitted at the wrong escaping layer

Problem:

SQL Server supplies the pass-through query as a T-SQL string literal. [`StringLiteralValue`](https://github.com/apache/shardingsphere/blob/6e5ec36170022fbc1b53010fde73ecf439420141/parser/sql/statement/core/src/main/java/org/apache/shardingsphere/sql/parser/statement/core/value/literal/impl/StringLiteralValue.java#L33-L35) removes the outer quotes but retains doubled inner apostrophes, and the [SQL Server visitor](https://github.com/apache/shardingsphere/blob/6e5ec36170022fbc1b53010fde73ecf439420141/parser/sql/engine/dialect/sqlserver/src/main/java/org/apache/shardingsphere/sql/parser/engine/sqlserver/visitor/statement/SQLServerStatementVisitor.java#L610-L612) forwards that value unchanged.

The new scanners operate as if they received decoded inner SQL. Consequently, production text such as `WHERE Note = ''GroupName''` exposes `GroupName` to the scanner as an identifier outside a string. Likewise, `''UNION ALL''` is misclassified as a set operation. The current [direct parser tests](https://github.com/apache/shardingsphere/blob/6e5ec36170022fbc1b53010fde73ecf439420141/features/encrypt/core/src/test/java/org/apache/shardingsphere/encrypt/rewrite/token/generator/assignment/EncryptOpenQueryPassThroughSQLTest.java#L138-L161) use already-decoded single quotes and therefore do not cover the production representation.

The reverse path has the same boundary problem: [`quotePhysicalColumnName`](https://github.com/apache/shardingsphere/blob/6e5ec36170022fbc1b53010fde73ecf439420141/features/encrypt/core/src/main/java/org/apache/shardingsphere/encrypt/rewrite/token/generator/assignment/EncryptOpenQueryPassThroughSQL.java#L606-L617) permits apostrophes, while [`EncryptOpenQuerySQLToken.toString()`](https://github.com/apache/shardingsphere/blob/6e5ec36170022fbc1b53010fde73ecf439420141/features/encrypt/core/src/main/java/org/apache/shardingsphere/encrypt/rewrite/token/pojo/EncryptOpenQuerySQLToken.java#L40-L42) wraps rewritten SQL without doubling inserted apostrophes. A physical name such as `foo'bar` therefore produces a malformed outer literal.

This escaping requirement is visible in Microsoft’s [OPENQUERY examples](https://learn.microsoft.com/en-us/sql/t-sql/functions/openquery-transact-sql?view=sql-server-ver17), which use doubled apostrophes inside the outer literal.

Impact:

Valid supported queries containing string predicates can be rejected, and valid physical identifiers containing apostrophes can produce syntactically broken rewritten SQL.

Required Change:

Model the boundary explicitly: decode one T-SQL string-literal layer before lexing the pass-through query and escape it exactly once when emitting the outer literal. Add production-path rewrite tests using doubled predicate literals and a physical column name containing an apostrophe.

### P1 — Target-table discovery can fail open on valid delimited identifiers

Problem:

[`findFromKeywordIndexIfPresent`](https://github.com/apache/shardingsphere/blob/6e5ec36170022fbc1b53010fde73ecf439420141/features/encrypt/core/src/main/java/org/apache/shardingsphere/encrypt/rewrite/token/generator/assignment/EncryptOpenQueryPassThroughSQL.java#L427-L451) tracks string literals but not bracketed or double-quoted identifiers. For valid SQL such as:

```sql
SELECT [FROM], GroupName FROM dbo.Department
```

it treats the first `[FROM]` as the clause delimiter. Current-head diagnostics confirm that [`findTableName`](https://github.com/apache/shardingsphere/blob/6e5ec36170022fbc1b53010fde73ecf439420141/features/encrypt/core/src/main/java/org/apache/shardingsphere/encrypt/rewrite/token/generator/assignment/EncryptOpenQueryPassThroughSQL.java#L83-L93) then returns empty. [`EncryptOpenQueryUtils.findEncryptTable`](https://github.com/apache/shardingsphere/blob/6e5ec36170022fbc1b53010fde73ecf439420141/features/encrypt/core/src/main/java/org/apache/shardingsphere/encrypt/rewrite/token/generator/assignment/EncryptOpenQueryUtils.java#L76-L82) cannot identify the encrypt table, so the [decorator exits without rewriting](https://github.com/apache/shardingsphere/blob/6e5ec36170022fbc1b53010fde73ecf439420141/features/encrypt/core/src/main/java/org/apache/shardingsphere/encrypt/rewrite/context/EncryptSQLRewriteContextDecorator.java#L54-L58).

Bracketed reserved words are valid delimited identifiers according to Microsoft’s [identifier rules](https://learn.microsoft.com/en-us/sql/relational-databases/databases/database-identifiers?view=sql-server-ver17).

Impact:

A valid encrypted target silently bypasses encryption rewrite instead of being rewritten or rejected, leaving logical columns and plaintext assignments in the SQL sent downstream.

Required Change:

Use one delimiter-aware lexical model for both target discovery and later validation, covering strings, brackets, double quotes, and escaped delimiters. Once an encrypted `OPENQUERY` target is identified, every unsupported shape must fail closed. Add a full decorator/rewrite regression test with `[FROM]` in the projection.

### P1 — The documented narrow single-SELECT grammar is not enforced

Problem:

The documentation limits support to `SELECT <columns> FROM ... [WHERE ...]` and explicitly excludes expressions and multiple statements. However, [`validateColumnIdentifier`](https://github.com/apache/shardingsphere/blob/6e5ec36170022fbc1b53010fde73ecf439420141/features/encrypt/core/src/main/java/org/apache/shardingsphere/encrypt/rewrite/token/generator/assignment/EncryptOpenQueryPassThroughSQL.java#L217-L240) accepts numeric and keyword expressions such as `1` or `NULL`, while [`validateRemainder`](https://github.com/apache/shardingsphere/blob/6e5ec36170022fbc1b53010fde73ecf439420141/features/encrypt/core/src/main/java/org/apache/shardingsphere/encrypt/rewrite/token/generator/assignment/EncryptOpenQueryPassThroughSQL.java#L242-L258) rejects only a subset of table-source and set-operation forms.

Current-head diagnostics confirm that all of these are accepted:

```sql
SELECT 1 FROM dbo.Department
SELECT GroupName FROM dbo.Department; DELETE FROM dbo.Other
SELECT GroupName FROM dbo.Department ORDER BY DepartmentID
```

This contradicts the [documented supported shape and exclusions](https://github.com/apache/shardingsphere/blob/6e5ec36170022fbc1b53010fde73ecf439420141/docs/document/content/features/encrypt/limitations.en.md#L11-L33).

Impact:

The rewrite can generate an update against a rowset that does not expose the required physical column, or forward additional clauses/statements after partially rewriting only the first projection.

Required Change:

Validate the complete supported single-statement grammar before rewriting: every projection item must be an allowed identifier, and only explicitly supported table hints and an optional `WHERE` remainder may follow the table reference. Reject statement terminators, additional clauses, numeric/keyword expressions, and all unrecognized trailing forms. Cover these counterexamples through the full rewrite boundary.

### P1 — Validation and pass-through rewrite are skipped when SET has no encrypt column

Problem:

[`generateSQLTokens`](https://github.com/apache/shardingsphere/blob/6e5ec36170022fbc1b53010fde73ecf439420141/features/encrypt/core/src/main/java/org/apache/shardingsphere/encrypt/rewrite/token/generator/assignment/EncryptAssignmentTokenGenerator.java#L105-L124) invokes the composed pass-through rewrite only when `hasEncryptAssignment` becomes true. Because validation of the inner query is performed inside that rewrite, a statement such as:

```sql
UPDATE OPENQUERY(server,
  'SELECT GroupName, DepartmentID FROM dbo.Department WHERE GroupName IS NOT NULL')
SET DepartmentID = 5
```

is neither rewritten nor rejected, even though the documented contract explicitly rejects encrypted predicates.

Impact:

Logical encrypted columns and predicates can be sent unchanged to the remote table, causing invalid SQL or incorrect row selection.

Required Change:

Validate and rewrite the pass-through target whenever an encrypted target table is identified, independently of whether the outer assignment itself targets an encrypted column. If non-encrypted assignments are outside the supported contract, reject them explicitly instead of silently skipping the rewrite. Add integration coverage for this branch.

## Multi-Round Comparison

- Fixed from the first review round: parameterized assignment rewriting, multiple-assignment token composition, derived-query parameter handling, and `Optional` method parameters.
- Fixed or improved from the second round: explicit documentation and release notes were added; comma sources, joins, apply operators, set operations, and encrypted predicates are rejected when the parser is reached.
- Partially fixed: replacing regex rewriting with structured scanning is the right direction, but escaping-layer handling, identifier-aware discovery, complete grammar validation, and fail-closed behavior remain incomplete.
- The issues above apply to the current head only; resolved earlier findings are not repeated as blockers.

## Review Details

- Review Focus: Code Correctness Review. CI not reviewed by request.
- Reviewed Scope: all 20 changed files at head `6e5ec36170022fbc1b53010fde73ecf439420141`, compared with base `6ecd9c15b0ea089c6036fa4d1776098a19b03cb1`; authenticated GitHub and local changed-file inventories matched.
- Root-Cause Assessment: the PR correctly recognizes that `OPENQUERY` is represented as a function table and therefore bypasses the existing simple-table encrypt rewrite path. The implementation reaches that boundary, but the current parser/emitter and conditional validation do not preserve it safely.
- Verification:
  - `./mvnw -pl features/encrypt/core -DskipITs -Dtest=EncryptOpenQueryPassThroughSQLTest,EncryptOpenQueryUtilsTest,EncryptAssignmentTokenGeneratorTest,EncryptAssignmentTokenGeneratorOpenQueryUnsupportedShapeTest,EncryptUpdateAssignmentTokenGeneratorTest,EncryptSQLRewriteContextDecoratorTest test -B -T1C` — exit 0, 68 tests.
  - `./mvnw -pl features/encrypt/core,test/it/rewriter -Dtest=EncryptSQLRewriterIT -Dsurefire.failIfNoSpecifiedTests=false test -B -T1C` — exit 0, 208 tests.
  - Latest-head diagnostic probes reproduced the escaping mismatch, `[FROM]` discovery bypass, incomplete grammar acceptance, and unescaped emitted apostrophe.
- Not Reviewed: GitHub Actions/check logs and live SQL Server/OLE DB provider execution. The verdict is code-scope only and does not depend on those items because each blocker is reproducible in the current Java paths before provider execution.
- Release Note / User Docs: Present and appropriately narrowed, but the implementation does not yet enforce the documented boundary.
- Evidence Gaps: None affecting the verdict.

ClaireLytt:
@terrymanu 
Thanks for the review. The four P1s are addressed as follows:

→ **Wrong escaping layer** — Code: `EncryptOpenQueryPassThroughSQL.decodeTSqlStringLiteralEscaping()` / `parse()` / `findTableName()` decode one T-SQL string-literal layer (`''` → `'`) before lexing; `EncryptOpenQuerySQLToken.toString()` escapes once on emit (`'` → `''`). Tests: doubled predicate literals (`''GroupName''` / `''UNION ALL''`) and physical name `foo'bar` (`[foo''bar]`) in unit tests and `update.xml`.

→ **Target-table discovery fail-open on delimited identifiers** — Code: `findFromKeywordIndexIfPresent()` now skips bracketed and double-quoted identifiers (including escaped delimiters) before treating `FROM` as the clause keyword. Tests: `SELECT [FROM], GroupName FROM ...` discovery/rewrite unit tests and decorator/`update.xml` regression.

→ **Documented narrow single-SELECT grammar not enforced** — Code: `validateColumnIdentifier()` rejects numeric/keyword projections (`1`, `NULL`); `validateRemainder()` rejects `;`, `ORDER BY`, `GROUP BY`, and `HAVING`. Docs: `limitations.en.md` / `limitations.cn.md` updated. Tests: counterexamples in `EncryptOpenQueryPassThroughSQLTest` and unsupported-shape decorator tests.

→ **Validation/rewrite skipped when SET has no encrypt column** — Code: `generateOpenQueryUpdateTokens()` always runs `appendComposedOpenQuerySQLToken()` once an encrypt table is identified, independent of encrypt assignments. Tests: In `EncryptAssignmentTokenGeneratorOpenQueryUnsupportedShapeTest.java`

terrymanu (changes requested):
### Summary

The overall direction is correct, and the latest head fixes several issues raised in earlier rounds. However, two correctness paths can still generate invalid SQL or silently bypass encryption, and the new public collection accessor exposes mutable rule state.

### Issues

#### P1 — Encrypted assignments need not exist in the OPENQUERY target rowset

**Problem:** [`generateOpenQueryUpdateTokens`](https://github.com/apache/shardingsphere/blob/f1855a143bc6a142ada2041365f0e7c0d80d1bb4/features/encrypt/core/src/main/java/org/apache/shardingsphere/encrypt/rewrite/token/generator/assignment/EncryptAssignmentTokenGenerator.java#L105-L120) rewrites every encrypted outer assignment, while [`rewrite`](https://github.com/apache/shardingsphere/blob/f1855a143bc6a142ada2041365f0e7c0d80d1bb4/features/encrypt/core/src/main/java/org/apache/shardingsphere/encrypt/rewrite/token/generator/assignment/EncryptOpenQueryPassThroughSQL.java#L131-L139) only replaces columns already present in the inner projection.

For example:

`UPDATE OPENQUERY(server, 'SELECT DepartmentID FROM dbo.Department') SET GroupName = 'x'`

If `GroupName` is encrypted, the outer assignment becomes an assignment to its cipher column, but the OPENQUERY rowset still exposes only `DepartmentID`. SQL Server requires the updated column to exist in the target table, view, or rowset. [Microsoft UPDATE documentation](https://learn.microsoft.com/en-us/sql/t-sql/queries/update-transact-sql?view=sql-server-ver17)

**Impact:** A statement accepted by the new rewrite path produces invalid physical SQL at the linked provider instead of valid rewritten SQL or a deterministic unsupported-shape error. The parameter path also rewrites the value before that downstream failure.

**Required Change:** Validate that every encrypted `SET` target is exposed by the rewritten OPENQUERY projection, or safely add all required physical projection columns. Add production-path tests for both literal and parameter assignments where the assigned encrypted column is omitted from the original `SELECT` list.

#### P1 — Valid SQL Server lexical forms can still make encrypted-table discovery fail open

**Problem:** [`findFromKeywordIndexIfPresent`](https://github.com/apache/shardingsphere/blob/f1855a143bc6a142ada2041365f0e7c0d80d1bb4/features/encrypt/core/src/main/java/org/apache/shardingsphere/encrypt/rewrite/token/generator/assignment/EncryptOpenQueryPassThroughSQL.java#L479-L513) skips strings and delimited identifiers, but not SQL comments. Therefore, `SELECT GroupName /* FROM */ FROM dbo.Department` selects the `FROM` inside the comment. Table extraction then returns empty, [`findEncryptTable`](https://github.com/apache/shardingsphere/blob/f1855a143bc6a142ada2041365f0e7c0d80d1bb4/features/encrypt/core/src/main/java/org/apache/shardingsphere/encrypt/rewrite/token/generator/assignment/EncryptOpenQueryUtils.java#L76-L82) returns empty, and the [decorator exits without rewriting](https://github.com/apache/shardingsphere/blob/f1855a143bc6a142ada2041365f0e7c0d80d1bb4/features/encrypt/core/src/main/java/org/apache/shardingsphere/encrypt/rewrite/context/EncryptSQLRewriteContextDecorator.java#L54-L84). SQL Server permits comments inside statements. [Microsoft block-comment documentation](https://learn.microsoft.com/en-us/sql/t-sql/language-elements/slash-star-comment-transact-sql?view=sql-server-ver17)

The regular-identifier readers also accept only letters, digits, and `_` ([parser lines 433–457 and 515–536](https://github.com/apache/shardingsphere/blob/f1855a143bc6a142ada2041365f0e7c0d80d1bb4/features/encrypt/core/src/main/java/org/apache/shardingsphere/encrypt/rewrite/token/generator/assignment/EncryptOpenQueryPassThroughSQL.java#L433-L457)). SQL Server regular identifiers may contain `$`, `@`, and `#`, so a configured table such as `Depart$ment` is truncated during discovery and likewise misses the encrypt rule. [Microsoft identifier documentation](https://learn.microsoft.com/en-us/sql/relational-databases/databases/database-identifiers?view=sql-server-ver17)

**Impact:** Valid SQL targeting an encrypted table can silently forward logical column names and plaintext values rather than being rewritten or rejected.

**Required Change:** Use one quote-, delimiter-, comment-, and SQL Server regular-identifier-aware lexical model for discovery and validation. Once the statement targets an encrypted table, unsupported syntax must fail closed rather than return “not an encrypt table.” Add production-path regression tests for comments before the real `FROM` and regular identifiers containing `$`, `@`, or `#`.

#### P2 — `getEncryptColumns()` exposes mutable rule state

**Problem:** The new public [`EncryptTable.getEncryptColumns()`](https://github.com/apache/shardingsphere/blob/f1855a143bc6a142ada2041365f0e7c0d80d1bb4/features/encrypt/core/src/main/java/org/apache/shardingsphere/encrypt/rule/table/EncryptTable.java#L77-L84) returns `columns.values()`. Java specifies that this collection is backed by the map and that removal operations remove the corresponding mappings. [Java `Map.values()` contract](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/Map.html#values())

This was also reproduced against the repository’s exact `java-util` dependency: calling `values().clear()` reduced the backing `CaseInsensitiveMap` size to zero.

**Impact:** Any caller can remove encrypt-column definitions through the returned collection, after which the rule silently stops recognizing those encrypted columns.

**Required Change:** Avoid exposing the collection or return a non-removable view/snapshot whose mutation contract cannot alter the rule. Add a direct focused test protecting the public contract, as required by the repository’s [public-type testing rule](https://github.com/apache/shardingsphere/blob/f1855a143bc6a142ada2041365f0e7c0d80d1bb4/AGENTS.md#L42-L43).

### Multi-Round Comparison

- **Resolved:** The latest head addresses the earlier parameter-plaintext path, overlapping token generation, optional-parameter handling, SQL literal escaping, quoted `FROM` tokens, additional grammar exclusions, and validation when `SET` contains no encrypted column. See the [first](https://github.com/apache/shardingsphere/pull/39156#pullrequestreview-4713365295), [second](https://github.com/apache/shardingsphere/pull/39156#pullrequestreview-4732447526), and [third](https://github.com/apache/shardingsphere/pull/39156#pullrequestreview-4751768173) review rounds and the [latest author update](https://github.com/apache/shardingsphere/pull/39156#issuecomment-5082228686).

- **Partially resolved:** Encrypted-target discovery now handles bracketed and double-quoted `FROM` tokens, but comments and valid regular-identifier symbols still cause the same fail-open class of behavior.

- **Newly identified:** The outer encrypted assignment is not checked against the OPENQUERY projection, and the new `EncryptTable` accessor exposes its backing map through a mutable values view.

### Review Details

- **Reviewed Scope:** All 20 changed files at immutable head `f1855a143bc6a142ada2041365f0e7c0d80d1bb4`; local triple-dot scope from merge-base `cb50781bc2a08ba20d7b9a66ec5e017fc2bf17c5` matched GitHub’s changed-file list. Production entry points, literal and parameter paths, unit tests, rewriter integration cases, documentation, release notes, prior review rounds, linked [issue #30227](https://github.com/apache/shardingsphere/issues/30227), and applicable Microsoft SQL Server contracts were reviewed.
- **Not Reviewed:** GitHub Actions/check results and a live SQL Server linked-provider execution. This limits runtime compatibility evidence but does not weaken the code-local counterexamples above.
- **Verification:** The review inventory and full-coverage ledger validated successfully with all 20 files classified and a final adversarial pass producing zero additional findings. The mutable-view behavior was reproduced with the exact repository dependency. Maven was not run because the active checkout was not the immutable PR head; build/test claims therefore remain author-provided rather than reviewer-verified.
- **Release Notes / Documentation:** Both are present and appropriately describe the intended narrow scope. They do not identify comments or SQL Server-valid `$`, `@`, and `#` regular identifiers as unsupported, and documentation alone would not make silent encryption bypass acceptable.

ClaireLytt:
@terrymanu Thank you for your review and patience!
**P1 — Encrypted assignments need not exist in the OPENQUERY target rowset**  
→ Code: `EncryptAssignmentTokenGenerator.generateOpenQueryUpdateTokens()` — for each encrypted SET column, call `EncryptOpenQueryPassThroughSQL.containsSelectColumn()` and throw `UnsupportedEncryptSQLException` when the column is missing from the inner SELECT list.  
→ Unit tests: `EncryptAssignmentTokenGeneratorOpenQueryUnsupportedShapeTest.assertGenerateSQLTokenWithOpenQueryEncryptAssignmentOmittedFromSelectExpectsException`; `EncryptOpenQueryPassThroughSQLTest.assertContainsSelectColumn`.

**P1 — Valid SQL Server lexical forms can still make encrypted-table discovery fail open**  
→ Code: `EncryptOpenQueryPassThroughSQL.findFromKeywordIndexIfPresent()` — skip `/* ... */` and `--` comments when locating `FROM`; `parse()` strips comments from the SELECT list via `stripSqlComments()` before validation/rewrite. Regular-identifier readers use `isRegularIdentifierCharacter()` / `isRegularIdentifierStartCharacter()` to accept `$`, `@`, and `#`.  
→ Unit tests: `findTableNameArguments` covers comments and `$`/`@`/`#` table names; rewrite tests for block/line-comment `FROM`.  
→ IT (`update.xml` + `query-with-cipher.yaml`): double-quoted/`FROM` string literal/block-comment/line-comment cases; `Depart$ment` / `Depart@ment` / `Depart#ment` table configs and rewrite assertions.

**P2 — getEncryptColumns() exposes mutable rule state**  
→ Code: `EncryptTable.getEncryptColumns()` — return `Collections.unmodifiableCollection(new LinkedList<>(columns.values()))` instead of `columns.values()`.  
→ Unit tests: `EncryptTableTest.assertGetEncryptColumnsReturnsConfiguredColumns`; `assertGetEncryptColumnsMutationDoesNotAlterRuleState` (`clear()` throws `UnsupportedOperationException` and rule state is preserved).

terrymanu (changes requested):
### Summary

**Reason:** I do not think we should continue adding isolated fixes to this PR. The original root-cause analysis is correct: `OPENQUERY` is represented as a `FunctionTableSegment`, while the existing Encrypt rewrite path recognizes only regular tables. However, safely rewriting the SQL inside `OPENQUERY` first requires clear contracts for the remote provider dialect, the mapping between remote tables and Encrypt metadata, and ownership of the inner SQL parser. None of these prerequisites is currently established. I suggest pausing the implementation and reopening the scope and design discussion.

### Issues

#### P1 — The remote dialect and metadata ownership for OPENQUERY are undefined

**Problem:**

Microsoft explicitly states that an OPENQUERY string is executed by the OLE DB provider of the specified linked server. The official examples even use an Oracle linked server. Linked servers are also explicitly designed to support heterogeneous data sources such as Oracle:

- [OPENQUERY documentation](https://learn.microsoft.com/en-us/sql/t-sql/functions/openquery-transact-sql?view=sql-server-ver17)
- [Linked servers documentation](https://learn.microsoft.com/en-us/sql/relational-databases/linked-servers/linked-servers-database-engine?view=sql-server-ver17)

The current implementation does not use the linked-server argument or any provider information:

- [`EncryptOpenQueryUtils.findEncryptTable`](https://github.com/apache/shardingsphere/blob/fe5d012f5387a2d29185864c92506845526a97b3/features/encrypt/core/src/main/java/org/apache/shardingsphere/encrypt/rewrite/token/generator/assignment/EncryptOpenQueryUtils.java#L76-L82) extracts only the terminal table name from the inner SQL and uses it to look up an `EncryptTable` in the current database.
- [`EncryptOpenQueryPassThroughSQL`](https://github.com/apache/shardingsphere/blob/fe5d012f5387a2d29185864c92506845526a97b3/features/encrypt/core/src/main/java/org/apache/shardingsphere/encrypt/rewrite/token/generator/assignment/EncryptOpenQueryPassThroughSQL.java#L34-L740) introduces a 776-line SQL Server-style scanner, validator, and rewriter inside the Encrypt module.
- [`quotePhysicalColumnName`](https://github.com/apache/shardingsphere/blob/fe5d012f5387a2d29185864c92506845526a97b3/features/encrypt/core/src/main/java/org/apache/shardingsphere/encrypt/rewrite/token/generator/assignment/EncryptOpenQueryPassThroughSQL.java#L731-L735) always emits SQL Server bracket-quoted column names into the remote query.
- The new tests all use an untyped `MyLinkedServer` and expect bracket quoting, for example in [`update.xml`](https://github.com/apache/shardingsphere/blob/fe5d012f5387a2d29185864c92506845526a97b3/test/it/rewriter/src/test/resources/scenario/encrypt/case/query-with-cipher/dml/update/update.xml#L119-L125).

For an Oracle or another non-SQL Server linked provider, this can therefore emit invalid `[cipher_column]` syntax. Different linked servers may also contain same-named tables, yet the implementation can apply the current database’s Encrypt rule to them without distinguishing the server or provider.

Documenting a “narrow SELECT ... FROM shape” does not resolve this issue because the implementation can neither identify the remote provider nor verify that the remote table belongs to the selected Encrypt metadata.

The scope growth also shows that this is no longer a local rewrite fix:

- First review round: 13 files, `+465/-54`
- Second review round: 16 files, `+1340/-68`
- Third review round: 20 files, `+2054/-68`
- Fourth review round: 20 files, `+2277/-68`
- Current head: 21 files, `+2497/-68`

Most of this expansion is not unrelated churn. It was added to address legitimate lexical, syntactic, and fail-open problems exposed by the current approach. However, that is also evidence that the implementation has crossed into a new SQL parsing and remote-metadata contract rather than remaining a small Encrypt rewrite change.

**Impact:** Without explicit linked-provider and remote-metadata contracts, the implementation can rewrite the wrong remote table, emit SQL for the wrong dialect, or continue silently bypassing Encrypt rewrite on additional lexical forms. The 111 focused tests pass, but they demonstrate only that the hard-coded model is internally consistent; they do not establish correctness for OPENQUERY’s provider-dependent semantics.

**Discussion Needed:** Please pause the current implementation and first reach agreement on:

1. Whether support is limited to SQL Server-to-SQL Server linked servers and, if so, how that restriction is configured, detected, and enforced.
2. How linked server, catalog, schema, and table identity map to the correct Encrypt metadata instead of matching only the terminal table name.
3. Once the provider dialect is known, whether inner SQL parsing belongs in the existing `DatabaseType`-driven parser/rewriter infrastructure, a new explicit extension contract, or another established owner. Continuing to extend an independent SQL scanner inside Encrypt should not be the default.
4. Whether the current work should be split into a design/infrastructure PR followed by an Encrypt integration PR, with a narrower and independently verifiable subtask under issue #30227.

I suggest applying the `type: discussion` label and pausing further patch-level refinement until these questions are resolved.

### Multi-Round Comparison

- Most previously reported issues—parameter encryption, overlapping tokens, string escaping, identifier handling, comments, unsupported shapes, and mutable collection exposure—have been addressed in subsequent commits.
- Each round exposed another parsing boundary. This means the earlier assessment that the direction was settled and only implementation details remained is no longer sufficient.
- The latest review’s leading-comment fail-open case and missing SQL-token test remain valid implementation-level observations, but they should not be the next primary tasks until provider, metadata, and parser ownership are agreed.

### Review Details

- **Reviewed Scope:** All 21 changed files at the latest head `fe5d012f5387a2d29185864c92506845526a97b3`; base and merge-base `39e7d4851fdbf9e1be5ca9d05897a8aa14db1a40`. The extracted base/head source snapshot matched GitHub `/pulls/39156/files` exactly. The complete production path, unit tests, rewrite integration cases, configuration, English and Chinese limitation documentation, release note, linked issue, and all five public review rounds were inspected.
- **SQL / Dialect Scope:** The outer dialect is SQL Server. I reviewed the SQL Server OPENQUERY grammar and visitor, `FunctionTableSegment`, the existing `DatabaseType`-driven parser contract, and Microsoft’s OPENQUERY, UPDATE, and linked-server documentation. The inner dialect cannot be determined from the current model, which is the direction-level blocker described above.
- **Not Reviewed Scope:** GitHub Actions/check runs, live SQL Server linked-server/OLE DB execution, and runtime compatibility with each possible remote provider. These omissions do not affect the conclusion because the provider and metadata contracts are already missing before provider execution. No required public GitHub evidence remained inaccessible.
- **Verification:** `./mvnw -pl features/encrypt/core -DskipITs -Dspotless.skip=true -Dtest=EncryptSQLRewriteContextDecoratorTest,EncryptAssignmentTokenGeneratorOpenQueryUnsupportedShapeTest,EncryptAssignmentTokenGeneratorTest,EncryptOpenQueryPassThroughSQLTest,EncryptOpenQueryUtilsTest,EncryptUpdateAssignmentTokenGeneratorTest,EncryptTableTest -Dsurefire.failIfNoSpecifiedTests=false test` exited with `0`; 111 tests passed. PR metadata, files, commits, comments, and reviews were refreshed before publication, and the head remained unchanged. The full-coverage ledger validated successfully, and the final adversarial pass produced `0` new findings.
- **Release Note / User Docs:** The release note and both limitation documents are present, but they describe only the accepted inner-query shape. They do not define the linked-provider or remote Encrypt-metadata contract and therefore do not resolve the blocker above.

ClaireLytt:
@terrymanu Thanks for the review.
Thanks for raising this. I agree this is a design/contract issue, not another patch-level fix. I’m pausing further Encrypt-local OPENQUERY scanner expansion until we align on the points below.

### 1) SQL Server → SQL Server only?
**Proposal for v1: yes.**

OPENQUERY SQL runs on the linked-server provider. The current code assumes SQL Server lexing and always emits `[cipher_column]`, so v1 should support only SQL Server → SQL Server with a documented narrow T-SQL `SELECT ... FROM` shape. Heterogeneous/unknown providers are out of scope.

- **Configured:** explicit linked-server contract, e.g. `openQuery.linkedServers.MyLinkedServer.databaseType: SQLServer` (do not infer from the name).
- **Detected:** resolve `OPENQUERY(linkedServer, ...)` against that contract before rewrite.
- **Enforced:** unconfigured / non-SQLServer / unsupported → fail closed with `UnsupportedEncryptSQLException`; never silently skip rewrite or emit SQL Server quoting for an unknown provider.

### 2) Metadata mapping?
**Proposal: stop matching only the terminal table name.**

Current flow (`inner SQL → table name → findEncryptTable(tableName)`) can mis-apply Encrypt rules across linked servers/catalogs/schemas with the same table name.

For v1, identify the remote table as:

`linkedServer + optional catalog + schema + table`

and map it to an Encrypt logical table via config. Rewrite only when that mapping exists; if only a bare table name is available and no linked-server mapping exists, fail closed instead of falling back to `findEncryptTable("Department")`.

### 3) Who owns inner SQL parsing?
**Proposal: Encrypt should not keep growing an independent SQL scanner.**

`EncryptOpenQueryPassThroughSQL` has already become a SQL Server lexical/syntactic engine. Prefer:

1. reuse `DatabaseType`-driven parser/rewriter infrastructure where practical; or
2. add an explicit dialect contract (e.g. `OpenQueryDialectHandler`) with a SQL Server-only first implementation.

Encrypt should only apply encryption on top: logical→physical mapping, SET tokens, and fail-closed checks. In short: dialect/provider owns parsing/quoting; Encrypt owns encryption rewrite.

### 4) Split the work?
**Proposal: yes, under #30227.**

1. **Design/contract** — SQL Server-only scope, linked-server config, remote-table mapping, fail-closed rules  
2. **Infrastructure** — dialect/provider ownership + SQL Server handler/tests  
3. **Encrypt integration** — OPENQUERY UPDATE rewrite on the agreed contract  

I’d prefer to keep this PR open as draft/discussion rather than close it, then either reduce it to the design/contract part or migrate the agreed subset into the split PRs.

Which direction does the community prefer?

- **A)** document the SQL Server-only contract first, then a narrowed implementation  
- **B)** land infrastructure ownership first, defer Encrypt integration  

I personally **prefer B**: land infrastructure ownership first, and defer Encrypt integration.

Reason: the main review concern is remote dialect/metadata ownership, not another Encrypt-local patch. Landing config, detection, fail-closed enforcement, and dialect ownership first makes the later Encrypt rewrite independently verifiable and avoids growing the Encrypt-local scanner further.

I also hope we can open a new design/infrastructure issue under #30227 (linked from this PR) to track that work cleanly, instead of continuing design discussion only inside this large implementation PR.

If the community agrees with B, the next steps would be:

* Mark this PR as draft/discussion and pause further Encrypt-local scanner expansion.
* Open a new issue for OPENQUERY remote dialect/metadata ownership and the SQL Server-only contract.
* Align the v1 **contract** there: SQL Server → SQL Server only, linked-server config, remote-table → Encrypt mapping, and fail-closed rules.
* Implement an **Infrastructure PR**: linked-server/provider config, SQL Server dialect handler (or reuse of DatabaseType-driven infrastructure), detection/enforcement, and focused tests.
* After that lands, open an **Encrypt integration PR** for OPENQUERY UPDATE rewrite on top of the agreed contract, migrating any agreed subset from this PR.

Please let me know whether this direction looks good, and whether opening that follow-up issue under #30227 is acceptable.

terrymanu:
Thanks for the thoughtful response and for pausing further scanner changes.

**Current status: Retained — Needs Discussion.**

I agree with **B as the implementation sequence**, with one qualification: the contract must be agreed before starting an infrastructure PR. The proposed `openQuery.linkedServers...` configuration and `OpenQueryDialectHandler` are possible design options, not yet accepted contracts.

The current head still performs terminal-table-name lookup in [`EncryptOpenQueryUtils`](https://github.com/apache/shardingsphere/blob/fe5d012f5387a2d29185864c92506845526a97b3/features/encrypt/core/src/main/java/org/apache/shardingsphere/encrypt/rewrite/token/generator/assignment/EncryptOpenQueryUtils.java#L76-L82) and owns the inner parser in [`EncryptOpenQueryPassThroughSQL`](https://github.com/apache/shardingsphere/blob/fe5d012f5387a2d29185864c92506845526a97b3/features/encrypt/core/src/main/java/org/apache/shardingsphere/encrypt/rewrite/token/generator/assignment/EncryptOpenQueryPassThroughSQL.java#L34-L740), so the direction-level concern remains.

Please open a focused design issue linked to [#30227](https://github.com/apache/shardingsphere/issues/30227) and this PR. It should first settle:

1. The supported provider boundary and how it is authoritatively detected.
2. The ownership and source of linked-server and remote-table metadata.
3. How remote identity maps to Encrypt metadata.
4. Whether existing parser infrastructure can be reused or a new extension contract is justified.
5. Fail-closed behavior, configuration compatibility, and runtime verification.

After that contract is accepted, the implementation can be split into infrastructure work followed by Encrypt integration.

Once the design issue is linked, I recommend closing this PR rather than keeping the current implementation as a long-lived draft, because its core ownership model is being reconsidered. The useful test scenarios can be migrated into the later PRs.

ClaireLytt:
@terrymanu 
The proposal in https://github.com/apache/shardingsphere/issues/39313 outlines the design specifications for SQL Server’s OPENQUERY encryption rewriting feature. Please review and confirm the feasibility of this approach to facilitate further development efforts. Once confirmed the contract, I will close this pr. Thanks a lot!

terrymanu:
Thanks for opening and linking #39313. This moves the design discussion to the right place, but the contract has not yet been accepted. The provider boundary, remote-table metadata ownership and mapping, parser ownership, and fail-closed behavior still require community agreement.

**Current status: Retained — Needs Discussion.**

Let’s continue the design discussion in #39313. Since the current implementation’s core ownership model is being reconsidered, please close this PR now rather than wait for the contract to be finalized. Once the contract is accepted, the implementation can proceed through focused PRs based on the agreed design, reusing the valuable test scenarios from this PR where appropriate.

Issue #24982, opened 2023-04-03
Can not show logic tables when I config same actual table name in different storage unit


strongduanmu:
### Which version of ShardingSphere did you use?

[c20da72](https://github.com/apache/shardingsphere/commit/c20da72e9ecbe8302b7fcc0eb1a60dead97e4854)

### Which project did you use? ShardingSphere-JDBC or ShardingSphere-Proxy?

ShardingSphere-Proxy

### Expected behavior

Show all logic tables when I config same actual table name in different storage unit

### Actual behavior

Just show the first logic table now.

```
mysql> show tables like 't_order0';
+-----------------------+------------+
| Tables_in_sharding_db | Table_type |
+-----------------------+------------+
| t_order0              | BASE TABLE |
+-----------------------+------------+
1 row in set (0.00 sec)
```

### Reason analyze (If you can)

Then ShardingRule#findTableRuleByActualTable method just consider actual table now. When I config same actual table name in different storage unit, and establish different mapping relationships for actual tables, the logic will be wrong.

![5hQgg2IQd9](https://user-images.githubusercontent.com/10829171/229473384-eab646ac-8b4d-4003-aac0-8ad3c4f77487.jpg)

### Steps to reproduce the behavior, such as: SQL to execute, sharding rule configuration, when exception occur etc.

```yaml
databaseName: sharding_db

dataSources:
  ds_0:
    url: jdbc:mysql://127.0.0.1:3306/demo_ds_0?serverTimezone=UTC&useSSL=false
    username: root
    password: 123456
    connectionTimeoutMilliseconds: 30000
    idleTimeoutMilliseconds: 60000
    maxLifetimeMilliseconds: 1800000
    maxPoolSize: 50
    minPoolSize: 1
  ds_1:
    url: jdbc:mysql://127.0.0.1:3306/demo_ds_1?serverTimezone=UTC&useSSL=false
    username: root
    password: 123456
    connectionTimeoutMilliseconds: 30000
    idleTimeoutMilliseconds: 60000
    maxLifetimeMilliseconds: 1800000
    maxPoolSize: 50
    minPoolSize: 1

rules:
- !SHARDING
  tables:
    t_order0:
      actualDataNodes: ds_0.t_order
    t_order1:
      actualDataNodes: ds_1.t_order
  defaultDatabaseStrategy:
    none:
  defaultTableStrategy:
    none:
```

And then execute following sqls:

```sql
-- execute in ds_0 and ds_1
[...]

PR #39157, opened 2026-07-16
Fix wrong logic table metadata when config same actual table name in different storage unit

ClaireLytt:
for https://github.com/apache/shardingsphere/issues/24982

## Summary

Fix wrong logic table metadata when configuring same actual table name in different storage units.

When two logical sharding tables map to the same physical table name indifferent data sources (e.g. `t_order0 -> ds_0.t_order` and `t_order1 -> ds_1.t_order`), only the first logical table is visible. `SHOW TABLES` loses `t_order1` because the metadata revision path cannot disambiguate which logical table owns which physical table.

### Root cause

1. `MetaDataLoader` loads `TableMetaData("t_order")` from each storage unitthen merges them, losing storage-unit provenance.
2. `ShardingTableNameReviser` calls `findShardingTableByActualTable("t_order")` which returns the first match — both entries are renamed to `t_order0`.
3. The aggregator deduplicates by revised name — `t_order1` disappears.

### Fix

Carry storage-unit provenance through the existing `TableMetaData` object so
the revision path can disambiguate:

- Add a transient `storageUnitName` field to `TableMetaData` (`@ToString(exclude = "storageUnitName")` to avoid false structural mismatches in the aggregator).
- Stamp `storageUnitName` from `MetaDataLoaderMaterial` in `MetaDataLoader.load()` immediately after loading.
- Add a default `revise(name, rule, storageUnitName)` overload to `TableNameReviser`; the default delegates to the existing two-arg method, keeping all non-sharding implementations unchanged.
- Override in `ShardingTableNameReviser` to call `findShardingTableByDataSourceAndActualTable()` for an exact
  `(dataSource, actualTable)` lookup.
- When the qualified lookup misses (e.g. readwrite-splitting / shadow composed mapper, where loaded provenance is a physical storage unit while sharding `actualDataNodes` use a logical data source name), fall back to the original two-arg `revise()` so composed-mapper semantics stay unchanged.
- The revised `TableMetaData` is a new instance — `storageUnitName` is consumed during revision and not propagated downstream.

### Scope

- Limited to schema metadata / `SHOW TABLES` visibility for direct storage-unit mappings.
- No general storage-qualified identity contract introduced.
- No changes to protocol/header, generated-column, index, foreign-key, or alias semantics.
- Composed-mapper topologies keep pre-existing table-name-only revision behavior via qualified-miss fallback.
- Existing `MetaDataReviseEntry` implementations (Single, Encrypt, Broadcast) and unrelated rules remain behaviorally unchanged.
- Storage-unit provenance is local to the metadata-loading/revision path.

## Test plan

- [x] `ShardingSameActualTableNameMetaDataVisibilityTest`: exercises the
  production loader-to-revision path via a registered `DialectMetaDataLoader`
  mock (no manual provenance injection); verifies both `t_order0` and
  `t_order1` remain visible.
- [x] Same test with `check-table-metadata-enabled=true`: verifies no false
  structural mismatch is reported.
- [x] `SchemaTableMetaDataAggregatorTest`: same-structure tables with different
  `storageUnitName` values do not throw
  `RuleAndStorageMetaDataMismatchedException` when check is enabled.
- [x] `ShardingTableNameReviserTest`: qualified hit, qualified-miss fallback to
  actual-table lookup, and null `storageUnitName` path.
---


ClaireLytt:
test in proxy after fixing this bug

terrymanu (changes requested):
### Summary

**Reason:** The storage-unit-qualified lookup is the right direction, but the latest head introduces one shared aggregation regression and leaves two production branches using ambiguous table identity.

### Issues

#### P1 — Storage-unit provenance becomes a metadata mismatch

**Problem:** [`TableMetaData`](https://github.com/apache/shardingsphere/blob/4d49eeea3e3b2b3e2d04dd209f30d8f2ce0c2524/database/connector/core/src/main/java/org/apache/shardingsphere/database/connector/core/metadata/data/model/TableMetaData.java#L33-L47) includes the new `storageUnitName` in Lombok-generated `toString()`. The loader assigns a different value to each storage unit, while [`SchemaTableMetaDataAggregator`](https://github.com/apache/shardingsphere/blob/4d49eeea3e3b2b3e2d04dd209f30d8f2ce0c2524/database/connector/core/src/main/java/org/apache/shardingsphere/database/connector/core/metadata/data/revise/SchemaTableMetaDataAggregator.java#L45-L60) uses `toString()` as its structural-equality check.

**Impact:** With documented property `check-table-metadata-enabled=true`, identical actual-table schemas from `ds_0` and `ds_1` are rejected as different. A current-head reproduction throws `RuleAndStorageMetaDataMismatchedException`. This also affects other users of the shared aggregator, including broadcast metadata.

**Required Change:** Please keep storage-unit provenance out of structural metadata comparison, or replace the `toString()` comparison with an explicit structural comparison. Add coverage for identical schemas loaded from different storage units with metadata checking enabled.

#### P1 — Qualified identity is used only for the table name

**Problem:** [`TableMetaDataReviseEngine`](https://github.com/apache/shardingsphere/blob/4d49eeea3e3b2b3e2d04dd209f30d8f2ce0c2524/infra/common/src/main/java/org/apache/shardingsphere/infra/metadata/database/schema/reviser/table/TableMetaDataReviseEngine.java#L73-L83) passes `storageUnitName` to `TableNameReviser`, but column, index, and constraint revision still receive only the ambiguous actual table name. [`ShardingMetaDataReviseEntry`](https://github.com/apache/shardingsphere/blob/4d49eeea3e3b2b3e2d04dd209f30d8f2ce0c2524/features/sharding/core/src/main/java/org/apache/shardingsphere/sharding/metadata/reviser/ShardingMetaDataReviseEntry.java#L48-L60) therefore selects those revisers through the original unqualified lookup.

**Impact:** For `t_order0 -> ds_0.t_order` and `t_order1 -> ds_1.t_order` with different generated-key columns, current head correctly names the second table `t_order1` but marks `t_order0`’s generated-key column in its metadata. The resulting logical table is internally inconsistent.

**Required Change:** Please propagate the storage-unit-qualified sharding-table identity through all sharding metadata revision branches, preferably resolving the owning `ShardingTable` once and reusing it. Add a regression test with different per-table metadata for the two storage units.

#### P1 — A qualified miss falls back to the ambiguous lookup

**Problem:** [`ShardingTableNameReviser`](https://github.com/apache/shardingsphere/blob/4d49eeea3e3b2b3e2d04dd209f30d8f2ce0c2524/features/sharding/core/src/main/java/org/apache/shardingsphere/sharding/metadata/reviser/table/ShardingTableNameReviser.java#L37-L44) calls the old unqualified lookup when a non-null storage-unit context has no matching sharding table.

**Impact:** If `ds_0.t_order` belongs to sharding table `t_order0` while `ds_1.t_order` belongs to another rule, the qualified lookup correctly returns empty, but the fallback still renames the `ds_1` table to `t_order0`. Because the metadata engine applies the sharding reviser to every loaded table, this preserves the original table-loss behavior in mixed-rule configurations.

**Required Change:** When storage-unit context is available and the qualified lookup misses, please preserve the original table name. Retain the unqualified fallback only when origin context is genuinely unavailable, if compatibility requires it. Add mixed-rule coverage for this branch.

### Review Details

- **Reviewed Scope:** All nine files from GitHub `/pulls/39157/files`: `RELEASE-NOTES.md`; the connector-core loader and metadata model; the infra-common reviser interface, engine, and test; and the sharding-core reviser and rule classes. Latest head: `4d49eeea3e3b2b3e2d04dd209f30d8f2ce0c2524`; base ref: `master`; GitHub-reported base SHA: `6ecd9c15b0ea089c6036fa4d1776098a19b03cb1`; local merge-base: `cb50781bc2a08ba20d7b9a66ec5e017fc2bf17c5`. The local triple-dot file list matched GitHub exactly.
- **Not Reviewed Scope:** GitHub Actions/check-runs and full Proxy/database E2E smoke were not reviewed. No dependency, packaging, license, or distribution files changed.
- **Verification:** The focused latest-head Maven run covering connector-core, infra-common, and sharding-core completed successfully with 89 tests and exit code 0. Compiled-bytecode inspection confirmed that `storageUnitName` participates in `TableMetaData.toString()`. Three bounded current-head reproductions confirmed the aggregation exception, cross-rule generated-column selection, and qualified-miss fallback described above. Public PR, issue, file, commit, comment, and review data were available; there is no relevant GitHub evidence gap.
- **Release Note / User Docs:** The release-note entry is present. Additional user documentation is not required because the intended fix introduces no configuration or migration change.

ClaireLytt:
@terrymanu Thank you for your review suggestion. Here are new changes.
**P1 — Storage-unit provenance becomes a metadata mismatch**  
→ `TableMetaData`: `@ToString(exclude = "storageUnitName")` so aggregator structural equality ignores provenance  
→ Test: `SchemaTableMetaDataAggregatorTest.assertAggregateWithSameStructureDifferentStorageUnitNames`

**P1 — Qualified identity is used only for the table name**  
→ `TableMetaDataRevisionContext` + `ShardingTableMetaDataRevisionContext`: resolve owning `ShardingTable` once and reuse  
→ `MetaDataReviseEntry.createTableMetaDataRevisionContext(...)` / `ShardingMetaDataReviseEntry`: factory + qualified `findShardingTable(...)`  
→ `TableMetaDataReviseEngine`: prefer context path for table/column/index/constraint revision  
→ `ColumnReviseEngine` / `IndexReviseEngine` / `ConstraintReviseEngine`: accept pre-bound revisers  
→ Test: `ShardingQualifiedTableMetaDataReviseTest.assertReviseWithDifferentGeneratedKeyColumnsPerStorageUnit`

**P1 — A qualified miss falls back to the ambiguous lookup**  
→ `ShardingTableNameReviser.revise(..., storageUnitName)`: on non-null storage unit + qualified miss, return `originalName` (no unqualified fallback)  
→ Test: `ShardingQualifiedTableMetaDataReviseTest.assertRevisePreservesOriginalNameWhenQualifiedLookupMisses`

terrymanu (changes requested):
### Summary

**Reason:** The implementation should pause until maintainers and the author agree whether this PR is a narrow fix for issue #24982 or the beginning of a general storage-qualified table identity design. The current patch changes central shared contracts without an agreed scope, owner, lifecycle, or completeness boundary. Continuing to add patch-level fixes would expand the architecture before those decisions are settled.

### Issues

#### P1 — A narrow metadata-display issue has become a shared identity architecture change

**Problem:** [Issue #24982](https://github.com/apache/shardingsphere/issues/24982) reports one externally visible symptom: when two logic tables use the same actual table name in different storage units, only the first logic table is shown.

The current PR changes 25 files across three core areas:

- [`TableMetaData`](https://github.com/apache/shardingsphere/blob/263bcf18ba245b7411b545665ac372d6c82e0d35/database/connector/core/src/main/java/org/apache/shardingsphere/database/connector/core/metadata/data/model/TableMetaData.java#L31-L47) becomes storage-aware.
- [`MetaDataReviseEntry`](https://github.com/apache/shardingsphere/blob/263bcf18ba245b7411b545665ac372d6c82e0d35/infra/common/src/main/java/org/apache/shardingsphere/infra/metadata/database/schema/reviser/MetaDataReviseEntry.java#L62-L162) gains a new revision-context contract and multiple storage-aware SPI methods.
- The shared table, column, index, and constraint revision engines gain new execution paths.
- [`ShardingRule`](https://github.com/apache/shardingsphere/blob/263bcf18ba245b7411b545665ac372d6c82e0d35/features/sharding/core/src/main/java/org/apache/shardingsphere/sharding/rule/ShardingRule.java#L118-L169) gains derived cross-rule data-source mapping state.
- Generated-column, index, and foreign-key metadata semantics are included even though the reported behavior is table visibility.

This is no longer only a localized bug fix. It defines how physical table identity is propagated through the shared metadata lifecycle.

**Impact:** A contract introduced here can affect every database dialect and every rule using the metadata revision SPI. If the larger scope is not intentional, the PR creates a permanent shared abstraction for one narrow symptom. If the larger scope is intentional, the linked issue and PR description do not define the complete requirements.

**Discussion Needed:** Before implementation continues, please agree on one of these scopes:

1. A narrowly bounded fix for the `SHOW TABLES` behavior, with an explicit maximum shared-layer impact and clear non-goals.
2. A general storage-qualified table identity design covering metadata loading, revision, aggregation, and reverse lookup across affected features.

If the second scope is selected, it should be tracked as an explicit architectural change and split into reviewable infrastructure and feature-level stages. If neither shared-layer change is justified for this issue, deferring the issue is preferable to merging a partial workaround.

#### P1 — Ownership and lifecycle of the qualified identity are not defined

**Problem:** The current patch distributes qualified identity across several owners:

- [`MetaDataLoader`](https://github.com/apache/shardingsphere/blob/263bcf18ba245b7411b545665ac372d6c82e0d35/database/connector/core/src/main/java/org/apache/shardingsphere/database/connector/core/metadata/data/loader/MetaDataLoader.java#L83-L97) assigns mutable storage provenance to `TableMetaData`.
- [`TableMetaDataReviseEngine`](https://github.com/apache/shardingsphere/blob/263bcf18ba245b7411b545665ac372d6c82e0d35/infra/common/src/main/java/org/apache/shardingsphere/infra/metadata/database/schema/reviser/table/TableMetaDataReviseEngine.java#L73-L105) chooses between a context path and a parallel storage-aware fallback path.
- `Optional.empty()` from the context factory represents both “this SPI does not support qualified contexts” and “the current qualified table has no owner”.
- [`ShardingQualifiedTableFinder`](https://github.com/apache/shardingsphere/blob/263bcf18ba245b7411b545665ac372d6c82e0d35/features/sharding/core/src/main/java/org/apache/shardingsphere/sharding/metadata/reviser/table/ShardingQualifiedTableFinder.java#L42-L69) performs additional topology interpretation.
- `ShardingRule` maintains another derived mapping model, although existing infrastructure such as [`DataNodes`](https://github.com/apache/shardingsphere/blob/263bcf18ba245b7411b545665ac372d6c82e0d35/infra/common/src/main/java/org/apache/shardingsphere/infra/datanode/DataNodes.java#L50-L60) already composes data-source mapper rules.

The resulting model does not establish whether storage identity is temporary loader provenance, durable table metadata, a qualified `DataNode`, or a rule-owned topology key.

**Impact:** Without one owner and lifecycle invariant, different revision branches and rules can interpret the same identity differently. The current alias first-match, layered mapper, and foreign-key target counterexamples are evidence of this unresolved model rather than isolated missing conditions.

**Discussion Needed:** Please settle these design questions before selecting implementation details:

- What is the canonical qualified identity: physical storage unit plus actual table, logical rule data source plus actual table, or a richer data-node identity?
- Which component owns conversion across Readwrite Splitting, Shadow, and other mapper layers?
- At which phase is the identity introduced, consumed, and discarded?
- What is the semantic difference between an SPI not supporting qualified identity and a qualified lookup finding no owner?
- Is this identity permitted to survive logical metadata aggregation?
- Should the shared SPI expose this concept, or should it remain internal to the metadata-loading/revision pipeline and Sharding?

#### P1 — The completeness boundary is undefined across adjacent reverse-lookup paths

**Problem:** The public discussion in [issue #24982](https://github.com/apache/shardingsphere/issues/24982) already identifies multiple consumers that reverse-map actual table names, including metadata result sets and protocol/header paths. The current PR focuses on schema metadata revision but also expands into generated columns, indexes, and constraints.

At the same time, the proposed model does not yet behave consistently for:

- two logic tables behind different data-source aliases;
- multiple composed `DataSourceMapperRuleAttribute` layers;
- foreign-key referenced tables with the same physical name on different storage units;
- qualified lookup misses;
- production loader-to-revision provenance handoff.

These are not being requested as additional patches in the current design. They demonstrate that the intended domain of the new identity contract has not been established.

**Impact:** There is no objective point at which the PR can be declared complete. It could close the narrow issue while leaving other callers with the same ambiguity, or it could continue expanding until it implicitly becomes a repository-wide identity redesign.

**Discussion Needed:** Please define an acceptance boundary covering:

- the exact user-facing entry points included in this PR;
- adjacent paths intentionally deferred to follow-up work;
- supported topology combinations;
- behavior for table names, generated columns, indexes, and foreign-key references;
- compatibility expectations for existing `MetaDataReviseEntry` implementations;
- the tests required at the shared-contract and feature-integration levels.

### Multi-Round Comparison

- The first public review identified concrete implementation defects in the initial narrow patch. The structural comparison and direct qualified-miss cases were addressed, while qualified metadata consistency was expanded into a revision context.
- The later public review treated the expanded implementation as an ordinary `Change Request`. That feedback mode is superseded by this review because the prerequisite architectural scope and ownership decisions are not publicly established.
- The previously reported alias, composed-mapper, foreign-key, API-surface, and loader-handoff findings remain relevant evidence, but the author should not continue fixing them within the current design until the discussion above is resolved.

### Next Steps

Please pause the current implementation and use `type: discussion` to confirm the intended scope, ownership, lifecycle, and staging strategy. After that decision:

- a narrow solution can return as a correspondingly small PR; or
- an accepted broader design can be split into explicit infrastructure and feature-level changes.

### Review Details

- **Reviewed Scope:** All 25 files reported by GitHub `/pulls/39157/files`, including 15 production Java files, 9 test files, and `RELEASE-NOTES.md`. Latest head: `263bcf18ba245b7411b545665ac372d6c82e0d35`; base ref: `master`; GitHub base SHA: `80e6541874d47b5814e3d1400c925d6ff3afa0ed`; local merge-base: `248623385cba6d03264e4b035966233e8a0d329e`. The local triple-dot file list matched GitHub exactly. The linked issue, previous public reviews, mapper composition, metadata loaders, and adjacent reverse-lookup consumers were also inspected.
- **Not Reviewed Scope:** GitHub Actions/check-runs and full Proxy/database E2E execution were not reviewed. No dependency, packaging, license, or distribution files changed.
- **Verification:** Focused current-head tests across connector-core, infra-common, and sharding-core completed with 41 tests and exit code 0. A bounded current-head alias counterexample produced the wrong logic-table result. Complete source traces also confirmed different semantics between the new one-hop mapping state and existing ordered mapper composition, and between qualified owner-table selection and unqualified foreign-key target selection. These results establish the unresolved design boundary; they are not requests to continue patching the current approach.
- **Release Note / User Docs:** The release-note entry is present. Additional user documentation is not required for the narrow reported bug. Documentation and compatibility requirements must be reconsidered if the broader shared identity contract is accepted.

ClaireLytt:
@terrymanu 
Thanks for the review. I agree the current patch has grown too broad. I will take **Option A: a narrowly bounded fix for the #24982 table-visibility symptom**.

**Scope**
- In scope: schema metadata / `SHOW TABLES` visibility when two logic tables share the same actual table name on different storage units.
- Out of scope: a general storage-qualified identity architecture, protocol/header reverse-lookup redesign, and repository-wide generated-column / index / FK semantic unification.

**Ownership / lifecycle**
Because this PR will not introduce a shared qualified-identity contract, the architectural ownership and lifecycle questions are out of scope here. I will keep any provenance handling pipeline-local and avoid a half-defined shared model.

**Acceptance boundary**
- Included entry points: the reported metadata visibility path for direct storage units such as `t_order0 -> ds_0.t_order` and `t_order1 -> ds_1.t_order`.
- Deferred paths: protocol/header reverse-lookup, multi-layer `DataSourceMapperRuleAttribute` topologies, FK reverse lookup, and other adjacent reverse-lookup callers.
- Behavior in this PR: fix logic-table visibility / naming for the supported topology above; no commitment to generated-column, index, or FK semantic unification.
- Compatibility: existing `MetaDataReviseEntry` implementations should remain unchanged unless required by this narrow path; rules that do not participate keep current behavior.
- Required tests for this PR:
  - Feature regression: with the same actual table name on different storage units, both logic tables remain visible in schema metadata / `SHOW TABLES`.
  - Shared-contract coverage only where this narrow path still touches shared code (for example, identical schemas from different storage units must not fail aggregation when `check-table-metadata-enabled=true`).
  - Deferred topologies and adjacent reverse-lookup paths are intentionally not required for merge of this PR.

I will revise the PR to match this narrowed scope and update the description accordingly. Please confirm if this boundary is acceptable.

terrymanu:
Thanks for clarifying the boundary.

**Current status: The previous needs-discussion finding is withdrawn at the requirements level. The current implementation concern remains until the PR head is revised to match this boundary.**

I agree with Option A:

- Limit this PR to schema metadata / `SHOW TABLES` visibility for direct storage units such as `t_order0 -> ds_0.t_order` and `t_order1 -> ds_1.t_order`.
- Do not introduce a general storage-qualified identity contract or change protocol/header, generated-column, index, foreign-key, alias, or composed-mapper semantics.
- Keep existing `MetaDataReviseEntry` implementations and unrelated rules behaviorally unchanged.
- Keep any required storage-unit provenance local to this metadata-loading/revision path.
- Add a regression test that exercises the production loader-to-revision path without manually injecting provenance, and verifies that both logical tables remain visible.
- Retain the aggregation regression for `check-table-metadata-enabled=true`.
- Update the PR description to record these scope limits and deferred paths.

The current head (`263bcf18ba245b7411b545665ac372d6c82e0d35`) still contains the broader shared revision context and related semantics, so please narrow the implementation accordingly. The deferred reverse-lookup paths can be handled in separate follow-up work.

ClaireLytt:
@terrymanu 
The latest E2E failures look unrelated to this PR’s metadata changes. The failing tests (e.g. GeneralDQLE2EIT) fail at connection setup with MySQL error Too many connections, before any query/metadata assertion runs. Could you please help to check it?

terrymanu (changes requested):
### Result

**Review Result: Mergeable**

Against the acceptance boundary the maintainer publicly agreed on 2026-08-10 (the narrow fix for issue #24982), the current head `0444a92` is a correct and minimal implementation; the deepest adversarial verification found no blocking defect passing the proof gate. However, this re-review solidified — with source evidence — two residual defects previously described only as "deferred," together with their precise trigger conditions; the merge decision should be made on that basis, see Coverage.

### Evidence

- **Main fix path (verified line by line).** `SchemaMetaDataReviseEngine` revises each table first and aggregates by revised name afterwards; `MetaDataLoader` stamps provenance on both the dialect-loading and default-loading paths (the refactor is control-flow-equivalent to the original, with SQLException propagation unchanged); `TableMetaDataReviseEngine` passes `originalMetaData.getStorageUnitName()` into the new default method; `ShardingTableNameReviser` performs the qualified `(ds, actualTable)` hit, falling back to the legacy unqualified lookup on miss or null provenance. `t_order`@ds_0 → `t_order0` and `t_order`@ds_1 → `t_order1`; both survive after `ShardingSchemaTableAggregationReviser` groups by revised name.
- **Composed-topology fallback (confirmed this round).** `DataNodes.getDataNodes` maps through `DataSourceMapperRuleAttribute` layers in rule order down to physical units, so under readwrite-splitting/shadow the provenance is a physical unit name while sharding `actualDataNodes` use a logical name → the qualified lookup necessarily misses → fallback to legacy semantics. This matches the PR description's claim.
- **Public API shape unchanged (verified down to Lombok semantics).** `storageUnitName` is a non-final field without `@NonNull`, so the `@RequiredArgsConstructor`-generated 5-arg constructor signature is unchanged; `@ToString(exclude=...)` has precedent in the same package (`ShardingTable`); the class is not `Serializable` and defines no `equals`/`hashCode`. The aggregator's `toString()`-based structural comparison is blind to provenance.
- **No regression of existing correct behavior (adversarial enumeration).** Scenarios where the qualified hit agrees with the unqualified result (regular sharding; same actual table name across units within one logic table) produce identical output; case mismatch, cross-protocol `normalize()`, three-tier `instance.db` names, unconfigured tables, and broadcast-adjacent paths all reduce to miss→fallback or equivalent results; the SPI-registry reflection swap in `MetaDataLoader` is safe under sequential test execution with `MetaDataLoaderTest` precedent; the only other `TableNameReviser` referencer, `CreateTablePushDownMetaDataRefresherTest`, uses a real lambda, so default-method delegation applies and it is unaffected by the signature change.
- **Multi-round reconciliation.** Round-1 P1#1 (aggregator false mismatch) is fixed with unit plus end-to-end tests; round-1 P1#2 and the round-2 needs-discussion were withdrawn/downgraded by the maintainer on 2026-08-10 with an explicit boundary; the current 12-file head touches neither `MetaDataReviseEntry`, the column/index/constraint engines, nor protocol paths, and `ShardingRule` only gains a query method reading the pre-existing `dataSourceToTablesMap`.

### Coverage

- **Reviewed head:** `0444a921f443cd05fdf1f7aeeffeab62c3d66f90`, base `master`@`3ff25b95698`; all 12 files from GitHub `/pulls/39157/files` re-reviewed hunk by hunk.
- **Residual defect 1 (explicitly deferred by the maintainer, quantified this round): column/index/constraint revision of the newly visible table binds to the wrong rule.** `getColumnGeneratedReviser`/`getIndexReviser`/`getConstraintReviser` in `ShardingMetaDataReviseEntry` still use `findShardingTableByActualTable` (unqualified first match; cedarsoftware `CaseInsensitiveMap` preserves insertion order → always hits `t_order0`). Precise trigger conditions: the two logic tables configure different keyGenerateColumns (`t_order1`'s column `generated` flag is computed from `t_order0`'s config, `ShardingColumnGeneratedReviser.java:31`); or ShardingSphere-generated index names exist (the anonymous-index candidate set is taken from the wrong logic table's metadata, `ShardingIndexReviser.java:47-56`); or FKs reference same-named actual tables. The issue's original reproduction (no key generator, PRIMARY KEY only) is unaffected. `t_order0`'s own metadata is byte-for-byte identical to pre-PR — no regression. The maintainer's 2026-08-10 words "Do not ... change ... generated-column, index, foreign-key ... semantics" explicitly forbid fixing this in the current PR, so it is not a blocker, but it should be tracked as a follow-up issue.
- **Residual defect 2 (non-blocking observation): the qualified lookup is case-sensitive.** `dataSourceToTablesMap` uses exact matching while the legacy unqualified lookup's `actualTables` is a `CaseInsensitiveSet`; when the configured case differs from the database-returned case, the qualified lookup misses → falls back to legacy behavior — the original bug still exists in such environments (no regression; the fix is silently inactive).
- **Minor:** the PR description calls the field "transient," but the code has no `transient` modifier (the class is not serializable; purely a documentation wording inaccuracy); incremental/DDL refresh paths still use the unqualified lookup (full load and incremental refresh may produce inconsistent naming for `t_order1` — an already-deferred adjacent reverse-lookup path).
- **Confirmed behavior change (the correct outcome of the fix):** with `check-table-metadata-enabled=true`, same-named actual tables with different structures change from a pre-PR false-positive exception (wrongly merging and comparing two logic tables) to each passing independently — exactly the semantics the maintainer required regression-test coverage for.
- **Discovery lenses:** root cause/behavior, blast radius/contracts, and tests/runtime/operations were all redone at strict level; 15 deep-sweep checks converged with zero new candidates; no candidate passed the five-condition proof gate as a blocker.
- **Verification boundary:** the head is not in the local object store and fetching it exceeds review authority, so no local test execution was performed; conclusions rest on deterministic source traces of the base commit and head files plus the author's reported full local build. The review ledger script could not initialize for the same reason and was replaced by manual per-file accounting. **This conclusion is code-scope only (Code Correctness Review); GitHub Actions/CI were not reviewed**; the E2E `Too many connections` failures discussed in the PR thread were not evaluated.

terrymanu (approved):

PR #38793, opened 2026-06-04
Remove redundant parameter replay invocations

FlyingZC:
Fixes #ISSUSE_ID.

Changes proposed in this pull request:
  -

---

Before committing this PR, I'm sure that I have checked the following options:
- [ ] My code follows the [code of conduct](https://shardingsphere.apache.org/community/en/involved/conduct/code/) of this project.
- [ ] I have self-reviewed the commit code.
- [ ] I have (or in comment I request) added corresponding labels for the pull request.
- [ ] I have passed maven check locally : `./mvnw clean install -B -T1C -Dmaven.javadoc.skip -Dmaven.jacoco.skip -e`.
- [ ] I have made corresponding changes to the documentation.
- [ ] I have added corresponding unit tests for my changes.
- [ ] I have updated the Release Notes of the current development version. For more details, see [Update Release Note](https://shardingsphere.apache.org/community/en/involved/contribute/contributor/)

PR #39161, opened 2026-07-16
Preserve JDBC setter semantics when replaying prepared statement parameters

### LINKED ISSUE #37504: Insertion Failure with BLOB Columns

somiljain2006:
Fixes #37504 

Preserve the original JDBC setter semantics during prepared statement replay by tracking the parameter-binding method independently of the logical parameter values. This fixes cases where stream and LOB-based parameters, such as Oracle BLOB values bound via setBlob(InputStream), were previously replayed using setObject(), resulting in driver-specific failures (for example, ORA-17004: invalid column type). Add regression tests covering replay of BLOB, stream, CLOB, typed object, null, and shifted parameter scenarios to ensure the original binding method is preserved during execution.

---

- [x] I have passed Maven check locally: `./mvnw clean install -B -T1C -Dmaven.javadoc.skip -Dmaven.jacoco.skip -e`.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -41,6 +41,7 @@
 1. Proxy: Fix MySQL BLOB data corruption when string-like prepared statement parameters target BLOB columns - [#39072](https://github.com/apache/shardingsphere/pull/39072)
 1. Agent: Fix wrong target class name in StaticMethodAdviceExecutor error logs - [#39077](https://github.com/apache/shardingsphere/pull/39077)
 1. Metadata: Fix Oracle metadata version comparison skipping identity and collation columns on 18c and later - [#39104](https://github.com/apache/shardingsphere/pull/39104)
+1. JDBC: Preserve JDBC setter semantics when replaying prepared statement parameters - [#39161](https://github.com/apache/shardingsphere/pull/39161)
 
 ### Enhancements
 
--- a/jdbc/src/main/java/org/apache/shardingsphere/driver/jdbc/adapter/AbstractPreparedStatementAdapter.java
+++ b/jdbc/src/main/java/org/apache/shardingsphere/driver/jdbc/adapter/AbstractPreparedStatementAdapter.java
@@ -17,12 +17,10 @@
 
 package org.apache.shardingsphere.driver.jdbc.adapter;
 
-import com.google.common.io.CharStreams;
 import lombok.Getter;
+import lombok.RequiredArgsConstructor;
 import org.apache.shardingsphere.driver.jdbc.unsupported.AbstractUnsupportedOperationPreparedStatement;
-import org.apache.shardingsphere.infra.exception.generic.UnknownSQLException;
 
-import java.io.IOException;
 import java.io.InputStream;
 import java.io.Reader;
 import java.math.BigDecimal;
@@ -39,6 +37,7 @@ import java.sql.Timestamp;
 import java.util.ArrayList;
 import java.util.Calendar;
 import java.util.List;
+import java.util.Objects;
 
 /**
  * Adapter for {@code PreparedStatement}.
@@ -48,14 +47,16 @@ public abstract class AbstractPreparedStatementAdapter extends AbstractUnsupport
     @Getter
     private final List<Object> parameters = new ArrayList<>();
     
+    private final List<ReplaySetParameter> setParameterContexts = new ArrayList<>();
+    
     @Override
     public final void setNull(final int parameterIndex, final int sqlType) {
-        setParameter(parameterIndex, null);
+        setParameter(parameterIndex, null, new ReplaySetParameter(ReplayMethod.SET_NULL, new Object[]{sqlType}));
     }
     
     @Override
     public final void setNull(final int parameterIndex, final int sqlType, final String typeName) {
-        setParameter(parameterIndex, null);
+        setParameter(parameterIndex, null, new ReplaySetParameter(ReplayMethod.SET_NULL_WITH_NAME, new Object[]{sqlType, typeName}));
     }
     
     @Override
@@ -140,32 +141,32 @@ public abstract class AbstractPreparedStatementAdapter extends AbstractUnsupport
     
     @Override
     public final void setBlob(final int parameterIndex, final Blob x) {
-        setParameter(parameterIndex, x);
+        setParameter(parameterIndex, x, new ReplaySetParameter(ReplayMethod.SET_BLOB, new Object[0]));
     }
     
     @Override
     public final void setBlob(final int parameterIndex, final InputStream x) {
-        setParameter(parameterIndex, x);
+        setParameter(parameterIndex, x, new ReplaySetParameter(ReplayMethod.SET_BLOB_STREAM, new Object[0]));
     }
     
     @Override
     public final void setBlob(final int parameterIndex, final InputStream x, final long length) {
-        setParameter(parameterIndex, x);
+        setParameter(parameterIndex, x, new ReplaySetParameter(ReplayMethod.SET_BLOB_STREAM_WITH_LENGTH, new Object[]{length}));
     }
     
     @Override
     public final void setClob(final int parameterIndex, final Clob x) {
-        setParameter(parameterIndex, x);
+        setParameter(parameterIndex, x, new ReplaySetParameter(ReplayMethod.SET_CLOB, new Object[0]));
     }
     
     @Override
     public final void setClob(final int parameterIndex, final Reader x) {
-        setParameter(parameterIndex, x);
+        setParameter(parameterIndex, x, new ReplaySetParameter(ReplayMethod.SET_CLOB_READER, new Object[0]));
     }
     
     @Override
     public final void setClob(final int parameterIndex, final Reader x, final long length) {
-        setParameter(parameterIndex, x);
+        setParameter(parameterIndex, x, new ReplaySetParameter(ReplayMethod.SET_CLOB_READER_WITH_LENGTH, new Object[]{length}));
     }
     
     @Override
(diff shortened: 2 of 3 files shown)


somiljain2006:
@terrymanu Can you review this pr?

terrymanu (changes requested):
### Correction

- Previous Finding: The ownership and supported scope of the replay state need to be reconsidered.
- Current Status: Retained
- Reason: The stricter latest-head review confirmed four independent risks. The stream lifecycle and equality/identity-based provenance problems further demonstrate that this is not an ordinary patch-level change request.

### Summary

- Feedback Mode: Needs Discussion
- Reason: The implementation attempts to reconstruct the original JDBC setter from parameter values at the final replay stage. However, parameter provenance may already have been lost during SQL rewrite, translation, routing, or batch collection. Adding more matching branches cannot make this contract provably correct.
- Recommendation: Pause the current implementation. Maintainers should first decide whether this should be a narrowly scoped LOB fix or a systemic binding-provenance model carried through rewrite and batch execution.

### Review Intensity

This PR changes `AbstractPreparedStatementAdapter`, which is part of the core parameter binding and replay path shared by ShardingSphere-JDBC prepared statements. It affects ordinary execution, SQL rewrite, routing, multi-data-source execution, statement reuse, and batch execution across database dialects.

A regression in this path can affect far more than the reported Oracle BLOB case. Because of that blast radius, I intentionally applied a higher review bar than I would for a localized dialect-specific fix. The concern is not the size of the diff by itself, but that the change modifies a central behavioral contract relied upon by multiple execution layers.

### Issues

#### [P1] Discussion Needed — Parameter provenance is already lost before replay

`matchReplaySetParameter` first matches the current position using `Objects.equals`, then searches for the first identical object reference. It has no rewrite mapping, transformation provenance, or per-batch binding snapshot:

- Rewrite can add and replace parameters through [StandardParameterBuilder](https://github.com/apache/shardingsphere/blob/ddcfe79ea246b9a5237157072c4e1951bc26a5c0/infra/rewrite/core/src/main/java/org/apache/shardingsphere/infra/rewrite/parameter/builder/impl/StandardParameterBuilder.java#L73-L89). Encrypt rewrite performs both replacement and insertion in [EncryptAssignmentParameterRewriter](https://github.com/apache/shardingsphere/blob/ddcfe79ea246b9a5237157072c4e1951bc26a5c0/features/encrypt/core/src/main/java/org/apache/shardingsphere/encrypt/rewrite/parameter/rewriter/EncryptAssignmentParameterRewriter.java#L97-L112).
- `Objects.equals` delegates to the first argument’s application-defined `equals` implementation when references differ, as documented by [Java SE 8](https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#equals-java.lang.Object-java.lang.Object-). Distinct but equal values can therefore inherit the wrong setter context, while custom `equals` implementations can introduce exceptions or side effects into parameter replay.
- If the same object occurs at multiple original indexes with different setters, the identity scan always selects the first context.
- A typed null shifted by an inserted parameter cannot be identity-matched and falls back to `setObject(index, null)`.
- Batch execution loses the information completely. [`addBatch()` immediately calls `clearParameters()`](https://github.com/apache/shardingsphere/blob/ddcfe79ea246b9a5237157072c4e1951bc26a5c0/jdbc/src/main/java/org/apache/shardingsphere/driver/jdbc/core/statement/ShardingSpherePreparedStatement.java#L324-L345), while [PreparedStatementParametersReplayCallback](https://github.com/apache/shardingsphere/blob/ddcfe79ea246b9a5237157072c4e1951bc26a5c0/jdbc/src/main/java/org/apache/shardingsphere/driver/executor/callback/replay/PreparedStatementParametersReplayCallback.java#L27-L36) receives only `List<Object>`.

Consequently, batch LOB and stream parameters still replay through `setObject`. The PR does not fix the reported class of problem for batch execution.

Before implementation continues, the design needs to establish which layer owns binding provenance and how added, replaced, removed, duplicated, and per-batch parameters retain or intentionally discard their original setter semantics.

#### [P1] Discussion Needed — Streams and readers have no repeatable replay lifecycle

The PR starts retaining the original `InputStream` and `Reader` objects. A single logical execution, however, may produce multiple physical prepared statements, and batch execution defers replay.

JDBC permits a driver to consume stream data as needed until EOF. It also explicitly distinguishes `setBlob(InputStream)` from `setBinaryStream` because the former informs the driver that the value must be sent as a BLOB. See the [Java SE 8 PreparedStatement contract](https://docs.oracle.com/javase/8/docs/api/java/sql/PreparedStatement.html#setBlob-int-java.io.InputStream-long-).

A general `InputStream` does not guarantee that it can be reset, as documented by [Java SE 8 InputStream](https://docs.oracle.com/javase/8/docs/api/java/io/InputStream.html#markSupported--). Replaying the same stream reference to multiple physical statements can therefore produce EOF, empty data, truncated data, or length errors after the first bind or execution.

The Mockito tests only verify which setter was invoked. They never consume the stream, so they cannot detect this failure mode.

Character streams introduce another regression. The PR replaces the previously materialized `String` with a raw `Reader`, which then enters `QueryContext`. If that parameter participates in a sharding condition, [ConditionValue](https://github.com/apache/shardingsphere/blob/ddcfe79ea246b9a5237157072c4e1951bc26a5c0/features/sharding/core/src/main/java/org/apache/shardingsphere/sharding/route/engine/condition/generator/ConditionValue.java#L57-L63) rejects it because it is not `Comparable`.

The design first needs to define:

1. The logical value used for routing and rewrite.
2. The physical payload used for JDBC binding.
3. How stream payloads are recreated, consumed, and released across multiple routes and batches.

Keeping one raw stream reference in a parallel list does not define those semantics.

#### [P1] Discussion Needed — The declared setter-preservation contract does not match the implementation

The PR title, description, and release note state that original JDBC setter semantics are preserved. The implementation only covers selected setters, and even those are incomplete:

- `setBlob(index, (Blob) null)` and `setClob(index, (Clob) null)` are present in the tests, but the replay `instanceof` checks necessarily fail for null and fall back to `setObject`. See the [replay switch](https://github.com/apache/shardingsphere/blob/ddcfe79ea246b9a5237157072c4e1951bc26a5c0/jdbc/src/main/java/org/apache/shardingsphere/driver/jdbc/adapter/AbstractPreparedStatementAdapter.java#L286-L330).
- The new test explicitly expects `setInt` to replay as `setObject`, rather than preserving `setInt`: [PreparedStatementAdapterTest](https://github.com/apache/shardingsphere/blob/ddcfe79ea246b9a5237157072c4e1951bc26a5c0/jdbc/src/test/java/org/apache/shardingsphere/driver/jdbc/adapter/PreparedStatementAdapterTest.java#L273-L295).
- Twenty supported setter overloads still become generic `SET_OBJECT`, including the `Date`, `Time`, and `Timestamp` overloads that accept a `Calendar`.
- The shifted-parameter test covers only a unique, non-null object identity. It does not cover typed nulls, repeated objects, rewrite replacement, translation, batch execution, or multi-route execution.

The contract must be decided before the implementation and release note can be evaluated consistently:

- Preserve only the Oracle `setBlob(InputStream)` case under explicitly limited conditions.
- Preserve every supported JDBC setter.
- Provide best-effort replay with documented fallback behavior.

These are materially different compatibility promises and require different designs and tests.

#### [P2] Discussion Needed — The hot path regresses from O(n) to worst-case O(n²) without performance evidence

Every setter invocation now creates a `ReplaySetParameter` and an argument array. Once parameters are shifted, each replayed value may scan the complete original parameter list through the [matching implementation](https://github.com/apache/shardingsphere/blob/ddcfe79ea246b9a5237157072c4e1951bc26a5c0/jdbc/src/main/java/org/apache/shardingsphere/driver/jdbc/adapter/AbstractPreparedStatementAdapter.java#L416-L432), producing worst-case O(n²) work.

The same path was simplified one month earlier in [PR #38793](https://github.com/apache/shardingsphere/pull/38793#issuecomment-4618456891) specifically to remove per-replay allocations and intermediate traversal and retain direct O(n) binding.

This PR reintroduces more state, allocations, and nonlinear matching into that core path without a benchmark or complexity guard.

If a systemic provenance design is selected, mapping should remain linear. If a narrow fix is selected, ordinary setters should not pay this cost.

### Review Details

- Review focus: JDBC correctness, rewrite and batch provenance, stream lifecycle, compatibility contract, API boundaries, and hot-path performance.
- CI not reviewed by request.
- Reviewed scope: PR #39161 latest head `ddcfe79ea246b9a5237157072c4e1951bc26a5c0`, base `master`, merge-base `28975f419b52b00954eb52cf88116f62479676a`.
- File inventory: GitHub `/pulls/39161/files` and the local triple-dot diff matched exactly, 3/3 files.
- Public discussion: 1 issue comment, 0 submitted reviews, and 0 inline review comments.
- Verification: Read all changed files and traced direct execution, cached replay, SQL rewrite, translation, multi-route execution, and batch execution. Checked the implementation against the Java SE 8 JDBC, `InputStream`, `Reader`, and `Objects` contracts.
- Not reviewed: CI/check runs, full Maven build, E2E suites, and reproduction against a real Oracle JDBC driver.
- Secondary hygiene: `git diff --check` found 25 trailing-whitespace occurrences in added lines. The new protected replay helper and enum also have no in-repository consumers. These are not the primary reasons for the non-mergeable verdict.

terrymanu:
The findings from the previous review are retained. There has been no implementation update since the review, and the design concerns remain unresolved.

The current approach attempts to reconstruct JDBC setter provenance during replay, after SQL rewrite and batch processing may already have discarded that information. It also does not define a safe, repeatable lifecycle for stream and reader parameters across multiple physical statements.

Addressing this requires a redesigned parameter-binding provenance model rather than incremental changes to the current implementation. I am closing this PR due to inactivity. If you would like to continue, please first discuss and agree on the ownership and lifecycle semantics in [#37504](https://github.com/apache/shardingsphere/issues/37504), then open a new PR based on the latest `master`.

PR #39193, opened 2026-07-20
Rename MCPResourceURIVariables

terrymanu:


terrymanu:
Do not need review because of simple rename

taojintianxia (approved):

PR #39194, opened 2026-07-20
Refactor MCP constructors with Lombok

terrymanu:
Replace redundant MCP constructors with RequiredArgsConstructor while preserving their access levels and builder behavior.


terrymanu:
Do not need review because of simple refactor

taojintianxia (approved):

PR #39204, opened 2026-07-21
Fix MySQL prepared statement parameter signedness in Proxy

### LINKED ISSUE #39203: Proxy: MySQL prepared statement decodes signed INT parameters as unsigned for UNSIGNED columns
5.5.4-SNAPSHOT 6621687546488d654546b2c8a9f8bfd66cd12dfc

ShardingSphere-Proxy

### Expected behavior
For a MySQL server-side prepared statement, the signedness used to decode a `COM_STMT_EXECUTE` parameter must come from the parameter type metadata sent by the client.

For example, when a client calls:

```java
preparedStatement.setInt(1, -1);
// UPDATE t_unsigned_passthrough SET int_payload = ? WHERE id = ?
```
the parameter is encoded as:

- type: `MYSQL_TYPE_LONG`
- client unsigned flag: `0x00`
- value bytes: `FF FF FF FF`

ShardingSphere-Proxy should decode the value as `-1` and forward it to the backend. MySQL should then reject the assignment to an `INT UNSIGNED` column with an out-of-range error, consistent with a direct MySQL connection.

### Actual behavior
ShardingSphere-Proxy uses the target column's `UNSIGNED` metadata, collected during `COM_STMT_PREPARE`, to decode the execute parameter.

As a result, the same payload is decoded as `4294967295` instead of `-1`:

```text
UPDATE t_unsigned_passthrough SET int_payload = ? WHERE id = ? ::: [4294967295, 106]
```

Because `4294967295` is a valid value for `INT UNSIGNED`, MySQL accepts the statement. The original client value `-1` is silently changed by Proxy.

This differs from both direct MySQL behavior and the behavior for `BIGINT UNSIGNED` with `setLong(-1L)`, where the value remains `-1` and MySQL correctly returns an out-of-range error.

### Minimal reproduction
```sql
CREATE TABLE t_unsigned_passthrough (
    id INT PRIMARY KEY,
    int_payload INT UNSIGNED,
    bigint_payload BIGINT UNSIGNED
);

INSERT INTO t_unsigned_passthrough (id, int_payload, bigint_payload)
VALUES (106, 0, 0);
```

```java
try (PreparedStatement preparedStatement = connection.prepareStatement(
        "UPDATE t_unsigned_passthrough SET int_payload = ? WHERE id = ?")) {
    preparedStatement.setInt(1, -1);
    preparedStatement.setInt(2, 106);
    preparedStatement.executeUpdate();
}
```

Expected result:

```text
Data truncation: Out of range value for column 'int_payload' at row 1
```

Actual result through ShardingSphere-Proxy:

```text
The update succeeds and int_payload becomes 4294967295.
```

### Root cause analysis

The MySQL `COM_STMT_EXECUTE` protocol defines each parameter type with a type code and a flag byte. The highest bit of the flag byte (`0x80`) indicates whether the client parameter value is unsigned:

- MySQL protocol reference: https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_com_stmt_execute.html

The current implementation already reads this client flag into `MySQLPreparedStatementParameterType`, but `MySQLComStmtExecutePacket#readParameters` does not use it when decoding numeric values.

Instead, it uses `MySQLColumnDefinitionFlag.UNSIGNED` from the target column metadata generated during `COM_STMT_PREPARE`.

These two metadata sources have different responsibilities:

- The client parameter unsigned flag describes how the bytes in `COM_STMT_EXECUTE` are encoded.
- The target column `UNSIGNED` attribute describes the valid range enforced by MySQL after the value is decoded.

Using target-column metadata to decode the client payload can change the value before it reaches MySQL.

### Proposed fix

1. Decode numeric parameter values using `MySQLPreparedStatementParameterType#getUnsignedFlag()`.
2. Treat the parameter as unsigned only when `(unsignedFlag & 0x80) == 0x80`.
3. Remove `parameterColumnDefinitionFlags` from `MySQLServerPreparedStatement`, because it should not be used by execute-parameter numeric decoding.
4. Keep `parameterColumnTypes`, because they are still required to preserve raw bytes for string-family parameters targeting BLOB columns.
5. Add protocol unit tests covering the same `FF FF FF FF` payload with:
   - client unsigned flag `0x00` -> `-1`
   - client unsigned flag `0x80` -> `4294967295L`

### Relationship to the previous BLOB fix

This issue was identified while adding non-BLOB regression coverage after the MySQL BLOB prepared-statement fix in https://github.com/apache/shardingsphere/pull/39072.

Both issues occur while Proxy reads `COM_STMT_EXECUTE` parameters, but they require different metadata:

- BLOB protection needs target column type metadata to distinguish binary and text handling.
- Numeric signedness decoding must use the client parameter flag from the execute packet.

### Impact

This is a data correctness issue in MySQL Proxy prepared statements:

- A negative `int` value can be silently converted into a large positive unsigned value.
- The backend database cannot reject the invalid assignment because Proxy has already changed the value.
- The issue affects MySQL binary-protocol prepared statement execution when a signed client numeric parameter targets an unsigned column.

RaigorJiang:
Fixes #39203.

  - Decode MySQL binary prepared statement numeric parameters using the unsigned flag carried by `COM_STMT_EXECUTE`, instead of the `UNSIGNED` flag of the target column metadata.
  - Prevent a signed `setInt(-1)` parameter from being decoded as `4294967295` before it reaches an `INT UNSIGNED` backend column.
  - Remove `parameterColumnDefinitionFlags` from `MySQLServerPreparedStatement`, because target column flags must not determine client parameter wire decoding.
  - Keep `parameterColumnTypes` for the existing BLOB handling path, so string-family parameters targeting BLOB columns continue to preserve raw bytes.
  - Add signed and unsigned numeric packet coverage using the same `FF FF FF FF` payload:
    - client unsigned flag `0x00` decodes to `-1`;
    - client unsigned flag `0x80` decodes to `4294967295L`.
  - Keep regression coverage for `STRING`, `VAR_STRING`, and `VARCHAR` parameters targeting `TINYBLOB`, `BLOB`, `MEDIUMBLOB`, and `LONGBLOB` columns.

Manual verification:
  - Verified `INT UNSIGNED <- setInt(-1)` through Proxy now forwards `-1` and receives the same `SQLState 22001` out-of-range error as a direct MySQL connection.
  - Verified valid unsigned numeric INSERT and UPDATE scenarios through Proxy.
  - Verified INSERT and UPDATE byte preservation for `TINYBLOB`, `BLOB`, `MEDIUMBLOB`, `LONGBLOB`, `LONG VARBINARY`, `BINARY`, and `VARBINARY`.

---

- [x] I have made corresponding changes to the documentation. No user documentation change is required for this bug fix.
- [x] I have updated the Release Notes of the current development version. For more details, see [Update Release Note](https://shardingsphere.apache.org/community/en/involved/contribute/contributor/).

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -33,6 +33,7 @@
 1. Proxy: Fix incorrect generated key handling for explicit auto-increment values - [#38810](https://github.com/apache/shardingsphere/pull/38810)
 1. Proxy: Fix microseconds decoded as nanoseconds in MySQL binary TIME value - [#39138](https://github.com/apache/shardingsphere/pull/39138)
 1. Proxy: Fix MySQL BLOB data corruption when string-like prepared statement parameters target BLOB columns - [#39072](https://github.com/apache/shardingsphere/pull/39072)
+1. Proxy: Fix MySQL prepared statement parameter signedness decoding - [#39204](https://github.com/apache/shardingsphere/pull/39204)
 1. JDBC & Proxy: Remove default MySQL prepared statement query properties when creating data sources - [#38593](https://github.com/apache/shardingsphere/pull/38593)
 1. Mode: Fix rule metadata not removed from memory after dropping rules in Etcd cluster mode - [#38561](https://github.com/apache/shardingsphere/pull/38561)
 1. Agent: Fix wrong target class name in StaticMethodAdviceExecutor error logs - [#39077](https://github.com/apache/shardingsphere/pull/39077)
--- a/database/protocol/dialect/mysql/src/main/java/org/apache/shardingsphere/database/protocol/mysql/packet/command/query/binary/execute/MySQLComStmtExecutePacket.java
+++ b/database/protocol/dialect/mysql/src/main/java/org/apache/shardingsphere/database/protocol/mysql/packet/command/query/binary/execute/MySQLComStmtExecutePacket.java
@@ -23,7 +23,6 @@ import org.apache.shardingsphere.database.protocol.mysql.constant.MySQLBinaryCol
 import org.apache.shardingsphere.database.protocol.mysql.constant.MySQLNewParametersBoundFlag;
 import org.apache.shardingsphere.database.protocol.mysql.packet.command.MySQLCommandPacket;
 import org.apache.shardingsphere.database.protocol.mysql.packet.command.MySQLCommandPacketType;
-import org.apache.shardingsphere.database.protocol.mysql.packet.command.query.MySQLColumnDefinitionFlag;
 import org.apache.shardingsphere.database.protocol.mysql.packet.command.query.binary.MySQLPreparedStatementParameterType;
 import org.apache.shardingsphere.database.protocol.mysql.packet.command.query.binary.execute.protocol.MySQLBinaryProtocolValue;
 import org.apache.shardingsphere.database.protocol.mysql.packet.command.query.binary.execute.protocol.MySQLBinaryProtocolValueFactory;
@@ -46,6 +45,8 @@ public final class MySQLComStmtExecutePacket extends MySQLCommandPacket {
     private static final int ITERATION_COUNT = 1;
     
     private static final int NULL_BITMAP_OFFSET = 0;
+    
+    private static final int UNSIGNED_FLAG = 0x80;
     
     private final MySQLPacketPayload payload;
     
@@ -94,24 +95,24 @@ public final class MySQLComStmtExecutePacket extends MySQLCommandPacket {
      *
      * @param paramTypes parameter type of values
      * @param longDataIndexes indexes of long data
-     * @param parameterFlags column definition flag of parameters
      * @param parameterColumnTypes column type of parameters from COM_STMT_PREPARE
      * @return parameter values
      * @throws SQLException SQL exception
      */
     public List<Object> readParameters(final List<MySQLPreparedStatementParameterType> paramTypes, final Set<Integer> longDataIndexes,
-                                       final List<Integer> parameterFlags, final List<MySQLBinaryColumnType> parameterColumnTypes) throws SQLException {
+                                       final List<MySQLBinaryColumnType> parameterColumnTypes) throws SQLException {
         List<Object> result = new ArrayList<>(paramTypes.size());
         for (int paramIndex = 0; paramIndex < paramTypes.size(); paramIndex++) {
             if (longDataIndexes.contains(paramIndex)) {
                 result.add(null);
                 continue;
             }
-            MySQLBinaryColumnType parameterType = paramTypes.get(paramIndex).getColumnType();
+            MySQLPreparedStatementParameterType parameter = paramTypes.get(paramIndex);
+            MySQLBinaryColumnType parameterType = parameter.getColumnType();
             MySQLBinaryProtocolValue binaryProtocolValue = MySQLBinaryProtocolValueFactory.getBinaryProtocolValue(parameterType);
             Object value = nullBitmap.isNullParameter(paramIndex)
                     ? null
-                    : readParameterValue(binaryProtocolValue, parameterColumnTypes, paramIndex, parameterType, parameterFlags);
+                    : readParameterValue(binaryProtocolValue, parameterColumnTypes, paramIndex, parameterType, parameter.getUnsignedFlag());
             value = decodeStringParameterValue(parameterColumnTypes, paramIndex, parameterType, value);
             result.add(value);
         }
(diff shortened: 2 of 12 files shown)


RaigorJiang:
### Summary

**Review Result: Mergeable**

Reason: The change correctly restores MySQL `COM_STMT_EXECUTE` signedness ownership to the client-supplied parameter type flag, rather than deriving it from the target column metadata returned during `COM_STMT_PREPARE`. The implementation remains narrowly scoped, preserves the prior BLOB byte-preservation path, removes only now-invalid cached target-column flag state, and includes focused regression coverage. This result is based on code scope and local verification only; CI not reviewed by request.

### Evidence

- The [MySQL `COM_STMT_EXECUTE` protocol](https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_com_stmt_execute.html) defines a type-and-flag pair for each newly bound parameter, with the flag byte's high bit representing unsignedness.
- `MySQLComStmtExecutePacket` now decodes numeric values from `MySQLPreparedStatementParameterType#getUnsignedFlag()` at `database/protocol/dialect/mysql/src/main/java/org/apache/shardingsphere/database/protocol/mysql/packet/command/query/binary/execute/MySQLComStmtExecutePacket.java:127`. This prevents an `INT` parameter carrying `0xFFFFFFFF` with client flag `0x00` from being decoded as unsigned merely because the target column is `INT UNSIGNED`.
- `MySQLComStmtExecuteExecutor` stores the full client parameter type object when type information is supplied and reuses that object for subsequent executions at `proxy/frontend/dialect/mysql/src/main/java/org/apache/shardingsphere/proxy/frontend/mysql/command/query/binary/execute/MySQLComStmtExecuteExecutor.java:108`. Both the new-type and cached-type paths therefore retain the client unsigned bit.
- `MySQLComStmtPrepareExecutor` still calculates and emits target column flags in the `COM_STMT_PREPARE` response, while no longer retaining those flags as execution-decoding state at `proxy/frontend/dialect/mysql/src/main/java/org/apache/shardingsphere/proxy/frontend/mysql/command/query/binary/prepare/MySQLComStmtPrepareExecutor.java:141`.
- The packet-level regression tests use the same `MYSQL_TYPE_LONG` payload bytes for signed and unsigned cases, differing only in the parameter flag: `0x00` decodes to `-1`; `0x80` decodes to `4294967295L` at `database/protocol/dialect/mysql/src/test/java/org/apache/shardingsphere/database/protocol/mysql/packet/command/query/binary/execute/MySQLComStmtExecutePacketTest.java:72`.
- The existing BLOB regression matrix remains covered for `STRING`, `VAR_STRING`, and `VARCHAR` against `TINYBLOB`, `BLOB`, `MEDIUMBLOB`, and `LONGBLOB` target columns at `database/protocol/dialect/mysql/src/test/java/org/apache/shardingsphere/database/protocol/mysql/packet/command/query/binary/execute/MySQLComStmtExecutePacketTest.java:110`.
- Public PR discussion and review endpoints contained no prior review feedback to reconcile.

### Review Details

- Review Focus: Code Correctness Review; CI not reviewed by request.
- Reviewed Scope: PR head `7a74fc1bec2fc96d57d89304e0769367cd5fbd07`, merge base `5702a14ade73a3b5eeb34872af672ef5a9e1f264`; the GitHub `/files` inventory matched the local triple-dot diff:
  - `RELEASE-NOTES.md`
  - `database/protocol/dialect/mysql/src/main/java/org/apache/shardingsphere/database/protocol/mysql/packet/command/query/binary/execute/MySQLComStmtExecutePacket.java`
  - `database/protocol/dialect/mysql/src/test/java/org/apache/shardingsphere/database/protocol/mysql/packet/command/query/binary/execute/MySQLComStmtExecutePacketTest.java`
  - `proxy/frontend/dialect/mysql/src/main/java/org/apache/shardingsphere/proxy/frontend/mysql/command/query/binary/MySQLServerPreparedStatement.java`
  - `proxy/frontend/dialect/mysql/src/main/java/org/apache/shardingsphere/proxy/frontend/mysql/command/query/binary/execute/MySQLComStmtExecuteExecutor.java`
  - `proxy/frontend/dialect/mysql/src/main/java/org/apache/shardingsphere/proxy/frontend/mysql/command/query/binary/prepare/MySQLComStmtPrepareExecutor.java`
  - `proxy/frontend/dialect/mysql/src/test/java/org/apache/shardingsphere/proxy/frontend/mysql/command/MySQLCommandPacketFactoryTest.java`
  - `proxy/frontend/dialect/mysql/src/test/java/org/apache/shardingsphere/proxy/frontend/mysql/command/admin/MySQLComResetConnectionExecutorTest.java`
  - `proxy/frontend/dialect/mysql/src/test/java/org/apache/shardingsphere/proxy/frontend/mysql/command/query/binary/MySQLComStmtSendLongDataExecutorTest.java`
  - `proxy/frontend/dialect/mysql/src/test/java/org/apache/shardingsphere/proxy/frontend/mysql/command/query/binary/execute/MySQLComStmtExecuteExecutorTest.java`
  - `proxy/frontend/dialect/mysql/src/test/java/org/apache/shardingsphere/proxy/frontend/mysql/command/query/binary/prepare/MySQLPreparedStatementMetadataFactoryTest.java`
  - `proxy/frontend/dialect/mysql/src/test/java/org/apache/shardingsphere/proxy/frontend/mysql/command/query/binary/reset/MySQLComStmtResetExecutorTest.java`
- Not Reviewed Scope: CI/workflow status, full distribution packaging, and external client-to-server integration environments were not reviewed.
- Verification:
  - `./mvnw -pl database/protocol/dialect/mysql,proxy/frontend/dialect/mysql -DskipITs -DskipIT -DskipE2E -Dtest=MySQLComStmtExecutePacketTest,MySQLComStmtExecuteExecutorTest,MySQLCommandPacketFactoryTest,MySQLComResetConnectionExecutorTest,MySQLComStmtSendLongDataExecutorTest,MySQLPreparedStatementMetadataFactoryTest,MySQLComStmtResetExecutorTest,MySQLComStmtPrepareExecutorTest -DfailIfNoTests=true -Dsurefire.failIfNoSpecifiedTests=false test -Pcheck -T1C` — exit code `0`; protocol packet tests: 26 passed, Proxy targeted tests: 55 passed.
  - `./mvnw spotless:apply -Pcheck -T1C` — exit code `0`.
  - `./mvnw checkstyle:check -Pcheck -T1C` — exit code `0`.
- Release Note / User Docs: `RELEASE-NOTES.md` includes an accurate Proxy bug-fix entry linked to `#39204`; no separate user documentation change is needed for this internal protocol-decoding correction.

FlyingZC (approved):

PR #39217, opened 2026-07-22
Fix incorrect return value of SQLFederationResultSetMetaData#getColumnClassName

### LINKED ISSUE #39121: 【SQL Federation】SQL Federation ResultSetMetaData#getColumnClassName returns SqlTypeName enum class instead of actual Java column class

Apache ShardingSphere-JDBC 5.5.3.

Related modules:

- `shardingsphere-jdbc`: 5.5.3
- `shardingsphere-jdbc-dialect-mysql`: 5.5.3
- `shardingsphere-sharding-mysql`: 5.5.3

Environment:

- Java 17
- MySQL dialect
- SQL Federation enabled
- Apache Calcite resolved as 1.40.0

ShardingSphere-JDBC.

### Expected behavior

For SQL Federation query results, `ResultSetMetaData#getColumnClassName(column)` should return the Java class name corresponding to the actual result column type.

For example:

- BIGINT id column: `java.lang.Long`
- VARCHAR name column: `java.lang.String`
- DECIMAL column: `java.math.BigDecimal`

### Actual behavior

`ResultSetMetaData#getColumnClassName(column)` returns:

```text
org.apache.calcite.sql.type.SqlTypeName
```

This happens for normal result columns such as `id BIGINT`.

When a mapper framework maps results to `Map<String, Object>`, it may use the reported column class name to select a type handler. Since ShardingSphere reports the enum class `SqlTypeName`, MyBatis/MyBatis-Plus chooses an enum type handler and tries to parse the actual database value as a `SqlTypeName` enum constant.

Stack trace excerpt:

```text
Caused by: org.apache.ibatis.executor.result.ResultMapException:
Error attempting to get column 'id' from result set.
Cause: java.lang.IllegalArgumentException:
No enum constant org.apache.calcite.sql.type.SqlTypeName.2076978610812633090

at org.apache.ibatis.type.BaseTypeHandler.getResult(BaseTypeHandler.java:88)
at com.baomidou.mybatisplus.core.handlers.CompositeEnumTypeHandler.getResult(CompositeEnumTypeHandler.java:62)
at org.apache.ibatis.executor.resultset.DefaultResultSetHandler.applyAutomaticMappings(DefaultResultSetHandler.java:582)

Caused by: java.lang.IllegalArgumentException:
No enum constant org.apache.calcite.sql.type.SqlTypeName.2076978610812633090
at java.lang.Enum.valueOf(Enum.java:273)
at org.apache.ibatis.type.EnumTypeHandler.getNullableResult(EnumTypeHandler.java:49)
```

I checked the ShardingSphere 5.5.3 source locally.

Source location:

```text
D:\source\shardingsphere\shardingsphere-5.5.3\kernel\sql-federation\core\src\main\java\org\apache\shardingsphere\sqlfederation\resultset\SQLFederationResultSetMetaData.java
```

Current implementation:

```java
@Override
public String getColumnClassName(final int column) {
    return resultColumnType.getFieldList().get(column - 1).getType().getSqlTypeName().getClass().getName();
}
```

`getSqlTypeName()` returns a Calcite enum value. Calling `getClass().getName()` on that enum value always returns:

```text
org.apache.calcite.sql.type.SqlTypeName
```

It does not return the actual Java class for the result column.

Use ShardingSphere-JDBC 5.5.3 with MySQL dialect and SQL Federation enabled.

Sanitized YAML, preserving the real rule structure:

```yaml
databaseName: logic_db

mode:
  type: Standalone
  repository:
    type: JDBC

transaction:
  defaultType: LOCAL

dataSources:
  ds_0:
    dataSourceClassName: com.zaxxer.hikari.HikariDataSource
    driverClassName: com.mysql.cj.jdbc.Driver
    jdbcUrl: jdbc:mysql://127.0.0.1:3306/demo_master?useUnicode=true&characterEncoding=utf-8&serverTimezone=Asia/Shanghai
    username: root
    password: password
  ds_1:
    dataSourceClassName: com.zaxxer.hikari.HikariDataSource
    driverClassName: com.mysql.cj.jdbc.Driver
    jdbcUrl: jdbc:mysql://127.0.0.1:3306/demo_sharding_1?useUnicode=true&characterEncoding=utf-8&serverTimezone=Asia/Shanghai
    username: root
    password: password

sqlFederation:
  sqlFederationEnabled: true
  allQueryUseSQLFederation: false
  executionPlanCache:
    initialCapacity: 2000
    maximumSize: 65535

props:
  sql-show: true

rules:
  - !SINGLE
    tables:
      - ds_0.mnt_base_user

  - !SHARDING
    tables: {}
    defaultDatabaseStrategy:
      complex:
        shardingColumns: tenant_id
        shardingAlgorithmName: custom_database_tenant_sharding_algorithm
    defaultTableStrategy:
      complex:
        shardingColumns: tenant_id
        shardingAlgorithmName: tenant_sharding_algorithm
    shardingAlgorithms:
      custom_database_tenant_sharding_algorithm:
        type: CLASS_BASED
        props:
          strategy: COMPLEX
          algorithmClassName: com.example.sharding.CustomDatabaseTenantIdShardingAlgorithm
          dbPrefix: ds_
          tenantIdFiled: tenant_id
      tenant_sharding_algorithm:
        type: CLASS_BASED
        props:
          strategy: COMPLEX
          algorithmClassName: com.example.sharding.CustomTableTenantIdShardingAlgorithm
          tenantIdFiled: tenant_id
```

Run a SQL Federation query:

```sql
SELECT so.id AS id, u.name AS record_user_name
FROM ims_sales_order so
LEFT JOIN mnt_base_user u ON so.record_user_id = u.id
WHERE so.tenant_id = 1
ORDER BY so.create_time DESC
LIMIT 20
```

Then inspect metadata:

```java
try (ResultSet rs = ps.executeQuery()) {
    ResultSetMetaData md = rs.getMetaData();
    System.out.println(md.getColumnClassName(1));
}
```

Expected:

```text
java.lang.Long
```

Actual:

```text
org.apache.calcite.sql.type.SqlTypeName
```

```java
try (Connection connection = dataSource.getConnection();
     PreparedStatement ps = connection.prepareStatement("""
         SELECT so.id AS id, u.name AS record_user_name
         FROM ims_sales_order so
         LEFT JOIN mnt_base_user u ON so.record_user_id = u.id
         WHERE so.tenant_id = 1
         ORDER BY so.create_time DESC
         LIMIT 20
         """);
     ResultSet rs = ps.executeQuery()) {

    ResultSetMetaData md = rs.getMetaData();

    // Actual in SQL Federation: org.apache.calcite.sql.type.SqlTypeName
    // Expected for BIGINT id: java.lang.Long
    System.out.println(md.getColumnClassName(1));
}
```

### Suggested fix

`SQLFederationResultSetMetaData#getColumnClassName` should map Calcite `SqlTypeName` or `RelDataType` to the correct JDBC Java class name.

Possible mapping examples:

- BIGINT -> `java.lang.Long`
- INTEGER -> `java.lang.Integer`
- VARCHAR/CHAR -> `java.lang.String`
- DECIMAL -> `java.math.BigDecimal`
- BOOLEAN -> `java.lang.Boolean`
- DATE/TIME/TIMESTAMP -> corresponding JDBC date/time classes
- Unknown fallback -> `java.lang.Object`

The method should not return `getSqlTypeName().getClass().getName()`, because that returns the enum implementation class, not the result column Java class.

PRAHLAD09-dev:
## Description

Fixes #39121.

## Changes proposed in this pull request

- Fix incorrect return value of `SQLFederationResultSetMetaData#getColumnClassName`.
- Return the corresponding Java class name based on the effective JDBC column type instead of the `SqlTypeName` enum class.
- Update the corresponding unit test to verify the corrected behavior.


terrymanu (changes requested):
### Summary

**Reason:** The patch fixes the enum-class symptom for `VARCHAR`, but still reports classes incompatible with actual SQL Federation `getObject` values for Java-backed and dialect-converted columns. The linked issue’s required regression coverage is also incomplete.

### Issues

#### P1 — Keep `getColumnClassName` consistent with `getObject`

**Problem:** [`getColumnClassName`](https://github.com/apache/shardingsphere/blob/6370fd1e713e2acdca835faef9e081bcf5c10965/kernel/sql-federation/core/src/main/java/org/apache/shardingsphere/sqlfederation/resultset/SQLFederationResultSetMetaData.java#L187-L228) derives the class solely from `getColumnType`. This conflicts with the JDBC contract, which requires a class compatible with values produced by `ResultSet#getObject`:

- A `JavaType` backed by `BigInteger` is forced to JDBC `BIGINT` at [lines 154-160](https://github.com/apache/shardingsphere/blob/6370fd1e713e2acdca835faef9e081bcf5c10965/kernel/sql-federation/core/src/main/java/org/apache/shardingsphere/sqlfederation/resultset/SQLFederationResultSetMetaData.java#L154-L160), then reported as `Long`, while SQL Federation preserves the `BigInteger` value through [`getObject`](https://github.com/apache/shardingsphere/blob/6370fd1e713e2acdca835faef9e081bcf5c10965/kernel/sql-federation/core/src/main/java/org/apache/shardingsphere/sqlfederation/resultset/SQLFederationResultSet.java#L364-L366). MySQL unsigned `BIGINT` is explicitly modeled as `BigInteger`.
- The [MySQL converter](https://github.com/apache/shardingsphere/blob/6370fd1e713e2acdca835faef9e081bcf5c10965/kernel/sql-federation/dialect/mysql/src/main/java/org/apache/shardingsphere/sqlfederation/mysql/MySQLSQLFederationColumnTypeConverter.java#L28-L42) converts Boolean values to `Integer` values (`0`/`1`) but converts their JDBC type to `VARCHAR`, so this patch reports `String`.

Neither `BigInteger`/`Long` nor `Integer`/`String` has the assignability relationship permitted by the [Java SE 8 JDBC contract](https://docs.oracle.com/javase/8/docs/api/java/sql/ResultSetMetaData.html#getColumnClassName-int-).

**Impact:** Frameworks selecting handlers from metadata can still choose an incompatible handler for valid SQL Federation results, reproducing the same failure category as #39121.

**Required Change:** Please resolve the reported class from the actual SQL Federation value/converter contract and cover at least the Java-backed `BigInteger` and MySQL Boolean/`ANY` paths, ensuring the reported class is compatible with `getObject`.

#### P2 — Complete the requested regression coverage

**Problem:** [`assertGetColumnClassName`](https://github.com/apache/shardingsphere/blob/6370fd1e713e2acdca835faef9e081bcf5c10965/kernel/sql-federation/core/src/test/java/org/apache/shardingsphere/sqlfederation/resultset/SQLFederationResultSetMetaDataTest.java#L353-L364) checks only one mocked `VARCHAR` case. The latest-head full-module JaCoCo report shows 1 covered and 11 missed branches in the new classifier. It does not reproduce the reported `BIGINT` path or exercise numeric, temporal, binary, fallback, Java-backed, or dialect-converted behavior.

The [linked issue’s accepted scope](https://github.com/apache/shardingsphere/issues/39121#issuecomment-5004476356) explicitly requests representative numeric, character, temporal, binary, and fallback coverage plus a ShardingSphere-JDBC SQL Federation metadata sentinel.

**Impact:** The current test allowed the confirmed class/value mismatches above and does not prove the root-cause path through the JDBC-facing result set.

**Required Change:** Please add parameterized focused coverage for the requested type groups and one ShardingSphere-JDBC SQL Federation test asserting both `getColumnClassName` and the corresponding `getObject` class/value.

### Review Details

- **Reviewed Scope:** Both changed files in `kernel/sql-federation/core`; latest head `6370fd1e713e2acdca835faef9e081bcf5c10965`; base `master` at `da096439aa116e5f15cebee32a01671ccb3e662c`; local merge-base `80e6541874d47b5814e3d1400c925d6ff3afa0ed`. The local triple-dot file list matched GitHub `/pulls/39217/files`.
- **Not Reviewed Scope:** GitHub Actions, check-runs, and workflow logs; database-backed JDBC/E2E execution; unrelated modules.
- **Verification:** Latest-head focused test: 37 tests passed, exit 0. Full `kernel/sql-federation/core` tests with JaCoCo: 259 tests passed, exit 0; new classifier branches: 1 covered, 11 missed. Module Spotless and Checkstyle checks both passed, exit 0. Required PR, issue, file, comment, and review evidence was accessible; no evidence-access gap affected this result.
- **Release Note / User Docs:** Not required; this is an internal bug fix with no configuration, API/SPI signature, migration, or operational change.

PRAHLAD09-dev:
Thanks for the review. I revisited the implementation based on your feedback and traced how SQLFederationResultSet#getObject() actually obtains its value.

getObject() eventually goes through getValue(), which applies DialectSQLFederationColumnTypeConverter#convertColumnValue(). So getColumnClassName() should describe the Java type after that dialect-specific conversion, rather than relying only on the JDBC type returned by getColumnType().

I updated the implementation accordingly:

Added a dialect hook for reporting the Java class produced by column value conversion.
For MySQL, BOOLEAN values are converted to Integer, so getColumnClassName() now reports java.lang.Integer for that case.
Kept the BigInteger case based on Calcite's original JavaType, since mapping it through JDBC BIGINT would incorrectly report Long even though the actual value remains a BigInteger.
For types without a special runtime conversion, the existing JDBC type-to-Java-class mapping remains the fallback.

I also added regression coverage for the BigInteger case, dialect-converted class behavior, the JDBC type mappings, and MySQL BOOLEAN -> Integer.

I verified the changes with the SQL federation core and MySQL tests, including the reactor build, and they pass successfully.

Updated in commit 4c05f420.

terrymanu (changes requested):
### Summary

**Reason:** The latest patch fixes the original enum-class symptom plus the `BigInteger` and MySQL `BOOLEAN` mismatches, but numeric metadata can still disagree with `getObject`, the new shared SPI uses a forbidden nullable return contract, and the requested JDBC-facing regression sentinel remains absent.

### Issues

#### P1 — Derive numeric class names from the actual SQL Federation result type

**Problem:** [`getColumnClassNameByType`](https://github.com/apache/shardingsphere/blob/801a63c8bf8ec78f86d58ef428eb2de67da9f246/kernel/sql-federation/core/src/main/java/org/apache/shardingsphere/sqlfederation/resultset/SQLFederationResultSetMetaData.java#L202-L238) reports:

- `TINYINT` and `SMALLINT` as `Integer`
- `FLOAT` as `Float`

An exact-head `SQLFederationResultSet` reproduction produced `Byte`, `Short`, and `Double` respectively from `getObject`, while metadata reported `Integer`, `Integer`, and `Float`. The new [unit expectations](https://github.com/apache/shardingsphere/blob/801a63c8bf8ec78f86d58ef428eb2de67da9f246/kernel/sql-federation/core/src/test/java/org/apache/shardingsphere/sqlfederation/resultset/SQLFederationResultSetMetaDataTest.java#L413-L427) encode those incompatible results.

This is consistent with Avatica’s type definitions, which map `TINYINT`, `SMALLINT`, and `FLOAT` to `byte`, `short`, and `double` in its Java row model, and map JDBC `FLOAT` to `Double` for `getObject`. See [Avatica 1.26.0 `SqlType`](https://github.com/apache/calcite-avatica/blob/a4e7987104dfcfd490d76bc30734d4ed92105c02/core/src/main/java/org/apache/calcite/avatica/SqlType.java#L45-L69) and its [JDBC object mapping](https://github.com/apache/calcite-avatica/blob/a4e7987104dfcfd490d76bc30734d4ed92105c02/core/src/main/java/org/apache/calcite/avatica/SqlType.java#L117-L155). The [JDBC contract](https://docs.oracle.com/javase/8/docs/api/java/sql/ResultSetMetaData.html#getColumnClassName-int-) permits `getObject` to return a subclass of the reported class, but `Byte`/`Integer`, `Short`/`Integer`, and `Double`/`Float` have no such relationship.

**Impact:** Frameworks selecting handlers from metadata can still choose an incompatible handler, reproducing the same failure category as #39121 for valid numeric result types.

**Required Change:** Please derive the reported class from the effective Calcite/SQL Federation Java result type, followed by any dialect value conversion, rather than relying solely on this JDBC-type switch. Update the focused tests to compare `getColumnClassName` with the class actually returned by `getObject`, including the `TINYINT`, `SMALLINT`, and `FLOAT` counterexamples.

#### P2 — Remove the nullable absence sentinel from the shared converter SPI

**Problem:** The new [`convertColumnValueClass`](https://github.com/apache/shardingsphere/blob/801a63c8bf8ec78f86d58ef428eb2de67da9f246/kernel/sql-federation/core/src/main/java/org/apache/shardingsphere/sqlfederation/resultset/converter/DialectSQLFederationColumnTypeConverter.java#L46-L54) contract returns `null` for absence, and the [MySQL implementation](https://github.com/apache/shardingsphere/blob/801a63c8bf8ec78f86d58ef428eb2de67da9f246/kernel/sql-federation/dialect/mysql/src/main/java/org/apache/shardingsphere/sqlfederation/mysql/MySQLSQLFederationColumnTypeConverter.java#L45-L51) repeats that sentinel. This directly conflicts with the repository rule that [method return values must not be null](https://github.com/apache/shardingsphere/blob/801a63c8bf8ec78f86d58ef428eb2de67da9f246/CODE_OF_CONDUCT.md#L61-L64).

**Impact:** The shared SPI now exposes an implicit nullable mode and requires every caller and future implementation to duplicate defensive null handling.

**Required Change:** Please use a non-null absence contract, such as `Optional<Class<?>>`, and cover both the converted and no-conversion outcomes in the concrete converter test.

#### P2 — Add the requested JDBC-facing regression sentinel

**Problem:** The linked issue’s accepted scope explicitly requests a ShardingSphere-JDBC SQL Federation test that asserts both the metadata class and corresponding `getObject` value. The latest file list contains only core and converter unit tests, and the new metadata tests construct mocked `RelDataType` and converter inputs without exercising both methods on the same federation result set. See the [maintainer scope in #39121](https://github.com/apache/shardingsphere/issues/39121#issuecomment-5004476356).

**Impact:** The compiler/result-type/converter/result-set integration remains unprotected. The incorrect numeric expectations above demonstrate that isolated mapping tests can pass while the externally visible metadata/value contract is still inconsistent.

**Required Change:** Please add the requested JDBC-facing SQL Federation sentinel, or an equally strong real-entry-path contract test, that obtains metadata and values from the same result set and covers the reported `BIGINT` scenario plus a representative converted or boundary numeric type.

### Positive Feedback

- The `SqlTypeName` enum-class result is removed.
- Java-backed `BigInteger` and MySQL `BOOLEAN → Integer` are now handled consistently.
- Representative character, temporal, binary, fallback, and numeric unit cases were added.

### Multi-Round Comparison

- **Previous P1 — Partially fixed:** The confirmed `BigInteger` and MySQL `BOOLEAN` mismatches are fixed. The earlier `ANY` subpoint is not retained as an independent blocker because a current production result-column path yielding `ANY` was not established. The numeric mismatch above is newly discovered but was already present in the earlier PR revision.
- **Previous P2 — Partially fixed:** Representative unit coverage was added, but the requested JDBC-facing metadata/value sentinel is still missing.
- **Newly introduced:** The nullable shared-SPI contract was added while addressing the previous review.

### Review Details

- **Reviewed Scope:** All five changed files in `kernel/sql-federation/core` and `kernel/sql-federation/dialect/mysql`; latest head `801a63c8bf8ec78f86d58ef428eb2de67da9f246`; base `master` at `da096439aa116e5f15cebee32a01671ccb3e662c`; local merge-base `80e6541874d47b5814e3d1400c925d6ff3afa0ed`. The local triple-dot file list matched GitHub `/pulls/39217/files`.
- **Not Reviewed Scope:** GitHub Actions, check-runs, workflow logs, database-backed JDBC/E2E execution, and unrelated modules.
- **Verification:** Exact-head core metadata test: 53 tests passed, exit 0. An initial standalone MySQL module run exited 1 because the installed core snapshot predated the new SPI method; bytecode inspection confirmed the stale dependency. The exact-head two-module reactor run then passed all 61 focused tests, exit 0. An exact-head same-result-set reproduction confirmed `TINYINT: Byte/Integer`, `SMALLINT: Short/Integer`, and `FLOAT: Double/Float` for `getObject` versus metadata. Authenticated PR, issue, file, comment, review, and commit evidence was accessible; no evidence-access gap affected this result.
- **Release Note / User Docs:** No user-facing release note or configuration documentation is required. The new SPI method does require a non-null contract and matching Javadoc as described above.

terrymanu (changes requested):
### Summary

**Reason:** The latest implementation direction correctly repairs the metadata/value-class mapping, but the regression coverage still does not complete the JDBC-facing and boundary-type scope explicitly accepted in #39121.

### Issues

#### P2 — Complete the accepted JDBC-facing regression coverage

**Problem:** The [maintainer-approved scope](https://github.com/apache/shardingsphere/issues/39121#issuecomment-5004476356) requests representative numeric, character, temporal, binary, and fallback coverage plus a ShardingSphere-JDBC SQL Federation metadata test.

The latest [`SQLFederationResultSetMetaDataTest`](https://github.com/apache/shardingsphere/blob/6ef18117e3725f265dcc77fa9a511dba51d72f5b/kernel/sql-federation/core/src/test/java/org/apache/shardingsphere/sqlfederation/resultset/SQLFederationResultSetMetaDataTest.java#L360-L411) covers `VARCHAR`, `BigInteger`, a mocked converter, and several numeric types, but no decimal, temporal, binary, or fallback result type.

The new [`SQLFederationResultSetTest`](https://github.com/apache/shardingsphere/blob/6ef18117e3725f265dcc77fa9a511dba51d72f5b/kernel/sql-federation/core/src/test/java/org/apache/shardingsphere/sqlfederation/resultset/SQLFederationResultSetTest.java#L588-L613) is useful, but it manually constructs `SQLFederationResultSet`. It does not execute the ShardingSphere-JDBC → SQL Federation compiler/processor → result-type → dialect SPI → result-set path that exposed #39121. No ShardingSphere-JDBC metadata test appears in the authoritative six-file scope.

**Impact:** The externally visible path remains unprotected even though earlier isolated mapping tests accepted incompatible classes. The new Calcite mapping, MySQL converter loading, and actual JDBC result values could drift independently without this test detecting it.

**Required Change:** Please add the requested ShardingSphere-JDBC SQL Federation regression that executes a real federated query and asserts both `getColumnClassName` and the corresponding `getObject` class/value. It should cover the reported MySQL `BIGINT` path and at least one dialect-converted path. Please also add representative decimal, temporal, binary, and fallback cases to the focused coverage.

### Positive Feedback

- [`getColumnClassName`](https://github.com/apache/shardingsphere/blob/6ef18117e3725f265dcc77fa9a511dba51d72f5b/kernel/sql-federation/core/src/main/java/org/apache/shardingsphere/sqlfederation/resultset/SQLFederationResultSetMetaData.java#L184-L193) now derives the class from Calcite’s Java row type and then applies the dialect conversion contract, which matches the JDBC requirement that the reported class describe `getObject` values.
- The implementation agrees with Calcite 1.40.0’s [`JavaTypeFactoryImpl#getJavaClass`](https://github.com/apache/calcite/blob/calcite-1.40.0/core/src/main/java/org/apache/calcite/jdbc/JavaTypeFactoryImpl.java#L303-L380), including `Byte`, `Short`, and `Double` for `TINYINT`, `SMALLINT`, and `FLOAT`.
- The shared SPI now uses a non-null `Optional` absence contract, and MySQL `BOOLEAN → Integer` behavior is covered directly.

### Multi-Round Comparison

- **Previous numeric-class mismatch — Fixed:** The latest implementation and same-result-set test now align `TINYINT`, `SMALLINT`, and `FLOAT` metadata with the actual `Byte`, `Short`, and `Double` values.
- **Previous nullable SPI contract — Fixed:** `convertColumnValueClass` now returns `Optional<Class<?>>`.
- **Previous JDBC-facing sentinel — Partially fixed:** A useful direct `SQLFederationResultSet` contract test was added, but the explicitly requested ShardingSphere-JDBC entry-path test and representative temporal, binary, decimal, and fallback cases remain absent.

### Review Details

- **Reviewed Scope:** All six changed files under `kernel/sql-federation/core` and `kernel/sql-federation/dialect/mysql`; latest head `6ef18117e3725f265dcc77fa9a511dba51d72f5b`; base `master` at `35a7054902d7bf5d241bf9805625b64b536219f0`; merge-base `80e6541874d47b5814e3d1400c925d6ff3afa0ed`. GitHub `/pulls/39217/files` matched the GitHub compare scope exactly. A local triple-dot file list was not recomputed because the latest head object was not present locally; the exact public head source snapshot was reviewed and tested instead.
- **Not Reviewed Scope:** GitHub Actions, check-runs, workflow logs, database-backed E2E execution, and unrelated modules.
- **Verification:** Exact-head `./mvnw -pl kernel/sql-federation/core,kernel/sql-federation/dialect/mysql -DskipITs -Dspotless.skip=true -Djacoco.skip=false test jacoco:report` passed with exit 0: 269 core tests and 10 MySQL tests. `getColumnClassName` had 5 covered and 1 missed branch. Exact-head `./mvnw -pl kernel/sql-federation/core,kernel/sql-federation/dialect/mysql spotless:check checkstyle:check -Pcheck -T1C` passed with exit 0. PR, issue, files, commits, comments, and reviews were accessible; no evidence-access gap affected this result.
- **Release Note / User Docs:** No user-facing release note or configuration documentation is required; this is an internal metadata bug fix with no configuration or migration change. The additive SPI method has a compatibility-preserving default.

terrymanu:
Please resolve checkstyle

terrymanu (changes requested):
### Result

Feedback Mode: Change Request

**Blocking Issues: 1**

The functional fix and regression coverage are correct, but the PR includes an accidental generated build transcript.

### Blocking Issues

#### P2 — Remove the generated `effective-pom.xml`

- **Evidence:** [`effective-pom.xml`](https://github.com/apache/shardingsphere/blob/93dc4b80a57942d60f3ae7ca373a7b20cfb6f6c6/effective-pom.xml#L1-L23) begins with Maven `[INFO]` output before the XML declaration and ends with a build summary. `xmllint --noout effective-pom.xml` fails at line 1.
- **Impact:** The PR adds 3,599 lines of generated, invalid XML containing environment-specific Maven output. It is unrelated to the fix, is not a source of truth, and would pollute the repository/source distribution.
- **Required Change:** Delete `effective-pom.xml`; do not regenerate or commit it.

### Coverage

- Reviewed head: `93dc4b80a57942d60f3ae7ca373a7b20cfb6f6c6`; base: `6f4173cb2782a63040ff8cbd23ee1e8d08671996`; merge-base: `80e6541874d47b5814e3d1400c925d6ff3afa0ed`.
- All nine GitHub-listed files were accounted for. The authoritative file list matched GitHub’s compare scope, and none of the existing affected paths changed independently on the current base.
- Reviewed clusters:
  - Result metadata/value-class resolution and shared SPI contract.
  - MySQL Boolean conversion.
  - Unit and JDBC-facing SQL Federation regression coverage.
  - Test dependency and generated-artifact impact.
- The previous correctness findings are resolved. [`getColumnClassName`](https://github.com/apache/shardingsphere/blob/93dc4b80a57942d60f3ae7ca373a7b20cfb6f6c6/kernel/sql-federation/core/src/main/java/org/apache/shardingsphere/sqlfederation/resultset/SQLFederationResultSetMetaData.java#L184-L194) now follows the JDBC `getObject` contract, using Calcite 1.40’s [Java result-type mapping](https://github.com/apache/calcite/blob/calcite-1.40.0/core/src/main/java/org/apache/calcite/jdbc/JavaTypeFactoryImpl.java#L303-L380) and the dialect conversion SPI. The [JDBC E2E sentinel](https://github.com/apache/shardingsphere/blob/93dc4b80a57942d60f3ae7ca373a7b20cfb6f6c6/test/e2e/driver/src/test/java/org/apache/shardingsphere/test/e2e/driver/federation/SQLFederationResultSetColumnClassNameTest.java#L106-L125) covers BIGINT and MySQL Boolean conversion.
- Exact-head focused verification passed: 122 core tests, 8 MySQL converter tests, and 2 JDBC E2E tests. Scoped Spotless and Checkstyle checks also passed.
- Completed root-cause/behavior, blast-radius/contracts, tests/runtime/operations, latest-delta, and final convergence passes. No unresolved evidence gap could change the blocker set.
- Code Correctness Review only; GitHub Actions, checks, and workflow logs were not reviewed.

terrymanu (approved):
### Result

**Review Result: Mergeable**

The current head correctly repairs the JDBC metadata/value-class contract. No blocking code-correctness issues remain.

### Evidence

- [`getColumnClassName`](https://github.com/apache/shardingsphere/blob/abf1a6cd6c9f2425fad4053606c6d27c0d57142f/kernel/sql-federation/core/src/main/java/org/apache/shardingsphere/sqlfederation/resultset/SQLFederationResultSetMetaData.java#L184-L193) now uses Calcite’s effective Java row type, boxes primitives, and applies dialect-specific value conversion first. This agrees with the [JDBC contract](https://docs.oracle.com/en/java/javase/17/docs/api/java.sql/java/sql/ResultSetMetaData.html#getColumnClassName(int)) and [Calcite 1.40.0 mapping](https://github.com/apache/calcite/blob/calcite-1.40.0/core/src/main/java/org/apache/calcite/jdbc/JavaTypeFactoryImpl.java#L173-L249).
- The [SPI addition](https://github.com/apache/shardingsphere/blob/abf1a6cd6c9f2425fad4053606c6d27c0d57142f/kernel/sql-federation/core/src/main/java/org/apache/shardingsphere/sqlfederation/resultset/converter/DialectSQLFederationColumnTypeConverter.java#L48-L56) is additive and compatibility-preserving; the [MySQL implementation](https://github.com/apache/shardingsphere/blob/abf1a6cd6c9f2425fad4053606c6d27c0d57142f/kernel/sql-federation/dialect/mysql/src/main/java/org/apache/shardingsphere/sqlfederation/mysql/MySQLSQLFederationColumnTypeConverter.java#L47-L50) reports `Integer` for converted Boolean values.
- Focused coverage includes representative numeric, character, decimal, temporal, binary, Java-backed, converted, and fallback types, plus the [JDBC-facing BIGINT/Boolean regression test](https://github.com/apache/shardingsphere/blob/abf1a6cd6c9f2425fad4053606c6d27c0d57142f/test/e2e/driver/src/test/java/org/apache/shardingsphere/test/e2e/driver/federation/SQLFederationResultSetColumnClassNameTest.java#L106-L125).
- Exact-head verification passed: 122 focused core tests, 8 MySQL converter tests, the complete core/MySQL module suites, 2 driver E2E tests, and scoped Spotless/Checkstyle checks. An independent JDBC federation probe confirmed matching metadata/value classes for BIGINT, Boolean, decimal, date, time, timestamp, binary, TINYINT, SMALLINT, and FLOAT.
- Commit [`abf1a6cd`](https://github.com/apache/shardingsphere/commit/abf1a6cd6c9f2425fad4053606c6d27c0d57142f) removes the previously reported generated `effective-pom.xml`.

### Coverage

- Reviewed head: `abf1a6cd6c9f2425fad4053606c6d27c0d57142f`; base: `6f4173cb2782a63040ff8cbd23ee1e8d08671996`; merge-base: `80e6541874d47b5814e3d1400c925d6ff3afa0ed`.
- All eight authoritative GitHub-listed files were accounted for and matched GitHub’s compare scope. The PR head was unavailable as a local Git object, so the exact public-head source archive was reviewed and tested instead of claiming a local triple-dot comparison.
- Reviewed clusters: metadata/value-class resolution, dialect conversion SPI and MySQL behavior, focused regression coverage, JDBC entry-path coverage, and E2E dependency impact.
- Completed root-cause/behavior, blast-radius/contracts, tests/runtime/operations, latest-delta, and final convergence passes.
- Code Correctness Review only; GitHub Actions, checks, and workflow logs were not reviewed.

PR #39298, opened 2026-07-31
Fix Oracle pre-12.2 column case sensitivity

terrymanu:
Use Oracle's default case-sensitive behavior for character columns when COLLATION metadata is unavailable before version 12.2.

Reuse the Oracle data type classification for non-character columns and preserve the existing 12.2+ collation handling.

### DIFF
--- a/database/connector/dialect/oracle/src/main/java/org/apache/shardingsphere/database/connector/oracle/metadata/data/loader/OracleMetaDataLoader.java
+++ b/database/connector/dialect/oracle/src/main/java/org/apache/shardingsphere/database/connector/oracle/metadata/data/loader/OracleMetaDataLoader.java
@@ -29,7 +29,9 @@ import org.apache.shardingsphere.database.connector.core.metadata.data.model.Tab
 import org.apache.shardingsphere.database.connector.core.metadata.database.datatype.DataTypeRegistry;
 import org.apache.shardingsphere.database.connector.core.metadata.database.enums.QuoteCharacter;
 import org.apache.shardingsphere.database.connector.core.metadata.database.enums.TableType;
+import org.apache.shardingsphere.database.connector.core.metadata.database.metadata.option.datatype.DialectDataTypeOption;
 import org.apache.shardingsphere.database.connector.core.type.DatabaseType;
+import org.apache.shardingsphere.database.connector.oracle.metadata.database.option.OracleDataTypeOption;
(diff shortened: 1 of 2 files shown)


terrymanu:
### Result

**Review Result: Mergeable**

The change correctly restores Oracle’s default case-sensitive behavior for pre-12.2 character columns while preserving non-character and 12.2+ collation behavior.

### Evidence

- Reviewed [PR #39298](https://github.com/apache/shardingsphere/pull/39298) at head `3ffa8940707413b65dfa03d68ec47d489ceab130`.
- [OracleMetaDataLoader.java:151](https://github.com/apache/shardingsphere/blob/3ffa8940707413b65dfa03d68ec47d489ceab130/database/connector/dialect/oracle/src/main/java/org/apache/shardingsphere/database/connector/oracle/metadata/data/loader/OracleMetaDataLoader.java#L151-L168) uses Oracle’s existing data-type classification only when collation metadata is unavailable.
- Oracle documents binary, case-sensitive comparison as the 12.1 default and states that per-column `COLLATION` metadata starts in 12.2: [12.1 comparison behavior](https://docs.oracle.com/database/121/NLSPG/ch5lingsort.htm), [12.2 `ALL_TAB_COLUMNS`](https://docs.oracle.com/en/database/oracle/oracle-database/12.2/refrn/ALL_TAB_COLUMNS.html).
- [OracleMetaDataLoaderTest.java:257](https://github.com/apache/shardingsphere/blob/3ffa8940707413b65dfa03d68ec47d489ceab130/database/connector/dialect/oracle/src/test/java/org/apache/shardingsphere/database/connector/oracle/metadata/data/loader/OracleMetaDataLoaderTest.java#L257-L279) covers Oracle 11.2, 12.0, 12.1, non-character columns, and existing 12.2+ collation cases.
- Focused verification passed: 23 tests, 0 failures, 0 errors.

### Coverage

- GitHub’s authoritative two-file list matched the local triple-dot scope from merge-base `1025a5b4b3e770ca36b8470e620ef2329cb769e1`.
- Reviewed the production change and its test as one Oracle metadata case-sensitivity cluster.
- Completed root-cause, consumer blast-radius, compatibility/session-state, test-validity, runtime, and convergence passes.
- No unresolved evidence gaps or blocking findings remain.
- This is a code-scope result only; CI was not reviewed.

taojintianxia (approved):

Issue #1265, opened 2018-09-17
RoundRobinMasterSlaveLoadBalanceAlgorithm throw an ArrayIndexOutOfBoundsException


feelwing1314:
For English only, other languages we will close it directly.

Please answer these questions before submitting your issue. Thanks!

Before submit a new issue, please check existed issue first, to make sure your issue is not a duplicated one.

### Which version of Sharding-Sphere do you using?
3.0.0.M3

### Which project do you using? Sharding-JDBC or Sharding-Proxy?
Sharding-JDBC

### Expected behavior
RoundRobinMasterSlaveLoadBalance always be OK

### Actual behavior
after roundrobin Integer.MAX_VALUE times,  will  throw an ArrayIndexOutOfBoundsException.

### Reason analyze
the AtomicInteger's getAndIncrement() may produce an negative numbers.

PR #39323, opened 2026-08-03
Fix ArrayIndexOutOfBoundsException in RoundRobinLoadBalanceAlgorithm when its counter overflows

### LINKED ISSUE #39319: [Bug] RoundRobinLoadBalanceAlgorithm throws ArrayIndexOutOfBoundsException after its counter overflows
## Bug Report

master (latest)

Both — `RoundRobinLoadBalanceAlgorithm` lives in `infra` and is the default `ROUND_ROBIN` load balancer used by readwrite-splitting.

### Expected behavior
`RoundRobinLoadBalanceAlgorithm.getTargetName()` keeps returning a valid target name for the lifetime of the process.

### Actual behavior
Once the internal counter passes `Integer.MAX_VALUE`, every call throws `ArrayIndexOutOfBoundsException` for any replica group whose size is not a power of two.

```
java.lang.ArrayIndexOutOfBoundsException: Index -2 out of bounds for length 3
    at RoundRobinLoadBalanceAlgorithm.getTargetName(RoundRobinLoadBalanceAlgorithm.java:36)
```

`RoundRobinLoadBalanceAlgorithm.java:36`:

```java
return availableTargetNames.get(Math.abs(count.getAndIncrement()) % availableTargetNames.size());
```

`Math.abs(Integer.MIN_VALUE)` is `Integer.MIN_VALUE` — `abs` cannot represent `2^31` in an `int`, so it returns the negative value unchanged. The remainder is then negative and `List.get` throws:

| group size | `Math.abs(Integer.MIN_VALUE) % size` | |
| --- | --- | --- |
| 2 | 0 | ok |
| 3 | **-2** | throws |
| 4 | 0 | ok |
| 5 | **-3** | throws |
| 8 | 0 | ok |

Power-of-two group sizes happen to yield 0 and survive, which is why a 2-replica setup never shows it. `getTargetName` is annotated `@HighFrequencyInvocation` and the counter is never reset, so a long-lived proxy handling read traffic reaches the wrap on its own.

There is prior art for this exact failure: #1265 reported the same `ArrayIndexOutOfBoundsException` in 2018 and was closed as completed. The fix at the time was a `count.compareAndSet(readDataSourceNames.size(), 0)` line that kept the counter bounded. That line was later removed in #17422 while the counter was being changed from a static map to an instance field, and the `Math.abs(...)` expression was kept. (For the record, that old line was itself racy under concurrency — several threads could step past `size` before any of them observed it — so the point is not that a good guard was lost, but that the current code has no bound on the counter at all.)

Set the counter to `Integer.MAX_VALUE` and call the algorithm twice against a 3-element list:

```java
LoadBalanceAlgorithm algorithm = TypedSPILoader.getService(LoadBalanceAlgorithm.class, "ROUND_ROBIN", new Properties());
Plugins.getMemberAccessor().set(
        RoundRobinLoadBalanceAlgorithm.class.getDeclaredField("count"), algorithm, new AtomicInteger(Integer.MAX_VALUE));
List<String> availableTargetNames = Arrays.asList("ds_1", "ds_2", "ds_3");
algorithm.getTargetName("ds", availableTargetNames);  // ok, index 1
algorithm.getTargetName("ds", availableTargetNames);  // ArrayIndexOutOfBoundsException: Index -2 out of bounds for length 3
```

### Example codes for reproduce this issue (such as a github link)

The snippet above runs as-is inside `RoundRobinLoadBalanceAlgorithmTest`. I have a fix and a regression test ready and will open a PR referencing this issue.

`Math.floorMod(count.getAndIncrement(), availableTargetNames.size())` is a one-line replacement that is identical to the current expression for every non-negative counter value — I verified that exhaustively for all counters in `[0, 5_000_000]` across group sizes 1..16 — and returns a valid index after the wrap instead of throwing.

PDGGK:
## Changes

`RoundRobinLoadBalanceAlgorithm.getTargetName` indexes with `Math.abs(count.getAndIncrement()) % availableTargetNames.size()`. `Math.abs(Integer.MIN_VALUE)` is `Integer.MIN_VALUE` — `abs` cannot represent `2^31` in an `int` — so once the counter wraps, the remainder is negative and `List.get` throws.

```
java.lang.ArrayIndexOutOfBoundsException: Index -2 out of bounds for length 3
    at RoundRobinLoadBalanceAlgorithm.getTargetName(RoundRobinLoadBalanceAlgorithm.java:36)
```

Replaced with `Math.floorMod(count.getAndIncrement(), availableTargetNames.size())`.

Only replica groups whose size is not a power of two are affected — for 2, 4 or 8 the negative remainder happens to be 0:

| group size | `Math.abs(Integer.MIN_VALUE) % size` | |
| --- | --- | --- |
| 2 | 0 | survives |
| 3 | **-2** | throws |
| 4 | 0 | survives |
| 5 | **-3** | throws |
| 8 | 0 | survives |

The existing test uses a two-element list, which is why this was never caught.

## Why `floorMod` is safe below the wrap

`Math.floorMod(i, size)` and `Math.abs(i) % size` are identical for every non-negative `i`, so nothing changes until the counter overflows. I checked that exhaustively rather than assuming it — all counters in `[0, 5_000_000]` against group sizes 1..16, 80,000,016 pairs, zero differences. The only behavioural change is at the wrap itself, where one target is selected twice in a row instead of an exception being thrown.

## Prior art

#1265 reported this same `ArrayIndexOutOfBoundsException` in 2018 and was closed as completed; the fix then was a `count.compareAndSet(readDataSourceNames.size(), 0)` line that kept the counter bounded. #17422 removed that line in 2022 while changing the counter from a static map to an instance field, and kept the `Math.abs(...)` expression.

To be accurate about it: that old line was itself racy — several threads could step past `size` before any of them observed it — so the point is not that a correct guard was lost, but that the counter has had no bound at all since then, and `getTargetName` is annotated `@HighFrequencyInvocation`.

## Testing

Added `assertGetAvailableTargetNameWhenCounterOverflows`, which seeds `count` with `Integer.MAX_VALUE` using the repo's `Plugins.getMemberAccessor()` idiom and calls the algorithm three times against a three-element list. It reproduces the exception on `master` and passes with this change.

- `RoundRobinLoadBalanceAlgorithmTest` — fails before, 2/2 after
- `infra/algorithm/type/load-balancer/type/round-robin` + `features/readwrite-splitting/core` with `-am` — 3404 tests, 0 failures, 0 errors
- `checkstyle:check` with `src/resources/checkstyle.xml` — clean

Fixes #39319

## Type
- [x] Bugfix

### DIFF
--- a/infra/algorithm/type/load-balancer/type/round-robin/src/main/java/org/apache/shardingsphere/infra/algorithm/loadbalancer/round/robin/RoundRobinLoadBalanceAlgorithm.java
+++ b/infra/algorithm/type/load-balancer/type/round-robin/src/main/java/org/apache/shardingsphere/infra/algorithm/loadbalancer/round/robin/RoundRobinLoadBalanceAlgorithm.java
@@ -33,7 +33,7 @@ public final class RoundRobinLoadBalanceAlgorithm implements LoadBalanceAlgorith
     @HighFrequencyInvocation
     @Override
     public String getTargetName(final String groupName, final List<String> availableTargetNames) {
-        return availableTargetNames.get(Math.abs(count.getAndIncrement()) % availableTargetNames.size());
+        return availableTargetNames.get(Math.floorMod(count.getAndIncrement(), availableTargetNames.size()));
     }
     
     @Override
--- a/infra/algorithm/type/load-balancer/type/round-robin/src/test/java/org/apache/shardingsphere/infra/algorithm/loadbalancer/round/robin/RoundRobinLoadBalanceAlgorithmTest.java
+++ b/infra/algorithm/type/load-balancer/type/round-robin/src/test/java/org/apache/shardingsphere/infra/algorithm/loadbalancer/round/robin/RoundRobinLoadBalanceAlgorithmTest.java
@@ -20,10 +20,12 @@ package org.apache.shardingsphere.infra.algorithm.loadbalancer.round.robin;
 import org.apache.shardingsphere.infra.algorithm.loadbalancer.spi.LoadBalanceAlgorithm;
 import org.apache.shardingsphere.infra.spi.type.typed.TypedSPILoader;
 import org.junit.jupiter.api.Test;
+import org.mockito.internal.configuration.plugins.Plugins;
 
 import java.util.Arrays;
 import java.util.List;
 import java.util.Properties;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import static org.hamcrest.Matchers.is;
 import static org.hamcrest.MatcherAssert.assertThat;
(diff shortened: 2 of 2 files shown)


PDGGK:
Heads-up on the one red check, `E2E - Agent with MySQL (jdbc, tracing, jaeger)` — it is a runner-side download failure, not this change.

It failed in the `Build Project` step before any source was compiled:

```
wget: Failed to fetch https://repo.maven.apache.org/maven2/org/apache/maven/apache-maven/3.9.14/apache-maven-3.9.14-bin.zip
##[error]Process completed with exit code 1.
```

Maven itself could not be downloaded, so nothing from this branch was built or run in that job. The other eleven `E2E - Agent with MySQL` shards are green, as are `CI`, `Check - CheckStyle`, `Check - Spotless` and `Check - License`.

I do not have permission to re-run the job. Happy to push an empty commit to re-trigger if that is the convention here, but I did not want to churn the branch unnecessarily.

For what it is worth, the diff is one expression in `RoundRobinLoadBalanceAlgorithm` plus one test method, and the load balancer is not on the observability-agent path.

terrymanu (commented):
Thanks for the contribution. This fix has already been merged into `master` via #39339, including the `Math.floorMod` change and a regression test for the counter overflow scenario.

Closing this PR as superseded by #39339.

PDGGK:
Thanks for confirming — checked `master` and the `floorMod` change plus the overflow test are both in. Good to have it closed off.

Issue #31481, opened 2024-05-31
Support parsing Doris sql


iamhucong:
## Background

Hi community.\
The ShardingSphere SQL parser engine helps users to parse SQL to create the AST (Abstract Syntax Tree) and visit the AST to get SQLStatement (Java Object). Currently, we are planning to enhance the support for Doris SQL parsing in ShardingSphere.

More details:
<https://shardingsphere.apache.org/document/current/en/reference/sharding/parse/>  

## Issue Background Explanation

The current issue involves using a custom web scraping script to fetch SQL cases from the official website of a corresponding database. These SQL cases are then passed to ShardingSphere's parsing engine for analysis. For SQL cases that fail to be parsed successfully, every 3 to 5 SQL cases are grouped together as an issue.
1. Since  SQL cases are obtained through web scraping, it cannot be guaranteed that all SQL cases are correct. Please follow the following process to handle this pull request (PR).
2. Some SQL cases may have already been fixed in other PRs. For cases that can already be executed successfully, simply leave a comment to ignore them.
3. If a SQL case can be executed successfully without any code changes, there is no need to add a corresponding test assert file.  

## Task

This issue is to support more Doris sql parsing, as follows:

```
LAG(expr, offset, default) OVER (partition_by_clause order_by_clause)
```

- [ ] [link](https://doris.apache.org/zh-CN/docs/sql-manual/sql-functions/window-functions/window-function-lag/)

```
WINDOW,FUNCTION,LAG
```

- [ ] [link](https://doris.apache.org/zh-CN/docs/sql-manual/sql-functions/window-functions/window-function-lag/)

```
BITXOR
```

- [ ] [link](https://doris.apache.org/zh-CN/docs/sql-manual/sql-functions/bitwise-functions/bitxor/)

```
ALTER WORKLOAD GROUP  "rg_name" PROPERTIES (     property_list )
```

- [ ] [link](https://doris.apache.org/zh-CN/docs/sql-manual/sql-statements/Data-Definition-Statements/Alter/ALTER-WORKLOAD-GROUP/)

```
ALTER, WORKLOAD , GROUP
```

- [ ] [link](https://doris.apache.org/zh-CN/docs/sql-manual/sql-statements/Data-Definition-Statements/Alter/ALTER-WORKLOAD-GROUP/)

```
ALTER DATABASE db_name SET DATA QUOTA quota
```

[...]

PR #39333, opened 2026-08-04
Support more Doris SQL parsing (ALTER WORKLOAD GROUP, quota with unit)

divyankshah:
## What's changed

Closes the remaining gaps from #31481. `LAG`, `BITXOR` and most of the listed `ALTER DATABASE` variants (`RENAME`, `SET DATA/REPLICA QUOTA`) already parsed correctly and had test coverage, so this PR only adds the two pieces that were actually missing:

1. **`ALTER WORKLOAD GROUP name PROPERTIES(...)`** - new grammar rule modeled on the existing `ALTER STORAGE POLICY ... PROPERTIES(...)` rule, plus the corresponding statement/segment classes and visitor logic.
2. **`ALTER DATABASE ... SET DATA QUOTA 100G`** (unit-suffixed quota values) - the `QUOTA` clause now reuses the grammar's existing `fileSizeLiteral` rule (already used elsewhere for `SIZE 100M`-style literals) instead of requiring a bare number, and the visitor converts the K/M/G/T suffix into bytes.

## Test plan

- Added `.sql` cases and expected-output assertions for both features, plus a regression case for the existing quota-without-unit syntax.
- Ran the full parser IT suite across all dialects (10k+ cases): 0 failures, 0 errors.


terrymanu (changes requested):
### Result

Feedback Mode: Change Request

**Blocking Issues: 1**

The quota change does not preserve Doris’s type-specific unit contract.

### Blocking Issues

#### 1. Keep data-volume units separate from replica and transaction quotas

- **Evidence:** [`DDLStatement.g4:330`](https://github.com/divyankshah/shardingsphere/blob/3dc73d9a6eb917e5be15be6e381ae973efce222c/parser/sql/engine/dialect/doris/src/main/antlr4/imports/doris/DDLStatement.g4#L326-L331) applies `fileSizeLiteral` to `DATA`, `REPLICA`, and `TRANSACTION`; [`DorisDDLStatementVisitor.java:369`](https://github.com/divyankshah/shardingsphere/blob/3dc73d9a6eb917e5be15be6e381ae973efce222c/parser/sql/engine/dialect/doris/src/main/java/org/apache/shardingsphere/sql/parser/engine/doris/visitor/statement/type/DorisDDLStatementVisitor.java#L369-L409) converts suffixes for every quota type. The reused lexer accepts only uppercase `K/M/G/T`. Doris 2.0 instead accepts `B`, `K/KB`, `M/MB`, `G/GB`, `T/TB`, and `P/PB` case-insensitively for data quotas, while replica and transaction quotas use plain integers ([`ParseUtil.java:29`](https://github.com/apache/doris/blob/432e71ab1a07cca0940509ca666dc67cd9ad1bf5/fe/fe-core/src/main/java/org/apache/doris/common/util/ParseUtil.java#L29-L96), [`AlterDatabaseQuotaStmt.java:76`](https://github.com/apache/doris/blob/432e71ab1a07cca0940509ca666dc67cd9ad1bf5/fe/fe-core/src/main/java/org/apache/doris/analysis/AlterDatabaseQuotaStmt.java#L76-L81)).
- **Impact:** Valid Doris statements such as `SET DATA QUOTA 1GB`, `1P`, or lowercase units are rejected, while invalid statements such as `SET REPLICA QUOTA 1G` are accepted and converted into a misleading numeric AST value.
- **Required Change:** Give `DATA QUOTA` a Doris-compatible, case-insensitive data-volume literal and conversion path; retain unsuffixed numeric values for `REPLICA` and `TRANSACTION`. Add regression cases covering the accepted unit forms and the type-specific boundary.

### Coverage

- Reviewed head: `3dc73d9a6eb917e5be15be6e381ae973efce222c`.
- Base: `master` at `959db03e1b23d36daca2fe4e03a874dee50a45f9`; merge base `2ccb352f8a634535aaeb905aaa868b7e4e6ecdf5`.
- All 15 authoritative changed files were accounted for; GitHub’s paginated file list matched its base-to-head comparison.
- Requirements checked: [issue #31481](https://github.com/apache/shardingsphere/issues/31481), Doris `ALTER DATABASE`, and [Doris 2.0 `ALTER WORKLOAD GROUP`](https://doris.apache.org/docs/2.0/sql-manual/sql-reference/Data-Definition-Statements/Alter/ALTER-WORKLOAD-GROUP).
- Behavior clusters reviewed: workload-group grammar/AST/test wiring; quota grammar/conversion/regression coverage.
- Root-cause, blast-radius/contracts, and tests/runtime/operations lenses completed. MySQL trunk equivalents were checked; no shared grammar file was modified.
- Unresolved gaps: none affecting the blocker set. Candidate Maven tests were not rerun because the public head was unavailable in local refs; the finding is established directly from the current public source contract.
- Code-scope review only; CI was not queried or reviewed.

divyankshah:
Hi @terrymanu, 

Thanks for the detailed review, this was a real bug and good catch which I missed.

Fixed in the latest commit. DATA QUOTA now goes through its own literal that matches Doris's actual analyzeDataVolumn behavior: B, K/KB, M/MB, G/GB, T/TB, P/PB, case insensitive. REPLICA and TRANSACTION quotas are back to plain integers only, matching analyzeReplicaNumber/analyzeTransactionNumber.

Added regression cases for the two letter units, petabyte, lowercase, and bare byte forms. I also checked directly against the parser that SET REPLICA QUOTA 1G and SET TRANSACTION QUOTA 1G now fail to parse as expected, since that's the boundary you flagged. I didn't add it as an automated negative test case since the unsupported SQL test harness is disabled for every dialect in this repo right now, not just Doris, so there's no active place to wire it in without touching unrelated infrastructure.

Let me know if you want that boundary covered any other different way.

terrymanu (changes requested):
### Correction

- Previous Finding: DATA quota units were incorrectly shared with replica and transaction quotas.
- Current Status: Withdrawn
- Reason: The latest head separates `DATA` from `REPLICA`/`TRANSACTION`, implements Doris-compatible unit conversion, and preserves integer-only replica and transaction quotas ([grammar](https://github.com/divyankshah/shardingsphere/blob/4e009a4c0c3a78f13bc73969f3eb9e93d99c5e17/parser/sql/engine/dialect/doris/src/main/antlr4/imports/doris/DDLStatement.g4#L326-L339), [visitor](https://github.com/divyankshah/shardingsphere/blob/4e009a4c0c3a78f13bc73969f3eb9e93d99c5e17/parser/sql/engine/dialect/doris/src/main/java/org/apache/shardingsphere/sql/parser/engine/doris/visitor/statement/type/DorisDDLStatementVisitor.java#L369-L424)).

### Result

Feedback Mode: Change Request

**Blocking Issues: 1**

The latest quota fix introduces a Doris-wide lexer regression for digit-leading identifiers.

### Blocking Issues

#### 1. Preserve identifiers that resemble data-quota literals

- **Evidence:** [`DATA_QUOTA_LITERAL`](https://github.com/divyankshah/shardingsphere/blob/4e009a4c0c3a78f13bc73969f3eb9e93d99c5e17/parser/sql/engine/dialect/doris/src/main/antlr4/imports/doris/Literals.g4#L22-L29) now captures values such as `1GB`, but the general [`identifier` rule](https://github.com/divyankshah/shardingsphere/blob/4e009a4c0c3a78f13bc73969f3eb9e93d99c5e17/parser/sql/engine/dialect/doris/src/main/antlr4/imports/doris/BaseRule.g4#L117-L128) does not accept that token. Doris 2.0’s scanner deliberately recognizes digit-leading names as identifiers, and its expression grammar admits them as column references ([scanner](https://github.com/apache/doris/blob/432e71ab1a07cca0940509ca666dc67cd9ad1bf5/fe/fe-core/src/main/jflex/sql_scanner.flex#L630-L633), [parser](https://github.com/apache/doris/blob/432e71ab1a07cca0940509ca666dc67cd9ad1bf5/fe/fe-core/src/main/cup/sql_parser.cup#L7146-L7149)). A compiled snapshot probe confirmed `SELECT 1GB FROM t` parses on the base as `IDENTIFIER_` with zero syntax errors, but the PR head lexes `1GB` as `DATA_QUOTA_LITERAL` and reports one syntax error.
- **Impact:** Previously accepted Doris column, table, alias, and other identifier uses matching the new quota-token pattern can no longer be parsed. This regression was introduced by the latest commit.
- **Required Change:** Restrict quota-literal handling to the quota path or otherwise preserve matching tokens as identifiers outside that path. Add a regression case such as `SELECT 1GB FROM t` while retaining the accepted DATA units and rejected unit-suffixed replica/transaction cases.

### Coverage

- Reviewed head: `4e009a4c0c3a78f13bc73969f3eb9e93d99c5e17`.
- Base: `master` at `959db03e1b23d36daca2fe4e03a874dee50a45f9`; merge base `2ccb352f8a634535aaeb905aaa868b7e4e6ecdf5`.
- All 16 authoritative changed files were accounted for; GitHub’s file list matched its base-to-head comparison.
- Requirements checked: [issue #31481](https://github.com/apache/shardingsphere/issues/31481), [Doris 2.0 ALTER WORKLOAD GROUP](https://doris.apache.org/docs/2.0/sql-manual/sql-reference/Data-Definition-Statements/Alter/ALTER-WORKLOAD-GROUP/), Doris quota parsing source, and existing LAG/BITXOR/ALTER DATABASE coverage.
- Behavior clusters reviewed: workload-group grammar/AST/assertion wiring; data-quota grammar/conversion/boundaries; latest-commit lexer blast radius.
- Root-cause, blast-radius/contracts, tests/runtime/operations, and final convergence passes completed. MySQL’s separate grammar and relevant repository documentation were checked.
- Verification: Doris parser package succeeded; `InternalDorisParserIT` passed 1,269 tests with zero failures or errors; focused lexer/parser probes confirmed the quota boundaries and identifier regression.
- Unresolved gaps: none affecting the blocker set.
- Code-scope review only; CI was not queried or reviewed.

divyankshah:
@terrymanu 

Fixed at root cause. The lexer token I added competed globally with the identifier token, since
ANTLR lexer rules aren't scoped to a parser context. I removed it. The quota rule now takes the
identifier token directly and parses the unit in the visitor, so 'Literals.g4' is unchanged from
master. The only lexer addition left on this branch is the WORKLOAD keyword, registered in
'identifierKeywordsUnambiguous', with a test covering workload as a column, table and
qualified name.

While auditing I found a bug of my own: the unit multiplication overflowed silently, so
'SET DATA QUOTA 99999999999P' produced '-6918654927547924480'. It now uses Math.multiplyExact
and reports out of range. Quota values also went straight to Long.parseLong on master, so
1.5 escaped as a raw NumberFormatException. That predates this PR, but I tightened it while
rewriting the method: both paths validate first and raise SQLParsingException, matching
parseIntegerToken.

Correcting something I said earlier: I claimed there was no active place for the boundary test.
That was wrong. 'DorisStatementVisitorTest' runs in CI, and the unsupported harness is live for
PostgreSQL, SQL92, SQLServer and Firebird. Doris just had no runner class, so its existing
unsupported.xml entry had never executed. I added the runner. The boundary is now covered in
both places.

Full parser suite passes at 10,256 tests, with checkstyle, RAT and spotless clean. Binder,
optimizer and SQL federation suites pass as a blast radius check.

One deliberate omission: SET DATA QUOTA 0 parses to 0 though Doris rejects values at or below
zero during analysis. I treated value range as semantic validation rather than parsing, happy to
enforce it here if you'd prefer.

terrymanu (dismissed):
### Result

**Review Result: Mergeable**

Both previously blocking findings are fixed and independently re-verified at the latest head `a0d33d4a3a907ca92d717123978450327d2f02fa`; the newer commits (quota overflow/malformed-value validation, the Doris unsupported-SQL runner, and the workload-keyword-as-identifier regression tests) introduce no new correctness issues. This is a code-scope review only; CI was not queried.

### Evidence

**Prior blocking issue 1 (quota unit contract, withdrawn by the maintainer at 4e009a4): Fixed.**
`parser/sql/engine/dialect/doris/src/main/antlr4/imports/doris/DDLStatement.g4:329-343` now separates `SET DATA QUOTA dataQuotaValue` from `SET (REPLICA | TRANSACTION) QUOTA NUMBER_`; the unit table in `DorisDDLStatementVisitor.java:414-437` covers `B`, `K/KB`, `M/MB`, `G/GB`, `T/TB`, `P/PB` case-insensitively, matching the Doris `ParseUtil.analyzeDataVolumn` behavior cited by the maintainer. Replica and transaction quotas remain plain integers, so `SET REPLICA QUOTA 1G` is rejected at the grammar level again.

**Prior blocking issue 2 (lexer hijacking digit-leading identifiers): Fixed.**
The head's changed-file list does not include `Literals.g4` (identical to master); the `DATA_QUOTA_LITERAL` token was removed, and `dataQuotaValue : FILESIZE_LITERAL | NUMBER_ | IDENTIFIER_` consumes the identifier token directly with unit parsing moved into the visitor. The baseline `IDENTIFIER_` rule (`Literals.g4:72`) already admits digit-leading tokens, so `SELECT 1GB FROM t` keeps its base behavior, and the regression case `select_digit_leading_column_doris` is added in both `sql/supported/dml/select.xml:26` and `case/dml/select.xml`.

**Latest-delta review, item by item (all pass the Finding Proof Gate with no blocker):**
- Overflow and malformed-value handling (`89b2694`): `Math.multiplyExact` plus the all-digit check throw `SQLParsingException`; the `(String, Object, int)` constructor is confirmed at `parser/sql/engine/core/.../exception/SQLParsingException.java:39`. `1.5` is rejected at the grammar level because `DECIMAL_NUM_` is not among the `dataQuotaValue` alternatives, consistent with the REPLICA path.
- Doris unsupported runner (`a2a2331`): `InternalUnsupportedDorisParserIT` matches the eight existing runner precedents (PostgreSQL, MySQL, Oracle, etc.) exactly; the harness asserts `assertThrows(SQLParsingException.class, parse(...))`. Enabling it activates the pre-existing Doris entry `unsupported_select_json_value_on_error_for_doris`, which is statically confirmed to still fail parsing: `BaseRule.g4:1018` allows only one `onEmptyError?` clause while that SQL contains both `null on empty` and `default 0 on error`.
- WORKLOAD keyword (`BaseRule.g4:582-584`, `DorisKeyword.g4:3319-3324`): adding it to `identifierKeywordsUnambiguous` does not conflict with the explicit token sequence of `alterWorkloadGroup`, and `DorisStatementVisitorTest.java` covers workload as a column name, a qualified name, and a table name.
- `alterWorkloadGroup : ALTER WORKLOAD GROUP identifier propertiesClause` exactly mirrors the existing `alterStoragePolicy` precedent (`DDLStatement.g4:168`); `identifier` includes `DOUBLE_QUOTED_TEXT`/`BQUOTA_STRING`, covering both the Doris documented `"rg_name"` and the unquoted `g1` forms. The new public classes mirror the `DorisAlterColocateGroupStatement`/`ColocateGroupSegment` precedents.
- The `visitAlterDatabase` branch restructure is behavior-equivalent for old scenarios: an all-digit DATA quota flows through `parseQuotaNumber` to the same value as the baseline `Long.parseLong`, and the RENAME/PROPERTIES branches are unaffected.
- Case indexes were hand-verified character by character (the three cases in `case/ddl/alter-workload-group.xml` match their SQL strings exactly); the `case/` and `sql/unsupported/` registries scan directories, so the new XML files are auto-loaded; the JAXB element `alter-workload-group` follows the `alter-colocate-group` precedent with no cross-dialect collision.
- The test file's static import order is reversed relative to repository convention, but `src/resources/checkstyle.xml` contains only `UnusedImports` with no import-order rule, so this is not an issue.

### Coverage

- Reviewed head: `a0d33d4a3a907ca92d717123978450327d2f02fa`; base is master@`959db03e1b23d36daca2fe4e03a874dee50a45f9` (local master contains that commit; baseline rules were read at that commit).
- Authoritative requirements: issue #31481 (this PR covers its ALTER WORKLOAD GROUP and unit-suffixed DATA QUOTA items; LAG/BITXOR/RENAME and the other variants were confirmed already covered by the maintainer's earlier rounds, and this re-review found no regression in the diff).
- All 20 authoritative changed files are accounted for, mapped to 4 behavior clusters: workload-group grammar/AST/assert/case wiring; DATA quota units/conversion/boundaries plus the unsupported runner; digit-leading identifier regression protection; WORKLOAD keyword as identifier. No churn-only files.
- All three discovery lenses (root cause and behavior, blast radius and contracts, tests/runtime/operations) are complete; a convergence pass over the latest commit (`a0d33d4`, test-only) produced zero new independent candidates.
- Dialect family: only the Doris branch grammar directory is modified; the MySQL trunk grammar is untouched; `statement/core` receives only one new additive segment class with no existing shared file changed.
- Official evidence: Doris ALTER-WORKLOAD-GROUP documentation page (syntax and the `properties("key"="value")` example); quota units per the apache/doris `ParseUtil.java`/`AlterDatabaseQuotaStmt.java` sources cited by the maintainer.
- Verification status: the head is not in local refs (Git is read-only; no fetch), so Maven was not rerun locally; all decisive facts were established statically from base sources, the diff, and official documentation, and the author publicly reports the full 10,256-case suite passing. The ledger/inventory scripts required a local head ref, so the GitHub authoritative file list was used instead with manual accounting.
- Unresolved gaps: none affecting the blocker set. This result is a Code Correctness Review, code-scope only; CI was not queried or reviewed.

terrymanu:
Please resolve conflicts and I will merge

divyankshah:
@terrymanu

Conflicts resolved by merging master.

terrymanu (approved):
### Result

**Review Result: Mergeable**

The current head `a0764d7` correctly implements both remaining #31481 gaps — `ALTER WORKLOAD GROUP ... PROPERTIES(...)` and unit-suffixed `SET DATA QUOTA` — with both previously blocking findings still fixed, no regression in adjacent Doris paths, and valid focused tests. This is a code-scope review only; CI was not queried.

### Evidence

**Prior blocker 1 (quota unit contract) — still fixed.**
`parser/sql/engine/dialect/doris/src/main/antlr4/imports/doris/DDLStatement.g4` splits `SET DATA QUOTA dataQuotaValue` from `SET (REPLICA | TRANSACTION) QUOTA NUMBER_`. The unit table in `DorisDDLStatementVisitor.java` (`getDataQuotaUnitBytes`) accepts `B`, `K/KB`, `M/MB`, `G/GB`, `T/TB`, `P/PB` case-insensitively and converts to bytes with `Math.multiplyExact`; replica/transaction quotas remain integer-only and are now rejected at the grammar level when unit-suffixed. This matches the official Doris behavior: the [ALTER DATABASE reference](https://doris.apache.org/docs/3.x/sql-manual/sql-statements/database/ALTER-DATABASE/) documents quota units `B/K/KB/M/MB/G/GB/T/TB/P/PB` plus plain byte counts, and integer replica quotas. Verified arithmetic of all five new expected values in `test/it/parser/src/main/resources/case/ddl/alter-database.xml` (e.g. `100G` → 107374182400, `1P` → 1125899906842624, `500B` → 500).

**Prior blocker 2 (lexer hijacking digit-leading identifiers) — still fixed.**
The head's changed-file list contains no `Literals.g4` change; `dataQuotaValue : FILESIZE_LITERAL | NUMBER_ | IDENTIFIER_` consumes existing tokens and unit parsing lives in the visitor. Since `FILESIZE_LITERAL` only matches `INT_NUM_ ('K'|'M'|'G'|'T')`, forms like `1GB`, `500B`, `100g` lex as baseline `IDENTIFIER_` (which admits digit-leading tokens), so ordinary identifiers are untouched. The IT guards `select_digit_leading_column_doris` (`SELECT 1GB FROM t`) and `select_workload_keyword_column_doris` cover this.

**ALTER WORKLOAD GROUP matches official Doris syntax.**
The [Doris 2.1 SQL manual](https://doris.apache.org/docs/2.1/sql-manual/sql-statements/cluster-management/compute-management/ALTER-WORKLOAD-GROUP/) defines `ALTER WORKLOAD GROUP <name> PROPERTIES (<key> = <value>, ...)`; the new rule `ALTER WORKLOAD GROUP identifier propertiesClause` reuses the shared `propertiesClause` rule (also used by master's `CREATE WORKLOAD GROUP` from #39518) and the existing `extractPropertiesSegment` helper. The new statement, segment, visitor method, JAXB expected types, assert, and `RootSQLParserTestCases` registration all follow the maintained `DorisAlterColocateGroupStatement` siblings. Start/stop indices in `alter-workload-group.xml` were hand-verified against the SQL texts.

**Quota validation hardening is safe and contained.**
`parseQuotaNumber`/`getDataQuotaValue` replace the raw `Long.parseLong` (which leaked `NumberFormatException` for `1.5` and silently wrapped overflow on master) with `SQLParsingException`, matching the existing `DorisPlayKeywordIdentifierTest` engine-test pattern; the 3-arg `SQLParsingException` constructor exists in the engine module. `1.5` lexes as a single `NUMBER_` token, so the rejection path is reachable. The only consumers of `AlterDatabaseStatement` quota fields are the parser visitor and IT asserts — no downstream engine impact.

**New unsupported-SQL runner is sound.**
`InternalUnsupportedDorisParserIT` follows the existing runner pattern. Besides the two new quota boundary cases, it activates exactly one previously dead Doris entry, `unsupported_select_json_value_on_error_for_doris`; traced against the head grammar, `onEmptyError : (NULL | ERROR | DEFAULT literals) ON (EMPTY | ERROR)` is a single optional occurrence, so the double `null on empty default 0 on error` clause genuinely fails to parse and the case remains correctly unsupported.

**Post-review merge delta (the only unreviewed-before-now changes).**
The two trailing merge commits with master preserve every PR feature relative to merge-base `743996c`. One conflict resolution is worth knowing, though not blocking: master's #39463 deleted `DorisStatementVisitorTest.java`, and the merge accepted the deletion, dropping that file's workload-as-identifier unit tests (column/table/qualified-name). The behavior they guarded is master-owned (WORKLOAD token and its identifier registration come from master's #39518; this PR no longer touches them), the column case survives as the `select_workload_keyword_column_doris` IT, and the quota validation tests survive in `DorisAlterDatabaseQuotaValueTest`, so no PR-owned coverage of owned behavior was lost. `DorisKeyword.g4` is now a churn-only change (two `DORIS ADDED` comment markers around the master-existing WORKLOAD rule).

### Coverage

Reviewed head `a0764d733f44`, base `master` at merge-base `743996c9e75`; authoritative scope is the GitHub changed-file list (19 files, +511/−6), all accounted for. Local refs do not contain the head (fetch not authorized), so no local triple-dot reproduction or ledger script was possible; the GitHub file list plus per-file patches served as the authoritative basis and the mechanical accounting was completed as a manual checklist (fallback recorded). Requirements: PR description, linked issue #31481 remaining gaps, and the maintainer's round-1 unit-contract boundary. Behavior clusters: (1) ALTER WORKLOAD GROUP parsing, (2) DATA quota units plus quota validation hardening, (3) unsupported-SQL enforcement, (4) lexer-regression guards for identifier behavior — every changed file mapped to at least one cluster, `DorisKeyword.g4` classified churn-only. All three discovery lenses (root cause/behavior, blast radius/contracts, tests/runtime/operations) completed; convergence pass over the post-`a0d33d4a` merge delta found no new independent candidate. Related-dialect check: changes are Doris-dialect-only; core statement model and shared grammar untouched; sibling Doris statement/assert patterns inspected for consistency. Official documentation used: Doris 2.1 ALTER WORKLOAD GROUP and 3.x ALTER DATABASE references. No unresolved evidence gap could change the blocker set; runtime verification is static (grammar/lexer/constructor checks and hand-verified expected values) since the head could not be built locally. CI was not reviewed; this result is code-scope only.

PR #39364, opened 2026-08-06
Allow position 0 for from-x and to-y in mask algorithms

### LINKED ISSUE #39362: MASK_FROM_X_TO_Y and KEEP_FROM_X_TO_Y reject the documented from-x and to-y value 0

master @ 99a0b47786f

Both

### Expected behavior

`from-x: 0` and `to-y: 0` are accepted by `MASK_FROM_X_TO_Y` and `KEEP_FROM_X_TO_Y`. Both `mask.en.md` and `mask.cn.md` define them as 0-based: `from-x | int | start position (from 0)`, `to-y | int | end position (from 0)`.

### Actual behavior

Initialization fails with `AlgorithmInitializationException: from-x must be a positive integer.`, so masking a value from its very first character cannot be configured at all.

`MaskFromXToYMaskAlgorithm.createFromX()/createToY()` (`features/mask/core/src/main/java/org/apache/shardingsphere/mask/algorithm/cover/MaskFromXToYMaskAlgorithm.java:55,60`) and `KeepFromXToYMaskAlgorithm` (same lines) call `MaskAlgorithmPropertiesChecker.checkPositiveInteger()`, which asserts `integerValue > 0` (`MaskAlgorithmPropertiesChecker.java:70`). That checker is right for the count-typed `first-n`/`last-m`, but `from-x`/`to-y` are positions and must allow 0. The runtime is already 0-based: `MaskFromXToYMaskAlgorithm.mask():80` uses `fromX` directly as a char array index.

```sql
CREATE MASK RULE t_user (
COLUMNS((NAME=phone, TYPE(NAME='MASK_FROM_X_TO_Y', PROPERTIES('from-x'='0', 'to-y'='2', 'replace-char'='*'))))
);
```

The equivalent YAML `!MASK` rule fails at startup for the same reason.

N/A

thswlsqls:
Fixes #39362.

  - `MASK_FROM_X_TO_Y` and `KEEP_FROM_X_TO_Y` reject `from-x: 0` / `to-y: 0` with `AlgorithmInitializationException`, although `mask.en.md` and `mask.cn.md` both document them as `start position (from 0)` / `end position (from 0)`.
  - The runtime is already 0-based: `MaskFromXToYMaskAlgorithm.mask()` uses `fromX` directly as a char array index, so position 0 works once initialization accepts it.
  - Add `MaskAlgorithmPropertiesChecker.checkNonNegativeInteger()` and switch the four position-typed call sites (`from-x`, `to-y` in both algorithms) to it.
  - Keep `checkPositiveInteger()` unchanged: `first-n` / `last-m` in `KeepFirstNLastMMaskAlgorithm` and `MaskFirstNLastMMaskAlgorithm` are counts, not positions, so `> 0` stays correct for them.
  - Negative values are still rejected; the existing `negative_from_X` / `negative_to_Y` init cases still pass.
  - Add checker unit tests for zero, positive, negative, missing key and non-integer input, plus `from-x=0` execute cases to both algorithm tests.
  - No documentation change: the docs already describe the intended contract, the code did not honour it.
  - Verified with `./mvnw install -pl features/mask/core -am` plus repository-wide spotless, checkstyle and apache-rat: 159 tests pass.

---

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -55,6 +55,7 @@
 1. Sharding: Fix AUTO_INTERVAL sharding failure under JVM default locales that use comma decimal separators - [#38806](https://github.com/apache/shardingsphere/pull/38806)
 1. Sharding: Compute the Snowflake key generator epoch in UTC instead of the JVM default timezone - [#38932](https://github.com/apache/shardingsphere/pull/38932)
 1. SQL Federation: Fix SQL Federation pagination binding for long LIMIT parameters - [#39237](https://github.com/apache/shardingsphere/pull/39237)
+1. Mask: Allow position 0 for from-x and to-y in MASK_FROM_X_TO_Y and KEEP_FROM_X_TO_Y - [#39364](https://github.com/apache/shardingsphere/pull/39364)
 
 ### Enhancements
 
--- a/features/mask/core/src/main/java/org/apache/shardingsphere/mask/algorithm/MaskAlgorithmPropertiesChecker.java
+++ b/features/mask/core/src/main/java/org/apache/shardingsphere/mask/algorithm/MaskAlgorithmPropertiesChecker.java
@@ -73,6 +73,24 @@ public final class MaskAlgorithmPropertiesChecker {
         }
     }
     
+    /**
+     * Check non-negative integer.
+     *
+     * @param props properties to be checked
+     * @param propKey properties key to be checked
+     * @param algorithm mask algorithm
+     * @throws AlgorithmInitializationException algorithm initialization exception
+     */
+    public static void checkNonNegativeInteger(final Properties props, final String propKey, final MaskAlgorithm<?, ?> algorithm) {
+        checkRequired(props, propKey, algorithm);
+        try {
+            int integerValue = Integer.parseInt(props.getProperty(propKey));
+            ShardingSpherePreconditions.checkState(integerValue >= 0, () -> new AlgorithmInitializationException(algorithm, "%s must be a non-negative integer.", propKey));
+        } catch (final NumberFormatException ex) {
+            throw new AlgorithmInitializationException(algorithm, "%s must be a valid integer number", propKey);
+        }
+    }
+    
     private static void checkRequired(final Properties props, final String requiredPropKey, final MaskAlgorithm<?, ?> algorithm) {
         ShardingSpherePreconditions.checkContainsKey(props, requiredPropKey, () -> new AlgorithmInitializationException(algorithm, "%s is required", requiredPropKey));
     }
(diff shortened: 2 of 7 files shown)


terrymanu (changes requested):
### Result

Feedback Mode: Change Request

**Blocking Issues: 1**

The production fix matches the documented zero-based contract, but the regression coverage does not protect `to-y=0`.

### Blocking Issues

1. `to-y=0` is not exercised through either algorithm’s public path.

   - **Evidence:** The new cases use `from-x=0, to-y=2` in both [KeepFromXToYMaskAlgorithmTest](https://github.com/apache/shardingsphere/blob/48f46805ffb59a245ba0dab76f49e115fdb5e200/features/mask/core/src/test/java/org/apache/shardingsphere/mask/algorithm/cover/KeepFromXToYMaskAlgorithmTest.java#L116-L127) and [MaskFromXToYMaskAlgorithmTest](https://github.com/apache/shardingsphere/blob/48f46805ffb59a245ba0dab76f49e115fdb5e200/features/mask/core/src/test/java/org/apache/shardingsphere/mask/algorithm/cover/MaskFromXToYMaskAlgorithmTest.java#L114-L125). The direct checker tests prove that the helper accepts zero, but not that either `createToY` call site uses it. In a controlled mutation, reverting both `to-y` validations to `checkPositiveInteger` still left all 59 affected tests passing.
   - **Impact:** A partial implementation or later regression can restore the reported `to-y=0` initialization failure while the complete affected test set remains green. This leaves part of [issue #39362](https://github.com/apache/shardingsphere/issues/39362) unprotected.
   - **Required Change:** Add public-path cases for both algorithms using `from-x=0, to-y=0`, asserting the masking result—for example, `KEEP_FROM_X_TO_Y("abc") -> "a**"` and `MASK_FROM_X_TO_Y("abc") -> "*bc"`.

### Coverage

- Reviewed head: `48f46805ffb59a245ba0dab76f49e115fdb5e200`
- Base: `ba53784e1721b9582ed00713797c9074b69c33a6`; merge-base: `99a0b47786fd854dcb0f6325b102f492579ef3d3`
- All seven authoritative GitHub files were accounted for across validation, runtime masking, regression tests, and release-note behavior clusters.
- Completed root-cause, blast-radius/contract, tests/runtime/operations, and final convergence passes.
- Exact reconstructed head: all seven changed files matched GitHub; mask-core tests passed (`159` tests, exit `0`).

terrymanu:
Hi @thswlsqls, thanks for your contribution and the detailed analysis of this issue.

Unfortunately, this change has been superseded: #39389, which addresses the same issue (#39362), was already merged into `master` on Aug 18 (merge commit `efd97c0`). The two PRs are semantically identical — both add a non-negative integer check to `MaskAlgorithmPropertiesChecker` and switch the `from-x` / `to-y` call sites in `MASK_FROM_X_TO_Y` and `KEEP_FROM_X_TO_Y` to it, with equivalent test coverage.

Since the underlying issue is now resolved, I'm closing this PR to avoid duplicate changes. The only delta not covered by #39389 is the release note entry, which we can add separately referencing #39389 if needed.

Thanks again for reporting and fixing this — feel free to pick up another issue anytime!

thswlsqls:
@terrymanu Thanks for the review and for the pointer to #39389 — closing makes sense, the two changes are equivalent. One thing carried over: the `to-y=0` public-path gap you flagged here still applies on master, where both algorithm tests use `from-x=0, to-y=2`, so reverting `createToY` to `checkPositiveInteger` would stay green. Happy to send a small follow-up adding the `to-y=0` cases and the release note if that's useful.

PR #39310, opened 2026-08-02
Fix Oracle column binding for unparenthesized function names

terrymanu:

PR #39385, opened 2026-08-08
Add SYSTEM_USER to PostgreSQL unparenthesized function names

### LINKED ISSUE #39384: PostgreSQL dialect omits SYSTEM_USER from unparenthesized function names

master @ d26c87dc

Both (kernel-level binder)

### Expected behavior

Under the PostgreSQL dialect, the niladic function `SYSTEM_USER` should bind as a function, like the sibling user-value functions `CURRENT_USER`, `SESSION_USER`, and `USER`. PostgreSQL 16 added `SYSTEM_USER` as a SQL standard keyword; it returns the authenticated identity and the authentication method used to establish the session.

### Actual behavior

`SYSTEM_USER` is missing from `PostgreSQLFunctionOption.UNPARENTHESIZED_FUNCTION_NAMES`, so the binder misclassifies it as a column reference and fails with:

```
ColumnNotFoundException: Unknown column 'SYSTEM_USER' in 'field list'.
```

`PostgreSQLFunctionOption` (database/connector/dialect/postgresql/.../metadata/database/option/PostgreSQLFunctionOption.java lines 31-33) lists 11 names and omits `SYSTEM_USER`.

`ColumnSegmentBinder.isUnparenthesizedFunction()` (infra/binder/core/.../expression/type/ColumnSegmentBinder.java line 149) reads this set via `.contains()` to decide function vs column. When the lookup fails, `bind()` throws at line 102.

`MySQLFunctionOption` and `SQLServerFunctionOption` already include `SYSTEM_USER`.

Unlike SQL92, the PostgreSQL grammar does not declare `SYSTEM_USER` as a keyword token, so it already lexes as an identifier and parses as a column projection. The failure is at bind time only, and no grammar change is required.

Bind `SELECT SYSTEM_USER` against a PostgreSQL-typed schema; the identifier is treated as a column rather than a niladic function.

### Additional notes

Two points worth a maintainer's decision:

1. `OpenGaussDatabaseMetaData#getFunctionOption` returns `new PostgreSQLFunctionOption()`, so openGauss inherits this change. openGauss forked from PostgreSQL 9.2, which predates `SYSTEM_USER` by roughly a decade. Should openGauss get its own `OpenGaussFunctionOption`, following the existing `OpenGaussSchemaOption` and `OpenGaussDataTypeOption`? I have kept the fix minimal and not done this.

2. `infra/binder/core` has only `shardingsphere-database-connector-hive` on its test classpath, so binder tests naming any other dialect silently resolve to `DefaultFunctionOption`, whose unparenthesized name set is empty. A test for this bug cannot fail correctly without adding the dialect. Adding `postgresql` at test scope follows `infra/common` and `kernel/sql-federation/executor`.

Related: #39099 is the same gap in the SQL92 dialect.

viminizer:
Fixes #39384

## Problem

PostgreSQL 16 added the SQL standard `SYSTEM_USER` niladic function, but `PostgreSQLFunctionOption` does not list it. `ColumnSegmentBinder` therefore classifies a bare `SYSTEM_USER` as a column reference and throws `ColumnNotFoundException: Unknown column 'SYSTEM_USER' in 'field list'.`

`MySQLFunctionOption` and `SQLServerFunctionOption` already include it.

## Changes

- Add `SYSTEM_USER` to `PostgreSQLFunctionOption`, and the matching assertion to `PostgreSQLFunctionOptionTest`.
- Add a binder test covering `SYSTEM_USER` in a PostgreSQL projection.
- Add `shardingsphere-database-connector-postgresql` at test scope to `infra/binder/core`.
- Add a parser IT case for `SELECT SYSTEM_USER` under PostgreSQL.

## On the pom change

`infra/binder/core` previously had only `shardingsphere-database-connector-hive` on its test classpath. Any binder test naming another dialect silently resolved to `DefaultFunctionOption`, whose unparenthesized name set is empty, so a test for this bug could not fail correctly without the dialect present. This follows the same test-scope pattern already used in `infra/common` and `kernel/sql-federation/executor`.

## No grammar change needed

Unlike SQL92 (#39099 / #39102), the PostgreSQL grammar does not declare `SYSTEM_USER` as a keyword token, so it already lexes as an identifier and parses as a column projection. The added parser IT case passes without touching any `.g4` file, and documents that difference.

## Open question

`OpenGaussDatabaseMetaData#getFunctionOption` returns `new PostgreSQLFunctionOption()`, so openGauss inherits `SYSTEM_USER` from this change. openGauss forked from PostgreSQL 9.2, which predates the function. Should openGauss get its own `OpenGaussFunctionOption`, following the existing `OpenGaussSchemaOption` and `OpenGaussDataTypeOption`? I kept this PR minimal and did not do it; happy to add it if preferred.

## Verification

- `./mvnw test -pl infra/binder/core,database/connector/dialect/postgresql` — 385 tests, 0 failures
- `./mvnw test -pl test/it/parser -Dtest=InternalPostgreSQLParserIT` — 1152 tests, 0 failures
- `./mvnw checkstyle:check -Pcheck` and `./mvnw spotless:check -Pcheck` on both modules — clean
- Binder test confirmed failing on master before the fix

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -37,6 +37,7 @@
 1. SQL Parser: Fix No value specified for parameter exception when sql is 'INSERT INTO tableName ON CONFLICT  DO UPDATE set  WHERE ' - [#38668](https://github.com/apache/shardingsphere/pull/38668)
 1. SQL Binder: Add DialectFunctionOption to handle wrong skip column bind in ColumnSegmentBinder - [#38350](https://github.com/apache/shardingsphere/pull/38350)
 1. SQL Binder: Fix wrong bind info when order by refer column from with temporary table - [#38353](https://github.com/apache/shardingsphere/pull/38353)
+1. SQL Binder: Bind PostgreSQL SYSTEM_USER as a niladic function instead of a column - [#39385](https://github.com/apache/shardingsphere/pull/39385)
 1. Metadata: Fix MySQL metadata loading fallback when JDBC catalog is null for named tables - [#38855](https://github.com/apache/shardingsphere/pull/38855)
 1. Metadata: Fix Oracle metadata version comparison skipping identity and collation columns on 18c and later - [#39104](https://github.com/apache/shardingsphere/pull/39104)
 1. Metadata: Fix wrong logic table metadata when config same actual table name in different storage unit - [#39157](https://github.com/apache/shardingsphere/pull/39157)
--- a/database/connector/dialect/postgresql/src/main/java/org/apache/shardingsphere/database/connector/postgresql/metadata/database/option/PostgreSQLFunctionOption.java
+++ b/database/connector/dialect/postgresql/src/main/java/org/apache/shardingsphere/database/connector/postgresql/metadata/database/option/PostgreSQLFunctionOption.java
@@ -30,7 +30,7 @@ public final class PostgreSQLFunctionOption implements DialectFunctionOption {
     
     private static final Collection<String> UNPARENTHESIZED_FUNCTION_NAMES = new CaseInsensitiveSet<>(Arrays.asList(
             "CURRENT_CATALOG", "CURRENT_DATE", "CURRENT_ROLE", "CURRENT_SCHEMA", "CURRENT_TIME",
-            "CURRENT_TIMESTAMP", "CURRENT_USER", "LOCALTIME", "LOCALTIMESTAMP", "SESSION_USER", "USER"));
+            "CURRENT_TIMESTAMP", "CURRENT_USER", "LOCALTIME", "LOCALTIMESTAMP", "SESSION_USER", "SYSTEM_USER", "USER"));
     
     @Override
     public Collection<String> getUnparenthesizedFunctionNames() {
--- a/database/connector/dialect/postgresql/src/test/java/org/apache/shardingsphere/database/connector/postgresql/metadata/database/option/PostgreSQLFunctionOptionTest.java
+++ b/database/connector/dialect/postgresql/src/test/java/org/apache/shardingsphere/database/connector/postgresql/metadata/database/option/PostgreSQLFunctionOptionTest.java
@@ -37,6 +37,7 @@ class PostgreSQLFunctionOptionTest {
         assertTrue(functionOption.getUnparenthesizedFunctionNames().contains("LOCALTIME"));
         assertTrue(functionOption.getUnparenthesizedFunctionNames().contains("LOCALTIMESTAMP"));
         assertTrue(functionOption.getUnparenthesizedFunctionNames().contains("SESSION_USER"));
+        assertTrue(functionOption.getUnparenthesizedFunctionNames().contains("SYSTEM_USER"));
         assertTrue(functionOption.getUnparenthesizedFunctionNames().contains("USER"));
     }
 }
--- a/infra/binder/core/pom.xml
+++ b/infra/binder/core/pom.xml
@@ -62,5 +62,11 @@
             <version>${project.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.shardingsphere</groupId>
+            <artifactId>shardingsphere-database-connector-postgresql</artifactId>
+            <version>${project.version}</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 </project>
--- a/infra/binder/core/src/test/java/org/apache/shardingsphere/infra/binder/engine/segment/dml/expression/type/ColumnSegmentBinderTest.java
+++ b/infra/binder/core/src/test/java/org/apache/shardingsphere/infra/binder/engine/segment/dml/expression/type/ColumnSegmentBinderTest.java
@@ -377,4 +377,20 @@ class ColumnSegmentBinderTest {
         ColumnSegment actual = ColumnSegmentBinder.bind(columnSegment, SegmentType.PROJECTION, bindContext, LinkedHashMultimap.create(), LinkedHashMultimap.create());
         assertThat(actual, is(columnSegment));
     }
+    
+    @Test
+    void assertBindUnparenthesizedSystemUserWithTableContextForPostgreSQL() {
+        Multimap<CaseInsensitiveString, TableSegmentBinderContext> tableBinderContexts = LinkedHashMultimap.create();
+        ColumnSegment boundOrderIdColumn = new ColumnSegment(0, 0, new IdentifierValue("order_id"));
+        boundOrderIdColumn.setColumnBoundInfo(new ColumnSegmentBoundInfo(new TableSegmentBoundInfo(new IdentifierValue("foo_db"), new IdentifierValue("foo_schema")),
+                new IdentifierValue("t_order"), new IdentifierValue("order_id"), TableSourceType.PHYSICAL_TABLE));
+        tableBinderContexts.put(CaseInsensitiveString.of("t_order"),
+                new SimpleTableSegmentBinderContext(Collections.singleton(new ColumnProjectionSegment(boundOrderIdColumn)), TableSourceType.PHYSICAL_TABLE));
+        SelectStatement selectStatement = mock(SelectStatement.class);
+        when(selectStatement.getDatabaseType()).thenReturn(TypedSPILoader.getService(DatabaseType.class, "PostgreSQL"));
+        SQLStatementBinderContext bindContext = new SQLStatementBinderContext(mock(ShardingSphereMetaData.class), "foo_db", new HintValueContext(), selectStatement);
+        ColumnSegment columnSegment = new ColumnSegment(0, 0, new IdentifierValue("SYSTEM_USER"));
+        ColumnSegment actual = ColumnSegmentBinder.bind(columnSegment, SegmentType.PROJECTION, bindContext, tableBinderContexts, LinkedHashMultimap.create());
+        assertThat(actual, is(columnSegment));
+    }
 }
--- a/infra/binder/core/src/test/java/org/apache/shardingsphere/infra/binder/engine/segment/dml/expression/type/ColumnSegmentBinderTest.java
+++ b/infra/binder/core/src/test/java/org/apache/shardingsphere/infra/binder/engine/segment/dml/expression/type/ColumnSegmentBinderTest.java
@@ -367,4 +367,14 @@ class ColumnSegmentBinderTest {
         assertTrue(actual.getOwner().isPresent());
         assertThat(actual.getOwner().get().getIdentifier().getValue(), is("EXCLUDED"));
     }
+    
+    @Test
+    void assertBindUnparenthesizedSystemUserForPostgreSQL() {
+        SelectStatement selectStatement = mock(SelectStatement.class);
+        when(selectStatement.getDatabaseType()).thenReturn(TypedSPILoader.getService(DatabaseType.class, "PostgreSQL"));
+        SQLStatementBinderContext bindContext = new SQLStatementBinderContext(mock(ShardingSphereMetaData.class), "foo_db", new HintValueContext(), selectStatement);
+        ColumnSegment columnSegment = new ColumnSegment(0, 0, new IdentifierValue("SYSTEM_USER"));
+        ColumnSegment actual = ColumnSegmentBinder.bind(columnSegment, SegmentType.PROJECTION, bindContext, LinkedHashMultimap.create(), LinkedHashMultimap.create());
+        assertThat(actual, is(columnSegment));
+    }
 }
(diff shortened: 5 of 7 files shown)


terrymanu (changes requested):
### Result

**Blocking Issues: 1**

The change recognizes `SYSTEM_USER` only when no matching table column is found. That conflicts with PostgreSQL 16 syntax and can affect routing correctness.

### Blocking Issues

1. **PostgreSQL 16 `SYSTEM_USER` can still be bound as a table or sharding column**
   - **Evidence:** PostgreSQL 16 defines unquoted `SYSTEM_USER` as a reserved, unparenthesized system function; quoted identifiers remain identifiers ([system functions](https://www.postgresql.org/docs/16/functions-info.html), [keyword classification](https://www.postgresql.org/docs/16/sql-keywords-appendix.html), [quoted identifiers](https://www.postgresql.org/docs/16/sql-syntax-lexical.html)). The PR deliberately parses it as a [`ColumnProjectionSegment`](https://github.com/viminizer/shardingsphere/blob/e35f50ae658efe460852aa35524c48e7e28b0a25/test/it/parser/src/main/resources/case/dml/select.xml#L15114-L15118). [`ColumnSegmentBinder`](https://github.com/apache/shardingsphere/blob/d26c87dc8cac45d910264dba727d54eb105d8532/infra/binder/core/src/main/java/org/apache/shardingsphere/infra/binder/engine/segment/dml/expression/type/ColumnSegmentBinder.java#L94-L103) resolves a case-insensitive matching table column before consulting the function-name set. The added binder test covers only the no-table-context case.
   - **Impact:** If a table has a quoted column such as `"system_user"`, an unquoted predicate like `SYSTEM_USER = 'x'` means the session function to PostgreSQL 16, but ShardingSphere can bind it to that physical column. [`WhereClauseShardingConditionEngine`](https://github.com/apache/shardingsphere/blob/d26c87dc8cac45d910264dba727d54eb105d8532/features/sharding/core/src/main/java/org/apache/shardingsphere/sharding/route/engine/condition/engine/WhereClauseShardingConditionEngine.java#L113-L125) may then derive an incorrect sharding condition and prune valid routes.
   - **Discussion Needed:** Establish version-correct semantics before merging: PostgreSQL 16 unquoted `SYSTEM_USER` must not bind to a physical column, while quoted `"SYSTEM_USER"` must remain a column. PostgreSQL 15 compatibility also needs consideration because it does not provide this function. Add a regression covering the same-named-column and routing boundary once that contract is decided.

### Coverage

- Reviewed public head `e35f50ae658efe460852aa35524c48e7e28b0a25`; all 7 changed files accounted for.
- Inspected parser, binder, metadata-option, routing, PostgreSQL-version, openGauss-sharing, dependency, test, and release-note effects.
- XML validation passed for the changed POM and parser case files.
- Code correctness review only; CI was not reviewed.

viminizer:
I built both branches and measured this against PostgreSQL 16. You are right about the semantics.

### PostgreSQL behaviour

PostgreSQL 16.14:

```
CREATE TABLE t (id int, system_user text);     -- ERROR: syntax error at or near "system_user"
CREATE TABLE t (id int, "system_user" text);   -- ok
INSERT INTO t VALUES (1, 'x');

SELECT system_user FROM t;                     -- the function
SELECT * FROM t WHERE system_user = 'x';       -- 0 rows, though the column holds 'x'
SELECT t.system_user FROM t;                   -- x, the column
SELECT "system_user" FROM t;                   -- x, the column

SELECT word, catcode FROM pg_get_keywords() WHERE word = 'system_user';   -- system_user | R
```

On PostgreSQL 15.18 the same script behaves differently: `system_user` is not a keyword, the unquoted column is legal, and `SELECT system_user FROM t` returns the column. So only the bare unqualified form changed meaning between 15 and 16.

### What this PR changes

Proxy built from `master` and from this branch. PostgreSQL 16.14 with scram authentication, so `SYSTEM_USER` is `scram-sha-256:postgres`. `t_order` sharded over two tables on a `"system_user"` column, 4 rows in the first shard and 6 in the second. `t_order_all` holds the same 10 rows unsharded as the reference.

| query | PostgreSQL 16 | master | this PR |
|---|---|---|---|
| `SELECT SYSTEM_USER` | `scram-sha-256:postgres` | `ERROR: Unknown column 'SYSTEM_USER' in 'field list'.` | `scram-sha-256:postgres` |
| `SELECT SYSTEM_USER FROM t_order` | function | function | function |
| `SELECT "system_user" FROM t_order` | column | column | column |
| `... WHERE SYSTEM_USER = 'x'` | 0 | 0 | 0 |
| `... WHERE SYSTEM_USER = 'scram-sha-256:postgres'` | 10 | 4 | 4 |
| `... WHERE "system_user" = 'owner_3'` | 1 | 1 | 1 |
| `SELECT count(*) FROM t_order` | 10 | 10 | 10 |
| `... WHERE "system_user" = SYSTEM_USER` | 0 | 0 | 0 |

Only the first row differs between `master` and this PR.

### The routing case

It happens, and it happens the same way on master:

```
Logic SQL:  SELECT count(*) FROM t_order WHERE SYSTEM_USER = 'scram-sha-256:postgres';
Actual SQL: ds_0 ::: SELECT count(*) FROM t_order_0 WHERE SYSTEM_USER = 'scram-sha-256:postgres';
```

One shard instead of two, so 4 rows instead of 10.

It does not come from the new entry. `isUnparenthesizedFunction` is only checked when no column matched:

```java
if (!inputColumnSegment.isPresent() && isUnparenthesizedFunction(segment, parentSegmentType, binderContext)) {
    return segment;
}
```

When a `"system_user"` column exists, the column matches and this branch never runs. That is why master returns the same 4 rows. The list has one consumer, `ColumnSegmentBinder`, so table names cannot be affected by it either.

Two things limit it. The proxy sends the predicate to PostgreSQL unchanged, so when the predicate is false every shard returns nothing and reading one shard gives the same answer as reading both. And the natural shape does not prune at all:

```
Logic SQL:  SELECT count(*) FROM t_order WHERE "system_user" = SYSTEM_USER;
Actual SQL: ds_0 ::: SELECT count(*) FROM t_order_0 WHERE "system_user" = SYSTEM_USER
                     UNION ALL SELECT count(*) FROM t_order_1 WHERE "system_user" = SYSTEM_USER
```

`ConditionValueGeneratorFactory.generate` needs a literal or a parameter to build a sharding value, and `SYSTEM_USER` is neither, so the query goes to both shards.

The only query I found that returns a wrong answer is the one above, and it is artificial: it compares `SYSTEM_USER` against a hardcoded copy of its own value. If you know one an application would actually send, I will test it.

### Test coverage

You are right that the binder test only covered the no-table case. I added one where a table is in scope without a matching column, which is the more realistic query. The case where the table does have a `"system_user"` column depends on which contract you pick, so I left it out.

### Deciding the contract

Adding `SYSTEM_USER` to the grammar, as `CURRENT_USER` already is, would give correct PostgreSQL 16 semantics and close the routing case. The cost is PostgreSQL 15 and openGauss: one grammar serves every server version, so reserving the word removes a column name that works today. Keeping the resolution in the binder is what makes the change safe for those versions, because a real column always wins.

A version-aware option needs plumbing that does not exist. The proxy reads the storage node version at startup and keeps it as a display string in `database/protocol/core`, which only the proxy uses, while the binder is in `infra` and shared with ShardingSphere-JDBC. Some metadata loaders call `getDatabaseMajorVersion()` on the live connection, so the value is reachable in both modes, but nothing stores it on `ShardingSphereDatabase` or `StorageUnit`. And `getFunctionOption()` is resolved from `DatabaseType` alone. Making it version-aware means storing the version on the database metadata and changing `DialectFunctionOption`, which has ten implementations.

So:

1. Merge this as it stands. It fixes `SELECT SYSTEM_USER` and changes nothing else. A same-named column keeps winning, exactly as on master, and I file that as its own issue.
2. Put `SYSTEM_USER` in the grammar instead. PostgreSQL 16 becomes correct in every case, including the routing one, but a `system_user` column stops working on PostgreSQL 15 and openGauss.
3. Carry the server version into the database metadata first, then let the function list depend on it. Correct on every version, but it is a much larger change and this PR waits for it.

Tell me which one you want and I will add the matching test, since the expected result is different in each.

terrymanu:
Please resolve conflicts

viminizer:
Rebased onto `master` at 0d6b991878d, conflicts resolved.

Both conflicts were in the parser test resources, where `master` had grown the same lists:

- `test/it/parser/src/main/resources/sql/supported/dml/select.xml`: `master` widened `select_ltrim_function` to `db-types="MySQL,Doris"` and added `select_element_at` and `select_array_subscript`. Kept all of those and placed `select_system_user_postgresql` above them.
- `test/it/parser/src/main/resources/case/dml/select.xml`: `master` appended the Chinese-identifier cases at the tail. Kept them and moved the `select_system_user_postgresql` case after them.

No change to the fix itself. The diff against `master` is unchanged in substance: `SYSTEM_USER` added to `UNPARENTHESIZED_FUNCTION_NAMES`, its unit test, the binder test with table context, the parser test case, and the release note.

Verified locally on the rebased head: `InternalPostgreSQLParserIT` 1177 tests pass, `ColumnSegmentBinderTest` 20 pass, `PostgreSQLFunctionOptionTest` 1 passes.

terrymanu (approved):
### Correction

**Previous Finding:** PostgreSQL 16 `SYSTEM_USER` could still bind to a same-named column and cause incorrect routing.

**Current Status:** Withdrawn

**Reason:** [`ColumnSegmentBinder`](https://github.com/apache/shardingsphere/blob/cff829f28691777a909009788dbb8f95d3521fde/infra/binder/core/src/main/java/org/apache/shardingsphere/infra/binder/engine/segment/dml/expression/type/ColumnSegmentBinder.java#L106-L109) resolves a matching column before consulting the function option. This PR only adds a set member, so that precedence and routing behavior are unchanged. The author’s PostgreSQL 16 reproduction also observed identical same-column routing on `master` and this PR.

### Result

**Review Result: Mergeable**

The current head correctly fixes the reported bare `SELECT SYSTEM_USER` binding failure without introducing a confirmed regression.

### Evidence

- PostgreSQL 16 documents `system_user` as a niladic system-information function and classifies it as reserved. PostgreSQL 15 does not provide the function. [PostgreSQL 16 system information functions](https://www.postgresql.org/docs/16/functions-info.html), [PostgreSQL 16 keywords](https://www.postgresql.org/docs/16/sql-keywords-appendix.html), [PostgreSQL 15 system information functions](https://www.postgresql.org/docs/15/functions-info.html).
- [`PostgreSQLFunctionOption`](https://github.com/apache/shardingsphere/blob/cff829f28691777a909009788dbb8f95d3521fde/database/connector/dialect/postgresql/src/main/java/org/apache/shardingsphere/database/connector/postgresql/metadata/database/option/PostgreSQLFunctionOption.java#L31-L37) now recognizes `SYSTEM_USER` case-insensitively.
- The binder regression tests cover both an empty table context and a table without a matching column. Existing binder coverage preserves the deliberate column-first behavior established by [#39310](https://github.com/apache/shardingsphere/pull/39310).
- PostgreSQL 15 and openGauss compatibility was traced: valid same-named columns still bind first; only already-invalid missing-identifier cases can move from binder rejection to backend rejection.
- Current-head verification passed:
  - `PostgreSQLFunctionOptionTest`: 1 test.
  - `ColumnSegmentBinderTest`: 20 tests.
  - `InternalPostgreSQLParserIT`: 1,177 tests.
  - XML validation for the changed POM and parser resources.

### Coverage

- Candidate: public PR [#39385](https://github.com/apache/shardingsphere/pull/39385).
- Effective head: `cff829f28691777a909009788dbb8f95d3521fde`.
- Base and merge-base: `0d6b991878dec031cf13077dbf192c339ba3fb77`.
- The authoritative pull-files and compare endpoints agreed on all seven changed files; every materialized file matched its GitHub blob SHA.
- Files accounted for:
  - `RELEASE-NOTES.md`
  - `database/connector/dialect/postgresql/.../PostgreSQLFunctionOption.java`
  - `database/connector/dialect/postgresql/.../PostgreSQLFunctionOptionTest.java`
  - `infra/binder/core/pom.xml`
  - `infra/binder/core/.../ColumnSegmentBinderTest.java`
  - `test/it/parser/.../case/dml/select.xml`
  - `test/it/parser/.../sql/supported/dml/select.xml`
- Behavior clusters reviewed: function metadata and binder resolution; test-classpath provisioning; parser representation; release-note consistency.
- Dialects checked: PostgreSQL and the related openGauss implementation that reuses `PostgreSQLFunctionOption`.
- Root-cause, blast-radius/contracts, tests/runtime/operations, and final convergence passes completed with no unresolved gaps.
- Mandatory local style verification:
  - Modules: `database/connector/dialect/postgresql,infra/binder/core`.
  - `spotless:check checkstyle:check -Pcheck -T1C`: exit `0`.
  - Spotless covered the three changed Java files and `infra/binder/core/pom.xml`.
  - Checkstyle covered all three changed Java files.
  - Spotless was not applicable to `RELEASE-NOTES.md` or the parser XML resources; repository configuration covers Java, POMs, and GitHub workflow YAML.
  - Checkstyle was not applicable to the Markdown, POM, or parser XML files; its configured source scope is Java.
- Code-correctness review only. GitHub Actions and CI status were not reviewed; the completed local Checkstyle and Spotless verification is independent of CI.

PR #39389, opened 2026-08-09
[ISSUE-39362] Fix from-x and to-y rejecting 0 as a valid position

### LINKED ISSUE #39362: MASK_FROM_X_TO_Y and KEEP_FROM_X_TO_Y reject the documented from-x and to-y value 0

YALI58:
### Bug Description

`from-x` and `to-y` in `MASK_FROM_X_TO_Y` and `KEEP_FROM_X_TO_Y` are 0-based positions, but the validation uses `checkPositiveInteger` which rejects 0. This prevents masking from the very first character.

### Fix

1. Added `checkNotNegativeInteger` to `MaskAlgorithmPropertiesChecker` that allows 0.
2. Changed `MaskFromXToYMaskAlgorithm` and `KeepFromXToYMaskAlgorithm` to use `checkNotNegativeInteger` for `from-x` and `to-y`.

Closes #39362

### DIFF
--- a/features/mask/core/src/main/java/org/apache/shardingsphere/mask/algorithm/MaskAlgorithmPropertiesChecker.java
+++ b/features/mask/core/src/main/java/org/apache/shardingsphere/mask/algorithm/MaskAlgorithmPropertiesChecker.java
@@ -73,6 +73,24 @@ public final class MaskAlgorithmPropertiesChecker {
         }
     }
     
+    /**
+     * check not negative integer.
+     *
+     * @param props properties to be checked
+     * @param propKey properties key to be checked
+     * @param algorithm mask algorithm
+     * @throws AlgorithmInitializationException algorithm initialization exception
+     */
+    public static void checkNotNegativeInteger(final Properties props, final String propKey, final MaskAlgorithm<?, ?> algorithm) {
+        checkRequired(props, propKey, algorithm);
+        try {
+            int integerValue = Integer.parseInt(props.getProperty(propKey));
+            ShardingSpherePreconditions.checkState(integerValue >= 0, () -> new AlgorithmInitializationException(algorithm, "%s must be a not negative integer.", propKey));
+        } catch (final NumberFormatException ex) {
+            throw new AlgorithmInitializationException(algorithm, "%s must be a valid integer number", propKey);
+        }
+    }
+    
     private static void checkRequired(final Properties props, final String requiredPropKey, final MaskAlgorithm<?, ?> algorithm) {
         ShardingSpherePreconditions.checkContainsKey(props, requiredPropKey, () -> new AlgorithmInitializationException(algorithm, "%s is required", requiredPropKey));
     }
--- a/features/mask/core/src/main/java/org/apache/shardingsphere/mask/algorithm/cover/KeepFromXToYMaskAlgorithm.java
+++ b/features/mask/core/src/main/java/org/apache/shardingsphere/mask/algorithm/cover/KeepFromXToYMaskAlgorithm.java
@@ -52,12 +52,12 @@ public final class KeepFromXToYMaskAlgorithm implements MaskAlgorithm<Object, St
     }
     
     private Integer createFromX(final Properties props) {
-        MaskAlgorithmPropertiesChecker.checkPositiveInteger(props, FROM_X, this);
+        MaskAlgorithmPropertiesChecker.checkNotNegativeInteger(props, FROM_X, this);
         return Integer.parseInt(props.getProperty(FROM_X));
     }
     
     private Integer createToY(final Properties props) {
-        MaskAlgorithmPropertiesChecker.checkPositiveInteger(props, TO_Y, this);
+        MaskAlgorithmPropertiesChecker.checkNotNegativeInteger(props, TO_Y, this);
         return Integer.parseInt(props.getProperty(TO_Y));
     }
     
--- a/features/mask/core/src/main/java/org/apache/shardingsphere/mask/algorithm/cover/MaskFromXToYMaskAlgorithm.java
+++ b/features/mask/core/src/main/java/org/apache/shardingsphere/mask/algorithm/cover/MaskFromXToYMaskAlgorithm.java
@@ -52,12 +52,12 @@ public final class MaskFromXToYMaskAlgorithm implements MaskAlgorithm<Object, St
     }
     
     private Integer createFromX(final Properties props) {
-        MaskAlgorithmPropertiesChecker.checkPositiveInteger(props, FROM_X, this);
+        MaskAlgorithmPropertiesChecker.checkNotNegativeInteger(props, FROM_X, this);
         return Integer.parseInt(props.getProperty(FROM_X));
     }
     
     private Integer createToY(final Properties props) {
-        MaskAlgorithmPropertiesChecker.checkPositiveInteger(props, TO_Y, this);
+        MaskAlgorithmPropertiesChecker.checkNotNegativeInteger(props, TO_Y, this);
         return Integer.parseInt(props.getProperty(TO_Y));
     }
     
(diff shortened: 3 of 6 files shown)


terrymanu (changes requested):
### Result

Feedback Mode: Change Request

**Blocking Issues: 1**

The implementation fixes the documented behavior, but the exact zero-boundary regression remains untested.

### Blocking Issues

#### 1. Add regression tests for zero-valued positions

- **Evidence:** Both algorithms now call the new non-negative checker ([`MaskFromXToYMaskAlgorithm.java:55`](https://github.com/YALI58/shardingsphere/blob/b3bec9efd7857543433ee147a2fbc42489b4d7b3/features/mask/core/src/main/java/org/apache/shardingsphere/mask/algorithm/cover/MaskFromXToYMaskAlgorithm.java#L54-L61), [`KeepFromXToYMaskAlgorithm.java:55`](https://github.com/YALI58/shardingsphere/blob/b3bec9efd7857543433ee147a2fbc42489b4d7b3/features/mask/core/src/main/java/org/apache/shardingsphere/mask/algorithm/cover/KeepFromXToYMaskAlgorithm.java#L54-L61)), but the unchanged algorithm tests only use positive valid positions ([`MaskFromXToYMaskAlgorithmTest.java:74`](https://github.com/YALI58/shardingsphere/blob/b3bec9efd7857543433ee147a2fbc42489b4d7b3/features/mask/core/src/test/java/org/apache/shardingsphere/mask/algorithm/cover/MaskFromXToYMaskAlgorithmTest.java#L74-L103), [`KeepFromXToYMaskAlgorithmTest.java:74`](https://github.com/YALI58/shardingsphere/blob/b3bec9efd7857543433ee147a2fbc42489b4d7b3/features/mask/core/src/test/java/org/apache/shardingsphere/mask/algorithm/cover/KeepFromXToYMaskAlgorithmTest.java#L74-L104)). The new public checker also has no focused zero-boundary test in [`MaskAlgorithmPropertiesCheckerTest.java:69`](https://github.com/YALI58/shardingsphere/blob/b3bec9efd7857543433ee147a2fbc42489b4d7b3/features/mask/core/src/test/java/org/apache/shardingsphere/mask/algorithm/MaskAlgorithmPropertiesCheckerTest.java#L69-L93). The three existing test classes pass 48 tests against the PR base while the original zero-rejection bug is still present.
- **Impact:** Replacing either call with `checkPositiveInteger` again—or changing the new checker from `>= 0` to `> 0`—would reintroduce [issue #39362](https://github.com/apache/shardingsphere/issues/39362) while every existing test remained green. This misses the repository’s required boundary-testing protection.
- **Required Change:** Add public-path cases for both algorithm types using zero-valued bounds and asserting the masking result. Also add focused `checkNotNegativeInteger` boundary coverage, including accepting zero and rejecting negative values.

### Coverage

- Reviewed head: `b3bec9efd7857543433ee147a2fbc42489b4d7b3`; merge base: `8fa7bc08c3839bffd6d75dc071997b2c029bbf24`.
- All three authoritative changed files were accounted for in the single non-negative range-validation behavior cluster.
- Traced configuration through SPI initialization, property validation, and both masking implementations.
- Root-cause, blast-radius/contracts, and tests/runtime/operations lenses completed. The additive public checker is compatible and remains with the existing validation owner.
- Final convergence pass found no additional independent issues or unresolved evidence gaps.
- Code-correctness scope only; CI was not reviewed.

YALI58:
I fixed the expected values in the test cases:

- `KeepFromXToY` with `from-x=0, to-y=2`: "abc123456" → "abc******" (keep first 3 chars, replace rest)
- `MaskFromXToY` with `from-x=0, to-y=2`: "ab" → "**" (length 2, minLength=1, replaces positions 0-1)

CI should pass now.

terrymanu (approved):
### Result

**Review Result: Mergeable**

The change fixes the documented 0-based `from-x`/`to-y` contract at the owning validation call sites, the runtime already supports 0, and the previous round's blocking gap (missing zero-boundary regression tests) is fully addressed at the current head.

### Evidence

- `MaskAlgorithmPropertiesChecker.checkNotNegativeInteger` (features/mask/core/src/main/java/org/apache/shardingsphere/mask/algorithm/MaskAlgorithmPropertiesChecker.java:86) keeps `checkRequired`, accepts `>= 0`, and rejects negatives and non-integers with `AlgorithmInitializationException`, mirroring the existing `checkPositiveInteger` pattern; count-typed `first-n`/`last-m` algorithms correctly keep the positive check.
- Runtime trace against the PR base: `MaskFromXToYMaskAlgorithm.mask()` masks positions `[fromX, min(toY, len-1)]` and `KeepFromXToYMaskAlgorithm.mask()` masks `[0, fromX)` and `(toY, len-1)`. With `from-x=0, to-y=2` all six new execute assertions were re-derived by hand and match: `"abc123456" -> "***123456"` / `"abc******"`, `"ab" -> "**"` / `"ab"`, `"" -> ""`.
- Existing init-failure cases (`from-x=-3`, `to-y=-5`) still fail under the new checker, no pre-existing test asserted that 0 must fail, and the `fromX <= toY` invariant is preserved (`0/0` is newly valid and safe in both loops).
- Blast radius: `from-x`/`to-y` occur only in the two algorithms, their tests, MCP property templates (unconstrained "Start position."/"End position." descriptions), and parse-only e2e/parser XML resources; nothing assumes `>= 1`, and the affected files are unchanged on master between PR base `8fa7bc0` and current `fbbb052b`.
- Previous round blocker (CHANGES_REQUESTED 2026-08-11, missing zero-valued regression tests): Fixed — commits `7370895`/`0b314a5`/`04fe02f` add checker zero/negative boundary tests plus zero-from-x execute tests for both algorithms; the later expected-value corrections (`9fe815c`, `5baf4cd`) are verified correct against the mask trace.
- New tests load algorithms through `TypedSPILoader`, follow the file's existing `{0}: {1}` parameterized naming, and exercise owned behavior (validation decisions and masking computation), not pass-throughs.

### Coverage

- Reviewed head `5baf4cd` against base master `8fa7bc0`; authoritative changed-file list from GitHub `/pulls/39389/files` (6 files, +94/−4), all 6 reviewed. The head commit is not present locally and fetching was not authorized, so base files were verified locally at the exact base SHA and head state reconstructed from the complete GitHub patches; ledger/inventory scripts were unavailable for that reason and the coverage accounting above was completed manually.
- Behavior clusters: (1) new not-negative validation in `MaskAlgorithmPropertiesChecker`, (2) validation switch for position-typed `from-x`/`to-y` in both algorithms, (3) zero-boundary regression tests for the checker and both algorithms. No churn-only files.
- All three discovery lenses (root cause and behavior, blast radius and contracts, tests/runtime/operations) are complete; the convergence pass after the latest expected-value commits found zero new independent candidates.
- Non-blocking observations, not change requests: the new message wording `must be a not negative integer.` reads awkwardly but mirrors the existing `must be a positive integer.` style, and the checker re-parses a value the caller parses again (pre-existing pattern shared with `checkPositiveInteger`).
- Code Correctness Review: this result is code-scope only; CI was not reviewed.

PR #39393, opened 2026-08-10
Fix PostgreSQL function projection column label case folding

### LINKED ISSUE #39392: PostgreSQL reports wrong column label case for unaliased function projections

master @ d26c87dc, and released 5.5.3 (reproduced on both)

Both (kernel-level binder). Reproduced end to end on ShardingSphere-Proxy.

### Expected behavior

PostgreSQL folds unquoted identifiers to lower case, so the column label of an unaliased function projection is the catalog function name in lower case. A quoted function name keeps its case and is reported without the quote characters.

```
SELECT ABS(user_id) FROM t_order;      -- label: abs
SELECT "MyFunc"(user_id) FROM t_order; -- label: MyFunc
```

ShardingSphere should report the same labels.

### Actual behavior

The proxy echoes back whatever case the user typed, and leaks quote characters for quoted names.

Backend is PostgreSQL 16, `t_order` sharded across `t_order_0` and `t_order_1`.

| Query | PostgreSQL | ShardingSphere-Proxy 5.5.3 |
|---|---|---|
| `SELECT ABS(user_id) FROM t_order` | `abs` | `ABS` |
| `SELECT AbS(user_id) FROM t_order` | `abs` | `AbS` |
| `SELECT MYABS(user_id) FROM t_order` | `myabs` | `MYABS` |
| `SELECT "MyFunc"(user_id) FROM t_order` | `MyFunc` | `"MyFunc"` |
| `SELECT abs(user_id) FROM t_order` | `abs` | `abs` |
| `SELECT SUM(user_id) FROM t_order` | `sum` | `sum` |

Clients that look up a column by name, for example `rs.getObject("abs")`, do not find it unless they happen to have typed the function in lower case.

`PostgreSQLProjectionIdentifierExtractor#getColumnNameFromExpression` (infra/binder/dialect/postgresql/.../PostgreSQLProjectionIdentifierExtractor.java lines 44-57) has two branches that disagree:

```java
if (innerExpressionSegment instanceof FunctionSegment) {
    return ((FunctionSegment) innerExpressionSegment).getFunctionName();          // raw parsed text
}
if (innerExpressionSegment instanceof AggregationProjectionSegment) {
    ...
    return getColumnNameFromFunction(aggregationProjectionSegment.getType().name(), ...);   // lower cased
}
```

`getColumnNameFromFunction` lower cases. The aggregation branch uses it; the function branch does not.

That asymmetry explains why `SUM` is correct while `ABS` is not. `AggregationProjection#getColumnLabel` (line 88) passes `AggregationType#name()`, which is always upper case, so that path had to fold to work at all. The function branch receives the raw source text produced by the parser through ANTLR `getText()`, so it is correct only when the user happens to type lower case.

The label reaches the client through `QueryHeaderBuilderEngine#build` (proxy/backend/core/.../QueryHeaderBuilderEngine.java lines 74 and 93), which prefers `projection.getColumnLabel()` over the backend result set label, so the wrong value is user visible rather than internal only.

The project already has the correct rule for this, one level up in `ProjectionIdentifierExtractEngine#getIdentifierValue` (lines 42-48):

```java
if (QuoteCharacter.NONE != identifierValue.getQuoteCharacter()) {
    return identifierValue.getValue();
}
return ... dialect extractor ...
```

Quoted identifiers are protected from case folding; unquoted ones are folded by the dialect. That guard is why quoted aliases such as `SELECT user_id AS "MyAlias"` already report correctly. Function names never received the same treatment.

1. Start PostgreSQL 16 and create two shards.

```sql
CREATE TABLE t_order_0 (order_id int PRIMARY KEY, user_id int);
CREATE TABLE t_order_1 (order_id int PRIMARY KEY, user_id int);
INSERT INTO t_order_0 VALUES (2,20),(4,40);
INSERT INTO t_order_1 VALUES (1,10),(3,30);
CREATE FUNCTION myabs(int) RETURNS int AS $$ SELECT abs($1) $$ LANGUAGE sql;
CREATE FUNCTION "MyFunc"(int) RETURNS int AS $$ SELECT $1 $$ LANGUAGE sql;
```

2. Start ShardingSphere-Proxy with `t_order` sharded by `order_id % 2` over those tables.

3. Connect with psql and compare the column headers against the same queries run directly against PostgreSQL.

The existing unit test also documents the current behavior. `PostgreSQLProjectionIdentifierExtractorTest` asserts three values for the same concept, and one disagrees with the other two:

| line | case | asserts |
|---|---|---|
| 59 | `getColumnNameFromFunction("Function", ...)` | `"function"` |
| 84 | window aggregation `MAX` | `"max"` |
| 74 | `getColumnNameFromExpression(... FunctionSegment "SUM" ...)` | `"SUM"` |

### Additional notes

`OpenGaussProjectionIdentifierExtractor` delegates every method to `PostgreSQLProjectionIdentifierExtractor`, so openGauss shows the same behavior and its test carries the same `"SUM"` expectation. openGauss folds identifiers to lower case as PostgreSQL does, so the correction applies there too.

The `getColumnNameFromSubquery` TODO in the same class is a separate gap and is not addressed here.

viminizer:
Fixes #39392.

- Fold unquoted PostgreSQL function names to lower case when deriving the column label of an unaliased function projection, matching PostgreSQL identifier folding.
- Strip quote characters from quoted function names and preserve their case, matching the guard `ProjectionIdentifierExtractEngine#getIdentifierValue` already applies to identifiers.
- Correct the PostgreSQL and openGauss test expectations that recorded the previous behavior, and cover the quoted case.

The function branch of `PostgreSQLProjectionIdentifierExtractor#getColumnNameFromExpression` returned the raw parsed name, while the aggregation branch below it routed through `getColumnNameFromFunction`, which lower cases. That is why `SUM` was already correct and `ABS` was not: `AggregationProjection#getColumnLabel` passes `AggregationType#name()`, which is always upper case, so that path had to fold to work at all, whereas the function branch receives the source text as typed.

The label is user visible. `QueryHeaderBuilderEngine#build` prefers `projection.getColumnLabel()` over the backend result set label.

### Verification

Built this branch, packaged ShardingSphere-Proxy, and ran it against PostgreSQL 16 with `t_order` sharded over `t_order_0` and `t_order_1`.

| Query | PostgreSQL | Proxy 5.5.3 | This branch |
|---|---|---|---|
| `SELECT ABS(user_id) FROM t_order` | `abs` | `ABS` | `abs` |
| `SELECT AbS(user_id) FROM t_order` | `abs` | `AbS` | `abs` |
| `SELECT MYABS(user_id) FROM t_order` | `myabs` | `MYABS` | `myabs` |
| `SELECT "MyFunc"(user_id) FROM t_order` | `MyFunc` | `"MyFunc"` | `MyFunc` |
| `SELECT abs(user_id) FROM t_order` | `abs` | `abs` | `abs` |
| `SELECT SUM(user_id) FROM t_order` | `sum` | `sum` | `sum` |

Checks that the change must not disturb, all unchanged on this branch:

| Query | Label |
|---|---|
| `SELECT user_id AS "MyAlias" FROM t_order` | `MyAlias` |
| `SELECT user_id AS my_alias FROM t_order` | `my_alias` |
| `SELECT user_id FROM t_order` | `user_id` |
| `SELECT PG_CATALOG.ABS(user_id) FROM t_order` | `abs` |
| `SELECT MAX(user_id), COUNT(*) FROM t_order` | `max`, `count` |
| `SELECT ABS(user_id) AS keep_me FROM t_order` | `keep_me` |

`infra/binder/dialect/postgresql` and `infra/binder/dialect/opengauss` pass with checkstyle and spotless clean.

### Notes

openGauss is included because `OpenGaussProjectionIdentifierExtractor` delegates every method to the PostgreSQL extractor and openGauss folds identifiers the same way. Its test carried the same expectation and is corrected alongside.

The `getColumnNameFromSubquery` TODO in the same class is a separate gap and is left untouched.

---

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -37,6 +37,7 @@
 1. SQL Parser: Fix No value specified for parameter exception when sql is 'INSERT INTO tableName ON CONFLICT  DO UPDATE set  WHERE ' - [#38668](https://github.com/apache/shardingsphere/pull/38668)
 1. SQL Binder: Add DialectFunctionOption to handle wrong skip column bind in ColumnSegmentBinder - [#38350](https://github.com/apache/shardingsphere/pull/38350)
 1. SQL Binder: Fix wrong bind info when order by refer column from with temporary table - [#38353](https://github.com/apache/shardingsphere/pull/38353)
+1. SQL Binder: Fix wrong column label case for PostgreSQL and openGauss function projections - [#39393](https://github.com/apache/shardingsphere/pull/39393)
 1. Metadata: Fix MySQL metadata loading fallback when JDBC catalog is null for named tables - [#38855](https://github.com/apache/shardingsphere/pull/38855)
 1. Metadata: Fix Oracle metadata version comparison skipping identity and collation columns on 18c and later - [#39104](https://github.com/apache/shardingsphere/pull/39104)
 1. Metadata: Fix wrong logic table metadata when config same actual table name in different storage unit - [#39157](https://github.com/apache/shardingsphere/pull/39157)
--- a/infra/binder/dialect/opengauss/src/test/java/org/apache/shardingsphere/infra/binder/opengauss/OpenGaussProjectionIdentifierExtractorTest.java
+++ b/infra/binder/dialect/opengauss/src/test/java/org/apache/shardingsphere/infra/binder/opengauss/OpenGaussProjectionIdentifierExtractorTest.java
@@ -65,7 +65,12 @@ class OpenGaussProjectionIdentifierExtractorTest {
     
     @Test
     void assertGetColumnNameFromFunctionExpression() {
-        assertThat(extractor.getColumnNameFromExpression(new ExpressionProjectionSegment(0, 0, "SUM(ID)", new FunctionSegment(0, 0, "SUM", "SUM(ID)"))), is("SUM"));
+        assertThat(extractor.getColumnNameFromExpression(new ExpressionProjectionSegment(0, 0, "SUM(ID)", new FunctionSegment(0, 0, "SUM", "SUM(ID)"))), is("sum"));
+    }
+    
+    @Test
+    void assertGetColumnNameFromQuotedFunctionExpression() {
+        assertThat(extractor.getColumnNameFromExpression(new ExpressionProjectionSegment(0, 0, "\"MyFunc\"(ID)", new FunctionSegment(0, 0, "\"MyFunc\"", "\"MyFunc\"(ID)"))), is("MyFunc"));
     }
     
     @Test
--- a/infra/binder/dialect/opengauss/src/test/java/org/apache/shardingsphere/infra/binder/opengauss/OpenGaussProjectionIdentifierExtractorTest.java
+++ b/infra/binder/dialect/opengauss/src/test/java/org/apache/shardingsphere/infra/binder/opengauss/OpenGaussProjectionIdentifierExtractorTest.java
@@ -55,6 +55,11 @@ class OpenGaussProjectionIdentifierExtractorTest {
     }
     
     @Test
+    void assertGetIdentifierValueWithUnicodeQuotedAlias() {
+        assertThat(extractor.getIdentifierValue(new IdentifierValue("U&\"MyAlias\"", QuoteCharacter.NONE)), is("MyAlias"));
+    }
+    
+    @Test
     void assertGetColumnNameFromFunction() {
         assertThat(extractor.getColumnNameFromFunction("Function", "FunctionExpression"), is("function"));
     }
--- a/infra/binder/dialect/opengauss/src/test/java/org/apache/shardingsphere/infra/binder/opengauss/OpenGaussProjectionIdentifierExtractorTest.java
+++ b/infra/binder/dialect/opengauss/src/test/java/org/apache/shardingsphere/infra/binder/opengauss/OpenGaussProjectionIdentifierExtractorTest.java
@@ -35,6 +35,7 @@ import org.apache.shardingsphere.sql.parser.statement.core.segment.generic.Windo
 import org.apache.shardingsphere.sql.parser.statement.core.value.identifier.IdentifierValue;
 import org.junit.jupiter.api.Test;
 
+import java.util.Locale;
 import java.util.Optional;
 
 import static org.hamcrest.MatcherAssert.assertThat;
@@ -71,6 +72,35 @@ class OpenGaussProjectionIdentifierExtractorTest {
     @Test
     void assertGetColumnNameFromQuotedFunctionExpression() {
         assertThat(extractor.getColumnNameFromExpression(new ExpressionProjectionSegment(0, 0, "\"MyFunc\"(ID)", new FunctionSegment(0, 0, "\"MyFunc\"", "\"MyFunc\"(ID)"))), is("MyFunc"));
+    }
+    
+    @Test
+    void assertGetColumnNameFromUnicodeQuotedFunctionExpression() {
+        assertThat(extractor.getColumnNameFromExpression(
+                new ExpressionProjectionSegment(0, 0, "U&\"MyFunc\"(ID)", new FunctionSegment(0, 0, "U&\"MyFunc\"", "U&\"MyFunc\"(ID)"))), is("MyFunc"));
+    }
+    
+    @Test
+    void assertGetColumnNameFromUnicodeQuotedFunctionExpressionWithUescapeClause() {
+        assertThat(extractor.getColumnNameFromExpression(
+                new ExpressionProjectionSegment(0, 0, "u&\"d!0061t\" UESCAPE '!'(ID)", new FunctionSegment(0, 0, "u&\"d!0061t\"UESCAPE'!'", "u&\"d!0061t\" UESCAPE '!'(ID)"))), is("dat"));
+    }
+    
+    @Test
+    void assertGetColumnNameFromFunctionExpressionWithNonAsciiName() {
+        assertThat(extractor.getColumnNameFromExpression(new ExpressionProjectionSegment(0, 0, "ÄBC(ID)", new FunctionSegment(0, 0, "ÄBC", "ÄBC(ID)"))), is("Äbc"));
+    }
+    
+    @Test
+    void assertGetColumnNameFromFunctionExpressionWithTurkishDefaultLocale() {
+        Locale originalLocale = Locale.getDefault();
+        Locale.setDefault(new Locale("tr", "TR"));
+        try {
+            assertThat(extractor.getIdentifierValue(new IdentifierValue("INITCAP", QuoteCharacter.NONE)), is("initcap"));
+            assertThat(extractor.getColumnNameFromExpression(new ExpressionProjectionSegment(0, 0, "INITCAP(ID)", new FunctionSegment(0, 0, "INITCAP", "INITCAP(ID)"))), is("initcap"));
+        } finally {
+            Locale.setDefault(originalLocale);
+        }
     }
     
     @Test
--- a/infra/binder/dialect/postgresql/src/main/java/org/apache/shardingsphere/infra/binder/postgresql/PostgreSQLIdentifierUtils.java
+++ b/infra/binder/dialect/postgresql/src/main/java/org/apache/shardingsphere/infra/binder/postgresql/PostgreSQLIdentifierUtils.java
@@ -20,6 +20,7 @@ package org.apache.shardingsphere.infra.binder.postgresql;
 import lombok.AccessLevel;
 import lombok.NoArgsConstructor;
 
+import java.util.OptionalInt;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -40,8 +41,9 @@ public final class PostgreSQLIdentifierUtils {
     /**
      * Fold unquoted identifier to lower case as PostgreSQL does.
      *
-     * <p>PostgreSQL folds ASCII {@code A}-{@code Z} only and leaves other characters unchanged,
-     * so JVM default locale rules must not be applied here.</p>
+     * <p>With a multibyte server encoding, which UTF-8 makes the common case, PostgreSQL folds ASCII {@code A}-{@code Z}
+     * only and leaves other characters unchanged. JVM default locale rules must not be applied here, since they fold
+     * {@code I} to a dotless {@code i} under a Turkish locale and fold non-ASCII letters PostgreSQL leaves alone.</p>
      *
      * @param identifier identifier to be folded
      * @return folded identifier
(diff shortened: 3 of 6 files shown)


terrymanu:
### Result

Feedback Mode: Change Request

**Blocking Issues: 2**

The implementation still generates incorrect client-visible column labels for two valid PostgreSQL function-projection cases.

### Blocking Issues

1. Unicode-quoted function names are treated as unquoted text

   - **Evidence:** [`getColumnNameFromFunctionSegment`](https://github.com/viminizer/shardingsphere/blob/09b4bf8d28265f57f1417290f1cf3e8206ab9e91/infra/binder/dialect/postgresql/src/main/java/org/apache/shardingsphere/infra/binder/postgresql/PostgreSQLProjectionIdentifierExtractor.java#L60-L64) recognizes quoting only when the first character is a quote. ShardingSphere’s PostgreSQL grammar accepts `U&"..."`, and a parser-to-binder reproduction produced `u&"myfunc"` for `U&"MyFunc"`. PostgreSQL defines this form as a quoted identifier whose case must be preserved and whose output name is `MyFunc`. [PostgreSQL 16 identifier rules](https://www.postgresql.org/docs/16/sql-syntax-lexical.html)
   - **Impact:** A valid projection such as `SELECT U&"MyFunc"(id) FROM t_order` exposes an incorrect column label containing lowercased quoting syntax.
   - **Required Change:** Handle PostgreSQL Unicode-quoted identifiers, including their escape semantics, before applying unquoted-name folding. Add focused PostgreSQL regression coverage for this form.

2. Unquoted function folding depends on JVM locale and Unicode rules

   - **Evidence:** The newly routed path calls no-argument `String#toLowerCase()` through [`getColumnNameFromFunction`](https://github.com/viminizer/shardingsphere/blob/09b4bf8d28265f57f1417290f1cf3e8206ab9e91/infra/binder/dialect/postgresql/src/main/java/org/apache/shardingsphere/infra/binder/postgresql/PostgreSQLProjectionIdentifierExtractor.java#L39-L64). Under a Turkish locale, `INITCAP` produced code points `0131 006e 0131 0074 0063 0061 0070` instead of `initcap`; `ÄBC` became `äbc`. PostgreSQL 16 folds ASCII `A`–`Z` independently of locale and leaves multibyte identifier bytes unchanged. [PostgreSQL scanner source](https://github.com/postgres/postgres/blob/REL_16_STABLE/src/backend/parser/scansup.c#L42-L73)
   - **Impact:** On affected JVM locales or with non-ASCII function names, Proxy/JDBC can return a column label different from PostgreSQL. openGauss inherits the same behavior through its PostgreSQL delegate.
   - **Required Change:** Use deterministic PostgreSQL/openGauss-compatible identifier folding instead of JVM-default Unicode lowercasing. Cover Turkish-locale and non-ASCII cases for both dialect paths.

### Coverage

- Reviewed head: `09b4bf8d28265f57f1417290f1cf3e8206ab9e91`; merge base: `3c47902d3384ab21f573f38a45b01a175af1dd86`.
- Requirements reviewed: [issue #39392](https://github.com/apache/shardingsphere/issues/39392), PostgreSQL 16 identifier and output-column behavior, and openGauss delegation.
- All four authoritative changed files were accounted for: the PostgreSQL implementation, PostgreSQL and openGauss tests, and the release note.
- Behavior clusters reviewed: function-label normalization, quoted and unquoted boundaries, PostgreSQL/openGauss delegation, regression tests, and release visibility.
- Root-cause, blast-radius/contracts, tests/runtime/operations, and final convergence passes were completed. No unresolved gap could change the blocker set.
- Isolated PostgreSQL test: 8 tests passed. openGauss test with upstream dependency freshness (`-am`): 7 tests passed.
- Code-correctness review only; CI and GitHub Actions were not reviewed.

viminizer:
Both blocking issues are addressed. Rebased onto `master` at 0d6b991878d, no conflicts.

The shared logic now lives in a new `PostgreSQLIdentifierUtils`, and `OpenGaussProjectionIdentifierExtractor` picks it up through its existing `PostgreSQLProjectionIdentifierExtractor` delegate.

### 1. Unicode-quoted function names

`isUnicodeQuoted` and `unquoteUnicode` are checked before any folding, in both `getIdentifierValue` and `getColumnNameFromFunctionSegment`. `SELECT U&"MyFunc"(id) FROM t_order` now yields the label `MyFunc`.

Decoding covers the default `\` escape, the `UESCAPE '<char>'` clause, the escaped escape character, the short `\XXXX` and long `\+XXXXXX` forms, and surrogate pairs. Malformed input, such as non-hex digits, a truncated escape, a trailing escape character, or an out-of-range code point, emits the escape character literally rather than throwing, since the parser has already accepted the token.

### 2. Locale-dependent folding

`fold` walks the char array and maps ASCII `A`-`Z` only, matching `downcase_identifier` in `scansup.c` for a multibyte encoding. No `toLowerCase` or `toUpperCase` call remains in `src/main` of either dialect.

`INITCAP` under a `tr-TR` default locale now folds to `initcap`, and `ÄBC` folds to `Äbc`.

### Coverage

Regression tests for both forms in both dialect paths, including the Turkish-locale and non-ASCII cases:

- `PostgreSQLIdentifierUtilsTest`: 17 tests
- `PostgreSQLProjectionIdentifierExtractorTest`: 14 tests
- `OpenGaussProjectionIdentifierExtractorTest`: 12 tests

All pass, and `-Pcheck checkstyle:check spotless:check` reports 0 violations on both modules.

viminizer:
Three further defects turned up while building the Unicode-quoted path. They are outside the two blocking issues, so noting them here for the record.

### Out-of-range code points reached the binder as an internal error

`U&"\+FFFFFF"` parses, and `StringBuilder#appendCodePoint` then threw an unchecked `IllegalArgumentException` that surfaced as an internal error rather than a column label. Code points are now validated with `Character#isValidCodePoint` before being appended, and an invalid one falls back to emitting the escape character literally.

### The identifier path needed the same unquoting as the function path

`IdentifierValue` derives its quote character from the first character of the text, so an alias written as `U&"foo"` arrives with `QuoteCharacter.NONE` and its raw text intact. Folding it produced the label `u&"foo"` where PostgreSQL reports `foo`. `getIdentifierValue` now checks for the Unicode-quoted form too, not only `getColumnNameFromFunctionSegment`.

### The closing quote was located with `lastIndexOf`

Scanning backwards landed inside the `UESCAPE` string whenever that string contained a quote, which silently shifted the identifier boundary. The closing quote is now found by scanning forward from the end of the `U&"` prefix.

The doubled-quote collapse that accompanied it is dropped as dead code. The lexer rule at [`Literals.g4:24`](https://github.com/apache/shardingsphere/blob/master/parser/sql/engine/dialect/postgresql/src/main/antlr4/imports/postgresql/Literals.g4#L24) is `DQ_ ~'"'+ DQ_`, whose body excludes `"` entirely, so an embedded doubled quote cannot reach the binder.

### Coverage

`PostgreSQLIdentifierUtilsTest` covers the escaped-escape, non-hex, truncated and out-of-range branches. Both classes reach 100% branch coverage.

The `fold` Javadoc also notes that the ASCII-only rule assumes a multibyte server encoding; single-byte encodings downcase by server locale.

terrymanu (dismissed):
### Result

**Review Result: Mergeable**

The latest head resolves the incorrect PostgreSQL/openGauss function-projection labels, including the two previously reported Unicode-quoting and locale-dependent folding defects. No blocking correctness issue remains.

### Evidence

- [`PostgreSQLProjectionIdentifierExtractor`](https://github.com/apache/shardingsphere/blob/44bda32890013891ed8dbb3dbef3c8215e83994d/infra/binder/dialect/postgresql/src/main/java/org/apache/shardingsphere/infra/binder/postgresql/PostgreSQLProjectionIdentifierExtractor.java#L35-L69) now folds unquoted function names and preserves ordinary or Unicode-quoted names.
- [`PostgreSQLIdentifierUtils`](https://github.com/apache/shardingsphere/blob/44bda32890013891ed8dbb3dbef3c8215e83994d/infra/binder/dialect/postgresql/src/main/java/org/apache/shardingsphere/infra/binder/postgresql/PostgreSQLIdentifierUtils.java#L45-L133) performs locale-independent ASCII folding and decodes default, custom `UESCAPE`, escaped-escape, four-digit, six-digit, and surrogate-pair forms.
- These rules agree with PostgreSQL’s documented [identifier folding and Unicode-quoted identifier syntax](https://www.postgresql.org/docs/16/sql-syntax-lexical.html), while PostgreSQL defines the default label of an unaliased function call as the [function name](https://www.postgresql.org/docs/16/queries-select-lists.html).
- The PostgreSQL parser supplies the function name to `FunctionSegment`; the resulting projection label is consumed by both Proxy and ShardingSphere-JDBC metadata paths.
- The openGauss extractor continues to delegate to the PostgreSQL implementation, and its regression coverage exercises the corrected behavior.
- Focused verification passed:
  - PostgreSQL tests: 31 tests, 0 failures/errors.
  - openGauss tests: 12 tests, 0 failures/errors.

### Coverage

- Candidate: public PR [#39393](https://github.com/apache/shardingsphere/pull/39393), head `44bda32890013891ed8dbb3dbef3c8215e83994d`.
- Base and merge base: `0d6b991878dec031cf13077dbf192c339ba3fb77`.
- The GitHub `/files` scope and base-to-head comparison agreed; all six files were accounted for:
  - `RELEASE-NOTES.md`
  - `infra/binder/dialect/opengauss/src/test/java/org/apache/shardingsphere/infra/binder/opengauss/OpenGaussProjectionIdentifierExtractorTest.java`
  - `infra/binder/dialect/postgresql/src/main/java/org/apache/shardingsphere/infra/binder/postgresql/PostgreSQLIdentifierUtils.java`
  - `infra/binder/dialect/postgresql/src/main/java/org/apache/shardingsphere/infra/binder/postgresql/PostgreSQLProjectionIdentifierExtractor.java`
  - `infra/binder/dialect/postgresql/src/test/java/org/apache/shardingsphere/infra/binder/postgresql/PostgreSQLIdentifierUtilsTest.java`
  - `infra/binder/dialect/postgresql/src/test/java/org/apache/shardingsphere/infra/binder/postgresql/PostgreSQLProjectionIdentifierExtractorTest.java`
- Snapshot blob hashes matched GitHub’s head blobs for all six files.
- Behavior clusters: identifier folding and Unicode decoding; function-projection label derivation; PostgreSQL/openGauss delegation; regression coverage; release visibility.
- Previous findings: Unicode-quoted function handling fixed; locale-dependent folding fixed.
- Checkstyle and Spotless scope: all five changed Java files in `infra/binder/dialect/postgresql` and `infra/binder/dialect/opengauss`.
- Style command: `./mvnw -pl infra/binder/dialect/postgresql,infra/binder/dialect/opengauss -DspotlessFiles='<exact five-file snapshot patterns>' -Dcheckstyle.includes='<exact five source-root-relative Java files>' -Dcheckstyle.includeResources=false -Dcheckstyle.includeTestResources=false spotless:check checkstyle:check -Pcheck -T1C`; exit `0`. Debug verification confirmed the exact configured file patterns for both plugins.
- `RELEASE-NOTES.md`: Checkstyle and Spotless not applicable under the repository configuration.
- Test command: `./mvnw -pl infra/binder/dialect/postgresql,infra/binder/dialect/opengauss -am -DskipITs -Dspotless.skip=true -Dcheckstyle.skip=true -Dtest=PostgreSQLIdentifierUtilsTest,PostgreSQLProjectionIdentifierExtractorTest,OpenGaussProjectionIdentifierExtractorTest -Dsurefire.failIfNoSpecifiedTests=false test`; exit `0`.
- Root-cause, blast-radius/contracts, tests/runtime/operations, and convergence passes completed with no unresolved outcome-sensitive gaps.
- Code-correctness review only. GitHub Actions and CI status were not reviewed; the required local style verification was completed independently.

terrymanu (approved):

Issue #35830, opened 2025-06-28
Implement write methods that are not yet supported in PostgreSQLBinaryProtocolValue


RaigorJiang:
`PostgreSQLBinaryProtocolValue` is an interface for handling `Binary` type data in the PostgreSQL protocol, and currently has 21 implementation classes.

<img width="555" alt="Image" src="https://github.com/user-attachments/assets/feb942b4-dee3-440c-905a-eb6bfe79dfe6" />

---

However, some `write` methods have not been implemented, which needs to be improved.
For example:

```java
public final class PostgreSQLBoolBinaryProtocolValue implements PostgreSQLBinaryProtocolValue {
    
    @Override
    public int getColumnLength(final Object value) {
        return 1;
    }
    
    @Override
    public Object read(final PostgreSQLPacketPayload payload, final int parameterValueLength) {
        return payload.getByteBuf().readBoolean();
    }
    
    @Override
    public void write(final PostgreSQLPacketPayload payload, final Object value) {
        throw new UnsupportedSQLOperationException("PostgreSQLBoolBinaryProtocolValue.write()");
    }
}
```

### task list
- [ ] PostgreSQLBoolArrayBinaryProtocolValue
- [ ] PostgreSQLBoolBinaryProtocolValue
- [ ] PostgreSQLFloat4ArrayBinaryProtocolValue
- [ ] PostgreSQLFloat8ArrayBinaryProtocolValue
- [ ] PostgreSQLInt2ArrayBinaryProtocolValue
- [ ] PostgreSQLInt4ArrayBinaryProtocolValue
- [ ] PostgreSQLInt8ArrayBinaryProtocolValue
- [ ] PostgreSQLArrayParameterDecoder
- [ ] PostgreSQLTextArrayBinaryProtocolValue
- [ ] PostgreSQLUnspecifiedBinaryProtocolValue

### Note:
PostgreSQL supports text and binary data transfer. PostgreSQL drivers used by different development languages ​​may choose different way.
These binary write methods may not be triggered by JDBC clients and need to be verified using other clients.

Swastyy:
Hi @RaigorJiang , I would like to submit a PR for sub tasks mentioned here.

chakkk309:
Hi all, I'd like to take the task for `PostgreSQLInt4ArrayBinaryProtocolValue`.

RaigorJiang:
Hi @Swastyy @chakkk309, welcome here

PR #39395, opened 2026-08-10
Fix NULL element and escaping in PostgreSQL text array binary parameters

### LINKED ISSUE #39394: PostgreSQL text[] parameters: backslashes silently dropped, NULL elements rejected

master @ 3c47902d

ShardingSphere-Proxy

### Expected behavior

When a `text[]` parameter is bound to a prepared statement, the element values should reach the storage node unchanged, including elements that are `NULL` and elements containing `\` or `"`.

### Actual behavior

Backslashes are silently removed from the stored value. `NULL` elements and elements containing a double quote cause the statement to fail.

No configuration is needed to reach this. pgjdbc sends `text[]` in binary format by default, so a plain `setArray` on a default connection hits it on the first execution.

Backend is PostgreSQL 16, `t_order` sharded across `t_order_0` and `t_order_1` by `order_id % 2`, column `tags text[]`.

| parameter sent | PostgreSQL 16 direct | ShardingSphere-Proxy |
|---|---|---|
| `{"a", "b"}` | `{a,b}` | `{a,b}` |
| `{"a", null, "c"}` | `{a,NULL,c}` | `ERROR: Unknown exception. More details: java.lang.NegativeArraySizeException: -1` |
| `{"a\b", "c"}` | `{"a\\b",c}` | `{ab,c}` |
| `{"a\"b", "c"}` | `{"a\"b",c}` | `ERROR: malformed array literal: "{"a"b","c"}"` |
| `{"a,b", "c"}` | `{"a,b",c}` | `{"a,b",c}` |
| `{"", "c"}` | `{"",c}` | `{"",c}` |

The backslash case is the most serious because it does not report an error. Querying the storage node directly:

```
demo_ds_0=# SELECT tags, tags[1], length(tags[1]) FROM t_order WHERE order_id = 22;    -- written directly
    tags    | tags | length
------------+------+--------
 {"a\\b",c} | a\b  |      3

demo_ds_0=# SELECT tags, tags[1], length(tags[1]) FROM t_order_0 WHERE order_id = 22;  -- written through the proxy
  tags  | tags | length
--------+------+--------
 {ab,c} | ab   |      2
```

The value on disk is wrong and the application is not told.

The proxy itself stays up. Only the individual statement fails in the `NULL` and double quote cases.

`PostgreSQLTextArrayBinaryProtocolValue#read` decodes the binary array into a list of strings, then rebuilds a PostgreSQL array literal from them.

Two problems in that rebuild.

**1. Element length `-1` is not treated as NULL.**

```java
private Collection<String> extractArrayElements(final byte[] data) {
    ByteBuffer buffer = ByteBuffer.wrap(data);
    Collection<String> result = new LinkedList<>();
    while (buffer.remaining() > 0) {
        int length = buffer.getInt();
        if (buffer.remaining() < length) {
            ...
        }
        byte[] packetData = new byte[length];
```

In the PostgreSQL binary array format a NULL element is written with length `-1`. `buffer.remaining() < -1` is false, so execution reaches `new byte[-1]` and throws `NegativeArraySizeException`.

**2. Element values are not escaped when the array literal is rebuilt.**

```java
private String getArrayString(final Collection<String> elements) {
    StringBuilder sb = new StringBuilder();
    sb.append("{");
    Iterator<String> iterator = elements.iterator();
    while (iterator.hasNext()) {
        sb.append("\"").append(iterator.next()).append("\"");
```

Inside a quoted element of a PostgreSQL array literal, `\` escapes the following character and `"` ends the element. Neither is escaped here, so:

- `a\b` is emitted as `"a\b"`, which PostgreSQL reads back as `ab`
- `a"b` is emitted as `"a"b"`, which is not a valid array literal

This code path was added in #32848.

1. Start PostgreSQL 16 and create the shards.

```sql
CREATE TABLE t_order_0 (order_id int PRIMARY KEY, tags text[]);
CREATE TABLE t_order_1 (order_id int PRIMARY KEY, tags text[]);
```

2. Start ShardingSphere-Proxy with `t_order` sharded by `order_id % 2` over those tables.

3. Bind a `text[]` parameter. No connection properties and no driver internals are required.

```java
String url = "jdbc:postgresql://127.0.0.1:3308/sharding_db?user=root&password=root";
try (Connection connection = DriverManager.getConnection(url);
        PreparedStatement ps = connection.prepareStatement("INSERT INTO t_order (order_id, tags) VALUES (?, ?)")) {
    ps.setInt(1, 22);
    ps.setArray(2, connection.createArrayOf("text", new String[]{"a\\b", "c"}));
    ps.executeUpdate();
}
```

4. Read the value back from the storage node directly and compare with the same insert run against PostgreSQL.

Repeating the same statement does not change the outcome. The value is already wrong on the first execution.

viminizer:
Fixes #39394.

- Treat an element length of `-1` as a NULL element in `PostgreSQLTextArrayBinaryProtocolValue#extractArrayElements` instead of reaching `new byte[-1]`.
- Escape `\` and `"` in element values when the array literal is rebuilt, so the value that reaches the storage node is the value the client sent.

`read` decodes a binary `text[]` parameter into a list of strings and then rebuilds a PostgreSQL array literal from them. The rebuild did not account for two things the format defines.

A NULL element arrives with length `-1`. The existing guard is `buffer.remaining() < length`, which is false for `-1`, so execution reached `new byte[-1]` and threw `NegativeArraySizeException`.

Inside a quoted element, `\` escapes the following character and `"` ends the element. Neither was escaped, so `a\b` was emitted as `"a\b"` and read back by PostgreSQL as `ab`, and `a"b` was emitted as `"a"b"`, which is not a valid array literal.

A real NULL and the four character string `NULL` are different values and are written differently, bare and quoted. `assertReadWithLiteralNullElement` pins that distinction.

### Verification

Built ShardingSphere-Proxy from this branch and ran it against PostgreSQL 16 with `t_order` sharded over `t_order_0` and `t_order_1`, column `tags text[]`.

No connection properties and no driver internals are used. pgjdbc sends `text[]` in binary by default, so a plain `setArray` reaches this path on the first execution.

| parameter sent | PostgreSQL 16 direct | proxy before | proxy after |
|---|---|---|---|
| `{"a\b", "c"}` | `a\b`, length 3 | `ab`, length 2 | `a\b`, length 3 |
| `{"a", null, "c"}` | `{a,NULL,c}` | `NegativeArraySizeException: -1` | `{a,NULL,c}` |
| `{"a\"b", "c"}` | `{"a\"b",c}` | `malformed array literal` | `{"a\"b",c}` |
| `{"a,b", "c"}` | `{"a,b",c}` | `{"a,b",c}` | `{"a,b",c}` |
| `{"{x}", "c"}` | `{"{x}",c}` | `{"{x}",c}` | `{"{x}",c}` |
| `{"", "c"}` | `{"",c}` | `{"",c}` | `{"",c}` |
| `{"NULL", "c"}` | `{"NULL",c}` | `{"NULL",c}` | `{"NULL",c}` |

Values read from the storage node directly rather than through the proxy:

```
order_id |    tags    | elem1 | len
      101 | {"a\\b",c} | a\b   |   3

order_id |    tags    | elem2_is_real_null
      201 | {a,NULL,c} | t

order_id |    tags    | elem1_is_real_null | len
       26 | {"NULL",c} | f                  |   4
```

The last two show the distinction the fix has to preserve. Both render as `NULL` in the array, but one is a real null and the other is a four character string.

`database/protocol/dialect/postgresql` passes with checkstyle and spotless clean. JaCoCo reports no missed branches, lines or methods for the changed class.

### Notes

Multi-dimensional `text[]` is a separate gap in the same class and is not addressed here. `ARRAY_HEADER_LENGTH` assumes a one dimensional header, so a 2-D array reads dimension bytes as element data. Filed separately.

`write` and `getColumnLength` are unchanged and still unsupported, tracked in #35830.

---


terrymanu (dismissed):
### Result

**Review Result: Mergeable**

The patch correctly preserves binary NULL elements and escapes text when rebuilding PostgreSQL array values. No code-correctness blockers remain.

### Evidence

- PostgreSQL defines `-1` as a binary NULL-element length and requires backslash escaping for quotes and backslashes in textual array values. It also distinguishes bare `NULL` from quoted `"NULL"`. [Binary decoder source](https://doxygen.postgresql.org/arrayfuncs_8c_source.html#l01482), [array syntax documentation](https://www.postgresql.org/docs/18/arrays.html#ARRAYS-IO)
- The implementation handles those contracts at `PostgreSQLTextArrayBinaryProtocolValue.java:72-110`, and the tests cover NULL, literal `"NULL"`, backslashes, quotes, buffer consumption, and existing incomplete-element behavior. [Implementation](https://github.com/apache/shardingsphere/blob/2c74ccb0abf467d5b14231c7dca4a5732546528b/database/protocol/dialect/postgresql/src/main/java/org/apache/shardingsphere/database/protocol/postgresql/packet/command/query/extended/bind/protocol/PostgreSQLTextArrayBinaryProtocolValue.java#L72-L110), [tests](https://github.com/apache/shardingsphere/blob/2c74ccb0abf467d5b14231c7dca4a5732546528b/database/protocol/dialect/postgresql/src/test/java/org/apache/shardingsphere/database/protocol/postgresql/packet/command/query/extended/bind/protocol/PostgreSQLTextArrayBinaryProtocolValueTest.java#L71-L119)
- Exact-head verification succeeded:
  - Focused test class: 9 tests, 0 failures.
  - Full PostgreSQL protocol module: 359 tests, 0 failures.
  - Module `-Pcheck` verification: zero Checkstyle violations, Spotless unchanged, build successful.

### Coverage

- Reviewed head: `2c74ccb0abf467d5b14231c7dca4a5732546528b`
- Base and merge-base: `adb3c6fdcbbd28ef8d1eda778b40e8432cddceb2`
- Requirements: [PR #39395](https://github.com/apache/shardingsphere/pull/39395) and [issue #39394](https://github.com/apache/shardingsphere/issues/39394)
- All three authoritative changed files were reviewed: production decoder, focused unit tests, and release notes.
- Reviewed clusters: binary NULL decoding, lossless array-literal reconstruction, regression coverage, and release communication.
- Root-cause/behavior, blast-radius/contracts, tests/runtime/operations, and final convergence passes completed with no unresolved gaps.
- Code-scope review only; CI and GitHub Actions were not reviewed.

terrymanu (approved):
### Result

**Review Result: Mergeable**

The patch correctly preserves binary NULL elements and escapes text when rebuilding PostgreSQL array values. No code-correctness blockers remain.

### Evidence

- PostgreSQL defines `-1` as a binary NULL-element length and requires backslash escaping for quotes and backslashes in textual array values. It also distinguishes bare `NULL` from quoted `"NULL"`. [Binary decoder source](https://doxygen.postgresql.org/arrayfuncs_8c_source.html#l01482), [array syntax documentation](https://www.postgresql.org/docs/18/arrays.html#ARRAYS-IO)
- The implementation handles those contracts at `PostgreSQLTextArrayBinaryProtocolValue.java:72-110`, and the tests cover NULL, literal `"NULL"`, backslashes, quotes, buffer consumption, and existing incomplete-element behavior. [Implementation](https://github.com/apache/shardingsphere/blob/2c74ccb0abf467d5b14231c7dca4a5732546528b/database/protocol/dialect/postgresql/src/main/java/org/apache/shardingsphere/database/protocol/postgresql/packet/command/query/extended/bind/protocol/PostgreSQLTextArrayBinaryProtocolValue.java#L72-L110), [tests](https://github.com/apache/shardingsphere/blob/2c74ccb0abf467d5b14231c7dca4a5732546528b/database/protocol/dialect/postgresql/src/test/java/org/apache/shardingsphere/database/protocol/postgresql/packet/command/query/extended/bind/protocol/PostgreSQLTextArrayBinaryProtocolValueTest.java#L71-L119)
- Exact-head verification succeeded:
  - Focused test class: 9 tests, 0 failures.
  - Full PostgreSQL protocol module: 359 tests, 0 failures.
  - Module `-Pcheck` verification: zero Checkstyle violations, Spotless unchanged, build successful.

### Coverage

- Reviewed head: `2c74ccb0abf467d5b14231c7dca4a5732546528b`
- Base and merge-base: `adb3c6fdcbbd28ef8d1eda778b40e8432cddceb2`
- Requirements: [PR #39395](https://github.com/apache/shardingsphere/pull/39395) and [issue #39394](https://github.com/apache/shardingsphere/issues/39394)
- All three authoritative changed files were reviewed: production decoder, focused unit tests, and release notes.
- Reviewed clusters: binary NULL decoding, lossless array-literal reconstruction, regression coverage, and release communication.
- Root-cause/behavior, blast-radius/contracts, tests/runtime/operations, and final convergence passes completed with no unresolved gaps.
- Code-scope review only; CI and GitHub Actions were not reviewed.

PR #39401, opened 2026-08-11
Fix COM_STMT_SEND_LONG_DATA handling in MySQL Proxy

terrymanu:


terrymanu (commented):
### Result

**Review Result: Mergeable**

No blocking issues were found within the Code Correctness Review scope. The implementation matches MySQL’s long-data concatenation, charset conversion, error persistence, and reset lifecycle.

### Evidence

- Connector/J 8.4 sends `Blob`, `Clob`, `InputStream`, and `Reader` values through `COM_STMT_SEND_LONG_DATA`, potentially split across multiple packets. [Connector/J implementation](https://github.com/mysql/mysql-connector-j/blob/8.4.0/src/main/core-impl/java/com/mysql/cj/ServerPreparedQuery.java#L221-L229)
- MySQL 5.7.44, 8.0.36, and 8.4.0 concatenate all bytes before charset conversion, preventing multibyte characters from being corrupted at packet boundaries. [MySQL 8.4 implementation](https://github.com/mysql/mysql-server/blob/mysql-8.4.0/sql/item.cc#L4114-L4143)
- When `max_allowed_packet` is exceeded, MySQL stores the prepared-statement error and reports it on subsequent executions. `COM_STMT_RESET` clears that state. [Error persistence](https://github.com/mysql/mysql-server/blob/mysql-8.4.0/sql/sql_prepare.cc#L2053-L2069), [execute check](https://github.com/mysql/mysql-server/blob/mysql-8.4.0/sql/sql_prepare.cc#L2863-L2870), [reset handling](https://github.com/mysql/mysql-server/blob/mysql-8.4.0/sql/sql_prepare.cc#L1946-L1976)
- The PR’s 64 MiB limit matches the Proxy’s exposed `@@max_allowed_packet=67108864` value. [Proxy system variable](https://github.com/apache/shardingsphere/blob/65b1bf3ba6a5a286384f917cb9c9650c0adc1a9f/proxy/backend/dialect/mysql/src/main/java/org/apache/shardingsphere/proxy/backend/mysql/handler/admin/executor/sysvar/MySQLSystemVariable.java#L422)
- Character parameters are decoded after all chunks are combined. BLOB, NULL, and unknown parameter types remain `byte[]`. Valid long data is cleared after an execution attempt, while an oversized-data error remains until reset. [State implementation](https://github.com/apache/shardingsphere/blob/65b1bf3ba6a5a286384f917cb9c9650c0adc1a9f/proxy/frontend/dialect/mysql/src/main/java/org/apache/shardingsphere/proxy/frontend/mysql/command/query/binary/MySQLServerPreparedStatement.java#L78-L173)
- Commands for one connection are serialized by disabling `autoRead` until `WriteCompleteEvent`. The volatile state publishes changes between executor threads, so the `HashMap` has no concurrent access path. [Command scheduling](https://github.com/apache/shardingsphere/blob/65b1bf3ba6a5a286384f917cb9c9650c0adc1a9f/proxy/frontend/core/src/main/java/org/apache/shardingsphere/proxy/frontend/state/impl/OKProxyState.java#L37-L47)
- Local verification:
  - Four affected test classes: 26 tests passed.
  - MySQL Proxy frontend module: 176 tests passed.
  - Spotless passed.
  - Checkstyle reported zero violations.

### Coverage

- Reviewed head: `65b1bf3ba6a5a286384f917cb9c9650c0adc1a9f`
- Merge base: `6ededf25820a16210baf72cd33b150968eeaac26`
- The authoritative GitHub file list matches the local triple-dot diff: 8/8 files reviewed.
- Covered data ingestion and concatenation, type conversion, size boundaries, sticky error state, successful and failed execution cleanup, reset behavior, thread visibility, memory usage, and the no-long-data fast path.
- Root Cause, Blast Radius, and Tests/Runtime review lenses were completed, followed by a full convergence pass. No unresolved evidence gaps remain.
- Docker/E2E was not run because the protocol behavior is established by official MySQL sources and the production state paths are covered by module tests.
- This is a code-scope result only; CI status was not reviewed.

taojintianxia (approved):

PR #39403, opened 2026-08-11
Fix MySQL SHOW CREATE TABLE replacement for special characters

iamhucong:
- Treat actual and logic table names as literal values when rewriting MySQL `SHOW CREATE TABLE` output.
- Cover dollar signs in both sharding and binding table names with a regression test.

MySQL permits `$` in unquoted identifiers, but `String.replaceFirst` interprets its first argument as a regular expression and its second argument as a replacement template. Quoting both values preserves the existing first-occurrence behavior without treating valid identifier characters as regex syntax.

Verification:
- `./mvnw -pl features/sharding/dialect/mysql -DskipITs test`

Documentation and release notes are not changed because this restores existing `SHOW CREATE TABLE` behavior without changing configuration or public APIs.

### DIFF
--- a/features/sharding/dialect/mysql/src/main/java/org/apache/shardingsphere/sharding/merge/mysql/type/MySQLShardingShowCreateTableMergedResult.java
+++ b/features/sharding/dialect/mysql/src/main/java/org/apache/shardingsphere/sharding/merge/mysql/type/MySQLShardingShowCreateTableMergedResult.java
@@ -37,6 +37,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Optional;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 /**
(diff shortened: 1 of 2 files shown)


iamhucong:
### Code Correctness Review

**Review Result: Mergeable**

The patch fixes the root cause while preserving first-occurrence replacement semantics. No blocking code-correctness issues were found.

Evidence:
- MySQL permits `$` in unquoted identifiers: [MySQL 8.4 Reference Manual §11.2](https://dev.mysql.com/doc/refman/8.4/en/identifiers.html).
- Both regex patterns and replacement values are quoted for sharding and binding table names.
- The regression test exercises both replacements with `$` and verifies the complete rewritten DDL.
- Focused tests (7), module tests (21), Spotless, and Checkstyle passed; target-class JaCoCo CLASS/LINE/BRANCH coverage is 100%.

Reviewed head: `a638bcef973867cc7dfd71ec537d6615aa54ca1b`. This is a code-scope review; CI was not reviewed.

strongduanmu (approved):

FlyingZC (approved):

PR #39406, opened 2026-08-11
Implement MySQLPacketPayload.writeInt6()

terrymanu:
Write six-byte fixed-length integers in little-endian order and add wire-format coverage for high-bit and maximum values.

### DIFF
--- a/database/protocol/dialect/mysql/src/main/java/org/apache/shardingsphere/database/protocol/mysql/payload/MySQLPacketPayload.java
+++ b/database/protocol/dialect/mysql/src/main/java/org/apache/shardingsphere/database/protocol/mysql/payload/MySQLPacketPayload.java
@@ -129,7 +129,8 @@ public final class MySQLPacketPayload implements PacketPayload {
      * @param value 6 byte fixed length integer
      */
     public void writeInt6(final long value) {
-        // TODO
+        byteBuf.writeIntLE((int) value);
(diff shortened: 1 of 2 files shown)


terrymanu:
### Result

**Review Result: Mergeable**

No blocking code-correctness issues were found. The implementation follows MySQL’s six-byte little-endian integer format.

### Evidence

- Root cause: `writeInt6()` had remained an empty TODO implementation since 2018.
- Implementation: It writes the lower 32 bits followed by the upper 16 bits, matching the [MySQL protocol specification](https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_basic_dt_integers.html).
- Tests: The asymmetric high-bit value and maximum 48-bit value cover byte order, truncation, and output length.
- Local verification: All 42 focused tests and 443 module tests passed. Spotless and Checkstyle also passed.

### Coverage

- Reviewed revision: `6c99852e3a0a32ef52faeaf057227c5d17a7581d`
- Changed scope: One production class and one test class, matching the GitHub PR file list.
- Impact: There are currently no in-repository production callers of `writeInt6()`, so the immediate runtime impact is low. However, this fixes a public MySQL protocol-writing capability that external callers may use.
- Reviewed root cause, protocol contract, callers, compatibility, performance, test validity, and resource lifecycle.
- This was a code-correctness review only; CI was not reviewed.

taojintianxia (approved):

PR #39408, opened 2026-08-11
refactor: replace registry-based lock with Redis native lock in GlobalClockTransactionHook [FaaFyfxR9WAQrL7FcAgEHJvztd8cVMxvjHRS55rw1nwH]

### LINKED ISSUE #35041: Refactor the lock implementation in GlobalClockTransactionHook
## Feature Request

Refactor the lock implementation in GlobalClockTransactionHook.

1.  Avoid using registry-based locks (e.g., ZooKeeper) because acquiring a lock might timeout in beforeCommit, but the afterCommit could still attempt to release it. Eliminate unnecessary unlock operations. 

Corresponding FIXME:
```
FIXME: If timeout occurs when a lock is required but TSO is not assigned, the commit will still proceed. The solution is to replace the registry center's lock with a Redis-based lock implementation.
```

2. Move the locking logic to the TSO component (e.g., implement locking/unlocking via Redis's native lock in RedisTSOProvider). Corresponding TODO:

```
// TODO: Use Redis lock instead of registry center's lock.
```

waterWang:
Fixes #35041

### Changes

1. **GlobalClockProvider**: Added `tryLock()`/`unlock()` default methods (no-op by default) for backward compatibility.
2. **RedisTSOProvider**: Implemented Redis native distributed lock using `SET NX PX` with a unique instance token. The lock is acquired in `tryLock()` and released in `unlock()`.
3. **GlobalClockTransactionHook**: Call `tryLock()` in `beforeCommit()` and `unlock()` in `afterCommit()`. If lock acquisition fails, throw `SQLException` to prevent commit without TSO assignment.
4. **ProxyBackendTransactionManager**: Removed the `ExclusiveOperatorEngine` registry-based lock (ZooKeeper) that had a 200ms timeout issue. Added the missing `afterCommit()` hook call after the actual commit.
5. Removed `TransactionCommitOperation` class (no longer needed).

### Why

The registry-based lock (ZooKeeper) could timeout (200ms) and the commit would proceed without a proper TSO assignment. Redis native lock is faster and more reliable. The TODO/FIXME comments in the code explicitly asked for this migration.


terrymanu:
Please fix GitHub action first

PR #39412, opened 2026-08-12
Parse PostgreSQL column NOT NULL constraints

iamhucong:
Fixes PostgreSQL `CREATE TABLE` column nullability metadata.

### Problem

The PostgreSQL grammar accepts column-level `NOT NULL`, but `PostgreSQLDDLStatementVisitor` always constructed `ColumnDefinitionSegment` with `notNull=false`. Consumers therefore could not distinguish `NOT NULL` from nullable or explicitly `NULL` columns.

PostgreSQL documents `NOT NULL` as a column constraint: https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-NOT-NULL

### Change

- Detect a PostgreSQL column constraint containing both `NOT` and `NULL`.
- Preserve `notNull=false` for an explicit `NULL` constraint.
- Add one focused parser case that asserts both boundaries.

### Verification

- RED: the new PostgreSQL parser case failed with `expected true, actual false` before the visitor change.
- `InternalPostgreSQLParserIT`: 1,152 tests passed.
- PostgreSQL parser module: 5 tests passed.
- Spotless: passed.
- Checkstyle: 0 violations.


iamhucong:
### Result

**Review Result: Mergeable**

The current implementation fixes the loss of PostgreSQL `CREATE TABLE` column-level `NOT NULL` metadata. The relevant boundaries are covered, and no code-correctness blockers were found.

### Evidence

- Reviewed head: `c13f08636f74c7db175cd7ed64b48141f624732b`
- `PostgreSQLDDLStatementVisitor` sets `notNull=true` only when the same column constraint contains both `NOT` and `NULL`.
- An explicit `NULL` constraint remains `notNull=false`, so the match is not over-broad.
- PostgreSQL's official `CREATE TABLE` documentation confirms that `NOT NULL` disallows null values while `NULL` permits them and is the default: https://www.postgresql.org/docs/current/sql-createtable.html
- The parser case exercises the public parsing path and verifies both AST boundaries:
  - `order_id INT NOT NULL` -> `not-null="true"`
  - `user_id INT NULL` -> `not-null="false"`
- Local verification:
  - `InternalPostgreSQLParserIT`: 1,152 tests passed.
  - PostgreSQL parser module: 5 tests passed.
  - Spotless passed.

### Coverage

- The reviewed scope matches the complete three-file GitHub change list: the PostgreSQL DDL visitor, SQL input case, and expected parser baseline.
- The review covered the root cause, AST generation path, metadata consumers, adjacent explicit-`NULL` behavior, assertion validity, and regression risk.
- PostgreSQL is an independent trunk dialect in this repository. openGauss has a separate visitor and is unchanged; no shared Oracle, MySQL, or SQLServer parser code is affected.
- No grammar, public API, SPI, dependency, configuration, or generated-source contract changes are present.
- Root-cause, blast-radius, tests/runtime, and final convergence passes found no unresolved evidence gaps.
- This is a code-correctness review only; CI status was not reviewed.

FlyingZC (approved):

PR #39414, opened 2026-08-12
Parse SQLServer column NOT NULL constraints

iamhucong:
Fixes SQLServer `CREATE TABLE` column nullability metadata.

### Problem

The SQLServer grammar accepts column-level `NULL` and `NOT NULL`, but `SQLServerDDLStatementVisitor` always constructed `ColumnDefinitionSegment` with `notNull=false`.

Microsoft documents `{ NULL | NOT NULL }` as part of a SQLServer column definition: https://learn.microsoft.com/en-us/sql/t-sql/statements/create-table-transact-sql

### Change

- Detect a SQLServer column option containing both `NOT` and `NULL`.
- Preserve `notNull=false` for an explicit `NULL` option.
- Add one focused parser case that asserts both boundaries.

### Verification

- RED: the new SQLServer parser case was the only failure among 771 tests before the visitor change (`expected true, actual false`).
- `InternalSQLServerParserIT`: 771 tests passed.
- SQLServer parser module: passed.
- Spotless: passed.
- Checkstyle: 0 violations.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -18,6 +18,7 @@
 1. SQL Parser: Preserve unary NOT as NotExpression for scalar-subquery table extraction in PostgreSQL - [#38187](https://github.com/apache/shardingsphere/pull/38187)
 1. SQL Parser: Fix wrong parameter index parse in MySQL, Doris - [#38624](https://github.com/apache/shardingsphere/pull/38624)
 1. SQL Parser: Fix reversed parameter marker order for openGauss `LIMIT offset, row-count` - [#39242](https://github.com/apache/shardingsphere/pull/39242)
+1. SQL Parser: Preserve SQLServer `CREATE TABLE` column nullability metadata - [#39414](https://github.com/apache/shardingsphere/pull/39414)
 1. SQL Parser: Fix No value specified for parameter exception when sql is 'INSERT INTO tableName ON CONFLICT  DO UPDATE set  WHERE ' - [#38668](https://github.com/apache/shardingsphere/pull/38668)
 1. SQL Binder: Add DialectFunctionOption to handle wrong skip column bind in ColumnSegmentBinder - [#38350](https://github.com/apache/shardingsphere/pull/38350)
 1. SQL Binder: Fix wrong bind info when order by refer column from with temporary table - [#38353](https://github.com/apache/shardingsphere/pull/38353)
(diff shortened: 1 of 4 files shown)


iamhucong:
### Result

**Review Result: Mergeable**

The change correctly preserves SQLServer `CREATE TABLE` column nullability metadata. No code-correctness blockers were found.

### Evidence

- Reviewed public head: `afd23f005f266ee4af1f5c5768e346815391dc61`.
- `SQLServerDDLStatementVisitor` sets `notNull=true` only when one `columnDefinitionOption` contains both `NOT` and `NULL`.
- Explicit `NULL` and omitted nullability remain `false`.
- SQLServer grammar already models the relevant alternatives as `NOT? NULL`; no grammar expansion is needed.
- Microsoft Learn documents `[CONSTRAINT constraint_name] { NULL | NOT NULL }` in SQLServer column definitions: https://learn.microsoft.com/en-us/sql/t-sql/statements/create-table-transact-sql
- The focused parser case exercises the public parsing path and asserts both boundaries:
  - `user_id INT NULL` → `not-null="false"`
- `ColumnDefinitionAssert` reads these expected values and compares them directly with `ColumnDefinitionSegment.isNotNull()`.
  - `InternalSQLServerParserIT`: 771 tests passed.
  - SQLServer parser module passed.
  - Spotless passed.

### Coverage

- The authoritative GitHub scope contains three files, all accounted for:
  - `parser/sql/engine/dialect/sqlserver/src/main/java/org/apache/shardingsphere/sql/parser/engine/sqlserver/visitor/statement/type/SQLServerDDLStatementVisitor.java`
- The local triple-dot file list matches GitHub at merge base `881898ca445598de81bfc711d32acd7c7957e5ad`.
- Reviewed behavior cluster: SQLServer `CREATE TABLE` column-level nullability metadata.
- Root cause, visitor entry path, grammar representation, binder propagation, metadata consumers, explicit-`NULL` boundary, omitted-nullability behavior, expected baseline, and assertion path were checked.
- SQLServer is an independent dialect visitor. No trunk/branch dialect inherits this implementation; MySQL, Oracle, and PostgreSQL use separate equivalent dialect-specific extraction, while SQL92 and openGauss remain separate pre-existing implementations.
- Repository documentation and examples contain no conflicting SQLServer parser contract. No public API, SPI, grammar, dependency, configuration, packaging, or generated-source contract changes are present.
- Root-cause, blast-radius, tests/runtime, and final full-scope convergence passes completed with no unresolved evidence gaps.
- This result is code-scope only; CI was not reviewed.

strongduanmu (dismissed):

FlyingZC (approved):

PR #39418, opened 2026-08-12
fix: add SYSTEM_USER to PostgreSQL unparenthesized function names [FaaFyfxR9WAQrL7FcAgEHJvztd8cVMxvjHRS55rw1nwH]

### LINKED ISSUE #39384: PostgreSQL dialect omits SYSTEM_USER from unparenthesized function names

waterWang:
Fixes #39384

## Changes

Added `SYSTEM_USER` to `PostgreSQLFunctionOption.UNPARENTHESIZED_FUNCTION_NAMES` so that the PostgreSQL dialect correctly binds the niladic function `SYSTEM_USER` instead of misclassifying it as a column reference.

## Root Cause

`PostgreSQLFunctionOption` (line 31-33) lists 11 unparenthesized function names and omits `SYSTEM_USER`. `ColumnSegmentBinder.isUnparenthesizedFunction()` reads this set via `.contains()` to decide function vs column. When the lookup fails, `bind()` throws `ColumnNotFoundException`.

`MySQLFunctionOption` and `SQLServerFunctionOption` already include `SYSTEM_USER`.

## Verification

- `SELECT SYSTEM_USER` against a PostgreSQL-typed schema now correctly binds as a niladic function
- No grammar change required (SYSTEM_USER already lexes as an identifier in PostgreSQL)
- openGauss inherits this change via `OpenGaussDatabaseMetaData` which returns `new PostgreSQLFunctionOption()`

Signed-off-by: waterWang <waterWang@users.noreply.github.com>

### DIFF
--- a/database/connector/dialect/postgresql/src/main/java/org/apache/shardingsphere/database/connector/postgresql/metadata/database/option/PostgreSQLFunctionOption.java
+++ b/database/connector/dialect/postgresql/src/main/java/org/apache/shardingsphere/database/connector/postgresql/metadata/database/option/PostgreSQLFunctionOption.java
@@ -30,7 +30,7 @@ public final class PostgreSQLFunctionOption implements DialectFunctionOption {
     
     private static final Collection<String> UNPARENTHESIZED_FUNCTION_NAMES = new CaseInsensitiveSet<>(Arrays.asList(
             "CURRENT_CATALOG", "CURRENT_DATE", "CURRENT_ROLE", "CURRENT_SCHEMA", "CURRENT_TIME",
-            "CURRENT_TIMESTAMP", "CURRENT_USER", "LOCALTIME", "LOCALTIMESTAMP", "SESSION_USER", "USER"));
+            "CURRENT_TIMESTAMP", "CURRENT_USER", "LOCALTIME", "LOCALTIMESTAMP", "SESSION_USER", "SYSTEM_USER", "USER"));
     
     @Override
     public Collection<String> getUnparenthesizedFunctionNames() {
--- a/database/connector/dialect/postgresql/src/test/java/org/apache/shardingsphere/database/connector/postgresql/metadata/database/option/PostgreSQLFunctionOptionTest.java
+++ b/database/connector/dialect/postgresql/src/test/java/org/apache/shardingsphere/database/connector/postgresql/metadata/database/option/PostgreSQLFunctionOptionTest.java
@@ -37,6 +37,7 @@ class PostgreSQLFunctionOptionTest {
         assertTrue(functionOption.getUnparenthesizedFunctionNames().contains("LOCALTIME"));
         assertTrue(functionOption.getUnparenthesizedFunctionNames().contains("LOCALTIMESTAMP"));
         assertTrue(functionOption.getUnparenthesizedFunctionNames().contains("SESSION_USER"));
+        assertTrue(functionOption.getUnparenthesizedFunctionNames().contains("SYSTEM_USER"));
         assertTrue(functionOption.getUnparenthesizedFunctionNames().contains("USER"));
     }
 }


terrymanu (changes requested):
### Result

**Blocking Issues: 1**

The patch fixes the narrow no-table reproducer, but does not model PostgreSQL 16’s `SYSTEM_USER` semantics correctly.

### Blocking Issues

#### 1. `SYSTEM_USER` remains an ordinary column expression

**Evidence:** PostgreSQL 16 defines `SYSTEM_USER` as a [reserved keyword](https://github.com/postgres/postgres/blob/REL_16_STABLE/src/include/parser/kwlist.h#L422) and its grammar constructs a [`system_user` function call](https://github.com/postgres/postgres/blob/REL_16_STABLE/src/backend/parser/gram.y#L15326-L15338); the [official manual](https://www.postgresql.org/docs/16/functions-info.html) documents the unparenthesized expression. At the reviewed head, PostgreSQL’s [`reservedKeyword` and `functionExprCommonSubexpr`](https://github.com/apache/shardingsphere/blob/303b4deb795598b5ade8ec47108853409730e1e2/parser/sql/engine/dialect/postgresql/src/main/antlr4/imports/postgresql/BaseRule.g4#L30-L108) still omit `SYSTEM_USER`. Exact-head-equivalent parser verification consequently produced `ColumnProjectionSegment` for `SYSTEM_USER`, versus `ExpressionProjectionSegment` for `CURRENT_USER`.

The added metadata entry is only consulted after column resolution in [`ColumnSegmentBinder`](https://github.com/apache/shardingsphere/blob/303b4deb795598b5ade8ec47108853409730e1e2/infra/binder/core/src/main/java/org/apache/shardingsphere/infra/binder/engine/segment/dml/expression/type/ColumnSegmentBinder.java#L97-L103). Therefore, when table metadata contains a quoted `system_user` column, unquoted `SELECT SYSTEM_USER FROM t` is bound as that physical column even though PostgreSQL 16 evaluates the reserved function. The added [membership assertion](https://github.com/apache/shardingsphere/blob/303b4deb795598b5ade8ec47108853409730e1e2/database/connector/dialect/postgresql/src/test/java/org/apache/shardingsphere/database/connector/postgresql/metadata/database/option/PostgreSQLFunctionOptionTest.java#L28-L41) cannot detect this path.

**Impact:** Column-bound processing can treat the function result as table data—for example, [`MaskMergedResult`](https://github.com/apache/shardingsphere/blob/303b4deb795598b5ade8ec47108853409730e1e2/features/mask/core/src/main/java/org/apache/shardingsphere/mask/merge/dql/MaskMergedResult.java#L52-L70) consumes that bound table/column identity. This leaves issue [#39384](https://github.com/apache/shardingsphere/issues/39384) incomplete for valid PostgreSQL 16 schemas.

**Discussion Needed:** Establish whether PostgreSQL ≥16 syntax should be owned by the lexer/parser or handled through a version-aware fallback, including compatibility with older PostgreSQL versions. Also resolve openGauss separately: it inherits [`PostgreSQLFunctionOption`](https://github.com/apache/shardingsphere/blob/303b4deb795598b5ade8ec47108853409730e1e2/database/connector/dialect/opengauss/src/main/java/org/apache/shardingsphere/database/connector/opengauss/metadata/database/OpenGaussDatabaseMetaData.java#L84-L87), while its official documentation describes [`system_user()`](https://docs.opengauss.org/en/docs/3.1.1/docs/Developerguide/system-information-functions.html) as parenthesized and its [keyword table](https://docs.opengauss.org/en/docs/latest-lite/sql_reference/keywords.html) does not classify `SYSTEM_USER` as an openGauss keyword. The agreed model needs parser-to-binder behavioral coverage, including same-name column handling.

### Coverage

- Reviewed head: `303b4deb795598b5ade8ec47108853409730e1e2`; base: `c7bf30fc66598831afe237f71f8a0e8e7f93146b`.
- Accounted for both authoritative GitHub files and linked requirement #39384; public comments and reviews were empty.
- Reviewed clusters: PostgreSQL function classification and its assertion-only test.
- Traced parser → projection → column binder → result-processing paths; checked PostgreSQL, openGauss, and the cited SQL92 issue #39099.
- Completed root-cause, blast-radius/contracts, tests/runtime, and final convergence passes. No evidence gap remains that could change the blocker set.
- Verification: scoped PostgreSQL parser compilation succeeded; runtime parser comparison confirmed the projection-type mismatch.
- Code-correctness review only. GitHub Actions and CI were not reviewed.

PR #39423, opened 2026-08-12
Parse Firebird column NOT NULL constraints

iamhucong:
Fix Firebird `CREATE TABLE` column nullability metadata.

### Problem

The Firebird grammar accepts column-level `NOT NULL`, but `FirebirdDDLStatementVisitor` always constructed `ColumnDefinitionSegment` with `notNull=false`.

Firebird 5.0 documents `NOT NULL` as a column constraint: https://firebirdsql.org/file/documentation/chunk/en/refdocs/fblangref50/fblangref50-ddl-table.html

### Change

- Detect `NOT NULL` in the existing `dataTypeOption` AST.
- Add a focused Firebird parser case covering `NOT NULL` and default nullable behavior.

### Verification

- RED: the new case was the only failure among 234 Firebird parser tests before the visitor change.
- `InternalFirebirdParserIT`: 234 tests passed.
- Firebird parser module passed.
- Spotless passed.
- `git diff --check` passed.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -21,6 +21,7 @@
 1. SQL Parser: Preserve PostgreSQL `CREATE TABLE` column nullability metadata - [#39412](https://github.com/apache/shardingsphere/pull/39412)
 1. SQL Parser: Preserve SQLServer `CREATE TABLE` column nullability metadata - [#39414](https://github.com/apache/shardingsphere/pull/39414)
 1. SQL Parser: Preserve MySQL `ALTER TABLE MODIFY COLUMN` nullability metadata - [#39421](https://github.com/apache/shardingsphere/pull/39421)
+1. SQL Parser: Preserve Firebird `CREATE TABLE` column nullability metadata - [#39423](https://github.com/apache/shardingsphere/pull/39423)
 1. SQL Parser: Fix No value specified for parameter exception when sql is 'INSERT INTO tableName ON CONFLICT  DO UPDATE set  WHERE ' - [#38668](https://github.com/apache/shardingsphere/pull/38668)
 1. SQL Binder: Add DialectFunctionOption to handle wrong skip column bind in ColumnSegmentBinder - [#38350](https://github.com/apache/shardingsphere/pull/38350)
 1. SQL Binder: Fix wrong bind info when order by refer column from with temporary table - [#38353](https://github.com/apache/shardingsphere/pull/38353)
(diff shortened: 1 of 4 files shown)


iamhucong:
### Result

**Review Result: Mergeable**

The change correctly preserves Firebird `CREATE TABLE` column-level `NOT NULL` metadata. I found no blocking code-correctness issues.

### Evidence

- Reviewed public head: `8bb8d67ea5d3c7a1a2b0d55048d098486dd31073`.
- Before this change, `FirebirdDDLStatementVisitor.visitColumnDefinition()` always constructed `ColumnDefinitionSegment` with `notNull=false`.
- The new condition sets `notNull=true` only when the same `dataTypeOption` contains both `NOT` and `NULL`.
- The existing Firebird grammar already models this column constraint as `NOT? NULL`; no grammar change is required.
- The Firebird 5.0 Language Reference confirms that columns are nullable by default and that `NOT NULL` is a column-only constraint: https://firebirdsql.org/file/documentation/chunk/en/refdocs/fblangref50/fblangref50-ddl-table.html
- The Firebird-only parser case exercises the public parser path and verifies both boundaries:
  - `user_id INT` with no nullability clause → `not-null="false"`
- `CreateTableStatementAssert` delegates each column to `ColumnDefinitionAssert`, which directly compares `ColumnDefinitionSegment.isNotNull()`.
  - RED: the new case was the only failure among 234 Firebird parser tests before the visitor fix.
  - GREEN: all 234 Firebird parser tests passed.
  - The Firebird parser module passed.
  - Spotless passed.

### Coverage

- Accounted for all 4 files in the authoritative GitHub changed-file list:
  - `parser/sql/engine/dialect/firebird/src/main/java/org/apache/shardingsphere/sql/parser/engine/firebird/visitor/statement/type/FirebirdDDLStatementVisitor.java`
  - `RELEASE-NOTES.md`
- The local triple-dot file list matched GitHub; merge base: `d8720e1dd15cfdcd5d31753d1397c74e020c9e6c`.
- Reviewed the Firebird DDL visitor entry path, grammar AST, default-nullable boundary, binder propagation, metadata consumers, test assertion path, segment bounds, and the 5.5.4 release-note link.
- `FirebirdDDLStatementVisitor` is an independent final dialect visitor. SQL92, openGauss, and other dialects use separate visitors and are unaffected by this change.
- No public API, SPI, grammar, dependency, configuration, packaging, or generated-source contract changes.
- Root-cause, blast-radius, tests/runtime/operations, and final full-scope convergence reviews completed with no unresolved evidence gaps.
- This is a code-scope review only; CI was not reviewed.

FlyingZC (approved):

PR #39425, opened 2026-08-12
Parse openGauss column nullability constraints

iamhucong:
Fix openGauss `CREATE TABLE` column nullability metadata.

### Problem

The openGauss grammar accepts column-level `NOT NULL` and `NULL`, but `OpenGaussDDLStatementVisitor` always constructed `ColumnDefinitionSegment` with `notNull=false`.

The openGauss 6.0 SQL Reference documents `NOT NULL` as disallowing null values and `NULL` as allowing them: https://docs.opengauss.org/en/docs/6.0.0/docs/SQLReference/create-table.html

### Change

- Detect `NOT NULL` in the existing `columnConstraint` AST.
- Add a focused openGauss parser case covering explicit `NOT NULL` and `NULL` behavior.

### Verification

- RED: the new case was the only failure among 923 openGauss parser tests before the visitor change.
- `InternalOpenGaussParserIT`: 923 tests passed.
- openGauss parser module: 3 tests passed.
- Spotless passed.
- `git diff --check` passed.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -22,6 +22,7 @@
 1. SQL Parser: Preserve SQLServer `CREATE TABLE` column nullability metadata - [#39414](https://github.com/apache/shardingsphere/pull/39414)
 1. SQL Parser: Preserve MySQL `ALTER TABLE MODIFY COLUMN` nullability metadata - [#39421](https://github.com/apache/shardingsphere/pull/39421)
 1. SQL Parser: Preserve Firebird `CREATE TABLE` column nullability metadata - [#39423](https://github.com/apache/shardingsphere/pull/39423)
+1. SQL Parser: Preserve openGauss `CREATE TABLE` column nullability metadata - [#39425](https://github.com/apache/shardingsphere/pull/39425)
 1. SQL Parser: Fix No value specified for parameter exception when sql is 'INSERT INTO tableName ON CONFLICT  DO UPDATE set  WHERE ' - [#38668](https://github.com/apache/shardingsphere/pull/38668)
 1. SQL Binder: Add DialectFunctionOption to handle wrong skip column bind in ColumnSegmentBinder - [#38350](https://github.com/apache/shardingsphere/pull/38350)
 1. SQL Binder: Fix wrong bind info when order by refer column from with temporary table - [#38353](https://github.com/apache/shardingsphere/pull/38353)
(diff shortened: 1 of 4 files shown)


iamhucong:
### Result

**Review Result: Mergeable**

The change correctly preserves openGauss `CREATE TABLE` column-level `NOT NULL` and `NULL` metadata. I found no blocking code-correctness issues.

### Evidence

- Reviewed public head: `3213f60f933fa80763d2bdc08b30186cf7f22b12`.
- Before this change, `OpenGaussDDLStatementVisitor.visitColumnDefinition()` always constructed `ColumnDefinitionSegment` with `notNull=false`.
- The new condition sets `notNull=true` only when the same `columnConstraintOption` contains both `NOT` and `NULL`; explicit `NULL` remains `false`.
- The existing openGauss grammar already models this syntax as `NOT? NULL`; no grammar change is required.
- The openGauss 6.0 SQL Reference confirms that `NOT NULL` disallows null values and `NULL` allows them as the default setting: https://docs.opengauss.org/en/docs/6.0.0/docs/SQLReference/create-table.html
- The openGauss-only parser case exercises the public parser path and verifies both boundaries:
  - `user_id INT NULL` → `not-null="false"`
- `ColumnDefinitionAssert` directly compares `ColumnDefinitionSegment.isNotNull()`, and the binder preserves this value.
- The implementation matches the maintained PostgreSQL sibling visitor for the same grammar structure.
  - RED: the new case was the only failure among 923 openGauss parser tests before the visitor fix.
  - GREEN: all 923 openGauss parser tests passed.
  - The openGauss parser module passed all 3 tests.
  - Spotless passed.

### Coverage

- Accounted for all 4 files in the authoritative GitHub changed-file list:
  - `parser/sql/engine/dialect/opengauss/src/main/java/org/apache/shardingsphere/sql/parser/engine/opengauss/visitor/statement/type/OpenGaussDDLStatementVisitor.java`
  - `RELEASE-NOTES.md`
- The local triple-dot file list matched GitHub; merge base: `e1ec6c953014b346f7786d91cd9edc2e2c7583ab`.
- Reviewed the openGauss DDL visitor entry path, grammar AST, `NOT NULL`/`NULL` boundaries, binder propagation, metadata consumers, test assertion path, segment bounds, PostgreSQL sibling implementation, and the 5.5.4 release-note link.
- `OpenGaussDDLStatementVisitor` is an independent dialect visitor. Other dialects use separate visitors and are unaffected by this change.
- No public API, SPI, grammar, dependency, configuration, packaging, or generated-source contract changes.
- Root-cause, blast-radius, tests/runtime/operations, and final full-scope convergence reviews completed with no unresolved evidence gaps.
- This is a code-scope review only; CI was not reviewed.

FlyingZC (approved):

Issue #14432, opened 2021-12-30
 In the sharding table scenario, the result obtained by avg(distinct intField) is wrong


weichaocst:
## Bug Report

**For English only**, other languages ​​will not accept.

Before report a bug, make sure you have:

-Searched open and closed [GitHub issues](https://github.com/apache/shardingsphere/issues).
-Read documentation: [ShardingSphere Doc](https://shardingsphere.apache.org/document/current/en/overview).

Please pay attention on issues you submitted, because we maybe need more details.
If no response anymore and we cannot reproduce it on current information, we will **close it**.

Please answer these questions before submitting your issue. Thanks!

### Which version of ShardingSphere did you use?
5.0.0
### Which project did you use? ShardingSphere-JDBC or ShardingSphere-Proxy?
ShardingSphere-JDBC
### Expected behavior
The expected result is 100
### Actual behavior
The result obtained is 1
### Reason analyze (If you can)
Table Structure:
CREATE TABLE IF NOT EXISTS t_order (order_id BIGINT NOT NULL AUTO_INCREMENT, user_id INT NOT NULL, intField int(11) DEFAULT NULL, address_id BIGINT NOT NULL, status VARCHAR(50), PRIMARY KEY (order_id))
Fragmentation rules:
spring.shardingsphere.rules.sharding.tables.t_order.actual-data-nodes=ds.t_order_$->{0..3}
spring.shardingsphere.rules.sharding.tables.t_order.table-strategy.standard.sharding-column=order_id
spring.shardingsphere.rules.sharding.tables.t_order.table-strategy.standard.sharding-algorithm-name=t-order-inline

spring.shardingsphere.rules.sharding.tables.t_order.key-generate-strategy.column=order_id
spring.shardingsphere.rules.sharding.tables.t_order.key-generate-strategy.key-generator-name=snowflake

### Steps to reproduce the behavior, such as: SQL to execute, sharding rule configuration, when exception occur etc.
I have a physical table t_order divided into four sub-tables, and there is an intField field of type int in the four sub-tables. There are a total of 100 intField field values ​​of 100 data in the four sub-tables, but logical SQL is executed through ShardingSphere-JDBC: select avg(distinct intField) from t_order, the result obtained is 1, and the expected value should be 100.
### Example codes for reproduce this issue (such as a github link).

strongduanmu:
Hi @weichaocst, thank you for your feedback. Can you try master branch? We have fixed this bug recently.

[...]

PR #39429, opened 2026-08-12
Fix incorrect AVG(DISTINCT) merge result

Gimini-3:
Related to #14432.

  - Increment the `AVG(DISTINCT)` count by one for each newly accepted distinct value instead of adding the value itself.
  - Keep a single `Set` for distinct-value tracking and remove the unreachable zero-count guard.
  - Add focused unit tests for null inputs, cross-shard duplicate values, and the empty merge result.

Root cause:
The sharding rewrite path emits the raw distinct expression for the derived COUNT and SUM projections of `AVG(DISTINCT)`. The merge unit therefore receives the distinct value in both positions. Adding the first value to `count` produces incorrect results such as `(1 + 2) / (1 + 2) = 1.0000` instead of `1.5000`.

Verification:
- `DistinctAverageAggregationUnitTest`: 5 tests passed.
- JaCoCo target coverage: class 100%, line 100%, branch 100%.
- `./mvnw -pl features/sharding/core -Pcheck spotless:check -DskipTests`

---

- [ ] I have passed maven check locally: `./mvnw clean install -B -T1C -Dmaven.javadoc.skip -Dmaven.jacoco.skip -e`.
- [ ] I have updated the Release Notes of the current development version. For more details, see [Update Release Note](https://shardingsphere.apache.org/community/en/involved/contribute/contributor/).

### DIFF
--- a/features/sharding/core/src/main/java/org/apache/shardingsphere/sharding/merge/dql/groupby/aggregation/DistinctAverageAggregationUnit.java
+++ b/features/sharding/core/src/main/java/org/apache/shardingsphere/sharding/merge/dql/groupby/aggregation/DistinctAverageAggregationUnit.java
@@ -21,9 +21,9 @@ import lombok.RequiredArgsConstructor;
 
 import java.math.BigDecimal;
 import java.math.RoundingMode;
-import java.util.Collection;
 import java.util.LinkedHashSet;
 import java.util.List;
+import java.util.Set;
 
 /**
  * Distinct average aggregation unit.
@@ -35,30 +35,28 @@ public final class DistinctAverageAggregationUnit implements AggregationUnit {
     
     private BigDecimal sum;
     
-    private final Collection<Comparable<?>> countValues = new LinkedHashSet<>();
-    
-    private final Collection<Comparable<?>> sumValues = new LinkedHashSet<>();
+    private final Set<Comparable<?>> distinctValues = new LinkedHashSet<>();
     
     @Override
     public void merge(final List<Comparable<?>> values) {
         if (null == values || null == values.get(0) || null == values.get(1)) {
             return;
         }
-        if (countValues.add(values.get(0)) && sumValues.add(values.get(0))) {
+        if (distinctValues.add(values.get(0))) {
             if (null == count) {
                 count = BigDecimal.ZERO;
             }
             if (null == sum) {
                 sum = BigDecimal.ZERO;
             }
-            count = count.add(new BigDecimal(values.get(0).toString()));
+            count = count.add(BigDecimal.ONE);
             sum = sum.add(new BigDecimal(values.get(1).toString()));
         }
     }
     
     @Override
     public Comparable<?> getResult() {
-        if (null == count || BigDecimal.ZERO.compareTo(count) == 0) {
+        if (null == count) {
             return count;
         }
         // TODO use metadata to fetch float number precise for database field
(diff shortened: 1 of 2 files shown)


terrymanu (approved):
### Result
**Review Result: Mergeable**

The PR correctly fixes the `AVG(DISTINCT)` merge arithmetic: `count` now increments by one per newly accepted distinct value instead of accumulating the value itself. Root cause, fix, and tests were verified end to end against the current code; no blocking issue was found. This is a code-scope-only conclusion; CI was not reviewed.

### Evidence
- Root cause confirmed. For `AVG(DISTINCT x)`, `ProjectionEngine.appendAverageDistinctDerivedProjection` (infra/binder/core/.../projection/engine/ProjectionEngine.java:154-168) derives COUNT and SUM `AggregationDistinctProjection`s that carry the raw inner expression `x`. `ShardingProjectionsTokenGenerator.getDerivedProjectionText` (features/sharding/core/.../ShardingProjectionsTokenGenerator.java:106-110) rewrites both derived columns to `x AS AVG_DERIVED_COUNT_n` / `x AS AVG_DERIVED_SUM_n`, and `ShardingRemoveTokenGenerator` drops GROUP BY so shards return distinct rows. Both merge paths therefore call `merge([x, x])` (GroupByMemoryMergedResult.java:91-101, GroupByStreamMergedResult.java:100-107). Under the old code `count` summed the values, so `count == sum` and every non-empty result collapsed to `1.0000`, matching issue #14432 (expected 100, got 1).
- Fix is correct. Since both value positions hold the same per-row distinct value, one `LinkedHashSet` keyed on `values.get(0)` preserves the old deduplication exactly (the old second `sumValues` set always returned the same `add` result, so the `&&` was dead logic). `count.add(BigDecimal.ONE)` yields `count >= 1` whenever non-null, which makes the removed `BigDecimal.ZERO.compareTo(count) == 0` guard genuinely unreachable. The unit now composes the behavior of the sibling units `DistinctCountAggregationUnit` (count distinct values) and `DistinctSumAggregationUnit` (sum distinct values), consistent with the aggregation-unit family.
- Behavior delta check. The only externally visible difference beyond the fix is the single-distinct-value-`0` case, which previously returned scale-0 `0` via the zero-count guard and now returns `0.0000` through the normal divide path; numerically equal and consistent with the scale-4 convention of all other AVG results. Null handling (`values == null`, `values.get(0) == null`, `values.get(1) == null`) and `sum.divide(count, 4, RoundingMode.HALF_UP)` are unchanged. No public API, SPI, signature, or architecture change; the unit is reachable only through `AggregationUnitFactory.create(AVG, true, ...)`.
- Tests are valid and discriminating. `assertGetResultWithDistinctValues` merges `(1,1)`, `(2,2)`, `(1,1)` and asserts `1.5000`; the old code produces `1.0000`, so the test fails before the fix. The parameterized null-input cases and `assertGetResultWithoutMergedValues` protect the null guards in `merge` and `getResult`. Naming (`<ProductionClassName>Test`), `name = "{0}"`, JUnit 5, and ASF header follow repository conventions. The existing empty-result E2E case (`AVG(DISTINCT order_id) ... WHERE 1 = 2`) never invokes the merge unit, so no snapshot or E2E baseline is affected.
- Scope verification. The head `f6a10a3` is a fork commit not present locally; the patch was applied to the locally available base `840c9f080dc` and the reconstructed file's blob hash `e0c20c7a2a0e2` matches the GitHub diff index line, so the reviewed candidate is byte-for-byte the PR head.

### Coverage
Reviewed head `f6a10a3bfcd614208d04aac3553c1be14c37bdde` (1 commit, base master `840c9f080dc`), linked issue #14432, and the authoritative GitHub changed-file list: `features/sharding/core/src/main/java/.../aggregation/DistinctAverageAggregationUnit.java` and the new `DistinctAverageAggregationUnitTest.java` — both mapped to the single behavior cluster "AVG(DISTINCT) merge semantics". All three discovery lenses (root cause and behavior; blast radius and contracts; tests, runtime, and operations) completed across the full scope, including risk triage of contracts, compatibility, concurrency, performance, and generated/E2E artifacts. All candidates were classified; none passed the finding-proof gate (style-level `Set` vs `Collection` declaration and pre-existing `values.get(1)` parsing characteristics are non-blocking and not published as findings). No prior reviews or comments exist; no convergence-gap remains. This result is code-scope only and CI was not reviewed.

PR #39432, opened 2026-08-13
Parse PostgreSQL altered column nullability

iamhucong:
Fix PostgreSQL `ALTER TABLE ... ALTER COLUMN` nullability metadata.

### Problem

The PostgreSQL grammar accepts `SET NOT NULL` and `DROP NOT NULL`, but `PostgreSQLDDLStatementVisitor` always constructed `ColumnDefinitionSegment` with `notNull=false`.

PostgreSQL documents both forms in the official `ALTER TABLE` syntax: https://www.postgresql.org/docs/current/sql-altertable.html

### Change

- Detect `SET NOT NULL` in the existing `modifyColumnSpecification` AST.
- Add focused PostgreSQL parser cases covering `SET NOT NULL` and `DROP NOT NULL`.

### Verification

- RED: the `SET NOT NULL` case was the only failure among 1,154 PostgreSQL parser tests before the visitor change.
- `InternalPostgreSQLParserIT`: 1,154 tests passed.
- PostgreSQL parser module: 5 tests passed.
- Spotless passed.
- `git diff --check` passed.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -25,6 +25,7 @@
 1. SQL Parser: Preserve openGauss `CREATE TABLE` column nullability metadata - [#39425](https://github.com/apache/shardingsphere/pull/39425)
 1. SQL Parser: Preserve Hive `CREATE TABLE` column nullability metadata - [#39428](https://github.com/apache/shardingsphere/pull/39428)
 1. SQL Parser: Preserve SQLServer `ALTER TABLE ALTER COLUMN` nullability metadata - [#39430](https://github.com/apache/shardingsphere/pull/39430)
+1. SQL Parser: Preserve PostgreSQL `ALTER TABLE ALTER COLUMN` nullability metadata - [#39432](https://github.com/apache/shardingsphere/pull/39432)
 1. SQL Parser: Fix No value specified for parameter exception when sql is 'INSERT INTO tableName ON CONFLICT  DO UPDATE set  WHERE ' - [#38668](https://github.com/apache/shardingsphere/pull/38668)
 1. SQL Binder: Add DialectFunctionOption to handle wrong skip column bind in ColumnSegmentBinder - [#38350](https://github.com/apache/shardingsphere/pull/38350)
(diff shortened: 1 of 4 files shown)


iamhucong:
### Code Correctness Review

**Review Result: Mergeable**

The change correctly fixes PostgreSQL `ALTER TABLE ... ALTER COLUMN ... SET/DROP NOT NULL` nullability metadata, with no code-correctness blockers found.

### Evidence

- Reviewed the latest public head: `d4a94f35876fab937ede0617c125783c2591b52c`.
- The PostgreSQL grammar already defines `modifyColumn (SET | DROP) NOT NULL`, while the visitor previously constructed every altered column with `notNull=false`.
- The new condition correctly preserves the relevant boundaries:
  - `SET NOT NULL` → `true`
  - `DROP NOT NULL` → `false`
  - other modify-column forms → `false`
- PostgreSQL documents that these forms change whether a column allows null values and provides both examples in the official [`ALTER TABLE` documentation](https://www.postgresql.org/docs/current/sql-altertable.html).
- The PostgreSQL-only parser cases directly assert the SET and DROP states. The existing TYPE case continues to protect adjacent behavior.
- The assertion path reaches `ColumnDefinitionSegment.isNotNull()`, and the binder preserves the metadata for downstream consumers.
- Verification completed:
  - RED: the new SET case was the only failure among 1,154 PostgreSQL parser tests before the visitor fix.
  - GREEN: all 1,154 PostgreSQL parser tests passed.
  - PostgreSQL parser module: 5 tests passed.
  - Spotless passed.

### Coverage

All four GitHub-listed files were reviewed, and the authoritative GitHub file list matched the local triple-dot scope from merge base `ba4d694eaec3797e7ea04bc975211195a9ef64a6`. The review covered the PostgreSQL visitor entry path, grammar AST, SET/DROP and adjacent TYPE boundaries, binder propagation, metadata consumers, parser assertion path, segment bounds, and the 5.5.4 release note. The related openGauss dialect uses its own grammar, visitor, and database type; the PostgreSQL-only fixtures do not change its behavior. No public API, SPI, grammar, dependency, configuration, packaging, or generated-source contract is changed.

This is a code-scope review; CI was not reviewed.

FlyingZC (approved):

PR #39448, opened 2026-08-13
Fix broadcast transaction outcome assertions

FlyingZC:


FlyingZC:
### Summary

**Review Result: Mergeable**

Reason: [PR #39448](https://github.com/apache/shardingsphere/pull/39448) correctly closes the previous same-connection-only assertion gap and precisely validates broadcast INSERT results. This result is code-scope only; CI was not reviewed by request.

### Evidence

- The rollback path now verifies through an independent connection that the inserted broadcast row is not visible after rollback.
- The commit path independently verifies that the row remains visible after commit.
- Both INSERT operations use `Statement.executeUpdate` and require the logical update count to equal `1`, while closing the statement through try-with-resources.
- JDBC and Proxy deliberately do not accumulate update counts for broadcast tables, so `1` is the stable logical affected-row result rather than the number of physical replicas.
- Focused execution confirmed that the INSERT reached all three physical `t_address` replicas and that the external rollback and commit assertions completed successfully.
- The latest PR contains two commits but has a clean net scope of one transaction E2E case file.

### Review Details

- **Reviewed Scope:** `BroadcastTableTransactionTestCase.java` at latest head `7aeb649d09b1342d2320775f4fd4a25050e2effb`; local merge-base `840c9f080dc0f39bb18c74d39543002205bbbb60`. The local triple-dot file list exactly matches GitHub `/pulls/39448/files`.
- **Not Reviewed Scope:** GitHub Actions and full PostgreSQL/openGauss/Proxy runtime matrices. Their adjacent update-count ownership paths were statically reviewed.
- **Verification:** Spotless check exit 0; Checkstyle exit 0; transaction `test-compile` exit 0; MySQL 5.7 JDBC LOCAL `BroadcastTableTransactionTestCase` passed 1/1.
- **Release Note / User Docs:** Not required because the PR changes only E2E assertions and does not modify public APIs, configuration, dependencies, protocols, or production behavior.

iamhucong (approved):

PR #39503, opened 2026-08-18
Fix statement close invalidating live result sets of other statements

### LINKED ISSUE #39189: Hibernate fails with "Operation not allowed after ResultSet closed" when using ShardingSphere JDBC
# Bug: Hibernate fails with "Operation not allowed after ResultSet closed" when using ShardingSphere JDBC

## Description

The application fails when executing a Spring Data JPA repository query. Hibernate throws a `JpaSystemException` while extracting data from the JDBC `ResultSet`.

The failing repository method:

```java
findByAssignedUserId
```

Error:

```
org.springframework.orm.jpa.JpaSystemException:
Could not extract column [1] from JDBC ResultSet
[Operation not allowed after ResultSet closed]
```

Root cause:

```
java.sql.SQLException: Operation not allowed after ResultSet closed
```

## Environment

- Spring Framework: 7.0.8
- Spring Data JPA: 4.0.6
- Hibernate ORM: 7.2.19.Final
- MySQL Connector/J: 9.7.0
- HikariCP: 7.0.2
- ShardingSphere JDBC: 5.5.3
- Database: MySQL
- Java: 21

## Stack Trace

```
Caused by: java.sql.SQLException:
Operation not allowed after ResultSet closed

at com.mysql.cj.jdbc.result.ResultSetImpl.checkClosed(ResultSetImpl.java:494)
at com.mysql.cj.jdbc.result.ResultSetImpl.getObject(ResultSetImpl.java:1280)
at com.mysql.cj.jdbc.result.ResultSetImpl.getLong(ResultSetImpl.java:882)
at com.zaxxer.hikari.pool.HikariProxyResultSet.getLong(HikariProxyResultSet.java:...)
at org.apache.shardingsphere.infra.executor.sql.execute.result.query.impl.driver.jdbc.type.stream.JDBCStreamQueryResult.getValue(JDBCStreamQueryResult.java:82)
at org.apache.shardingsphere.infra.merge.result.impl.stream.StreamMergedResult.getValue(StreamMergedResult.java:47)
at org.apache.shardingsphere.driver.jdbc.core.resultset.ShardingSphereResultSet.getLong(ShardingSphereResultSet.java:134)
at org.hibernate.type.descriptor.jdbc.BigIntJdbcType$2.doExtract(BigIntJdbcType.java:76)
```

## Steps to Reproduce

1. Configure a Spring Boot application with:
   - Spring Data JPA
   - Hibernate 7.2.x
   - ShardingSphere JDBC 5.5.3
   - MySQL Connector/J 9.7.0

2. Execute a repository query:

```java
List<Entity> findByAssignedUserId(Long assignedUserId);
```

3. Hibernate fails while hydrating the entity result.

## Expected Behavior

Hibernate should successfully read all columns from the JDBC `ResultSet`.

The `ResultSet` should remain open until Hibernate completes processing the query result.

## Actual Behavior

The JDBC `ResultSet` is closed before Hibernate finishes reading data:

```
ResultSet closed
        ↓
Hibernate tries to extract column value
        ↓
SQLException:
Operation not allowed after ResultSet closed
```

## Analysis

The failure appears to occur inside ShardingSphere JDBC streaming result handling:

```
org.apache.shardingsphere.infra.executor.sql.execute.result.query.impl.driver.jdbc.type.stream.JDBCStreamQueryResult
```

Possible compatibility issue between:

- Hibernate 7.2.x result processing
- ShardingSphere JDBC streaming merge implementation
- MySQL Connector/J 9.7 ResultSet lifecycle handling

## Additional Information

The exception happens during Hibernate entity loading:

```
EntitySelectFetchInitializer.resolveInstance
```

The application does not manually close the `ResultSet`.

## Request

Please investigate whether ShardingSphere JDBC 5.5.3 is fully compatible with Hibernate 7.2.x.

The ResultSet lifecycle should be reviewed to ensure the underlying JDBC ResultSet is not closed before Hibernate completes entity hydration.

AryamannSingh7:
Fixes #39189.

  - Gate the auto-commit connection cleanup in `AbstractStatementAdapter#handleAutoCommitAfterExecution` on there being no other statement still registered on the logical connection, so closing one statement no longer closes physical connections that still back another statement's live `ResultSet`.
  - Add `ShardingSphereConnection#hasRegisteredStatementManagers()` to expose that state.
  - Add two regression tests: one reproducing the failure, and one asserting the connections *are* still released once the last statement closes.

### Root cause

This is a 5.5.2 -> 5.5.3 regression introduced by #35834.

`AbstractStatementAdapter#close` calls `safeCloseAndUnregisterStatementManager`, which ends in `handleAutoCommitAfterExecution` -> `DriverDatabaseConnectionManager#clearCachedConnections()`. That closes **every** connection in `cachedConnections`, and that Multimap is owned by the logical connection and shared by all of its statements. Since a `StatementManager` is created per `Statement`, closing a secondary statement closes the physical connections still backing an outer statement's open `ResultSet`.

It is not a Hibernate problem — Hibernate just produces the access pattern that triggers it. It reproduces with plain JDBC and two statements on one connection, which is what the added test does.

The `connection.getAutoCommit()` gate on that call also explains both of the reporter's observations: 5.5.2 works because the path did not exist before #35834, and `@Transactional(readOnly = true)` works because inside a transaction `autoCommit` is false, so the clear never runs.

### Fix

`unregisterStatementManager` already runs immediately before the auto-commit handling, so at that point a non-empty registry means another statement is still open:

```java
protected final void handleAutoCommitAfterExecution(final ShardingSphereConnection connection) throws SQLException {
    if (connection.getAutoCommit() && !connection.hasRegisteredStatementManagers()) {
        connection.getDatabaseConnectionManager().clearCachedConnections();
    }
}
```

This preserves #35834's timely-release goal — the last statement to close still releases everything — and only skips the clear while another statement is genuinely live.

### Verification

Run on `jdbc` against current `master`:

- `mvn -pl jdbc test` — **1216 tests, 0 failures**, including the tests added by #39437 to `StatementAdapterTest` and `DriverDatabaseConnectionManagerTest`.
- `mvn -Pcheck -pl jdbc validate` — 0 Checkstyle violations; `spotless:check` clean.
- A/B checked both ways: with the guard reverted, `assertResultSetRemainsUsableAfterAnotherStatementCloses` fails while `assertCachedConnectionsClearedWhenLastStatementCloses` still passes — so the second test is genuinely guarding the release behaviour rather than being trivially satisfied.

### Two open questions

I raised these on the issue and am happy to go either way:

1. **Test placement** — these currently live in `ShardingSphereDriverTest`. Happy to move them to a dedicated multi-statement lifecycle test class instead, as suggested in the issue.
2. **Granularity** — using open statements as the proxy for "a result set may still be live" is deliberately conservative: a statement left open without a live result set will hold its connections until it closes. Tracking open result sets directly would be tighter but considerably more invasive. Glad to go that way if preferred.

Also happy to add an E2E case if this should be covered beyond the unit level.

---

- [x] I have passed maven check locally (exact scope listed under Verification above).
- [ ] I have made corresponding changes to the documentation. (No documentation impact — internal resource-lifecycle fix.)
- [x] I have updated the Release Notes of the current development version.
- [x] I have disclosed the tool and affected files or scope for any material AI assistance, as required by the [AI-assisted contribution policy].

**AI assistance:** Claude (Claude Code) was used while investigating and implementing this change, covering `AbstractStatementAdapter`, `ShardingSphereConnection` and `ShardingSphereDriverTest`. I have reviewed and verified all of it and can explain every line.

[AI-assisted contribution policy]: https://github.com/apache/shardingsphere/blob/master/AI_POLICY.md

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -37,6 +37,7 @@
 1. DistSQL: Use case-insensitive table name matching in broadcast create and drop executors - [#39200](https://github.com/apache/shardingsphere/pull/39200)
 1. JDBC: Fix stale generated values leaking into prepared statement executeBatch calls without pending batches - [#38160](https://github.com/apache/shardingsphere/pull/38160)
 1. JDBC: Fix MySQL-compatible typed string conversion for `ResultSet#getObject(index, Class<T>)` - [#38444](https://github.com/apache/shardingsphere/pull/38444)
+1. JDBC: Fix statement close invalidating live result sets of other statements on the same connection - [#39503](https://github.com/apache/shardingsphere/pull/39503)
 1. Proxy: Resolve MySQL prepared statement parameter columns for where clause - [#38382](https://github.com/apache/shardingsphere/pull/38382)
 1. Proxy: Resolve PostgreSQL prepared statement parameter types before bind - [#38384](https://github.com/apache/shardingsphere/pull/38384)
 1. Proxy: Resolve openGauss batch bind parameter types before reading values - [#38390](https://github.com/apache/shardingsphere/pull/38390)
--- a/jdbc/src/main/java/org/apache/shardingsphere/driver/jdbc/adapter/AbstractStatementAdapter.java
+++ b/jdbc/src/main/java/org/apache/shardingsphere/driver/jdbc/adapter/AbstractStatementAdapter.java
@@ -83,7 +83,7 @@ public abstract class AbstractStatementAdapter extends WrapperAdapter implements
     }
     
     protected final void handleAutoCommitAfterExecution(final ShardingSphereConnection connection) throws SQLException {
-        if (connection.getAutoCommit()) {
+        if (connection.getAutoCommit() && !connection.hasRegisteredStatementManagers()) {
             connection.getDatabaseConnectionManager().clearCachedConnections();
         }
     }
--- a/jdbc/src/main/java/org/apache/shardingsphere/driver/jdbc/core/connection/ShardingSphereConnection.java
+++ b/jdbc/src/main/java/org/apache/shardingsphere/driver/jdbc/core/connection/ShardingSphereConnection.java
@@ -340,4 +340,13 @@ public final class ShardingSphereConnection extends AbstractConnectionAdapter {
     public void unregisterStatementManager(final StatementManager statementManager) {
         statementManagers.remove(statementManager);
     }
+    
+    /**
+     * Judge whether any statement manager is still registered.
+     *
+     * @return whether any statement manager is still registered
+     */
+    public boolean hasRegisteredStatementManagers() {
+        return !statementManagers.isEmpty();
+    }
 }
--- a/jdbc/src/test/java/org/apache/shardingsphere/driver/ShardingSphereDriverTest.java
+++ b/jdbc/src/test/java/org/apache/shardingsphere/driver/ShardingSphereDriverTest.java
@@ -17,9 +17,12 @@
 
 package org.apache.shardingsphere.driver;
 
+import com.google.common.collect.Multimap;
+import org.apache.shardingsphere.driver.jdbc.core.connection.DriverDatabaseConnectionManager;
 import org.apache.shardingsphere.driver.jdbc.core.connection.ShardingSphereConnection;
 import org.apache.shardingsphere.infra.hint.HintManager;
 import org.junit.jupiter.api.Test;
+import org.mockito.internal.configuration.plugins.Plugins;
 
 import java.sql.Connection;
 import java.sql.Driver;
@@ -148,6 +151,61 @@ class ShardingSphereDriverTest {
             assertTrue(resultSet.next());
             assertThat(resultSet.getInt(1), is(1));
         }
+    }
+    
+    @Test
+    void assertResultSetRemainsUsableAfterAnotherStatementCloses() throws SQLException {
+        try (Connection connection = DriverManager.getConnection("jdbc:shardingsphere:classpath:config/driver/driver-fixture-h2-mysql.yaml")) {
+            try (Statement statement = connection.createStatement()) {
+                statement.execute("DROP TABLE IF EXISTS t_order");
+                statement.execute("CREATE TABLE t_order (order_id INT PRIMARY KEY, user_id INT)");
+                statement.execute("INSERT INTO t_order (order_id, user_id) VALUES (1, 101), (2, 102)");
+            }
+            assertTrue(connection.getAutoCommit());
+            try (Statement queryStatement = connection.createStatement()) {
+                ResultSet resultSet = queryStatement.executeQuery("SELECT order_id, user_id FROM t_order ORDER BY order_id");
+                assertTrue(resultSet.next());
+                assertThat(resultSet.getInt(1), is(1));
+                executeAndCloseSecondaryStatement(connection);
+                assertThat(resultSet.getInt(2), is(101));
+                assertTrue(resultSet.next());
+                assertThat(resultSet.getInt(1), is(2));
+                assertThat(resultSet.getInt(2), is(102));
+            }
+        }
+    }
+    
+    private void executeAndCloseSecondaryStatement(final Connection connection) throws SQLException {
+        try (
+                Statement secondaryStatement = connection.createStatement();
+                ResultSet secondaryResultSet = secondaryStatement.executeQuery("SELECT COUNT(1) FROM t_order")) {
+            assertTrue(secondaryResultSet.next());
+            assertThat(secondaryResultSet.getInt(1), is(2));
+        }
+    }
+    
+    @Test
+    void assertCachedConnectionsClearedWhenLastStatementCloses() throws SQLException, ReflectiveOperationException {
+        try (Connection connection = DriverManager.getConnection("jdbc:shardingsphere:classpath:config/driver/driver-fixture-h2-mysql.yaml")) {
+            try (Statement statement = connection.createStatement()) {
+                statement.execute("DROP TABLE IF EXISTS t_order");
+                statement.execute("CREATE TABLE t_order (order_id INT PRIMARY KEY, user_id INT)");
+                statement.execute("INSERT INTO t_order (order_id, user_id) VALUES (1, 101), (2, 102)");
+            }
+            Statement queryStatement = connection.createStatement();
+            try (ResultSet resultSet = queryStatement.executeQuery("SELECT order_id FROM t_order")) {
+                assertTrue(resultSet.next());
+                assertFalse(getCachedConnections(connection).isEmpty());
+            }
+            queryStatement.close();
+            assertTrue(getCachedConnections(connection).isEmpty());
+        }
+    }
+    
+    private Multimap<?, ?> getCachedConnections(final Connection connection) throws ReflectiveOperationException {
+        DriverDatabaseConnectionManager databaseConnectionManager = ((ShardingSphereConnection) connection).getDatabaseConnectionManager();
+        return (Multimap<?, ?>) Plugins.getMemberAccessor().get(
+                DriverDatabaseConnectionManager.class.getDeclaredField("cachedConnections"), databaseConnectionManager);
     }
     
     @Test


AryamannSingh7:
Could a committer add the `type: bug` and `in: JDBC` labels, matching #39189?

@terrymanu — this implements the fix for the regression you triaged on that issue. The two questions from [my analysis comment](https://github.com/apache/shardingsphere/issues/39189#issuecomment-5108865819) are still open (test-class placement, and whether the conservative open-statements proxy is acceptable); I've noted both in the description and I'm happy to reshape either in review rather than hold the PR on them.

terrymanu (approved):
### Correction

- **Previous Finding:** The PR could newly misbind a same-named `SYSTEM_USER` column and cause incorrect route pruning.
- **Reason:** Column resolution occurs before the unparenthesized-function fallback in [`ColumnSegmentBinder`](https://github.com/apache/shardingsphere/blob/aa93b3d419dfcbae73f1319baefc67ca9fec57a5/infra/binder/core/src/main/java/org/apache/shardingsphere/infra/binder/engine/segment/dml/expression/type/ColumnSegmentBinder.java#L93-L103). A matching column therefore bypasses the changed function-name set, as protected by the existing [column-precedence test](https://github.com/apache/shardingsphere/blob/aa93b3d419dfcbae73f1319baefc67ca9fec57a5/infra/binder/core/src/test/java/org/apache/shardingsphere/infra/binder/engine/segment/dml/expression/type/ColumnSegmentBinderTest.java#L253-L265). The author’s [PostgreSQL 16 base/head reproduction](https://github.com/apache/shardingsphere/pull/39385#issuecomment-5251177694) also shows identical same-column routing behavior. The underlying version-aware name-resolution limitation is pre-existing and is not caused by this PR.

### Result

**Review Result: Mergeable**

The current head fixes the exact `SELECT SYSTEM_USER` binder failure from #39384, preserves existing identifier precedence, and has no confirmed code-correctness blockers.

### Evidence

- PostgreSQL 16 added the `SYSTEM_USER` server variable, and documents bare `system_user` as session information. It is reserved in PostgreSQL 16 but not classified as a PostgreSQL keyword in version 15. [PostgreSQL 16 release notes](https://www.postgresql.org/docs/release/16.0/), [PostgreSQL 16 system functions](https://www.postgresql.org/docs/16/functions-info.html), [PostgreSQL 16 keywords](https://www.postgresql.org/docs/16/sql-keywords-appendix.html), [PostgreSQL 15 keywords](https://www.postgresql.org/docs/15/sql-keywords-appendix.html).
- The production change adds `SYSTEM_USER` to the established case-insensitive metadata set in [`PostgreSQLFunctionOption`](https://github.com/apache/shardingsphere/blob/aa93b3d419dfcbae73f1319baefc67ca9fec57a5/database/connector/dialect/postgresql/src/main/java/org/apache/shardingsphere/database/connector/postgresql/metadata/database/option/PostgreSQLFunctionOption.java#L31-L37). The fallback still applies only after column, outer-column, variable, pivot, and model-column lookup fails; quoted identifiers remain excluded.
- The added [binder scenarios](https://github.com/apache/shardingsphere/blob/aa93b3d419dfcbae73f1319baefc67ca9fec57a5/infra/binder/core/src/test/java/org/apache/shardingsphere/infra/binder/engine/segment/dml/expression/type/ColumnSegmentBinderTest.java#L297-L320) exercise both the reported no-table case and a table without a matching column using the real PostgreSQL SPI. A controlled negative run using the base `PostgreSQLFunctionOption` made both scenarios fail with `ColumnNotFoundException`.
- The PostgreSQL connector is added with [test scope only](https://github.com/apache/shardingsphere/blob/aa93b3d419dfcbae73f1319baefc67ca9fec57a5/infra/binder/core/pom.xml#L65-L70), activating the dialect metadata without changing runtime packaging.
- The parser case correctly records the existing [`ColumnProjectionSegment` representation](https://github.com/apache/shardingsphere/blob/aa93b3d419dfcbae73f1319baefc67ca9fec57a5/test/it/parser/src/main/resources/case/dml/select.xml#L15114-L15118); the established projection binder then delegates it to `ColumnSegmentBinder`.
- PostgreSQL 15 and openGauss compatibility were checked. Because valid columns and variables resolve before the fallback, and quoted identifiers are excluded, the shared function option does not remove a supported identifier path. The openGauss metadata sharing is visible in [`OpenGaussDatabaseMetaData`](https://github.com/apache/shardingsphere/blob/aa93b3d419dfcbae73f1319baefc67ca9fec57a5/database/connector/dialect/opengauss/src/main/java/org/apache/shardingsphere/database/connector/opengauss/metadata/database/OpenGaussDatabaseMetaData.java#L82-L87).
- Verification on an isolated snapshot of the exact head:
  - Initial focused test without upstream modules: exit 1 before candidate tests because of stale local reactor artifacts.
  - Focused tests with `-am`: exit 0; `PostgreSQLFunctionOptionTest` and all 16 `ColumnSegmentBinderTest` scenarios passed.
  - Full PostgreSQL connector and binder-core module tests: exit 0; binder-core reported 340 passing tests.
  - `InternalPostgreSQLParserIT`: exit 0; 1,152 tests passed.
  - Scoped Checkstyle and Spotless checks: exit 0.
  - XML validation for the changed POM and parser resources: exit 0.
- Performance risk is negligible: the change adds one member to an existing small static set without changing algorithmic complexity, call frequency, allocation, I/O, or synchronization.

### Coverage

- **Candidate:** Public PR #39385.
- **Reviewed head:** `aa93b3d419dfcbae73f1319baefc67ca9fec57a5`.
- **Base and merge-base:** `d26c87dc8cac45d910264dba727d54eb105d8532`.
- GitHub’s files endpoint and compare endpoint matched exactly: 7 files, 4 commits, 42 additions, and 1 deletion.
- All changed files were accounted for across four behavior clusters: PostgreSQL function metadata, binder regression/test activation, parser representation, and release documentation.
- Root-cause, blast-radius/contracts, tests/runtime/operations, and final convergence passes were completed.
- PostgreSQL 15/16 semantics, PostgreSQL parser behavior, openGauss metadata sharing, routing consumers, test dependency scope, and release-note impact were checked.
- Code Correctness Review only; CI and GitHub Actions were not reviewed.

terrymanu (approved):
### Result

**Review Result: Mergeable**

The latest head fixes the shared physical-connection lifecycle regression while preserving cleanup after the final statement closes.

### Evidence

- The current statement unregisters before the new registry check, so closing a secondary statement no longer clears connections backing another live result set: [`AbstractStatementAdapter.java:282`](https://github.com/apache/shardingsphere/blob/708f94dc66ea89066d17cd8a1b727015b428649f/jdbc/src/main/java/org/apache/shardingsphere/driver/jdbc/adapter/AbstractStatementAdapter.java#L282-L315), [`ShardingSphereConnection.java:349`](https://github.com/apache/shardingsphere/blob/708f94dc66ea89066d17cd8a1b727015b428649f/jdbc/src/main/java/org/apache/shardingsphere/driver/jdbc/core/connection/ShardingSphereConnection.java#L349-L350).
- The regression and final-statement cleanup paths both have focused coverage: [`ShardingSphereDriverTest.java:157`](https://github.com/apache/shardingsphere/blob/708f94dc66ea89066d17cd8a1b727015b428649f/jdbc/src/test/java/org/apache/shardingsphere/driver/ShardingSphereDriverTest.java#L157-L208).
- A/B verification confirmed causality: the exact head passes; disabling only the new guard reproduces the closed-result-set error.
- `./mvnw -pl jdbc test`: exit 0; 1,216 tests, 0 failures/errors.
- `./mvnw -Pcheck -pl jdbc validate`: exit 0; 0 Checkstyle violations.
- The release note accurately describes the user-visible fix: [`RELEASE-NOTES.md:40`](https://github.com/apache/shardingsphere/blob/708f94dc66ea89066d17cd8a1b727015b428649f/RELEASE-NOTES.md#L40).

### Coverage

- Candidate: public PR #39503, base `7e011c78d85e8caab924f2b50917bc91bd1dd8c0`, head `708f94dc66ea89066d17cd8a1b727015b428649f`.
- All four GitHub-listed files were reviewed and matched their authoritative head blob hashes.
- Behavior clusters: statement/connection lifecycle and regression coverage; release communication.
- Root-cause, blast-radius/contracts, tests/runtime/operations, and final convergence passes completed with no unresolved gaps.

PR #39519, opened 2026-08-19
Support parsing Doris BITOR, ARRAY_LAST, bitmap_to_string, and MAX(DISTINCT|ALL)...OVER

joelle-a-dev:
## Summary

Adds Doris SQL parser support for the items tracked in #31506:

- `BITOR(a, b)`: new bitwise function, following the existing `BITXOR` pattern.
- `ARRAY_LAST(lambda, array)`: new lambda expression grammar (`x -> expr`) and function rule, since Doris had no prior support for lambda-style array functions.
- `MAX([DISTINCT | ALL]) ... OVER (...)`: added the missing `ALL` qualifier alternative to aggregation functions.
- `HEX()`, `LOCATE()`, `bitmap_to_string()`: already parsed correctly via the generic function-call grammar, just added/extended test coverage for Doris.
- `INSTALL PLUGIN FROM`: already fully implemented, no changes needed.

While adding `OVER`-clause test coverage for `MAX`, found that Doris's visitor never attached window-clause info to aggregation projections at all, so any `MAX`/`SUM`/`COUNT`/etc. `... OVER (...)` silently lost its window segment. Added the missing `visitOverClause` and wired `setWindow(...)` into both aggregation projection branches, matching the existing MySQL visitor behavior. This also fixes two pre-existing Doris test cases that had been passing only because they never asserted the (missing) window item.

## Test plan

- [x] `InternalDorisParserIT` passes: 1292/1292, including all new and previously-affected cases
- [x] `checkstyle:check` passes on touched modules
- [x] `spotless:apply` applied, no outstanding formatting diffs


terrymanu (changes requested):
### Result

Feedback Mode: Change Request

**Blocking Issues: 4**

At head `ebea553b5705b5603467fc816fd452c921c5f8c7`, two new function paths corrupt previously correct ASTs, aggregate windows remain incomplete outside the DML visitor, and the new tests cannot detect partition/frame loss.

### Blocking Issues

1. **Handle aggregate `OVER` clauses in the shared Doris visitor**

   - **Evidence:** The shared [`createAggregationSegment`](https://github.com/apache/shardingsphere/blob/ebea553b5705b5603467fc816fd452c921c5f8c7/parser/sql/engine/dialect/doris/src/main/java/org/apache/shardingsphere/sql/parser/engine/doris/visitor/statement/DorisStatementVisitor.java#L1083-L1103) now visits `overClause`, but [`visitOverClause` and `visitWindowSpecification`](https://github.com/apache/shardingsphere/blob/ebea553b5705b5603467fc816fd452c921c5f8c7/parser/sql/engine/dialect/doris/src/main/java/org/apache/shardingsphere/sql/parser/engine/doris/visitor/statement/type/DorisDMLStatementVisitor.java#L582-L612) exist only on `DorisDMLStatementVisitor`. [CREATE/ALTER VIEW](https://github.com/apache/shardingsphere/blob/ebea553b5705b5603467fc816fd452c921c5f8c7/parser/sql/engine/dialect/doris/src/main/java/org/apache/shardingsphere/sql/parser/engine/doris/visitor/statement/type/DorisDDLStatementVisitor.java#L318-L344), [CTAS](https://github.com/apache/shardingsphere/blob/ebea553b5705b5603467fc816fd452c921c5f8c7/parser/sql/engine/dialect/doris/src/main/java/org/apache/shardingsphere/sql/parser/engine/doris/visitor/statement/type/DorisDDLStatementVisitor.java#L665-L705), and [EXPLAIN SELECT](https://github.com/apache/shardingsphere/blob/ebea553b5705b5603467fc816fd452c921c5f8c7/parser/sql/engine/dialect/doris/src/main/java/org/apache/shardingsphere/sql/parser/engine/doris/visitor/statement/type/DorisDALStatementVisitor.java#L706-L729) traverse their SELECT using DDL/DAL visitor instances. An exact-head probe of `CREATE VIEW v AS SELECT MAX(x) OVER (PARTITION BY y) FROM t` produced an aggregation with no window.
   - **Impact:** Valid inline aggregate windows embedded in views, CTAS, or EXPLAIN silently lose partition/order/frame metadata. The PR’s window fix therefore depends on the outer statement type.
   - **Required Change:** Move or delegate `visitOverClause` and `visitWindowSpecification` to `DorisStatementVisitor`, following the maintained [MySQL shared-visitor precedent](https://github.com/apache/shardingsphere/blob/ebea553b5705b5603467fc816fd452c921c5f8c7/parser/sql/engine/dialect/mysql/src/main/java/org/apache/shardingsphere/sql/parser/engine/mysql/visitor/statement/MySQLStatementVisitor.java#L1161-L1189), and add focused DDL/DAL embedding tests.

2. **Preserve BITOR operands as expressions**

   - **Evidence:** `BITOR` is routed into [`bitwiseFunction`](https://github.com/apache/shardingsphere/blob/ebea553b5705b5603467fc816fd452c921c5f8c7/parser/sql/engine/dialect/doris/src/main/antlr4/imports/doris/BaseRule.g4#L1103-L1106), whose [visitor wraps every `expr` using `each.getText()`](https://github.com/apache/shardingsphere/blob/ebea553b5705b5603467fc816fd452c921c5f8c7/parser/sql/engine/dialect/doris/src/main/java/org/apache/shardingsphere/sql/parser/engine/doris/visitor/statement/DorisStatementVisitor.java#L1203-L1210). Exact head/base probes showed:
     - `BITOR(3,5)`: String literals at head versus Integer literals at base.
     - `BITOR(foo,bar)`: String literals at head versus `ColumnSegment`s at base.
     - `BITOR(?,?)`: zero recorded markers at head versus two parameter-marker segments at base.
     
     The [new fixture covers only constants](https://github.com/apache/shardingsphere/blob/ebea553b5705b5603467fc816fd452c921c5f8c7/test/it/parser/src/main/resources/case/dml/select-special-function.xml#L5035-L5045), while Doris defines BITOR operands as [expressions](https://doris.apache.org/docs/dev/sql-manual/sql-functions/scalar-functions/bitwise-functions/bitor/).
   - **Impact:** Column dependencies and prepared-statement markers disappear, and numeric operands can become quoted character literals downstream.
   - **Required Change:** Retain the already-working generic-function route or visit both operands as `ExpressionSegment`s. Add numeric-type, column/nested-expression, and parameter-marker assertions.

3. **Keep ARRAY_LAST lambda representation consistent**

   - **Evidence:** The latest base already accepts lambda expressions through the [generic function grammar](https://github.com/apache/shardingsphere/blob/ebea553b5705b5603467fc816fd452c921c5f8c7/parser/sql/engine/dialect/doris/src/main/antlr4/imports/doris/BaseRule.g4#L1048-L1062). The new single-array special path instead creates a [String `LiteralExpressionSegment`](https://github.com/apache/shardingsphere/blob/ebea553b5705b5603467fc816fd452c921c5f8c7/parser/sql/engine/dialect/doris/src/main/java/org/apache/shardingsphere/sql/parser/engine/doris/visitor/statement/DorisStatementVisitor.java#L1213-L1219), and the [fixture codifies that type](https://github.com/apache/shardingsphere/blob/ebea553b5705b5603467fc816fd452c921c5f8c7/test/it/parser/src/main/resources/case/dml/select-special-function.xml#L5112-L5119). Exact head/base probes showed the lambda changing from `CommonExpressionSegment` to a String literal. Doris permits [one or more array arguments](https://doris.apache.org/docs/dev/sql-manual/sql-functions/scalar-functions/array-functions/array-last/); multi-array ARRAY_LAST falls back to the generic route and remains a common expression, matching the existing [ARRAY_FIRST precedent](https://github.com/apache/shardingsphere/blob/ebea553b5705b5603467fc816fd452c921c5f8c7/test/it/parser/src/main/resources/case/dml/select-special-function.xml#L5414-L5454).
   - **Impact:** The same lambda contract changes AST type solely according to array count, and executable lambda text is misclassified as a SQL literal.
   - **Required Change:** Preserve the generic `CommonExpressionSegment` representation, preferably by removing the redundant special route, and cover both single- and multi-array forms.

4. **Assert the partition and frame fields newly propagated by the PR**

   - **Evidence:** The PR copies partition/order/frame in [`visitOverClause`](https://github.com/apache/shardingsphere/blob/ebea553b5705b5603467fc816fd452c921c5f8c7/parser/sql/engine/dialect/doris/src/main/java/org/apache/shardingsphere/sql/parser/engine/doris/visitor/statement/type/DorisDMLStatementVisitor.java#L601-L610), but both [MAX fixtures](https://github.com/apache/shardingsphere/blob/ebea553b5705b5603467fc816fd452c921c5f8c7/test/it/parser/src/main/resources/case/dml/select-special-function.xml#L5144-L5158) declare empty window expectations despite `PARTITION BY`, while the [frame fixture](https://github.com/apache/shardingsphere/blob/ebea553b5705b5603467fc816fd452c921c5f8c7/test/it/parser/src/main/resources/case/dml/select-window.xml#L151-L162) asserts only ordering. [`ExpectedWindowItem`](https://github.com/apache/shardingsphere/blob/ebea553b5705b5603467fc816fd452c921c5f8c7/test/it/parser/src/main/java/org/apache/shardingsphere/test/it/sql/parser/internal/cases/parser/jaxb/segment/impl/window/ExpectedWindowItem.java#L34-L41) cannot represent a frame, and [`WindowClauseAssert`](https://github.com/apache/shardingsphere/blob/ebea553b5705b5603467fc816fd452c921c5f8c7/test/it/parser/src/main/java/org/apache/shardingsphere/test/it/sql/parser/internal/asserts/segment/window/WindowClauseAssert.java#L72-L97) skips partition checking when the expected collection is empty and never checks the frame. A controlled mutation deleting both new copies still passed all 1,326 Doris parser IT cases.
   - **Impact:** Partition or frame propagation can be removed while every changed aggregate-window test remains green.
   - **Required Change:** Populate the partition expectations and extend the expected model/assertion to verify frame contents, ensuring that independently dropping either field fails.

### Coverage

- Reviewed [PR #39519](https://github.com/apache/shardingsphere/pull/39519) at head `ebea553b5705b5603467fc816fd452c921c5f8c7` against base `b7054656712e4b94541b0a89bf10932b3290766c`, including the latest merge delta, linked [issue #31506](https://github.com/apache/shardingsphere/issues/31506), commits, authoritative eight-file list, and public discussion.
- Accounted for all grammar/visitor files and all four changed parser fixtures: `BaseRule.g4`, `DorisKeyword.g4`, `DorisStatementVisitor.java`, `DorisDMLStatementVisitor.java`, `case/dml/{select-special-function.xml,select-window.xml,select.xml}`, and `sql/supported/dml/select-special-function.xml`.
- Reviewed the BITOR, ARRAY_LAST, aggregate qualifier/window, and generic HEX/LOCATE/BITMAP_TO_STRING clusters through root-cause/behavior, blast-radius/contracts, and tests/runtime/operations lenses.
- Verification: exact-head Doris grammar compiled; `InternalDorisParserIT` passed 1,326/1,326; focused head/base probes reproduced the AST differences; the controlled window-field mutation also passed 1,326/1,326; affected XML and case mappings were valid.
- Unresolved gaps: none capable of changing the blocker set. A final full-scope convergence pass found no additional independent blocker.
- This is a code-scope review only. CI was not reviewed.

terrymanu (changes requested):
### Result

Feedback Mode: Change Request

**Blocking Issues: 1**

At head `874854278743eb730be2c1c6ed1e8c6c9dc13ffa`, all four blockers from the previous review are fixed and verified, but the latest commit adds a named-window expectation that binds to nothing, leaving the new `OVER w` propagation without any regression protection.

### Blocking Issues

1. **Assert the window name or remove the inert expectation**

   - **Evidence:** The updated [`select_window_named_partition_doris`](https://github.com/apache/shardingsphere/blob/874854278743eb730be2c1c6ed1e8c6c9dc13ffa/test/it/parser/src/main/resources/case/dml/select.xml#L7281-L7286) fixture now declares `<window-name>w</window-name>`, but [`ExpectedWindowItem`](https://github.com/apache/shardingsphere/blob/874854278743eb730be2c1c6ed1e8c6c9dc13ffa/test/it/parser/src/main/java/org/apache/shardingsphere/test/it/sql/parser/internal/cases/parser/jaxb/segment/impl/window/ExpectedWindowItem.java#L35-L46) maps only `partition-by`, `order-by`, and the newly added `frame`; it has no window-name element, and [`WindowClauseAssert.assertIs`](https://github.com/apache/shardingsphere/blob/874854278743eb730be2c1c6ed1e8c6c9dc13ffa/test/it/parser/src/main/java/org/apache/shardingsphere/test/it/sql/parser/internal/asserts/segment/window/WindowClauseAssert.java#L73-L84) never checks a window name. The case loader unmarshals with plain JAXB and no schema validation, so the unmapped element is silently dropped. No class in the assert framework handles a window name (verified by search).
   - **Impact:** The `setWindowName` propagation this PR adds in the new shared `visitOverClause` identifier branch can be deleted while every test stays green — the same mutation-blindness class blocked in the previous round, now on the window-name field. The fixture also misleads readers into believing the named-window path is covered.
   - **Required Change:** Add a `window-name` field to `ExpectedWindowItem` and assert it in `WindowClauseAssert.assertIs` alongside the new frame assertion (so dropping `setWindowName` fails `select_window_named_partition_doris`), or remove the inert element from the fixture.

### Coverage

- Reviewed [PR #39519](https://github.com/apache/shardingsphere/pull/39519) at head `874854278743eb730be2c1c6ed1e8c6c9dc13ffa` against base `b7054656712e4b94541b0a89bf10932b3290766c`, covering the authoritative GitHub 13-file list, all 3 commits, PR description, linked issue #31506 scope, and the prior CHANGES_REQUESTED review at `ebea553b`.
- Prior findings re-checked at the latest head: (1) shared-visitor OVER handling — fixed, `visitOverClause`/`visitWindowSpecification` moved to `DorisStatementVisitor`, all six Doris statement visitors extend it, and a CREATE VIEW embedding test with partition assertions was added; (2) BITOR operands — fixed, `visitBitwiseFunction` now uses `getExpressions`, with constant, column/nested-expression, and parameter-marker coverage (the shared change also affects BITXOR, whose existing fixture tolerates the representation change because the literal assert compares stringified values); (3) ARRAY_LAST — fixed, the special single-array route was removed entirely and both forms now use the generic route, matching the maintained `array_first` precedent; (4) partition/frame assertions — fixed, partition expectations populated in all four affected fixtures, `frame` support added to the expected model and assert, and the frame text now asserted with original spacing via `getOriginalText`.
- Behavior clusters reviewed through the root-cause/behavior, blast-radius/contracts, and tests/runtime lenses: BITOR function, ARRAY_LAST generic route, aggregate `ALL` qualifier, aggregate window propagation and shared-visitor scope, window assertion framework, Presto frame text, and the coverage-only HEX/LOCATE additions. All 13 files accounted for; no churn-only files.
- Dialect family: Doris sits in the MySQL family; the `(distinct | all)?` aggregation grammar matches the existing MySQL trunk grammar exactly, and no other dialect imports the Doris grammar or visitor (verified by search). The Presto `getOriginalText` change preserves original frame spacing where the previous `getText()` output was whitespace-stripped; it is a correct-direction fix with no existing fixture asserting Presto frame text, and no regression path found. Official Doris documentation checked for BITOR (integer lhs/rhs) and MAX; the terse `MAX(<expr>)` doc page does not contradict the ALL qualifier given the MySQL-family grammar precedent accepted in the prior round.
- Verification: exact-head Java/g4/XML files inspected via the GitHub API; all new and updated fixture indexes verified programmatically against their SQL strings (all match); mutation detection traced through the assert framework for partition (size and per-column asserts fail on loss) and frame (`assertNotNull` plus text compare fails on loss); a repository-wide join-scan confirmed every Doris aggregation+OVER case declares a window-item expectation. `InternalDorisParserIT` was not rerun locally because the fork head is not fetchable under this review's read-only Git constraint; static verification covered the decisive facts, and the PR's public test plan reports 1,292/1,292 passing at head.
- Unresolved gaps: none capable of changing the blocker set. A full-scope convergence pass after the latest commit found only the blocking issue above.
- This is a code-scope review only. CI was not reviewed.

terrymanu:
Please fix spotless

terrymanu (changes requested):
### Result

Feedback Mode: Change Request

**Blocking Issues: 4**

The current head, `b8b977f22d5124a6019c94f7c124575b2a982bf0`, still fails parser integration tests and Spotless, while the new MAX window semantics are not fully protected by tests.

### Blocking Issues

1. **Reconcile existing expectations with the stricter shared window assertions**

   - **Evidence:** [`WindowClauseAssert`](https://github.com/apache/shardingsphere/blob/b8b977f22d5124a6019c94f7c124575b2a982bf0/test/it/parser/src/main/java/org/apache/shardingsphere/test/it/sql/parser/internal/asserts/segment/window/WindowClauseAssert.java#L73-L112) now interprets omitted `window-name` and `frame` values as requiring the actual values to be null. Existing fixtures such as [`select-window.xml`](https://github.com/apache/shardingsphere/blob/b8b977f22d5124a6019c94f7c124575b2a982bf0/test/it/parser/src/main/resources/case/dml/select-window.xml#L77-L82), [`select-group-by.xml`](https://github.com/apache/shardingsphere/blob/b8b977f22d5124a6019c94f7c124575b2a982bf0/test/it/parser/src/main/resources/case/dml/select-group-by.xml#L486-L512), and [`select-hive.xml`](https://github.com/apache/shardingsphere/blob/b8b977f22d5124a6019c94f7c124575b2a982bf0/test/it/parser/src/main/resources/case/dml/select-hive.xml#L1104-L1111) omit fields that the visitors populate. At the exact head, Doris, MySQL, Hive, PostgreSQL, and openGauss fail 2, 6, 4, 1, and 1 cases respectively.
   - **Impact:** The parser integration tests cannot pass. These failures come from one shared assertion/fixture mismatch rather than fourteen production parser defects.
   - **Required Change:** Keep the strict assertions, add the correct window names and frames to every affected existing fixture, and rerun all internal parser integration tests.

2. **Make the MAX DISTINCT/ALL cases verify the actual AST subtype and parameters**

   - **Evidence:** [`DorisStatementVisitor`](https://github.com/apache/shardingsphere/blob/b8b977f22d5124a6019c94f7c124575b2a982bf0/parser/sql/engine/dialect/doris/src/main/java/org/apache/shardingsphere/sql/parser/engine/doris/visitor/statement/DorisStatementVisitor.java#L1089-L1101) now selects the DISTINCT subtype and populates aggregation parameters. However, [`ProjectionAssert`](https://github.com/apache/shardingsphere/blob/b8b977f22d5124a6019c94f7c124575b2a982bf0/test/it/parser/src/main/java/org/apache/shardingsphere/test/it/sql/parser/internal/asserts/segment/projection/ProjectionAssert.java#L196-L211) checks the DISTINCT expectation only when the actual object is already distinct and never compares aggregation parameters. The [new cases](https://github.com/apache/shardingsphere/blob/b8b977f22d5124a6019c94f7c124575b2a982bf0/test/it/parser/src/main/resources/case/dml/select-special-function.xml#L5240-L5262) contain no parameter expectations, and the JAXB field is mapped to an unused plural `<parameters>` element.
   - **Impact:** Returning an ordinary aggregation for `MAX(DISTINCT ...)`, or removing `getParameters().addAll(...)`, would still pass the new tests. Both fields are consumed by [`AggregationProjectionConverter`](https://github.com/apache/shardingsphere/blob/b8b977f22d5124a6019c94f7c124575b2a982bf0/kernel/sql-federation/compiler/src/main/java/org/apache/shardingsphere/sqlfederation/compiler/sql/ast/converter/segment/projection/impl/AggregationProjectionConverter.java#L77-L84) and [`ColumnExtractor`](https://github.com/apache/shardingsphere/blob/b8b977f22d5124a6019c94f7c124575b2a982bf0/parser/sql/statement/core/src/main/java/org/apache/shardingsphere/sql/parser/statement/core/extractor/ColumnExtractor.java#L266-L274).
   - **Required Change:** Assert the DISTINCT subtype symmetrically, map and compare aggregation parameters using the fixture schema, and add operand expectations to the MAX DISTINCT/ALL cases.

3. **Cover the new `OVER (base_window …)` branch**

   - **Evidence:** The latest commit propagates the parenthesized base-window name at line 1146 of [`visitOverClause`](https://github.com/apache/shardingsphere/blob/b8b977f22d5124a6019c94f7c124575b2a982bf0/parser/sql/engine/dialect/doris/src/main/java/org/apache/shardingsphere/sql/parser/engine/doris/visitor/statement/DorisStatementVisitor.java#L1138-L1150). The current named-window case uses direct `OVER w` and therefore returns through lines 1141–1143 instead ([fixture](https://github.com/apache/shardingsphere/blob/b8b977f22d5124a6019c94f7c124575b2a982bf0/test/it/parser/src/main/resources/case/dml/select.xml#L7310-L7317)). No repository case uses `OVER (w)` or `OVER (w …)`.
   - **Impact:** Removing the new propagation at line 1146 would leave all current tests passing.
   - **Required Change:** Add a Doris case using `OVER (w)` or `OVER (w ORDER BY …) ... WINDOW w AS (...)`, and assert the base-window name and inline window attributes.

4. **Fix the Spotless failures at the current head**

   - **Evidence:** `./mvnw -Pcheck -T1C -pl parser/sql/engine/dialect/doris,parser/sql/engine/dialect/presto,test/it/parser spotless:check` fails at the exact head and reports [`ExpectedWindowItem.java`](https://github.com/apache/shardingsphere/blob/b8b977f22d5124a6019c94f7c124575b2a982bf0/test/it/parser/src/main/java/org/apache/shardingsphere/test/it/sql/parser/internal/cases/parser/jaxb/segment/impl/window/ExpectedWindowItem.java#L35-L49) and [`WindowClauseAssert.java`](https://github.com/apache/shardingsphere/blob/b8b977f22d5124a6019c94f7c124575b2a982bf0/test/it/parser/src/main/java/org/apache/shardingsphere/test/it/sql/parser/internal/asserts/segment/window/WindowClauseAssert.java#L73-L112). This also matches the unresolved maintainer feedback posted after the current head: [“Please fix spotless”](https://github.com/apache/shardingsphere/pull/39519#issuecomment-5390390754).
   - **Impact:** The repository formatting gate fails, and explicit maintainer feedback remains unresolved.
   - **Required Change:** Apply the repository Spotless formatting to these two files and rerun the same check.

### Coverage

- Reviewed head: `b8b977f22d5124a6019c94f7c124575b2a982bf0`; base and merge base: `e343ca1c6d1c4c32e7c92533d1d8f1ed226acac7`. The head was unchanged at final verification.
- Requirements reviewed: [PR #39519](https://github.com/apache/shardingsphere/pull/39519), [Issue #31506](https://github.com/apache/shardingsphere/issues/31506), all prior formal reviews, the latest public comment, and the Doris documentation for [BITOR](https://doris.apache.org/docs/dev/sql-manual/sql-functions/scalar-functions/bitwise-functions/bitor), [ARRAY_LAST](https://doris.apache.org/docs/dev/sql-manual/sql-functions/scalar-functions/array-functions/array-last/), and [MAX window functions](https://doris.apache.org/docs/2.0/sql-manual/sql-functions/window-functions/WINDOW-FUNCTION-MAX).
- Files accounted for: 13/13 changed files—2 grammar files, 3 visitors, 2 assertion/JAXB files, and 6 SQL/expected-result fixtures.
- Behavior clusters reviewed: Doris functions and keywords, BITOR/BITXOR operands, MAX DISTINCT/ALL window ASTs, shared window assertions, embedded DDL window handling, Presto frame text, and all new fixtures.
- Discovery lenses completed: root cause, blast radius/contracts, tests/runtime, and convergence. All earlier findings concerning shared visitor ownership, BITOR operands, ARRAY_LAST structure, and window-field assertions were re-evaluated.
- Verification: `InternalDorisParserIT` reported 1391 tests and 2 failures; the primary dialect suites confirmed 14 related fixture failures; `InternalPrestoParserIT` passed 151/151; Spotless failed.
- This is a code-scope result only. GitHub CI was not queried or evaluated, and no PR code or remote state was modified.

terrymanu (approved):
### Correction

- **Previous Finding:** Add a Doris `OVER (base_window …)` test to cover the window-name propagation branch in `visitOverClause`.
- **Reason:** The finding incorrectly treated grammar reachability in ShardingSphere and the MySQL precedent as evidence of valid Doris syntax without first checking the official Doris grammar. Doris 2.0 supports only inline `OVER (...)` ([parser source](https://github.com/apache/doris/blob/432e71ab1a07cca0940509ca666dc67cd9ad1bf5/fe/fe-core/src/main/cup/sql_parser.cup#L6847-L6875)), and the current Doris parser explicitly states that `windowRef` is unsupported ([parser source](https://github.com/apache/doris/blob/e925e6ae1b2abe5d8d2c2ef1e1677ec14d014d58/fe/fe-sql-parser/src/main/antlr4/org/apache/doris/nereids/DorisParser.g4#L1796-L1805)). Therefore, the previous request to add this test was not justified.

### Result

**Review Result: Mergeable**

The BITOR, ARRAY_LAST, MAX ALL/DISTINCT, and ordinary aggregate-window fixes are ready to merge. Incorrect Doris named-window support already existed in the PR baseline, and this PR does not introduce the corresponding grammar. The cleanup should be completed in a separate follow-up PR.

### Evidence

- The baseline grammar already accepts `OVER identifier` and `OVER (identifier …)` ([`BaseRule.g4`](https://github.com/apache/shardingsphere/blob/e343ca1c6d1c4c32e7c92533d1d8f1ed226acac7/parser/sql/engine/dialect/doris/src/main/antlr4/imports/doris/BaseRule.g4#L1154-L1160)).
- The baseline already contains the invalid Doris named-window case `select_window_named_partition_doris` ([`select.xml`](https://github.com/apache/shardingsphere/blob/e343ca1c6d1c4c32e7c92533d1d8f1ed226acac7/test/it/parser/src/main/resources/sql/supported/dml/select.xml#L261)).
- The new `select_window_named_base_partition_doris` case does not expand the production grammar; it only covers an existing invalid syntax path. Per the maintainer’s decision, it should be cleaned up together with the pre-existing issue in a follow-up PR.
- Recommended follow-up PR:
  1. Audit and remove Doris named-window grammar paths.
  2. Remove `select_window_named_partition_doris`, `select_window_named_base_partition_doris`, and their assertions.
  3. Remove visitor window-name handling that no longer has a production path.
  4. Add rejection coverage for Doris named-window SQL and run the complete parser integration suite.

### Coverage

- Reviewed head: `a7425374f74e099716ca8c74b0d44e6b3d1457ce`; base and merge-base: `e343ca1c6d1c4c32e7c92533d1d8f1ed226acac7`.
- Accounted for all 17 changed files and re-evaluated every prior finding.
- Exact-head internal parser suites passed: 10,604 tests, 0 failures or errors, and 3 skipped.
- Scoped Spotless and Checkstyle checks passed.
- No remaining code-scope blocker. CI was not reviewed.

PR #39549, opened 2026-08-22
Fix binding table lookup for schema-qualified data nodes

terrymanu:
Normalize ShardingTable's private index keys while preserving schema information in actual data nodes.

### DIFF
--- a/features/sharding/core/src/main/java/org/apache/shardingsphere/sharding/rule/ShardingTable.java
+++ b/features/sharding/core/src/main/java/org/apache/shardingsphere/sharding/rule/ShardingTable.java
@@ -202,7 +202,7 @@ public final class ShardingTable {
                 throw new DataNodeGenerateException(each);
             }
             result.add(dataNode);
-            dataNodeIndexMap.put(dataNode, index);
+            dataNodeIndexMap.put(new DataNode(dataNode.getDataSourceName(), (String) null, dataNode.getTableName()), index);
             actualDataSourceNames.add(dataNode.getDataSourceName());
             addActualTable(dataNode.getDataSourceName(), dataNode.getTableName());
             index++;
--- a/features/sharding/core/src/test/java/org/apache/shardingsphere/sharding/rule/ShardingTableTest.java
+++ b/features/sharding/core/src/test/java/org/apache/shardingsphere/sharding/rule/ShardingTableTest.java
@@ -31,6 +31,7 @@ import org.apache.shardingsphere.sharding.spi.ShardingAlgorithm;
 import org.junit.jupiter.api.Test;
 
 import java.util.Arrays;
+import java.util.Collections;
 
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.is;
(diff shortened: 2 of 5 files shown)


terrymanu:
### Result
**Review Result: Mergeable**

Scoped per user instruction to the last commit of PR #39549 only: `4fa30547a33b` (Fix binding table lookup for schema-qualified data nodes). The commit fixes the root cause at the owning contract, and no blocking issue was found.

### Evidence

- Complete root-cause chain: `ShardingTable.generateDataNodes` parses data nodes of the form `ds.schema.table` and previously stored the schema-carrying `DataNode` into `dataNodeIndexMap`; `DataNode.equals` compares `schemaName` (infra/common/src/main/java/org/apache/shardingsphere/infra/datanode/DataNode.java:163), while the map's sole consumer `findActualTableIndex` always probes with a null-schema key (features/sharding/core/src/main/java/org/apache/shardingsphere/sharding/rule/ShardingTable.java:240). The lookup therefore missed and `BindingTableRule.getBindingActualTable` threw `ActualTableNotFoundException` (features/sharding/core/src/main/java/org/apache/shardingsphere/sharding/rule/BindingTableRule.java:65-68). The fix normalizes the map key to a schema-less `DataNode` (ShardingTable.java:205), matching both the other generation path (ShardingTable.java:186) and the probe key shape — a root-cause repair at the owner, not a workaround layer.
- No regression for existing scenarios: for schema-less data nodes the parsed `schemaName` is null, so the normalized key equals the original key and behavior is unchanged; the auto-table and empty-actualDataNodes paths already use schema-less keys.
- Controlled blast radius: `dataNodeIndexMap` is private with `@Getter(AccessLevel.NONE)`, and a repository-wide search confirms `findActualTableIndex` is its only consumer; the `actualDataNodes` list still carries schema, so `getDataNodeGroups` and similar behavior are unchanged, protected by the first assertion of the new test.
- Consumer paths verified: the route engine (features/sharding/core/src/main/java/org/apache/shardingsphere/sharding/route/engine/type/broadcast/ShardingTableBroadcastRouteEngine.java:100), SQL rewrite (features/sharding/core/src/main/java/org/apache/shardingsphere/sharding/rewrite/token/pojo/ShardingTokenUtils.java:50), and MySQL merge (features/sharding/dialect/mysql/src/main/java/org/apache/shardingsphere/sharding/merge/mysql/type/MySQLShardingShowCreateTableMergedResult.java:77) all pass schema-less actual table names through `ShardingRule.getLogicAndActualTablesFromBindingTable`, confirming the schema-less key domain is the correct contract.
- The new test `assertFindActualTableIndex` (features/sharding/core/src/test/java/org/apache/shardingsphere/sharding/rule/ShardingTableTest.java:67) necessarily fails before the fix (probe key and stored key differ in `schemaName`, returning -1 instead of 0) and passes after it, providing valid regression protection.
- Verification: `./mvnw -pl features/sharding/core -DskipITs -Dspotless.skip=true -Dtest=ShardingTableTest test`, exit 0, Tests run: 5, Failures: 0; `./mvnw test -pl features/sharding/core -DskipITs -Dspotless.skip=true`, exit 0, Tests run: 863, Failures: 0, Skipped: 1 (pre-existing skip).

### Coverage

- Reviewed candidate: PR head `4fa30547a33b` (local HEAD is identical; its parent is the PR's first commit `551bca020ba`). Per the user's explicit narrowing, coverage is the last commit's authoritative file list of 2 files (ShardingTable.java, ShardingTableTest.java), both fully reviewed; the PR's other 3 files (Oracle DDL grammar) belong to the first commit and were not reviewed per the user's instruction.
- Behavior clusters: 1 (dataNodeIndexMap key normalization); both files map to it; no churn-only files.
- Discovery lenses: root cause and behavior, blast radius and contracts, and tests/runtime/operations lenses plus the final convergence pass are all complete, with zero new independent candidates.
- Candidate evaluated and excluded: a pathological config listing the same data source and table name under two schemas makes normalized keys overwrite each other, but such a config always returned -1 before the fix (no working behavior to regress), and `getBindingActualTable` only reads `getActualDataNodes().get(index).getTableName()`, which yields the same table name — not a blocker.
- Unresolved gaps: none. CI was not queried or reviewed (Code Correctness Review); this verdict covers code-scope readiness of the last commit only. The PR's first commit is outside this review's scope.

taojintianxia (approved):

PR #39553, opened 2026-08-22
Fix schema loss when mapping data nodes

terrymanu:


terrymanu:
### Result

**Review Result: Mergeable**

The patch correctly preserves schema identity while mapping data-source names, with focused regression coverage and no confirmed correctness issues.

### Evidence

- [`DataNodeUtils.buildDataNode`](https://github.com/apache/shardingsphere/blob/3e616bc97e1b704f1c8f5cda804feed5e5c456fd/infra/common/src/main/java/org/apache/shardingsphere/infra/datanode/DataNodeUtils.java#L58-L66) now changes only the data-source coordinate and retains the original schema and table.
- The mapper contract supplies only data-source-name mappings, with no schema transformation contract.
- [The regression test](https://github.com/apache/shardingsphere/blob/3e616bc97e1b704f1c8f5cda804feed5e5c456fd/infra/common/src/test/java/org/apache/shardingsphere/infra/datanode/DataNodeUtilsTest.java#L49-L55) compares complete mapped nodes and fails against the previous schema-dropping implementation.
- Scoped Maven verification passed: 38 tests, 0 failures, 0 errors.

### Coverage

- Reviewed head: `3e616bc97e1b704f1c8f5cda804feed5e5c456fd`
- Base and merge-base: `7befb44ce57f620a3ab6e6dfa6feb6720cd94523`
- All 2 authoritative changed files were accounted for; GitHub and local triple-dot file lists matched.
- Requirement: “Fix schema loss when mapping data nodes”; no linked issue or public review discussion exists.
- Behavior cluster: schema-preserving data-source expansion through read/write-splitting and shadow mappings.
- Root-cause/behavior, blast-radius/contracts, tests/runtime/operations, risk triage, and final convergence passes completed with no unresolved gaps.
- This is a code-correctness result only; CI was not reviewed.

taojintianxia (approved):

PR #39558, opened 2026-08-22
Fix schema-qualified data node validation in Sharding DistSQL

terrymanu:


terrymanu:
### Result

**Review Result: Mergeable**

The fix is correct and minimal: `isValidDataNode` in the sharding DistSQL checker now accepts 3-segment (schema-qualified) data nodes, aligning with the existing parsing contract of the infra-level `DataNode` and removing a 2-segment restriction unique to the DistSQL path. The added test covers the regression path, and local verification passes.

### Evidence

- Change: `features/sharding/distsql/handler/src/main/java/org/apache/shardingsphere/sharding/distsql/handler/checker/ShardingTableRuleStatementChecker.java:209-215` changes `isValidDataNode` from accepting only 2 segments to accepting 2 or 3 segments.
- Contract alignment: `infra/common/src/main/java/org/apache/shardingsphere/infra/datanode/DataNode.java:56-62` natively parses `ds.schema.table` (3 segments) and rejects malformed structures; the YAML configuration path (`getRequiredDataSources(ShardingRuleConfiguration)`, lines 337-344 of the same checker) uses `new DataNode(...)` directly and never had the 2-segment restriction. The DistSQL checker was the only inconsistent path, and this PR removes that inconsistency.
- End-to-end path check: `isValidDataNode` has exactly two call sites (lines 204 and 510), serving the storage-unit existence check (`checkDataSources`) and the none-strategy table-name check (`checkStrategy`) respectively; CREATE and ALTER share the same `check` entry, so both are covered. Before the fix, the whole string `ds_0.foo_schema.t_order_0` was treated as a data source name and inevitably triggered `MissingRequiredStorageUnitsException`.
- Missed-surface scan: none of the 6 feature DistSQL handlers contains another dot-delimited 2-segment validation (the remaining ones are comma-based binding-table reference parsing), so there is no similar gap left behind.
- Test validity: `features/sharding/distsql/handler/src/test/java/.../checker/ShardingTableRuleStatementCheckerTest.java:314-319` runs `checkCreation` with `ds_${0..1}.foo_schema.t_order_0` and a none table strategy, covering both call sites; by code tracing, this case would inevitably fail before the fix due to the missing-storage-unit exception, so it provides real regression protection.
- Local verification: `./mvnw -pl features/sharding/distsql/handler -DskipITs -Dspotless.skip=true -Dtest=ShardingTableRuleStatementCheckerTest test` → exit 0, Tests run: 27, Failures: 0.
- Boundary check: 2-segment behavior is unchanged; 4 or more segments are still rejected; malformed input (for example an empty segment like `a..b.c`) fails before and after the fix, with the exception changing from `MissingRequiredStorageUnitsException` to the `DataNode`-owned `InvalidDataNodeFormatException`, which is semantically more accurate and not a regression.

### Coverage

- Reviewed head: `f9e875edf25` (base and merge-base both `6a9503300a7`, single commit, no review/comment history); local HEAD equals the PR head, and the local triple-dot file list matches the GitHub changed-file list exactly (2/2).
- Behavior clusters: A) schema-qualified data node validation semantics (production file); B) regression test (test file). Both files are mapped and risk-triaged; no churn-only files.
- All three discovery lenses completed (root cause and behavior / blast radius and contracts / tests and runtime); every candidate is classified with no blockers; the convergence pass after the latest delta found no new candidates.
- Evidence gaps: none. This conclusion is code-scope readiness only (Code Correctness Review); CI was not reviewed.
- GitHub access route: `GITHUB_TOKEN` from the interactive zsh environment → direct REST API.

taojintianxia (approved):

PR #39573, opened 2026-08-23
Support parsing Doris VARCHAR without field length and add Doris function parser IT cases

terrymanu:
Fixes #31484

PR #39575, opened 2026-08-23
Restrict optional field length to VARCHAR only for Doris dataType rule

terrymanu:


terrymanu:
### Result

**Review Result: Mergeable**

The PR narrows the over-broad `fieldLength` optionality introduced by #39573 in the Doris `dataType` rule down to the `VARCHAR` alternative only, restoring the pre-regression rejection of bare `CHAR VARYING` / `CHARACTER VARYING` while keeping bare `VARCHAR` and `VARCHAR(M)`, both officially supported by Doris. Grammar boundaries, dialect-family consistency, positive/negative test coverage, and local verification at the PR head all pass; no blocking issues were found.

### Evidence

- Scope and metadata: PR head `41c692ece76061b0af2809b5ba62145a4e8df2c5` (1 commit, 4 files, +15/-1, author terrymanu, empty PR body, no reviews/comments); base `75462d75c47` (master). The GitHub `/pulls/39575/files` list and the local merge-base triple-dot file list match exactly (4/4), with no scope gap.
- Root cause and fix mapping: #39573 (master `be8bf02e51c`) changed `(CHAR_VARYING | CHARACTER_VARYING | VARCHAR) fieldLength` to a blanket `fieldLength?`, newly accepting both bare varying aliases on master (the removed line in the PR diff is direct evidence of the base state). This PR splits the alternative in `parser/sql/engine/dialect/doris/src/main/antlr4/imports/doris/BaseRule.g4:1428-1431` into `(CHAR_VARYING | CHARACTER_VARYING) fieldLength charsetWithOptBinary?` and `VARCHAR fieldLength? charsetWithOptBinary?`, keeping the change inside the existing `DORIS CHANGED` branch-dialect markers per repository convention.
- Official semantics (Doris master `fe/fe-sql-parser/src/main/antlr4/org/apache/doris/nereids/DorisParser.g4`): `primitiveColType` contains `type=VARCHAR` with the length parentheses wholly optional (`(LEFT_PAREN ... RIGHT_PAREN)?`), so both bare `VARCHAR` and `VARCHAR(M)` are valid; the grammar contains no VARYING alias production at all, meaning real Doris does not accept `CHAR VARYING` with or without a length. The Doris 3.x official VARCHAR page (doris.apache.org/docs/3.x/sql-manual/basic-element/sql-data-types/string-type/VARCHAR/) defines only `VARCHAR(M)` with M in 1-65533 and likewise defines no varying alias. Hence: rejecting the bare aliases matches both real Doris and the MySQL trunk behavior; the with-length aliases are not official Doris syntax but were pre-existing ShardingSphere Doris parser behavior before #39573 (inherited from the MySQL trunk), and retaining them fits this PR's boundary of fixing only the regression without a breaking narrowing.
- Lexer and visitor: `CHAR_VARYING` / `CHARACTER_VARYING` are single tokens containing one space (`DorisKeyword.g4:306,314`); a multi-space input lexes as `CHAR` + `VARYING`, and the leftover token likewise fails the parse, so the boundary is consistent. Both new alternatives keep the `dataTypeName` label, which `DorisStatementVisitor` reads via the shared token field `ctx.dataTypeName`; no visitor change is required.
- Dialect family: the MySQL trunk `BaseRule.g4:830` still requires `fieldLength` for all three tokens and this PR touches no MySQL file; MariaDB reuses the MySQL parser with no separate grammar; the CAST path, the `NATIONAL VARCHAR` family, `LONG_CHAR_VARYING`, and `CHAR fieldLength?` adjacent alternatives are all unaffected.
- Test validity: two new Doris unsupported negative cases assert `SQLParsingException` (`unsupported_create_table_with_char_varying_without_length_for_doris` / `..._character_varying_...`, `test/it/parser/src/main/resources/sql/unsupported/unsupported.xml:32-33`); the mechanism is guaranteed by `SQLParserExecutor` converting an ErrorNode root or `ParseCancellationException` into `SQLParsingException`, and these cases would necessarily fail on the base grammar, so the regression protection is real. One new positive case `create_table_with_varying_alias_with_length_doris` (`CHAR VARYING(10)` + `CHARACTER VARYING(20)`) guards against accidentally dropping alias support; every start/stop index in its expected tree was verified character by character (`test/it/parser/src/main/resources/case/ddl/create-table.xml:2675-2684`). Bare `VARCHAR` and `VARCHAR(n)` are already covered by existing cases (e.g. `create_table_with_varchar_without_length_doris`), so no duplicates were added. A full scan of the test resources confirms no other Doris case uses a varying alias, so the narrowing breaks no existing scenario.
- Local verification (PR head, clean working tree): `./mvnw -pl parser/sql/engine/dialect/doris clean install -DskipTests` (with ANTLR regeneration) exited 0; `InternalDorisParserIT` passed 1349/1349 and `InternalUnsupportedDorisParserIT` passed 12/12, exit code 0.

### Coverage

- Reviewed head: `41c692ece76061b0af2809b5ba62145a4e8df2c5` (single-commit PR, no commits after creation, no prior review threads; the convergence pass covers the latest delta).
- Requirements and files: applicable requirements are the narrowing goal stated in the PR title and the regression context of #39573 (umbrella issue #31484 "Support parsing Doris sql"); all 4 authoritative GitHub files map to the single behavior cluster doris-datatype-fieldlength (1 production file reviewed, 3 test resources test-only-reviewed), with no churn-only files.
- Discovery lenses: root-cause, blast-radius, and tests-runtime lenses plus the convergence pass are all complete, with zero findings passing the Finding Proof Gate; candidates examined and dismissed include index miscalculation, adjacent-alternative regressions, CAST-path impact, missing visitor labels, breakage of existing cases, and the multi-space lexical boundary.
- Dialects and documentation: target dialect Doris (a branch dialect of the MySQL trunk); the MySQL trunk rule and the MariaDB reuse relationship were checked; official evidence is the Doris master DorisParser.g4 and the Doris 3.x official VARCHAR page; ShardingSphere repository docs/examples do not enumerate per-type fieldLength rules, so no updates are needed.
- This conclusion is a code-scope review only: GitHub Actions / CI state was not queried, awaited, or reported.

taojintianxia (approved):

PR #39603, opened 2026-08-26
Close every statement in JDBCDataRowEnumerator even when one of them fails

PDGGK:

  - Close every statement in `JDBCDataRowEnumerator.close()`, not just the ones before the first failure.

The close loop was inside the `try` block:

```java
public void close() {
    try {
        for (Statement each : statements) {
            each.close();
        }
        currentRow = null;
        throw new SQLWrapperException(ex);
    }
}
```

`statements` holds one statement per data source the federated query touched, so when the first one throws the loop aborts and the remaining statements — and the server side cursors behind them — are never closed. `currentRow` is left dangling too.

Each statement is now closed independently. The first failure is kept and later ones are attached to it with `addSuppressed`, following `MCPJdbcStatementExecutor#appendFailure`, so the `SQLWrapperException` a caller already sees is unchanged and nothing is silently dropped.

Two tests in `JDBCDataRowEnumeratorTest`:

  - `assertCloseClosesRemainingStatementsAfterFailure` — the first of three statements throws; the other two are still closed and the original exception still surfaces. Without the change this fails with `Wanted but not invoked: secondStatement.close()`.
  - `assertCloseSuppressesLaterFailures` — two failing statements; the first is the cause and the second is suppressed rather than lost.

The three existing tests are unchanged and still pass.

---

### DIFF
--- a/kernel/sql-federation/executor/src/main/java/org/apache/shardingsphere/sqlfederation/executor/enumerable/enumerator/jdbc/JDBCDataRowEnumerator.java
+++ b/kernel/sql-federation/executor/src/main/java/org/apache/shardingsphere/sqlfederation/executor/enumerable/enumerator/jdbc/JDBCDataRowEnumerator.java
@@ -97,7 +97,9 @@ public final class JDBCDataRowEnumerator implements Enumerator<Object> {
         if (null == previousFailure) {
             return failure;
         }
-        previousFailure.addSuppressed(failure);
+        if (previousFailure != failure) {
+            previousFailure.addSuppressed(failure);
+        }
         return previousFailure;
     }
 }
--- a/kernel/sql-federation/executor/src/main/java/org/apache/shardingsphere/sqlfederation/executor/enumerable/enumerator/jdbc/JDBCDataRowEnumerator.java
+++ b/kernel/sql-federation/executor/src/main/java/org/apache/shardingsphere/sqlfederation/executor/enumerable/enumerator/jdbc/JDBCDataRowEnumerator.java
@@ -79,13 +79,25 @@ public final class JDBCDataRowEnumerator implements Enumerator<Object> {
     
     @Override
     public void close() {
-        try {
-            for (Statement each : statements) {
+        SQLException failure = null;
+        for (Statement each : statements) {
+            try {
                 each.close();
+            } catch (final SQLException ex) {
+                failure = appendFailure(failure, ex);
             }
-            currentRow = null;
-        } catch (final SQLException ex) {
-            throw new SQLWrapperException(ex);
         }
+        currentRow = null;
+        if (null != failure) {
+            throw new SQLWrapperException(failure);
+        }
+    }
+    
+    private SQLException appendFailure(final SQLException previousFailure, final SQLException failure) {
+        if (null == previousFailure) {
+            return failure;
+        }
+        previousFailure.addSuppressed(failure);
+        return previousFailure;
     }
 }
--- a/kernel/sql-federation/executor/src/test/java/org/apache/shardingsphere/sqlfederation/executor/enumerable/enumerator/jdbc/JDBCDataRowEnumeratorTest.java
+++ b/kernel/sql-federation/executor/src/test/java/org/apache/shardingsphere/sqlfederation/executor/enumerable/enumerator/jdbc/JDBCDataRowEnumeratorTest.java
@@ -111,6 +111,22 @@ class JDBCDataRowEnumeratorTest {
     }
     
     @Test
+    void assertCloseContinuesWhenTwoStatementsThrowTheSameException() throws SQLException {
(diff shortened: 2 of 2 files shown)


terrymanu (changes requested):
### Result

Feedback Mode: Change Request

**Blocking Issues: 1**

The cleanup loop can still terminate before closing every statement.

### Blocking Issues

1. **Self-suppression can abort statement cleanup**

   **Evidence:** [`appendFailure`](https://github.com/apache/shardingsphere/blob/ebf8f32fd8a03faaf35628887d1d2701b5b8681e/kernel/sql-federation/executor/src/main/java/org/apache/shardingsphere/sqlfederation/executor/enumerable/enumerator/jdbc/JDBCDataRowEnumerator.java#L96-L101) unconditionally calls `previousFailure.addSuppressed(failure)`. If two statements throw the same `SQLException` instance, this attempts self-suppression. Java specifies that [`Throwable.addSuppressed`](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/Throwable.html#addSuppressed%28java.lang.Throwable%29) throws `IllegalArgumentException` in that case. The added test at `JDBCDataRowEnumeratorTest.java:114` uses two distinct exceptions and does not cover this path.

   **Impact:** The second failure aborts the loop, later statements remain open, `currentRow` is not cleared, and callers receive `IllegalArgumentException` instead of the established `SQLWrapperException`.

   **Required Change:** Avoid calling `addSuppressed` when both references identify the same exception, while continuing cleanup and preserving the first failure. Add a focused regression test where two closes throw one shared `SQLException` instance and a later statement is still closed.

### Coverage

- Candidate: public PR #39603 at `ebf8f32fd8a03faaf35628887d1d2701b5b8681e`.
- Authoritative scope: both GitHub-listed files reviewed.
- Behavior cluster: JDBC federation statement cleanup, failure aggregation, row-state cleanup, and focused tests.
- Completed root-cause, blast-radius/contracts, tests/runtime/operations, resource-lifecycle, exception-contract, and performance-risk review.
- Latest-head convergence pass found no additional independent issues or unresolved evidence gaps.

PDGGK:
You're right, and it's worse than the original problem: the self-suppression aborts the very loop this change added, so later statements stay open and the caller gets IllegalArgumentException instead of SQLWrapperException.

Fixed in 5c61e414 with a same-instance guard in appendFailure, plus a regression test where two statements throw one shared SQLException and a third is still closed. Reverting the guard fails that test with "Self-suppression not permitted" replacing SQLWrapperException, which is exactly the path you described.

mvn test -pl kernel/sql-federation/executor - 24 tests, all passing. Checkstyle clean.

terrymanu (approved):
### Result

**Review Result: Mergeable**

The current head closes every statement after an `SQLException`, preserves the first failure and established wrapper contract, and resolves the previous self-suppression blocker.

### Evidence

- [`JDBCDataRowEnumerator.java:81`](https://github.com/apache/shardingsphere/blob/5c61e414fab0509738b79f54f75053c5ae1dfa10/kernel/sql-federation/executor/src/main/java/org/apache/shardingsphere/sqlfederation/executor/enumerable/enumerator/jdbc/JDBCDataRowEnumerator.java#L81-L104) independently closes each statement, clears `currentRow`, wraps the first failure, suppresses later failures, and safely ignores duplicate exception instances.
- [`JDBCDataRowEnumeratorTest.java:98`](https://github.com/apache/shardingsphere/blob/5c61e414fab0509738b79f54f75053c5ae1dfa10/kernel/sql-federation/executor/src/test/java/org/apache/shardingsphere/sqlfederation/executor/enumerable/enumerator/jdbc/JDBCDataRowEnumeratorTest.java#L98-L142) covers continued cleanup, cleared row state, same-instance failures, and multiple distinct failures.
- The [previous change request](https://github.com/apache/shardingsphere/pull/39603#pullrequestreview-5056712399) is fixed by the identity guard and its focused regression test.
- `./mvnw test -pl kernel/sql-federation/executor`: exit 0; 24 tests passed.
- Scoped Checkstyle and Spotless verification: exit 0; no violations or formatting changes required.

### Coverage

- Candidate: public PR #39603 at `5c61e414fab0509738b79f54f75053c5ae1dfa10`; final head recheck was unchanged.
- Base and merge-base: `2e1eb7d8a1f1db529c2079138dcc8658636d8fcf`.
- Requirements: PR description; no linked issue.
- Authoritative scope: both GitHub-listed files reviewed.
- Behavior cluster: JDBC federation statement cleanup, failure aggregation, row-state cleanup, and focused regression tests.
- Completed root-cause, blast-radius/contracts, tests/runtime/operations, resource-lifecycle, exception-contract, shared-state, and performance-risk review.
- Full-scope convergence found no additional independent issues or unresolved evidence gaps.

PR #39617, opened 2026-08-27
Remove unsupported Doris named-window grammar and tests

joelle-a-dev:
## Summary

Doris has no WINDOW clause and no named-window OVER reference, only inline `OVER (...)`. This was inherited unchanged from the MySQL-derived Doris dialect baseline and flagged as a follow-up in review on #39519.

- Removed the `identifier`/named-window alternatives from `overClause` and `windowingClause`, and removed `windowClause`/`windowItem` (`WINDOW w AS (...)`) from the Doris grammar.
- Removed the corresponding dead visitor code (`visitWindowClause`,`visitWindowItem`, window-name handling in `visitOverClause` /`visitWindowSpecification` / `visitWindowingClause`).
- Removed the two named-window test cases and their assertions, and dropped `Doris` from a shared MySQL/Doris/Hive case that relied on the same unsupported syntax.
- Added rejection coverage for named-window SQL under Doris.

## Test plan

- [x] `InternalDorisParserIT`: 1388/1388
- [x] `InternalUnsupportedDorisParserIT`: 16/16
- [x] Full `test/it/parser` module: 0 failures, 0 errors
- [x] `checkstyle:check` / `spotless:check` clean on touched modules

### DIFF
--- a/parser/sql/engine/dialect/doris/src/main/antlr4/imports/doris/BaseRule.g4
+++ b/parser/sql/engine/dialect/doris/src/main/antlr4/imports/doris/BaseRule.g4
@@ -1163,11 +1163,11 @@ all
 // DORIS ADDED END
 
 overClause
-    : OVER (windowSpecification | identifier)
+    : OVER windowSpecification
     ;
 
 windowSpecification
-    : LP_ identifier? (PARTITION BY expr (COMMA_ expr)*)? orderByClause? frameClause? RP_
+    : LP_ (PARTITION BY expr (COMMA_ expr)*)? orderByClause? frameClause? RP_
     ;
 
 frameClause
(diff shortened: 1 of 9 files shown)


terrymanu:
Hi @joelle-a-dev, thanks for the PR and for the detailed write-up — your analysis was exactly right: Doris doesn't support a WINDOW clause or named-window OVER references, and that grammar was indeed inherited from the MySQL baseline.

Heads-up that this has already landed on master via #39622 (merged on Aug 28). It removes the same named-window grammar and visitor code, drops the same test cases, and adds rejection coverage for named-window SQL under Doris — so it's a superset of what this PR does, and that's also why you're seeing conflicts now.

Since there's nothing left to merge here, I'm closing this PR as superseded by #39622. Nothing wrong with the work at all — thanks again for contributing, and if you'd like to keep helping, feel free to grab another issue from the backlog. We'd love to see more PRs from you!

PR #39669, opened 2026-08-31
Enforce consistent Java import order with Spotless

terrymanu:


taojintianxia (approved):

terrymanu:
Format code with spotless, do not need to review

PR #39674, opened 2026-09-01
Decode MySQL TIME2 binlog values outside the LocalTime range

PDGGK:

  - Subtract the `0x800000` offset before unpacking a TIME2 value, so the sign survives, and return MySQL text form for the range `LocalTime` cannot hold.

`MySQLTime2BinlogProtocolValue` unpacks the 3-byte payload and hands the hour straight to `LocalTime.of`:

```java
int time = payload.getByteBuf().readUnsignedMedium();
...
int hour = (time >> 12) % (1 << 10);
return LocalTime.of(hour, minute, second).withNano(fractionalSeconds.getNanos());
```

MySQL `TIME` is a signed duration, not a time of day: its range is `-838:59:59` to `838:59:59`. The payload is the packed value offset by `0x800000` — the same sign bit `MySQLDatetime2BinlogProtocolValue` masks off with `datetime & (0x8000000000L - 1L)`, except that for `TIME` it carries information. Two consequences, both on values MySQL accepts in an ordinary column:

| column value | payload | current result |
|---|---|---|
| `16:08:04` | `0x810204` | `LocalTime` 16:08:04 |
| `24:00:00` | `0x818000` | `DateTimeException: Invalid value for HourOfDay (valid values 0 - 23): 24` |
| `100:08:04` | `0x864204` | `DateTimeException: ... 100` |
| `838:59:59` | `0xB46EFB` | `DateTimeException: ... 838` |
| `-16:08:04` | `0x7EFDFC` | `DateTimeException: ... 1007` |

The exception propagates out of `MySQLBinlogRowsEventPacket`, so a single such row stops the binlog stream for the whole migration or CDC job rather than affecting one column.

Those payloads are what a live server writes, not a derivation. Against `mysql:8.0.46` started with `--log-bin --binlog-format=ROW`, a `TIME` column accepts `100:08:04` and `-16:08:04`, and the resulting `binlog.000001` carries them as `86 42 04` and `7e fd fc` — the offset encoding above, negative value included.

The negative row also shows the second half of the problem: without the offset the sign bit is folded away by `% (1 << 10)`, which is why the hour comes out as 1007 rather than 16. Even if `LocalTime` accepted it, the value would be wrong.

Subtracting the offset first gives the signed packed value, from which the sign and the magnitude both fall out. `LocalTime` is kept for everything it can represent, so nothing that works today changes; the rest is returned as `[-]HH:mm:ss[.ffffff]`, which is what this same method already returns for zero (`MySQLTimeValueUtils.ZERO_OF_TIME`) and what `MySQLTimeBinlogProtocolValue` returns for every value of the pre-5.6.4 `TIME` type. On the CDC path a `String` reaches `ColumnValueConvertUtils` and comes out a `StringValue`, exactly as the zero value does today.

### Tests

Four cases added to `MySQLTime2BinlogProtocolValueTest`, using the payload encoding the existing tests already assert (`0x800000 | hour << 12 | minute << 6 | second`):

- `assertReadHourBeyondLocalTimeRange` — `100:08:04`
- `assertReadMaxHour` — `838:59:59`, the MySQL maximum
- `assertReadNegativeTime` — `-16:08:04`
- `assertReadNegativeTimeWithFraction` — `-16:08:04.010123`, so the fraction is not dropped on the new path

On `master` all four fail, each with the `DateTimeException` above; the six existing tests pass unchanged there and here, which is what pins that in-range decoding is untouched.

`mvn test -pl database/protocol/dialect/mysql` — 446 tests, and `-pl kernel/data-pipeline/dialect/mysql` — 97 tests, all passing. `checkstyle:check -Pcheck` clean.

The pre-5.6.4 `MySQLTimeBinlogProtocolValue` reads its payload unsigned as well, so negative values of that type decode wrong too. It produces a wrong string rather than an exception, and I did not want to assert the old format's on-disk sign convention without the same level of evidence, so it is left alone here.

---

### DIFF
--- a/database/protocol/dialect/mysql/src/main/java/org/apache/shardingsphere/database/protocol/mysql/packet/binlog/row/column/value/time/MySQLTime2BinlogProtocolValue.java
+++ b/database/protocol/dialect/mysql/src/main/java/org/apache/shardingsphere/database/protocol/mysql/packet/binlog/row/column/value/time/MySQLTime2BinlogProtocolValue.java
@@ -26,6 +26,7 @@ import java.time.LocalTime;
 
 /**
  * TIME2 type value of MySQL binlog protocol.
+ * The 3-byte value is the packed time offset by 0x800000, so it is signed: MySQL TIME ranges from -838:59:59 to 838:59:59.
  * Stored as 3-byte value The number of decimals for the fractional part is stored in the table metadata as a one byte value.
  * The number of bytes that follow the 3 byte time value can be calculated with the following formula: (decimals + 1) / 2
  *
@@ -37,16 +38,37 @@ import java.time.LocalTime;
  */
 public final class MySQLTime2BinlogProtocolValue implements MySQLBinlogProtocolValue {
     
+    private static final int INT_OFFSET = 0x800000;
+    
     @Override
     public Serializable read(final MySQLBinlogColumnDef columnDef, final MySQLPacketPayload payload) {
-        int time = payload.getByteBuf().readUnsignedMedium();
+        int packedTime = payload.getByteBuf().readUnsignedMedium() - INT_OFFSET;
         MySQLFractionalSeconds fractionalSeconds = new MySQLFractionalSeconds(columnDef.getColumnMeta(), payload);
-        if (0x800000 == time) {
+        if (0 == packedTime) {
             return MySQLTimeValueUtils.ZERO_OF_TIME;
         }
-        int hour = (time >> 12) % (1 << 10);
-        int minute = (time >> 6) % (1 << 6);
-        int second = time % (1 << 6);
-        return LocalTime.of(hour, minute, second).withNano(fractionalSeconds.getNanos());
+        int magnitude = Math.abs(packedTime);
+        int hour = (magnitude >> 12) % (1 << 10);
+        int minute = (magnitude >> 6) % (1 << 6);
+        int second = magnitude % (1 << 6);
+        return packedTime > 0 && hour < 24
+                ? LocalTime.of(hour, minute, second).withNano(fractionalSeconds.getNanos())
+                : formatOutOfLocalTimeRange(packedTime < 0, hour, minute, second, fractionalSeconds.getNanos());
+    }
+    
+    /**
+     * Formats a value that {@link LocalTime} cannot hold, which is any negative time and any time from 24:00:00 up to the MySQL maximum of 838:59:59.
+     * The text form matches what {@link MySQLTimeBinlogProtocolValue} returns for every value and what {@link MySQLTimeValueUtils#ZERO_OF_TIME} returns from this same method.
+     *
+     * @param negative whether the time is negative
+     * @param hour hour
+     * @param minute minute
+     * @param second second
+     * @param nanos nanoseconds, always a whole number of microseconds
+     * @return time in MySQL text form
+     */
+    private String formatOutOfLocalTimeRange(final boolean negative, final int hour, final int minute, final int second, final int nanos) {
+        String result = String.format("%s%02d:%02d:%02d", negative ? "-" : "", hour, minute, second);
+        return 0 == nanos ? result : result + String.format(".%06d", nanos / 1000);
     }
 }
--- a/database/protocol/dialect/mysql/src/test/java/org/apache/shardingsphere/database/protocol/mysql/packet/binlog/row/column/value/time/MySQLTime2BinlogProtocolValueTest.java
+++ b/database/protocol/dialect/mysql/src/test/java/org/apache/shardingsphere/database/protocol/mysql/packet/binlog/row/column/value/time/MySQLTime2BinlogProtocolValueTest.java
@@ -92,6 +92,35 @@ class MySQLTime2BinlogProtocolValueTest {
     }
     
     @Test
+    void assertReadHourBeyondLocalTimeRange() {
+        when(payload.getByteBuf()).thenReturn(byteBuf);
+        when(byteBuf.readUnsignedMedium()).thenReturn(0x800000 | (100 << 12) | (8 << 6) | 4);
+        assertThat(new MySQLTime2BinlogProtocolValue().read(columnDef, payload), is("100:08:04"));
+    }
+    
+    @Test
+    void assertReadMaxHour() {
+        when(payload.getByteBuf()).thenReturn(byteBuf);
+        when(byteBuf.readUnsignedMedium()).thenReturn(0x800000 | (838 << 12) | (59 << 6) | 59);
+        assertThat(new MySQLTime2BinlogProtocolValue().read(columnDef, payload), is("838:59:59"));
+    }
+    
+    @Test
+    void assertReadNegativeTime() {
+        when(payload.getByteBuf()).thenReturn(byteBuf);
+        when(byteBuf.readUnsignedMedium()).thenReturn(0x1000000 - (0x800000 | (16 << 12) | (8 << 6) | 4));
+        assertThat(new MySQLTime2BinlogProtocolValue().read(columnDef, payload), is("-16:08:04"));
+    }
+    
+    @Test
+    void assertReadNegativeTimeWithFraction() {
+        columnDef.setColumnMeta(6);
+        when(payload.getByteBuf()).thenReturn(byteBuf);
+        when(byteBuf.readUnsignedMedium()).thenReturn(0x1000000 - (0x800000 | (16 << 12) | (8 << 6) | 4), 10123);
+        assertThat(new MySQLTime2BinlogProtocolValue().read(columnDef, payload), is("-16:08:04.010123"));
+    }
+    
+    @Test
     void assertReadNullTimeWithFraction() {
         columnDef.setColumnMeta(4);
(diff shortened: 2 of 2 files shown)


terrymanu (changes requested):
### Result

Feedback Mode: Change Request

**Blocking Issues: 1**

Negative fractional `TIME2` values are still decoded incorrectly.

### Blocking Issues

1. Negative fractional values use an invalid test encoding

   - **Evidence:** MySQL 8.0.46 stores precision 5/6 values as one signed six-byte number, `nr + TIMEF_OFS`; lower precisions similarly reverse the fractional part for negative values ([MySQL source](https://github.com/mysql/mysql-server/blob/0a7df2e4693d8f10901a26034ae6257699356e30/mysys/my_time.cc#L1721-L1830)). For `-16:08:04.010123`, the resulting bytes are `7e fd fb ff d8 75`.
   - **Evidence:** The PR decodes the integer and fraction independently ([decoder lines 45–56](https://github.com/apache/shardingsphere/blob/a4164adc71f268fd2b76dfc914549c1f53aa579f/database/protocol/dialect/mysql/src/main/java/org/apache/shardingsphere/database/protocol/mysql/packet/binlog/row/column/value/time/MySQLTime2BinlogProtocolValue.java#L45-L56)). Executing it with those server-format bytes returns `-16:08:05.-412776` instead of `-16:08:04.010123`.
   - **Evidence:** The added test supplies `0x7efdfc` and `10123` independently ([test lines 115–120](https://github.com/apache/shardingsphere/blob/a4164adc71f268fd2b76dfc914549c1f53aa579f/database/protocol/dialect/mysql/src/test/java/org/apache/shardingsphere/database/protocol/mysql/packet/binlog/row/column/value/time/MySQLTime2BinlogProtocolValueTest.java#L115-L120)); that represents the whole-second integer encoding plus a positive fraction, not MySQL’s negative fractional encoding.
   - **Impact:** Legal negative fractional `TIME` column values propagate incorrect strings into migration and CDC records, leaving the PR’s stated fix incomplete.
   - **Required Change:** Decode negative fractional bytes according to MySQL’s signed `TIME2` format, including the integer carry and reversed fraction. Replace the mocked fixture with authentic server-format bytes and cover each fractional storage width: precision 1/2, 3/4, and 5/6.

### Coverage

- Candidate: public PR #39674 at `a4164adc71f268fd2b76dfc914549c1f53aa579f`.
- Base and merge-base: `81da8493d0fcc25a84b1259454c83a1d52eff50a`, the head commit’s sole parent.
- Scope: all two GitHub-listed files; the authoritative PR list matched the commit file list.
- Behavior cluster: signed MySQL `TIME2` decoding, returned representation, and regression tests.
- Reviewed lenses: root cause and boundaries; callers, contracts, and CDC/migration propagation; test validity, runtime behavior, compatibility, and performance.
- Verification: focused `MySQLTime2BinlogProtocolValueTest` run passed all 10 tests; an independent raw-payload execution against the compiled PR head reproduced the incorrect result.
- Public discussion: no issue comments, reviews, or review comments.
- This is a code-scope result. CI was not reviewed.

terrymanu (changes requested):
### Correction

- **Previous Finding:** Negative fractional `TIME2` values are decoded incorrectly.
- **Current Status:** Retained
- **Reason:** The public head is unchanged, and MySQL 8.0.46’s signed fractional encoding still contradicts the implementation and test fixture.
- The locale-dependent formatting issue below was already present at this head and was missed in the previous review.

### Result

Feedback Mode: Change Request

**Blocking Issues: 2**

The patch correctly handles whole-second signed values, but negative fractions remain corrupted and the new text representation depends on the JVM locale.

### Blocking Issues

1. Negative fractional values are not decoded as signed fixed-point values

   - **Evidence:** MySQL 8.0.46 applies an integer carry and complemented fractional bytes for precisions 1–4, while precisions 5–6 store the complete signed six-byte value plus `TIMEF_OFS` ([MySQL source, lines 1721–1829](https://github.com/mysql/mysql-server/blob/0a7df2e4693d8f10901a26034ae6257699356e30/mysys/my_time.cc#L1721-L1829)).
   - **Evidence:** The decoder subtracts the offset only from the first three bytes and then reads the fraction independently as unsigned nanos ([decoder lines 45–56](https://github.com/apache/shardingsphere/blob/a4164adc71f268fd2b76dfc914549c1f53aa579f/database/protocol/dialect/mysql/src/main/java/org/apache/shardingsphere/database/protocol/mysql/packet/binlog/row/column/value/time/MySQLTime2BinlogProtocolValue.java#L45-L56)).
   - **Evidence:** The test supplies `0x7efdfc` followed by positive `10123` ([test lines 115–120](https://github.com/apache/shardingsphere/blob/a4164adc71f268fd2b76dfc914549c1f53aa579f/database/protocol/dialect/mysql/src/test/java/org/apache/shardingsphere/database/protocol/mysql/packet/binlog/row/column/value/time/MySQLTime2BinlogProtocolValueTest.java#L115-L120)). The authentic precision-6 encoding of `-16:08:04.010123` is `7e fd fb ff d8 75`; reproducing the current arithmetic yields `-16:08:05.-412776`.
   - **Impact:** Legal negative fractional `TIME` values become incorrect migration and CDC record values.
   - **Required Change:** Decode the signed integer and fraction according to MySQL’s combined `TIME2` representation. Replace the mocked negative-fraction fixture with authentic sequential payload bytes and cover precision groups 1/2, 3/4, and 5/6.

2. The new MySQL text representation is locale-dependent

   - **Evidence:** Both new formatting calls use `String.format(String, Object...)` without a locale ([decoder lines 70–72](https://github.com/apache/shardingsphere/blob/a4164adc71f268fd2b76dfc914549c1f53aa579f/database/protocol/dialect/mysql/src/main/java/org/apache/shardingsphere/database/protocol/mysql/packet/binlog/row/column/value/time/MySQLTime2BinlogProtocolValue.java#L70-L72)). Java specifies that this overload uses the default locale ([Java SE 8 documentation](https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#format-java.lang.String-java.lang.Object...-)).
   - **Evidence:** An exact-overload reproduction with the default formatting locale set to Arabic emitted Arabic-Indic digit code points for `100:08:04`. The resulting `String` is passed through unchanged by the [binlog data handler](https://github.com/apache/shardingsphere/blob/a4164adc71f268fd2b76dfc914549c1f53aa579f/kernel/data-pipeline/dialect/mysql/src/main/java/org/apache/shardingsphere/data/pipeline/mysql/ingest/incremental/binlog/data/MySQLBinlogDataHandler.java#L42-L50) and becomes a CDC [`StringValue`](https://github.com/apache/shardingsphere/blob/a4164adc71f268fd2b76dfc914549c1f53aa579f/kernel/data-pipeline/scenario/cdc/core/src/main/java/org/apache/shardingsphere/data/pipeline/cdc/util/ColumnValueConvertUtils.java#L93-L95).
   - **Impact:** Identical binlog bytes produce different output depending on process locale, potentially corrupting CDC values or making migration sinks reject the time text.
   - **Required Change:** Format the new `TIME2` text path with locale-independent ASCII digits and add a focused test under a non-Latin default formatting locale, restoring the original locale afterward. Fixing the unrelated legacy `TIME` decoder is not required by this PR.

### Coverage

- Candidate: public PR #39674 at `a4164adc71f268fd2b76dfc914549c1f53aa579f`.
- Base and effective merge-base: `81da8493d0fcc25a84b1259454c83a1d52eff50a`, the head commit’s sole parent.
- Requirements: the PR title and description, including signed and negative fractional `TIME2` decoding; no separate linked issue was provided.
- Scope: all two GitHub-listed files. Their downloaded contents matched the advertised blob SHAs, and the PR file list matched the head commit’s complete file list.
- Behavior cluster: signed `TIME2` decoding, fallback text representation, and regression tests.
- Reviewed lenses: root cause and value boundaries; MySQL format compatibility and downstream propagation; test authenticity, runtime behavior, locale determinism, and performance/allocation risk.
- Verification: MySQL tag `mysql-8.0.46` resolved to `0a7df2e4693d8f10901a26034ae6257699356e30`; arithmetic and locale-focused JShell reproductions confirmed both failures. Maven was not run because the candidate head was not present locally; this does not leave an outcome-sensitive gap.
- Previous public feedback: one change-request review exists; there are no issue comments, review comments, follow-up commits, or author counter-evidence.
- This is a code-scope result. CI was not reviewed.

PDGGK:
@terrymanu thanks, both findings were right, and sorry this sat for so long.

- Negative fractional values: the decoder now combines the whole seconds and the fraction the way `my_time_packed_from_binary` does, with the carry and reversed fraction at precision 1–4 and the single signed six-byte value at 5–6. `7e fd fb ff d8 75` now decodes to `-16:08:04.010123`. The mocked fixture is replaced with sequential payload bytes for each storage width, including two rows from the table in MySQL's source (`7FFFFE.F6`, `7FFFFF.FF`), and each case checks that the payload is fully consumed.
- Locale: both format calls use `Locale.ROOT`. A test runs under a `th-TH-u-nu-thai` default format locale and restores the original afterwards.

Checking the combined value for zero also fixed `00:00:00.5`, which was coming back as `00:00:00`; it has its own case. I updated the description to match.

PR #39676, opened 2026-09-01
Decode MySQL binlog dates that have a zero month or a zero day

PDGGK:

  - Return MySQL text form from the `DATE`, `DATETIME` and `DATETIME2` binlog decoders when the date has a zero month or a zero day, instead of handing it to `LocalDate`.

MySQL stores a date with a zero month or a zero day whenever `NO_ZERO_IN_DATE` is not in `sql_mode`. Verified against a real server:

```
$ docker run -d mysql:8.0 --log-bin=binlog --binlog-format=ROW --sql-mode=''
$ mysql> INSERT INTO t_date VALUES (1, '2023-00-15', '2023-00-15 10:00:00'),
                                   (2, '2023-05-00', '2023-05-00 10:00:00'),
                                   (3, '2023-05-15', '2023-05-15 10:00:00');
$ mysql> SELECT id, d, dt FROM t_date;
+----+------------+---------------------+
| id | d          | dt                  |
+----+------------+---------------------+
|  1 | 2023-00-15 | 2023-00-15 10:00:00 |
|  2 | 2023-05-00 | 2023-05-00 10:00:00 |
|  3 | 2023-05-15 | 2023-05-15 10:00:00 |
+----+------------+---------------------+
```

Reading the resulting `binlog.000001` back, the row events carry those dates in the ordinary packed encodings — `0f ce 0f` and `a0 ce 0f` for the two `DATE` values, `99 ae de a0 00` and `99 b0 00 a0 00` for the two `DATETIME2` values. Running the decoders' own arithmetic over exactly those bytes:

| column value | payload | current result |
|---|---|---|
| `DATE` `2023-05-15` | `af ce 0f` | `java.sql.Date` 2023-05-15 |
| `DATE` `2023-00-15` | `0f ce 0f` | `DateTimeException: Invalid value for MonthOfYear (valid values 1 - 12): 0` |
| `DATE` `2023-05-00` | `a0 ce 0f` | `DateTimeException: Invalid value for DayOfMonth (valid values 1 - 28/31): 0` |
| `DATETIME2` `2023-00-15 10:00:00` | `99 ae de a0 00` | `DateTimeException: ... MonthOfYear ...: 0` |
| `DATETIME2` `2023-05-00 10:00:00` | `99 b0 00 a0 00` | `DateTimeException: ... DayOfMonth ...: 0` |

The exception propagates out of `MySQLBinlogRowsEventPacket`, so one such row stops the binlog stream for the whole migration or CDC job rather than affecting one column.

Each of these decoders already returns text for the *all*-zero date — `MySQLTimeValueUtils.ZERO_OF_DATE` and `DATETIME_OF_ZERO` — which needs exactly the same relaxed `sql_mode` to exist in a table at all. So the partially zero case is returned the same way, and the `java.sql` types are kept for every date they can hold, leaving anything that works today unchanged. On the CDC path a `String` reaches `ColumnValueConvertUtils` and comes out a `StringValue`, as the zero value already does.

The pre-5.6.4 `MySQLDatetimeBinlogProtocolValue` unpacks a plain decimal `YYYYMMDDHHMMSS`, so its zero-month and zero-day paths reach `LocalDateTime.of` the same way; it is fixed alongside, although a modern server writes `DATETIME2` and I could not produce a live payload for it.

### Tests

Six cases added, two per decoder, using the payloads above:

- `MySQLDateBinlogProtocolValueTest` — `assertReadDateWithZeroMonth`, `assertReadDateWithZeroDay`
- `MySQLDatetimeBinlogProtocolValueTest` — `assertReadDatetimeWithZeroMonth`, `assertReadDatetimeWithZeroDay`
- `MySQLDatetime2BinlogProtocolValueTest` — `assertReadWithZeroMonth`, `assertReadWithZeroDay`

On `master` all six fail with the `DateTimeException` above; the existing cases in those three classes pass unchanged there and here, which pins that complete dates are untouched.

`mvn test -pl database/protocol/dialect/mysql` — 448 tests, and `-pl kernel/data-pipeline/dialect/mysql` — 97 tests, all passing. `checkstyle:check -Pcheck` clean.

Related to #39674, which fixes the same shape of failure in the `TIME2` decoder. The two do not overlap and can be reviewed in either order.

---

### DIFF
--- a/database/protocol/dialect/mysql/src/main/java/org/apache/shardingsphere/database/protocol/mysql/packet/binlog/row/column/value/time/MySQLDateBinlogProtocolValue.java
+++ b/database/protocol/dialect/mysql/src/main/java/org/apache/shardingsphere/database/protocol/mysql/packet/binlog/row/column/value/time/MySQLDateBinlogProtocolValue.java
@@ -39,6 +39,11 @@ public final class MySQLDateBinlogProtocolValue implements MySQLBinlogProtocolVa
         int year = date / 16 / 32;
         int month = date / 32 % 16;
         int day = date % 32;
-        return 0 == date ? MySQLTimeValueUtils.ZERO_OF_DATE : Date.valueOf(LocalDate.of(year, month, day));
+        if (0 == date) {
+            return MySQLTimeValueUtils.ZERO_OF_DATE;
+        }
+        return MySQLTimeValueUtils.isCompleteDate(month, day)
+                ? Date.valueOf(LocalDate.of(year, month, day))
+                : MySQLTimeValueUtils.formatIncompleteDate(year, month, day);
     }
 }
--- a/database/protocol/dialect/mysql/src/main/java/org/apache/shardingsphere/database/protocol/mysql/packet/binlog/row/column/value/time/MySQLDatetime2BinlogProtocolValue.java
+++ b/database/protocol/dialect/mysql/src/main/java/org/apache/shardingsphere/database/protocol/mysql/packet/binlog/row/column/value/time/MySQLDatetime2BinlogProtocolValue.java
@@ -68,7 +68,9 @@ public final class MySQLDatetime2BinlogProtocolValue implements MySQLBinlogProto
         int minute = (int) ((time >> 6) % (1L << 6L));
         int second = (int) (time % (1L << 6L));
         MySQLFractionalSeconds fractionalSeconds = new MySQLFractionalSeconds(columnDef.getColumnMeta(), payload);
-        return Timestamp.valueOf(LocalDateTime.of(year, month, day, hour, minute, second, fractionalSeconds.getNanos()));
+        return MySQLTimeValueUtils.isCompleteDate(month, day)
+                ? Timestamp.valueOf(LocalDateTime.of(year, month, day, hour, minute, second, fractionalSeconds.getNanos()))
+                : MySQLTimeValueUtils.formatIncompleteDatetime(year, month, day, hour, minute, second, fractionalSeconds.getNanos());
     }
     
     private void skipFractionalSeconds(final MySQLBinlogColumnDef columnDef, final MySQLPacketPayload payload) {
--- a/database/protocol/dialect/mysql/src/main/java/org/apache/shardingsphere/database/protocol/mysql/packet/binlog/row/column/value/time/MySQLDatetimeBinlogProtocolValue.java
+++ b/database/protocol/dialect/mysql/src/main/java/org/apache/shardingsphere/database/protocol/mysql/packet/binlog/row/column/value/time/MySQLDatetimeBinlogProtocolValue.java
@@ -24,7 +24,6 @@ import org.apache.shardingsphere.database.protocol.mysql.payload.MySQLPacketPayl
 import java.io.Serializable;
 import java.sql.Timestamp;
 import java.time.LocalDateTime;
-import java.util.Date;
 
 /**
  * MySQL DATETIME binlog protocol value.
@@ -40,7 +39,7 @@ public final class MySQLDatetimeBinlogProtocolValue implements MySQLBinlogProtoc
         return 0L == datetime ? MySQLTimeValueUtils.DATETIME_OF_ZERO : readDateTime(datetime);
     }
     
-    private Date readDateTime(final long datetime) {
+    private Serializable readDateTime(final long datetime) {
         int date = (int) (datetime / 1000000L);
         int year = date / 10000;
         int month = (date % 10000) / 100;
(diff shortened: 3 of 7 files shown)


terrymanu (changes requested):
### Result

Feedback Mode: Change Request

**Blocking Issues: 2**

The core zero-date fallback is sound, but it produces locale-dependent protocol values and lacks coverage for fractional `DATETIME2` values.

### Blocking Issues

1. Make temporal text formatting locale-independent.

   - **Evidence:** [`MySQLTimeValueUtils.java:57-75`](https://github.com/apache/shardingsphere/blob/995cff271d43b3a6307861de84c38304771ce3d9/database/protocol/dialect/mysql/src/main/java/org/apache/shardingsphere/database/protocol/mysql/packet/binlog/row/column/value/time/MySQLTimeValueUtils.java#L57-L75) uses `String.format` without an explicit locale. Java specifies that this overload uses the JVM’s default `FORMAT` locale and localizes decimal digits ([`String.format`](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/String.html#format(java.lang.String,java.lang.Object...)), [`Formatter`](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/Formatter.html)). With an Arabic-format JVM, the three affected test classes produced six failures because values such as `2023-00-15` contained Arabic-Indic digits.
   - **Impact:** All three new decoder paths can return environment-dependent strings instead of the ASCII MySQL temporal representation. These strings propagate into binlog rows and then into migration JDBC values or CDC `StringValue` messages.
   - **Required Change:** Use locale-independent formatting, such as `Locale.ROOT`, for every added numeric format operation. Add a regression test using a non-Latin default format locale and restore the original locale after the test.

2. Cover partially zero `DATETIME2` values with non-zero fractional seconds.

   - **Evidence:** [`MySQLDatetime2BinlogProtocolValue.java:70-73`](https://github.com/apache/shardingsphere/blob/995cff271d43b3a6307861de84c38304771ce3d9/database/protocol/dialect/mysql/src/main/java/org/apache/shardingsphere/database/protocol/mysql/packet/binlog/row/column/value/time/MySQLDatetime2BinlogProtocolValue.java#L70-L73) adds a fallback that passes decoded nanoseconds to the new non-zero branch at [`MySQLTimeValueUtils.java:73-75`](https://github.com/apache/shardingsphere/blob/995cff271d43b3a6307861de84c38304771ce3d9/database/protocol/dialect/mysql/src/main/java/org/apache/shardingsphere/database/protocol/mysql/packet/binlog/row/column/value/time/MySQLTimeValueUtils.java#L73-L75). The new tests at [`MySQLDatetime2BinlogProtocolValueTest.java:87-97`](https://github.com/apache/shardingsphere/blob/995cff271d43b3a6307861de84c38304771ce3d9/database/protocol/dialect/mysql/src/test/java/org/apache/shardingsphere/database/protocol/mysql/packet/binlog/row/column/value/time/MySQLDatetime2BinlogProtocolValueTest.java#L87-L97) leave `columnMeta` at zero, so this branch is never exercised. Existing fractional tests use complete dates and therefore return `Timestamp`. MySQL supports `DATETIME` fractional precision from zero through six digits ([MySQL documentation](https://dev.mysql.com/doc/refman/8.4/en/fractional-seconds.html)).
   - **Impact:** Dropping or misformatting the fractional suffix—or failing to consume its payload bytes—would corrupt the value or subsequent row decoding without failing the current suite.
   - **Required Change:** Add a focused partial-zero `DATETIME2` case with a non-zero fractional payload, asserting the complete returned string and correct payload consumption.

### Coverage

- Candidate: public PR [#39676](https://github.com/apache/shardingsphere/pull/39676), Code Correctness Review.
- Reviewed base: `81da8493d0fcc25a84b1259454c83a1d52eff50a`.
- Reviewed head: `995cff271d43b3a6307861de84c38304771ce3d9`; the single commit’s parent equals the base.
- GitHub’s PR-files and commit endpoints agree on all seven changed files: four production classes and their three decoder test classes. The final head and file-list recheck was unchanged.
- Behavior clusters covered: `DATE`, legacy `DATETIME`, `DATETIME2` including fractional payload handling, shared text formatting, and propagation through migration and CDC consumers.
- Root-cause, blast-radius/contracts, tests/runtime/operations, protocol compatibility, performance, documentation, packaging, and convergence passes were completed. No unresolved evidence gap remains.
- Verification: focused affected tests passed `18/18`; the complete MySQL protocol module passed `448/448`; Checkstyle reported zero violations; Spotless reported zero changed files. The forced Arabic-format run failed `6/18`, confirming issue 1.
- Performance inspection found no credible regression on existing successful paths: they add only two integer comparisons and allocate no fallback string.
- This is a code-scope result only. CI and GitHub Actions were not reviewed.

PDGGK:
@terrymanu thanks for the review, and sorry for the delay.

- Locale: `formatIncompleteDate` and `formatIncompleteDatetime` now format with `Locale.ROOT`. `DATE`, `DATETIME` and `DATETIME2` each have a case under a `th-TH-u-nu-thai` default format locale that restores the original afterwards. Without the fix, all three return Thai digits.
- Fractional `DATETIME2`: added partial-zero cases with a non-zero fraction at precision 2, 4 and 6. They read real payload bytes and assert both the full string (e.g. `2023-00-15 10:00:00.123456`) and that the fraction bytes are consumed. I checked that they fail if the suffix is dropped or the fraction is not read on this path.

Description updated.

PR #39678, opened 2026-09-01
Preserve command execution failures when executor close fails

terrymanu:


terrymanu:
### Result

**Review Result: Mergeable**

The exception-propagation semantics must be fixed, but this exact private-method decomposition is not the only correct implementation. The proposed implementation is nevertheless minimal, correct, and avoids changing the SPI.

### Evidence

- The defect is real. The base implementation invokes `close()` from `finally`. Under [JLS 14.20.2](https://docs.oracle.com/javase/specs/jls/se17/html/jls-14.html#jls-14.20.2), if command execution fails with A and `close()` then fails with B, B replaces A. [Base implementation](https://github.com/apache/shardingsphere/blob/48730d8a405db5d65f570700357fadbafc5047fa/proxy/frontend/core/src/main/java/org/apache/shardingsphere/proxy/frontend/command/CommandExecutorTask.java#L106-L122)

- This affects the client response, not only diagnostics. The exception reaching `run()` is passed to `getErrorPacket`, so replacing A with B produces an error packet for the close failure instead of the command execution failure. [`run()` error path](https://github.com/apache/shardingsphere/blob/2cc89a8d219f89194f785173ea2d7d60de03aea2/proxy/frontend/core/src/main/java/org/apache/shardingsphere/proxy/frontend/command/CommandExecutorTask.java#L71-L89)

- All four execution/close combinations are correct:

  1. Execution succeeds and close succeeds: return normally.
  2. Execution fails and close succeeds: propagate the execution failure.
  3. Execution succeeds and close fails: propagate the close failure.
  4. Execution fails and close fails: propagate the execution failure and suppress the close failure.

  See [`CommandExecutorTask.java:106-145`](https://github.com/apache/shardingsphere/blob/2cc89a8d219f89194f785173ea2d7d60de03aea2/proxy/frontend/core/src/main/java/org/apache/shardingsphere/proxy/frontend/command/CommandExecutorTask.java#L106-L145).

- `handleException` still runs before `close()` for the exception types it previously handled. Its scope is not broadened to arbitrary `RuntimeException` or `Error`.

- Existing `Error` behavior is preserved: `run()` still wraps the original `Error` in a `RuntimeException`, while the close failure remains suppressed on the original `Error`.

- `close()` is called exactly once on every path after an executor is obtained. The success and failure close paths are mutually exclusive.

- The change does not modify `CommandExecutor`, public APIs, constructors, visibility, dependencies, or module boundaries. [`CommandExecutor` contract](https://github.com/apache/shardingsphere/blob/2cc89a8d219f89194f785173ea2d7d60de03aea2/proxy/frontend/spi/src/main/java/org/apache/shardingsphere/proxy/frontend/command/executor/CommandExecutor.java#L28-L44)

- The four added tests exercise the public `run()` entry point and cover close-only failure, dual `SQLException` failure, `RuntimeException` plus close failure, and `Error` plus close failure. They verify the primary exception, suppressed exception, ordering, cause, close count, and cleanup. [`CommandExecutorTaskTest.java:160-288`](https://github.com/apache/shardingsphere/blob/2cc89a8d219f89194f785173ea2d7d60de03aea2/proxy/frontend/core/src/test/java/org/apache/shardingsphere/proxy/frontend/command/CommandExecutorTaskTest.java#L160-L288)

- Separate tests for close throwing `RuntimeException` and `Error` are not required. Those types share the same multi-catch suppression branch already exercised with a close `SQLException`; duplicating the Mockito scenarios would not cover another production branch.

- The normal path adds no allocation, I/O, lock, loop, or algorithmic complexity.

- Read-only verification:

  - Focused tests: 10 passed, exit 0.
  - `proxy/frontend/core` tests: 89 passed, exit 0.
  - Spotless check: passed, exit 0.
  - Checkstyle: passed, exit 0.

#### Is this exact implementation absolutely necessary?

- Fixing the exception semantics is necessary. Otherwise, clients can receive the wrong primary error.
- This exact helper-method structure is not uniquely necessary. Explicitly tracking the primary failure inside a carefully structured `finally` could provide equivalent semantics.
- The current structure should still be retained: it is clearer than a stateful `finally`, preserves the `handleException`/`close` order, and avoids an unnecessary `AutoCloseable` SPI change.
- The `mockProxyContext` line reflow is not semantically necessary, but it is formatter-stable and does not justify a change request.

### Coverage

- Candidate: public PR Code Correctness Review.
- PR: [apache/shardingsphere#39678](https://github.com/apache/shardingsphere/pull/39678)
- Reviewed head: `2cc89a8d219f89194f785173ea2d7d60de03aea2`
- Base: `48730d8a405db5d65f570700357fadbafc5047fa`
- The authoritative GitHub file list matches the local triple-dot comparison:
  - `proxy/frontend/core/src/main/java/org/apache/shardingsphere/proxy/frontend/command/CommandExecutorTask.java`
  - `proxy/frontend/core/src/test/java/org/apache/shardingsphere/proxy/frontend/command/CommandExecutorTaskTest.java`
- Reviewed clusters: executor failure preservation, close lifecycle, client error conversion, and regression coverage.
- Completed Root Cause and Behavior, Blast Radius and Contracts, Tests/Runtime/Operations, and final convergence reviews.
- No blocking findings or outcome-sensitive evidence gaps remain.
- This is a code-scope result only; CI status was not reviewed.

taojintianxia (approved):

PR #39720, opened 2026-09-04
Preserve dotted Single table identity through load and unload

terrymanu:


terrymanu:
### Result

**Review Result: Mergeable**

No correctness, scope, performance, or test-quality blockers remain.

### Evidence

- Dotted table identities are preserved only for schema-unavailable storage types.
- Schema-aware, malformed, and unknown-storage multi-segment nodes remain invalid.
- UNLOAD removes exact runtime `DataNode` identities without confusing `bar` with `foo.bar`.
- Physical and aggregated data-source names resolve through existing metadata without JDBC access.
- Standalone reload uses refreshed runtime table mappings and retains the full dotted identity.
- No public API, configuration model, parser, dependency, or shared `DataNode` contract changed.

### Coverage

- Reviewed all six changed files against baseline `b6854a25e67d9cbbe46aa93d2b6b46924c475eae`.
- Focused tests passed: 63/63—loader 26, UNLOAD 18, standalone 19.
- Spotless and Checkstyle passed for all three affected modules.
- Full CI and E2E tests were not run.

taojintianxia (approved):

PR #39779, opened 2026-09-09
Fix transaction connection preparation before enlistment

terrymanu:


terrymanu:
### Result

**Review Result: Mergeable**

Task completion: **100%**.

The fix belongs in production code rather than the E2E tests. `transactionIsolation` and `readOnly` are now applied before ShardingSphere-managed XA resource enlistment, while the remaining connection properties are replayed afterward. This order is also compatible with older PostgreSQL JDBC drivers.

### Evidence

- pgjdbc 42.7.13 executes `BEGIN` in `XAResource.start(TMNOFLAGS)`. The previous implementation changed the isolation level after enlistment, resulting in an attempt to change it during an active transaction. [pgjdbc 42.7.13 source](https://github.com/pgjdbc/pgjdbc/blob/REL42.7.13/pgjdbc/src/main/java/org/postgresql/xa/PGXAConnection.java)
- The JDBC specification states that changing the isolation level during a transaction has implementation-defined behavior and that `setReadOnly` cannot be called during a transaction. [Java 21 Connection API](https://docs.oracle.com/en/java/javase/21/docs/api/java.sql/java/sql/Connection.html#setTransactionIsolation(int))
- The existing `ShardingSphereDistributedTransactionManager#getConnection` method was changed directly. No default compatibility overload was added, as agreed.
- The connection preparation callback is propagated through the XA, Seata, JDBC Driver, and Proxy paths. Connections are closed when preparation fails.
- **1,899 tests** across the seven affected modules passed with zero failures, errors, or skipped tests.
- Spotless, Checkstyle, RAT, test compilation, and the focused ordering and failure-cleanup tests all passed.
- The Docker-based E2E test was not run locally. No E2E test, build configuration, or dependency file was modified. The associated CI job should be rerun after the change is committed.

### Coverage

Baseline: `a326cb1db10124f6df09b8a20fa10a534bb8f3ed`. The final scope contains 20 files, with 351 additions and 61 deletions. No unrelated working-tree changes were introduced.

taojintianxia (approved):

PR #39781, opened 2026-09-09
Fix PostgreSQL XA auto-commit handling

terrymanu:


terrymanu:
### Result
**Review Result: Mergeable**

The fix follows the correct direction and keeps the scope minimal. PostgreSQL-specific behavior is handled by the existing `XAConnectionWrapper` SPI implementation. No database type is hard-coded in the XA core, and no new interface, overload, or abstraction is introduced.

### Evidence

- `PostgreSQLXAConnectionWrapper.java:40`: Disables auto-commit before XA enlistment only when the physical connection currently has auto-commit enabled.
- `XATransactionDataSource.java:127`: Records the connection state before and after SPI invocation and restores it in `afterCompletion` only when it changed from `true` to `false`.
- The transaction sequence remains: replay isolation level and read-only state → adjust auto-commit through the PostgreSQL SPI → enlist the XA resource.
- In pgjdbc 42.7.13, `start(TMNOFLAGS)` explicitly executes `BEGIN` while preserving the JDBC auto-commit flag: [PGXAConnection 42.7.13](https://github.com/pgjdbc/pgjdbc/blob/REL42.7.13/pgjdbc/src/main/java/org/postgresql/xa/PGXAConnection.java).
- pgjdbc 42.7.8 saves and restores the auto-commit state that existed before XA enlistment: [PGXAConnection 42.7.8](https://github.com/pgjdbc/pgjdbc/blob/REL42.7.8/pgjdbc/src/main/java/org/postgresql/xa/PGXAConnection.java). Restoring the state after transaction completion preserves compatibility with the older driver.
- `afterCompletion` is invoked after a transaction has committed or rolled back, which makes it the correct JTA lifecycle point for restoration: [Synchronization](https://jakarta.ee/specifications/transactions/1.3/apidocs/javax/transaction/synchronization).
- Focused unit tests: 13/13 passed.
- Full XA core module tests: 28/28 passed.
- PostgreSQL 12 with pgjdbc 42.7.13 E2E: 4/4 passed, covering JDBC, Proxy, and savepoints.
- PostgreSQL 12 with pgjdbc 42.7.8 JDBC E2E: 3/3 passed, covering Atomikos, Narayana, and savepoints.
- Spotless, Checkstyle, and RAT all passed with exit code 0.
- The final working tree contains changes to only the four agreed files: 77 insertions and 3 deletions.

### Coverage

- Candidate type: standalone local candidate.
- Baseline: `5c1478564a7812412748c10cdfbdb2309dcc7580`.
- `PostgreSQLXAConnectionWrapper.java:40` → disables auto-commit before PostgreSQL XA enlistment → fixes savepoint handling with the new driver → dialect-specific behavior belongs in the SPI implementation.
- `XATransactionDataSource.java:127` → restores SPI-modified state after transaction completion → preserves compatibility with both new and old drivers and prevents connection-state leakage → JTA lifecycle management belongs in this class.
- `PostgreSQLXAConnectionWrapperTest.java:52` → covers both auto-commit branches and invocation ordering → prevents regressions in the dialect-specific fix.
- `XATransactionDataSourceTest.java:126` → covers option replay, enlistment, completion-time restoration, and initially disabled auto-commit → prevents lifecycle and ordering regressions.
- Root-cause, blast-radius, test/runtime, and final convergence reviews were completed with no blocking findings.
- This result covers local code correctness only; remote CI status was not reviewed.

taojintianxia (approved):

PR #39782, opened 2026-09-09
Raise anonymous inner class length limit

terrymanu:


terrymanu:
### Result

**Review Result: Mergeable**

The Checkstyle `AnonInnerLength` limit has been increased from the default 20 lines to 30 lines. No blocking issues were found.

### Evidence

- `src/resources/checkstyle.xml:341` now explicitly defines `<property name="max" value="30" />`.
- The local Checkstyle 14.0.0 implementation confirms that the default limit is 20 and reports a violation when the anonymous inner class exceeds `max`.
- Targeted Checkstyle verification passed with zero violations, and the generated configuration confirms `max=30`.
- Apache RAT, XML validation, and `git diff --check` all passed.
- Spotless does not apply because its repository configuration does not process this XML file.
- The anonymous inner class in the current working tree is 20 lines long, so it already passed before this change. The reported 23-line version likely came from a different code revision. The new configuration permits 23 lines while still rejecting anything over 30 lines.

### Coverage

- Candidate type: standalone local change.
- Only `src/resources/checkstyle.xml` was changed.
- No Java files, tests, POM files, or other rules were modified.
- The change does not affect runtime functionality or performance.
- No unresolved issues remain.

taojintianxia (approved):

PR #39804, opened 2026-09-11
Optimize Codex policy harness verification

terrymanu:
- partition semantic cases into bounded concurrent batches
- evaluate each case once without retrying unchanged candidates
- add transport integrity and semantic stability checks
- support legacy and batched V0 baselines
- add deterministic regression and safety coverage


terrymanu:
For #39804

taojintianxia (approved):

PR #39882, opened 2026-09-17
Defer meta data refresh to transaction commit for DDL in PostgreSQL transaction

### LINKED ISSUE #35166: Metadata Not Refreshed After Executing DDL Within PostgreSQL Transaction
## Bug Report
After removing the restriction on executing DDL within PostgreSQL transactions in ShardingSphere(`org.apache.shardingsphere.proxy.backend.connector.ProxySQLExecutor#checkExecutePrerequisites`), the following issue is observed:

When a DDL statement that modifies metadata (e.g., ALTER TABLE t1 ADD COLUMN c1 INT) is executed within a transaction, the refresh metadata operation uses a new physical connection.
Since the transaction is not yet committed, the metadata changes from the DDL are not visible, leading to failure in refreshing metadata (e.g., new columns are not detected).

Master.

ShardingSphere-Proxy

### Expected behavior
After the DDL is executed in the postgresql transaction, the metadata information in ShardingSphere should be consistent with the table structure on PostgreSQL.

### Actual behavior
The DDL executed successfully in the transaction, but the corresponding metadata was not refreshed successfully on ShardingSphere.

Track all table names involved in metadata-changing DDL operations within the PostgreSQL transaction.
​​Defer metadata refresh​​ until the transaction is committed, then perform a unified metadata refresh for the affected tables.

1. Start a PostgreSQL transaction.
2. Execute a DDL statement within the transaction (e.g., ALTER TABLE t1 ADD COLUMN c1 INT).
3. After committing the transaction, query metadata again and observe that the new column is not refreshed.
​​Expected Behavior​​
Metadata should be automatically refreshed after transaction commit to ensure changes (e.g., new columns) take effect.

```sql
-- PostgreSQL test scenario
sharding_db=> begin;
BEGIN

-- Executing DDL within transaction. 
-- ShardingSphere Metadata refresh uses a new connection, thus cannot see uncommitted schema changes
sharding_db=> alter table t_order add column t1 int;  
ALTER TABLE
sharding_db=> commit;
COMMIT

-- Column t1 not refreshed automatically
sharding_db=> select * from t_order;  
 order_id | user_id | status | creation_date |
----------+---------+--------+---------------+----+----+----+
(0 rows)

-- Manual metadata refresh required to load t1 column
sharding_db=> refresh table metadata t_order;  
SUCCESS
sharding_db=> select * from t_order;  
 order_id | user_id | status | creation_date | t1 |
----------+---------+--------+---------------+----+----+----+----+
(0 rows)
```

rishi-rana:
Fixes #35166.

Meta data refresh runs immediately after execute, over a **new physical connection**, so it cannot see DDL that the current transaction has not committed yet. This PR defers the refresh to commit.

- Add `DeferredMetaDataRefreshContext`, a transaction-scoped queue of pending push down refreshes held by `ProxyDatabaseConnectionManager`. It stores the database name rather than a `ShardingSphereDatabase` reference, so the database is re-resolved when the refresh actually runs.
- `StandardDatabaseProxyConnector#doExecute` now enqueues the refresh when the session is in a transaction, and refreshes immediately otherwise. The existing `DDLCommitPolicy.COMMIT_CURRENT_TRANSACTION` path is unchanged: it commits first, so the refresh still runs immediately.
- `ProxyBackendTransactionManager` drains the queue after the underlying commit succeeds and before the connection context is cleared, discards it on rollback, records the queue position on `SAVEPOINT`, and rolls back to that position on `ROLLBACK TO SAVEPOINT`.
- Set `isSupportMetaDataRefreshInTransaction` to `true` for PostgreSQL. `ProxySQLExecutor#checkExecutePrerequisites` rejects metadata-changing DDL in a transaction while that flag is `false`, so the deferral mechanism and the flag have to change together.

Notes on scope:

- This covers XA and BASE as well. `ProxyBackendTransactionManager#commit` is the single commit entry point for all three transaction types, with the local/distributed split inside it, so the deferral is transaction-type agnostic. PostgreSQL already declared `isSupportDDLInXATransaction = true`, so this adds no new claim about XA beyond making the refresh correct.
- Rolling back to a savepoint must drop the refreshes recorded after it. `DropTablePushDownMetaDataRefresher` removes the table from meta data purely from the statement rather than reloading it, so keeping a rolled-back `DROP TABLE` in the queue would delete a still-existing table from meta data.
- The deferred refreshes are only rolled back when the savepoint rollback succeeded on every connection. If it failed, the DDL is still applied and its refresh must be kept.
- `DriverJDBCPushDownExecuteExecutor` and `DriverJDBCPushDownExecuteUpdateExecutor` have the same immediate-refresh shape, so ShardingSphere-JDBC appears to have the analogous behaviour. They never read `isSupportMetaDataRefreshInTransaction`, so this PR does not change them. Happy to open a separate issue if you would like that covered.

Tests:

- `DeferredMetaDataRefreshContextTest` covers the queue and savepoint semantics.
- `ProxyBackendTransactionManagerTest` covers the drain ordering relative to commit, the discard on rollback, and the savepoint mark and rollback wiring.
- `StandardDatabaseProxyConnectorTest` covers deferring instead of refreshing while in a transaction.
- `PostgreSQLDDLMetaDataRefreshTestCase` is a transaction E2E case running the `BEGIN` / `ALTER TABLE ADD COLUMN` / `COMMIT` sequence from the issue against Proxy, asserting the new column becomes visible without a manual `REFRESH TABLE METADATA`, and that a rolled back column does not. It compiles locally but I have not run the Docker-backed E2E suite on my machine, so please treat CI as the authority for it.

Local verification:

- `./mvnw -pl proxy/backend/core,database/connector/dialect/postgresql test` — 597 tests, 0 failures.
- `./mvnw -pl proxy/backend/core,database/connector/dialect/postgresql spotless:check checkstyle:check -Pcheck` — 0 violations.
- `./mvnw apache-rat:check -Pcheck` — passed.
- Each new test was confirmed to fail against the unfixed implementation before being accepted.

---

- [x] I have passed maven check locally : `./mvnw clean install -B -T1C -Dmaven.javadoc.skip -Dmaven.jacoco.skip -e`. I ran the targeted module builds, style checks and RAT listed above rather than the full reactor build; please let me know if you would like the full build result before review.
- [x] I have updated the Release Notes of the current development version.
- [x] I have disclosed the tool and affected files or scope for any material AI assistance, as required by the [AI-assisted contribution policy].

AI assistance disclosure, per `AI_POLICY.md`: this change was assisted by AI (Claude Code), covering the investigation, implementation and tests across all files in this PR, under my review and with the verification listed above.

[AI-assisted contribution policy]: https://github.com/apache/shardingsphere/blob/master/AI_POLICY.md


terrymanu (changes requested):
### Result

Feedback Mode: Change Request

**Blocking Issues: 2**

The deferred-refresh design causes a cross-dialect regression and cannot safely replay multiple transactional DDL statements.

### Blocking Issues

1. **[P1] Preserve immediate refresh for implicit-commit dialects**

   **Evidence:** [`StandardDatabaseProxyConnector.java:252`](https://github.com/apache/shardingsphere/blob/62a116f429ba317bbadb26bc3b74425f9dff8bee/proxy/backend/core/src/main/java/org/apache/shardingsphere/proxy/backend/connector/StandardDatabaseProxyConnector.java#L252-L268) now defers every metadata refresh whenever the logical session is in a transaction, without considering the backend dialect. [`MySQLDatabaseMetaData.java:85`](https://github.com/apache/shardingsphere/blob/62a116f429ba317bbadb26bc3b74425f9dff8bee/database/connector/dialect/mysql/src/main/java/org/apache/shardingsphere/database/connector/mysql/metadata/database/MySQLDatabaseMetaData.java#L85-L86) retains `NO_ADDITIONAL_COMMIT`, so ShardingSphere does not clear the transaction before entering this branch. On rollback, [`ProxyBackendTransactionManager.java:171`](https://github.com/apache/shardingsphere/blob/62a116f429ba317bbadb26bc3b74425f9dff8bee/proxy/backend/core/src/main/java/org/apache/shardingsphere/proxy/backend/connector/jdbc/transaction/ProxyBackendTransactionManager.java#L171-L181) discards the queued refresh. MySQL documents that ordinary `CREATE TABLE` and other listed DDL statements implicitly commit the active transaction. [MySQL 8.4 Reference Manual](https://dev.mysql.com/doc/refman/8.4/en/implicit-commit.html)

   **Impact:** With `BEGIN; CREATE TABLE t (...); ROLLBACK;`, MySQL commits the table creation, but this PR discards its metadata refresh. The table remains physically present while ShardingSphere metadata remains stale.

   **Required Change:** Defer only when the backend keeps DDL uncommitted and invisible to the refresh connection. Preserve immediate refresh for implicit-commit dialects and add a regression test covering MySQL DDL followed by rollback.

2. **[P1] Do not replay statement-local refreshers against the transaction’s final state**

   **Evidence:** [`DeferredMetaDataRefreshContext.java:50`](https://github.com/apache/shardingsphere/blob/62a116f429ba317bbadb26bc3b74425f9dff8bee/proxy/backend/core/src/main/java/org/apache/shardingsphere/proxy/backend/connector/jdbc/transaction/DeferredMetaDataRefreshContext.java#L50-L52) stores each original statement and [`DeferredMetaDataRefreshContext.java:87`](https://github.com/apache/shardingsphere/blob/62a116f429ba317bbadb26bc3b74425f9dff8bee/proxy/backend/core/src/main/java/org/apache/shardingsphere/proxy/backend/connector/jdbc/transaction/DeferredMetaDataRefreshContext.java#L87-L90) replays them after every DDL has committed. The existing create-table refresher reloads the statement’s original table name ([`CreateTablePushDownMetaDataRefresher.java:51`](https://github.com/apache/shardingsphere/blob/62a116f429ba317bbadb26bc3b74425f9dff8bee/mode/core/src/main/java/org/apache/shardingsphere/mode/metadata/refresher/pushdown/type/table/CreateTablePushDownMetaDataRefresher.java#L51-L53)); when that name no longer exists, the loader returns `null` ([`TableMetaDataRefresherLoader.java:89`](https://github.com/apache/shardingsphere/blob/62a116f429ba317bbadb26bc3b74425f9dff8bee/mode/core/src/main/java/org/apache/shardingsphere/mode/metadata/refresher/pushdown/type/table/TableMetaDataRefresherLoader.java#L89-L90)), which the standalone service dereferences ([`StandaloneMetaDataManagerPersistService.java:113`](https://github.com/apache/shardingsphere/blob/62a116f429ba317bbadb26bc3b74425f9dff8bee/mode/type/standalone/core/src/main/java/org/apache/shardingsphere/mode/manager/standalone/persist/service/StandaloneMetaDataManagerPersistService.java#L113-L116)). The one-storage-unit routing path allows both statements to reach the database without requiring refreshed metadata ([`SingleSQLRouter.java:52`](https://github.com/apache/shardingsphere/blob/62a116f429ba317bbadb26bc3b74425f9dff8bee/kernel/single/core/src/main/java/org/apache/shardingsphere/single/route/SingleSQLRouter.java#L52-L53)).

   **Impact:** For PostgreSQL, `BEGIN; CREATE TABLE t (...); ALTER TABLE t RENAME TO u; COMMIT;` commits successfully at the database, then replay attempts to load the now-absent `t`, fails before processing the rename, reports a commit error after the physical commit, and leaves `u` absent from ShardingSphere metadata.

   **Required Change:** Represent the final committed metadata effects rather than replaying unchanged statement-local refreshers. Coalesce operations or perform a transaction-level reload that handles renamed and removed objects, with regression coverage for sequences such as create-plus-rename and create-plus-drop.

### Coverage

- Candidate: public PR #39882, Code Correctness Review.
- Effective head: `62a116f429ba317bbadb26bc3b74425f9dff8bee`.
- Base: `master` at `86c1519e4d6deaf68e18ee47196e81da2e114f62`; merge base `101c3d56ccd7ee3c4a390a8f37550bed58a4751c`.
- GitHub’s 13-file list matched the independently reconstructed triple-dot scope.
- All behavior clusters were reviewed:
  - PostgreSQL capability activation and prerequisite tests.
  - Deferred enqueue, transaction lifecycle, savepoints, commit, and rollback.
  - Unit/E2E coverage and E2E registration.
  - Release-note and AI-assistance disclosure requirements.
- Root-cause, blast-radius/contracts, tests/runtime/operations, and final convergence passes completed. No unresolved evidence gaps remain.
- Mandatory style verification:
  - `./mvnw -pl database/connector/dialect/postgresql,proxy/backend/core,test/e2e/operation/transaction spotless:check checkstyle:check -Pcheck -T1C`
  - Exit code: `0`; Spotless covered 26, 243, and 60 Java files in the three modules, and Checkstyle reported zero violations in each.
  - Both checks covered all 11 changed Java files.
  - `RELEASE-NOTES.md` and `test/e2e/operation/transaction/src/test/resources/env/e2e-env.properties` are not governed by the configured Java/POM/workflow Spotless formats or Java Checkstyle scope.
- Focused verification passed:
  - PostgreSQL metadata and four proxy test classes: 116 tests, zero failures/errors.
  - Transaction E2E module test compilation: 60 sources compiled successfully.
  - Docker-backed E2E execution was not run; it is not outcome-sensitive to the two code-proven blockers.
- CI/Actions were not reviewed. This is a code-scope result only; local style verification is complete.

rishi-rana:
Thanks for both findings. Addressed in 5c5ea8bc04097599338aff374bb0b26ad0ce76bc.

`DeferredMetaDataRefreshContext` collects the tables touched by DDL as `(database, schema, table)` and calls `ContextManager#reloadTable` for each once the transaction ends. That reads the committed state of every affected table, and `persistTable` reconciles both directions: present in the database means the table is persisted, absent means it is dropped from meta data.

**1. Implicit-commit dialects**

Reloading committed state makes the outcome independent of how the backend treats DDL, so the implementation does not branch on dialect and no dialect capability is introduced. A rollback reloads the affected tables:

- MySQL `BEGIN; CREATE TABLE t; ROLLBACK;` reloads `t`, finds it present because the server committed it implicitly, and persists it.
- PostgreSQL `BEGIN; CREATE TABLE t; ROLLBACK;` reloads `t`, finds it absent, and leaves meta data unchanged.

Regression coverage is `MySQLDDLMetaDataRefreshTestCase`, an E2E case asserting that a column added inside a transaction and then rolled back is present in meta data on MySQL. It is an E2E rather than a unit test because the behaviour under test is the MySQL server's implicit commit, which needs a real MySQL backend to demonstrate. `ProxyBackendTransactionManagerTest` covers the mechanism, asserting that a rollback reloads and does so after the underlying rollback.

**2. Replaying statement-local refreshers**

Statements are not replayed, so a refresher cannot run against a table name that a later statement in the same transaction removed. `ALTER TABLE t RENAME TO u` contributes both names, because `AlterTableStatement#getTables` includes the rename target, so the reload drops `t` and persists `u`. `CREATE TABLE t; DROP TABLE t;` reloads `t`, finds it absent, and drops it. `ContextManager#reloadTable` logs rather than propagates `SQLException`, so a reload problem does not surface as a commit error after the physical commit has succeeded.

Savepoint tracking is not required, because reloading final state is already correct after a partial rollback.

**Verification**

- `./mvnw -pl proxy/backend/core,database/connector/dialect/postgresql test` - 592 tests, 0 failures.
- Transaction E2E on `mysql:5.7` - passed, including the new MySQL case.
- Transaction E2E on `postgres:12-alpine` - passed.
- `spotless:check` and `checkstyle:check -Pcheck` across the three changed modules - 0 violations; `apache-rat:check -Pcheck` - passed.

terrymanu (changes requested):
### Result

Feedback Mode: Change Request

**Blocking Issues: 3**

The deferred-refresh representation loses required metadata semantics, refreshes implicit-commit dialects too late, and is protected by tests that do not exercise a reachable production DDL path.

### Blocking Issues

1. **[P1] Preserve object type and identifier semantics in deferred refreshes**

   **Evidence:** [`refreshMetaData`](https://github.com/apache/shardingsphere/blob/5c5ea8bc04097599338aff374bb0b26ad0ce76bc/proxy/backend/core/src/main/java/org/apache/shardingsphere/proxy/backend/connector/StandardDatabaseProxyConnector.java#L259-L270) converts every supported metadata-changing DDL into `(database, schema, Collection<String> tableNames)`. The refresh engine also supports schema and view statements ([`PushDownMetaDataRefreshEngine.java:58`](https://github.com/apache/shardingsphere/blob/5c5ea8bc04097599338aff374bb0b26ad0ce76bc/mode/core/src/main/java/org/apache/shardingsphere/mode/metadata/refresher/pushdown/PushDownMetaDataRefreshEngine.java#L58-L61)), but schema statements have no table attribute, while view refreshers must update both table and view metadata ([`CreateViewPushDownMetaDataRefresher.java:43`](https://github.com/apache/shardingsphere/blob/5c5ea8bc04097599338aff374bb0b26ad0ce76bc/mode/core/src/main/java/org/apache/shardingsphere/mode/metadata/refresher/pushdown/type/view/CreateViewPushDownMetaDataRefresher.java#L43-L48)). The deferred context only invokes `reloadTable` ([`DeferredMetaDataRefreshContext.java:46`](https://github.com/apache/shardingsphere/blob/5c5ea8bc04097599338aff374bb0b26ad0ce76bc/proxy/backend/core/src/main/java/org/apache/shardingsphere/proxy/backend/connector/jdbc/transaction/DeferredMetaDataRefreshContext.java#L46-L60)).

   The conversion also discards identifier quoting: `TablesContext` stores only `IdentifierValue#getValue()` ([`TablesContext.java:69`](https://github.com/apache/shardingsphere/blob/5c5ea8bc04097599338aff374bb0b26ad0ce76bc/infra/binder/core/src/main/java/org/apache/shardingsphere/infra/binder/context/segment/table/TablesContext.java#L69-L76)), and `reloadTable(String)` recreates an unquoted identifier ([`ContextManager.java:220`](https://github.com/apache/shardingsphere/blob/5c5ea8bc04097599338aff374bb0b26ad0ce76bc/mode/core/src/main/java/org/apache/shardingsphere/mode/manager/ContextManager.java#L220-L236)). PostgreSQL’s identifier policy distinguishes quoted mixed-case identifiers from unquoted lower-case identifiers.

   **Impact:** `BEGIN; CREATE SCHEMA s; COMMIT` queues nothing, transactional view DDL never updates the view registry, and DDL against a quoted table such as `"MixedCase"` can take the “table missing” drop path instead of persisting the reloaded table.

   **Required Change:** Represent every supported deferred metadata effect with its object type and complete identifier semantics. Add transactional regression coverage for schema DDL, view DDL, and quoted PostgreSQL table names.

2. **[P1] Preserve immediate refresh for implicit-commit dialects**

   **Evidence:** MySQL retains `DDLCommitPolicy.NO_ADDITIONAL_COMMIT` while permitting metadata refresh in transactions ([`MySQLDatabaseMetaData.java:83`](https://github.com/apache/shardingsphere/blob/5c5ea8bc04097599338aff374bb0b26ad0ce76bc/database/connector/dialect/mysql/src/main/java/org/apache/shardingsphere/database/connector/mysql/metadata/database/MySQLDatabaseMetaData.java#L83-L86)). The shared connector nevertheless defers whenever the logical transaction flag remains set. Reloading occurs only during a later Proxy commit or rollback. The new E2E test itself documents that MySQL commits the DDL implicitly, but checks metadata only after calling `rollback` ([`MySQLDDLMetaDataRefreshTestCase.java:33`](https://github.com/apache/shardingsphere/blob/5c5ea8bc04097599338aff374bb0b26ad0ce76bc/test/e2e/operation/transaction/src/test/java/org/apache/shardingsphere/test/e2e/operation/transaction/cases/metadata/MySQLDDLMetaDataRefreshTestCase.java#L33-L62)).

   **Impact:** After MySQL physically commits `ALTER TABLE`, shared ShardingSphere metadata remains stale until the originating client eventually sends a logical commit or rollback. Other sessions can observe the committed database change while Proxy still holds the old metadata. Reloading after rollback fixes the previous stale endpoint but not this interval.

   **Required Change:** Refresh immediately when backend DDL has already committed, or restrict deferral to dialects whose DDL remains transactional. The MySQL E2E case should assert metadata correctness before the later rollback.

3. **[P2] Exercise a reachable DDL path and follow the test standards**

   **Evidence:** [`assertExecuteDefersMetaDataRefreshInTransaction`](https://github.com/apache/shardingsphere/blob/5c5ea8bc04097599338aff374bb0b26ad0ce76bc/proxy/backend/core/src/test/java/org/apache/shardingsphere/proxy/backend/connector/StandardDatabaseProxyConnectorTest.java#L316-L348) constructs an `InsertStatement` and forces `PushDownMetaDataRefreshEngine#isNeedRefresh()` to return `true`. Production returns `true` only for the listed DDL statement classes, so this combination is unreachable and cannot detect the failures above. The test also adds direct construction/static mocking without the required recorded reason, and [`DeferredMetaDataRefreshContextTest.java:59`](https://github.com/apache/shardingsphere/blob/5c5ea8bc04097599338aff374bb0b26ad0ce76bc/proxy/backend/core/src/test/java/org/apache/shardingsphere/proxy/backend/connector/jdbc/transaction/DeferredMetaDataRefreshContextTest.java#L52-L59) uses redundant `times(1)`, contrary to [`CODE_OF_CONDUCT.md:104`](https://github.com/apache/shardingsphere/blob/5c5ea8bc04097599338aff374bb0b26ad0ce76bc/CODE_OF_CONDUCT.md#L104-L110).

   **Impact:** The focused suite passes while the actual DDL extraction and deferred representation are broken.

   **Required Change:** Exercise the connector with real supported DDL contexts, cover the affected object and identifier cases, use the repository mock extension where applicable or record why it cannot apply, and replace `times(1)` with plain `verify`.

### Coverage

- Candidate: public PR #39882, Code Correctness Review.
- Effective head: `5c5ea8bc04097599338aff374bb0b26ad0ce76bc`.
- Base: `master` at `86c1519e4d6deaf68e18ee47196e81da2e114f62`; merge base `101c3d56ccd7ee3c4a390a8f37550bed58a4751c`.
- GitHub’s authoritative 14-file list matched the merge-base comparison and remained unchanged after verification.
- All behavior clusters were reviewed: PostgreSQL capability activation; deferred enqueue and transaction lifecycle; table/schema/view/index refresh contracts; identifier handling; MySQL implicit-commit compatibility; unit/E2E coverage and registration; release notes.
- Previous feedback was re-evaluated: statement replay is removed; the implicit-commit rollback endpoint is fixed, but the refresh-timing regression remains.
- Root-cause, blast-radius/contracts, tests/runtime/operations, and final convergence passes completed. No unresolved evidence gaps remain.
- Mandatory style verification:
  - Command: `./mvnw -pl database/connector/dialect/postgresql,proxy/backend/core,test/e2e/operation/transaction spotless:check checkstyle:check -Pcheck -T1C`
  - Exit code: `0`; Spotless checked 26, 243, and 61 Java files respectively, and Checkstyle reported zero violations.
  - Covered all 12 changed Java files:
    - `database/connector/dialect/postgresql/src/main/java/org/apache/shardingsphere/database/connector/postgresql/metadata/database/PostgreSQLDatabaseMetaData.java`
    - `database/connector/dialect/postgresql/src/test/java/org/apache/shardingsphere/database/connector/postgresql/metadata/database/PostgreSQLDatabaseMetaDataTest.java`
    - `proxy/backend/core/src/main/java/org/apache/shardingsphere/proxy/backend/connector/ProxyDatabaseConnectionManager.java`
    - `proxy/backend/core/src/main/java/org/apache/shardingsphere/proxy/backend/connector/StandardDatabaseProxyConnector.java`
    - `proxy/backend/core/src/main/java/org/apache/shardingsphere/proxy/backend/connector/jdbc/transaction/DeferredMetaDataRefreshContext.java`
    - `proxy/backend/core/src/main/java/org/apache/shardingsphere/proxy/backend/connector/jdbc/transaction/ProxyBackendTransactionManager.java`
    - `proxy/backend/core/src/test/java/org/apache/shardingsphere/proxy/backend/connector/ProxySQLExecutorTest.java`
    - `proxy/backend/core/src/test/java/org/apache/shardingsphere/proxy/backend/connector/StandardDatabaseProxyConnectorTest.java`
    - `proxy/backend/core/src/test/java/org/apache/shardingsphere/proxy/backend/connector/jdbc/transaction/DeferredMetaDataRefreshContextTest.java`
    - `proxy/backend/core/src/test/java/org/apache/shardingsphere/proxy/backend/connector/jdbc/transaction/ProxyBackendTransactionManagerTest.java`
    - `test/e2e/operation/transaction/src/test/java/org/apache/shardingsphere/test/e2e/operation/transaction/cases/metadata/MySQLDDLMetaDataRefreshTestCase.java`
    - `test/e2e/operation/transaction/src/test/java/org/apache/shardingsphere/test/e2e/operation/transaction/cases/metadata/PostgreSQLDDLMetaDataRefreshTestCase.java`
  - `RELEASE-NOTES.md` and `test/e2e/operation/transaction/src/test/resources/env/e2e-env.properties` are not governed by the configured Java Spotless or Checkstyle scopes.
- Focused verification passed: 100 unit tests, zero failures/errors; the transaction E2E module compiled 61 test sources.
- Docker-backed E2E execution was not run; it is not outcome-sensitive to the code-proven blockers.
- CI/Actions were not reviewed. This is a code-scope result only; mandatory local style verification is complete.

rishi-rana:
Thanks, all three addressed in 3c86b0011aed3d8a4b0b8e52baad4b5dd1fed6bd.

**Scope of this revision**

This revision changes behaviour for one case only: table DDL executed inside a transaction on PostgreSQL. `CREATE TABLE`, `ALTER TABLE`, `DROP TABLE` and `RENAME TABLE` are now permitted there and their meta data refresh is deferred to the end of the transaction.

Everything else is unchanged from master:

- PostgreSQL schema DDL and view DDL in a transaction are still rejected with `TableModifyInTransactionException`.
- Every other dialect keeps its existing permission rules and refreshes meta data immediately, because `isSupportTransactionalDDL` is `false` for all of them.
- Behaviour outside a transaction is untouched for every dialect.

**2. Immediate refresh for implicit-commit dialects**

`DialectTransactionOption` gains `isSupportTransactionalDDL`, set only for PostgreSQL. The refresh is deferred only when that capability is set, so MySQL, MariaDB, Oracle, H2, SQL Server and Firebird refresh immediately as before and their meta data does not go stale while the logical transaction is still open.

`MySQLDDLMetaDataRefreshTestCase` now asserts from a second connection that the altered column is in meta data while the first connection's transaction is still open, and again after its rollback.

**1. Object type and identifier semantics**

Deferral is restricted to table DDL. `ProxySQLExecutor` permits `CREATE TABLE`, `ALTER TABLE`, `DROP TABLE` and `RENAME TABLE` in a transaction when `isSupportTransactionalDDL` is set, and PostgreSQL keeps `isSupportMetaDataRefreshInTransaction` as `false`, so schema DDL and view DDL remain rejected in a transaction exactly as before this PR. Those statements therefore never reach the deferred context, which is why it reloads tables only.

Identifiers are preserved: the deferred context stores the parsed `IdentifierValue` of each table and calls `ContextManager#reloadTable(ShardingSphereDatabase, String, IdentifierValue)`, so a quoted mixed-case table is reloaded as the quoted identifier rather than an unquoted one. `DeferredMetaDataRefreshContextTest#assertReloadKeepsQuotedIdentifier` covers this, and the connector test below uses a quoted table name.

**3. Reachable DDL path and test standards**

`assertExecuteDefersMetaDataRefreshForTransactionalDDLDialect` builds a real `AlterTableStatement` through `CommonSQLStatementContext`, so `PushDownMetaDataRefreshEngine#isNeedRefresh` returns `true` from production logic rather than from a stub, and the table identifiers come from the real `TablesContext`. The table name is quoted, so the test also covers the identifier case above.

The remaining construction mock is `DatabaseTypeRegistry`, used to supply the `DialectTransactionOption` under test. PostgreSQL's `DialectDatabaseMetaData` SPI is not on `proxy/backend/core`'s test classpath, so the dialect capability is isolated at that boundary rather than by adding a module dependency for the test; `ProxySQLExecutorTest` isolates the same collaborator the same way.

`times(1)` is replaced with plain `verify`.

**Verification**

- `./mvnw -pl proxy/backend/core,database/connector/dialect/postgresql test` - 593 tests, 0 failures.
- Transaction E2E on `mysql:5.7` - passed.
- Transaction E2E on `postgres:12-alpine` - passed.
- `spotless:check` and `checkstyle:check -Pcheck` across all 11 touched modules - 0 violations; `apache-rat:check -Pcheck` - passed.

**On the intended scope**

Two questions, since both go beyond the reported case and I would rather agree the boundary than assume it.

Should the deferral eventually cover the remaining metadata-changing statements rather than table DDL alone? Doing so needs a deferred representation carrying object type, and a view-level reload on `ContextManager`, which is why this revision stops at tables.

Should other backends with transactional DDL opt in as well? SQL Server looks like the notable case: it permits metadata-changing DDL in a transaction today through `isSupportMetaDataRefreshInTransaction`, and its DDL is transactional, so the stale meta data described in #35166 appears to apply to it as it did to PostgreSQL. openGauss rejects such DDL in a transaction today, so it is unaffected either way. I left `isSupportTransactionalDDL` as `false` for both rather than change dialects the issue did not report, but I am happy to set it where you think it belongs, in this PR or a follow-up.

rishi-rana:
CI status note, updated after the latest run.

All checks pass on `cfbefd6f88e856b80fbc6b19a29d04237e918567`.

Two earlier red checks on this pull request turned out to be unrelated to it, recorded here in case anyone reads back through the history:

- `E2E - SQL (Stage 2) (proxy, Cluster, MySQL, db_tbl_sql_federation)` failed with `Unsupported SQL operation: MySQL BIN function` while the federation E2E cases still called `bin(...)`. That inconsistency was fixed upstream and both federation legs now pass.
- `E2E - pipeline on mysql:5.7` timed out in `IndexesMigrationE2EIT.assertNoUniqueKeyMigrationSuccess`, with the migration importer losing its JDBC statement while ZooKeeper dropped its connection. It passed on a re-run of the identical tree, so it looks like an intermittent failure in that suite.

linghengqian (commented):
I have opened https://github.com/apache/shardingsphere/pull/39913 .

terrymanu (changes requested):
### Result

**Blocking Issues: 2**

The PR changes a published transaction limitation without a settled support boundary, and the new dialect gate also affects a documented mixed-database path.

### Blocking Issues

1. **Agree on the PostgreSQL transaction DDL contract.**  
   **Evidence:** The maintained [LOCAL and XA limitations](https://github.com/apache/shardingsphere/blob/cfbefd6f88e856b80fbc6b19a29d04237e918567/docs/document/content/features/transaction/limitations.en.md#L10-L25) say Proxy rejects metadata-changing PostgreSQL DDL in transactions. This PR [permits four table DDL types](https://github.com/apache/shardingsphere/blob/cfbefd6f88e856b80fbc6b19a29d04237e918567/proxy/backend/core/src/main/java/org/apache/shardingsphere/proxy/backend/connector/ProxySQLExecutor.java#L152-L167), while its [release note](https://github.com/apache/shardingsphere/blob/cfbefd6f88e856b80fbc6b19a29d04237e918567/RELEASE-NOTES.md#L75) describes DDL more broadly.  
   **Impact:** Users receive conflicting guidance about which DDL is supported in LOCAL and XA transactions.  
   **Discussion Needed:** Confirm the exact supported statement and transaction types, then align the published contract with that decision.

2. **Resolve the PostgreSQL-protocol/MySQL-storage boundary.**  
   **Evidence:** [DB Gateway documents](https://github.com/apache/shardingsphere/blob/cfbefd6f88e856b80fbc6b19a29d04237e918567/docs/document/content/features/db-gateway/limitations.en.md#L5-L8) PostgreSQL protocol over MySQL storage. Both the [permission gate](https://github.com/apache/shardingsphere/blob/cfbefd6f88e856b80fbc6b19a29d04237e918567/proxy/backend/core/src/main/java/org/apache/shardingsphere/proxy/backend/connector/ProxySQLExecutor.java#L152-L167) and [deferral decision](https://github.com/apache/shardingsphere/blob/cfbefd6f88e856b80fbc6b19a29d04237e918567/proxy/backend/core/src/main/java/org/apache/shardingsphere/proxy/backend/connector/StandardDatabaseProxyConnector.java#L275-L280) use the statement dialect, although routing [identifies the physical storage dialect separately](https://github.com/apache/shardingsphere/blob/cfbefd6f88e856b80fbc6b19a29d04237e918567/infra/rewrite/core/src/main/java/org/apache/shardingsphere/infra/rewrite/engine/RouteSQLRewriteEngine.java#L201-L209). MySQL 5.7 [implicitly commits `ALTER TABLE`](https://docs.oracle.com/cd/E17952_01/mysql-5.7-en/implicit-commit.html).  
   **Impact:** A PostgreSQL-protocol transaction routed to MySQL can commit table DDL physically while Proxy defers its metadata refresh until the logical transaction ends. Other sessions can see stale metadata. The new MySQL E2E case covers a MySQL-protocol session, so it does not exercise this path.  
   **Discussion Needed:** Decide whether transactional DDL support is limited by the routed storage type, including mixed routes, and define the required metadata timing for that boundary.

### Coverage

- Public PR #39882, Code Correctness Review; effective head `cfbefd6f88e856b80fbc6b19a29d04237e918567`, base `7ea8b8ce0fbe42472b185a4f45abf84ff278547f`. GitHub’s authoritative 27-file list was accounted for using an exact-head source snapshot.
- Reviewed the dialect-option changes; Proxy permission, execution, deferred queue, commit and rollback paths; unit and E2E tests; E2E registration; and release note. Rechecked the earlier statement-replay, direct-MySQL, and quoted-identifier objections. Root-cause, contract/blast-radius, test/runtime, and final convergence passes found no further independent blocker.
- **Local style verification:** `./mvnw -pl database/connector/core,database/connector/dialect/firebird,database/connector/dialect/h2,database/connector/dialect/mariadb,database/connector/dialect/mysql,database/connector/dialect/opengauss,database/connector/dialect/oracle,database/connector/dialect/postgresql,database/connector/dialect/sqlserver,proxy/backend/core,test/e2e/operation/transaction spotless:check checkstyle:check -Pcheck -T1C` exited `0`. Spotless selected Java files in all 11 modules; Checkstyle reported zero violations in each. The covered 25 PR-impact Java files were:
  - `database/connector/core`: `DialectDatabaseMetaData.java`, `DialectTransactionOption.java`.
  - `database/connector/dialect`: `FirebirdDatabaseMetaData.java`, `H2DatabaseMetaData.java`, `MariaDBDatabaseMetaData.java`, `MySQLDatabaseMetaData.java`, `OpenGaussDatabaseMetaData.java`, `OracleDatabaseMetaData.java`, `PostgreSQLDatabaseMetaData.java`, `PostgreSQLDatabaseMetaDataTest.java`, `SQLServerDatabaseMetaData.java`.
  - `proxy/backend/core` production: `ProxyDatabaseConnectionManager.java`, `ProxySQLExecutor.java`, `StandardDatabaseProxyConnector.java`, `DeferredMetaDataRefreshContext.java`, `ProxyBackendTransactionManager.java`.
  - `proxy/backend/core` tests: `ProxySQLExecutorTest.java`, `StandardDatabaseProxyConnectorTest.java`, `DeferredMetaDataRefreshContextTest.java`, `ProxyBackendTransactionManagerTest.java`, `BeginTransactionProxyBackendHandlerTest.java`, `CommitProxyBackendHandlerTest.java`, `SetAutoCommitProxyBackendHandlerTest.java`.
  - `test/e2e/operation/transaction`: `MySQLDDLMetaDataRefreshTestCase.java`, `PostgreSQLDDLMetaDataRefreshTestCase.java`.
- Checkstyle and Spotless are **Not Applicable** to `RELEASE-NOTES.md` and `test/e2e/operation/transaction/src/test/resources/env/e2e-env.properties` under the configured Java and workflow-file scopes.
- Focused reactor tests passed: 101 tests, zero failures. An initial non-reactor compile used a stale locally installed core artifact; the corrected `-am` run passed. Docker-backed E2E was not run. This is a code-scope result; CI was not reviewed.
- Jev: the typesafe-ai skill was not used for this deep code review. Jev MCP was not called because quota or balance was unconfirmed: 0 calls and judgments, no effect on routing, prioritization, conclusions, or plan. Efficiency cannot be measured accurately; problems: none found.

rishi-rana:
Both points are fair. Here is the contract I propose, so the boundary is agreed before I change anything further.

**1. Supported statements and transaction types**

Proposed contract:

- Permitted inside a transaction: `CREATE TABLE`, `ALTER TABLE`, `DROP TABLE` and `RENAME TABLE`.
- Transaction types: LOCAL and XA.
- Still rejected with SQLSTATE `0A000`, unchanged from today: schema DDL, view DDL, index DDL, and any other metadata-changing statement.
- Backends: PostgreSQL only. openGauss keeps the current restriction because it does not set the new capability.

If that is the boundary you want, I will align the published contract with it:

- `docs/document/content/features/transaction/limitations.en.md` and `limitations.cn.md`: narrow the LOCAL bullet so it states that only the four table DDL statements are supported for PostgreSQL and that other metadata-changing DDL is still rejected, and update the XA bullet that refers to it.
- `RELEASE-NOTES.md`: change "DDL" to "table DDL" so the entry matches the code.

One question inside this point. Should the contract be written as "PostgreSQL" or as "backends whose DDL is transactional"? SQL Server is the case that decides the wording: it permits metadata-changing DDL in a transaction today through `isSupportMetaDataRefreshInTransaction`, and its DDL is transactional, so the stale meta data in #35166 appears to apply to it as well. I have left `isSupportTransactionalDDL` as `false` for it in this PR, and I am happy either to keep the contract PostgreSQL-only or to include SQL Server here or in a follow-up, whichever you prefer.

**2. Protocol dialect against routed storage dialect**

You are right that both decisions currently read the statement dialect, and that this is wrong for DB Gateway. Proposed rule:

- Resolve `isSupportTransactionalDDL` from the routed storage unit types, through `database.getResourceMetaData().getStorageUnits()` keyed by each route unit's actual data source name, in the same way routing already resolves the storage dialect.
- Permit and defer only when every routed storage unit is a backend whose DDL is transactional.
- If any routed storage unit is not, including a mixed route, behave exactly as today: the permission decision uses the existing options and the refresh happens immediately. A PostgreSQL-protocol session routed to MySQL storage therefore behaves as a MySQL session does.
- Use that same resolution for the permission gate and the deferral decision, so the two cannot disagree about a given statement.

`ProxySQLExecutor#checkExecutePrerequisites` is called after `generateExecutionContext`, so the route context is already available at that point and only its signature needs to widen.

For coverage I would add unit tests for a PostgreSQL-protocol session routed to MySQL storage, and for a mixed route, asserting that DDL is rejected rather than deferred in both.

Please confirm or correct these two, and I will implement and update the documentation accordingly.

PR #39901, opened 2026-09-20
Fix validation of dynamic database rule updates

terrymanu:


terrymanu:
### Result

**Review Result: Mergeable**

The local candidate fully fixes the runtime configuration contamination and missing partial-update validation issues. No blocking issues were found.

### Evidence

- Both alter and drop now locate the current configuration, create an in-memory structural copy through the YAML swapper, modify the candidate, and refresh it within the same synchronized critical section. The processor no longer receives the live configuration.
- Rule-item Bean Validation still runs before the candidate configuration is modified, and the existing `SQLException` wrapping boundary remains unchanged.
- Non-empty partial updates call `DatabaseRuleConfigurationCheckEngine.check` before `partialUpdate` and `updateConfiguration`. A Checker failure prevents every runtime-state update.
- Empty-configuration deletion and non-partial rebuild behavior remain unchanged.
- Focused tests exited with code 0: all 16 tests passed.
- Exact-file Spotless and Checkstyle checks exited with code 0: all four target files were selected, Spotless reported them clean, and Checkstyle reported zero violations.

### Coverage

- Candidate type: standalone local working-tree candidate.
- Baseline: `515be4a9e38a21b1cff0604aab2b492ee8dc0678`.
- Reviewed scope: the two production files and two test files specified by the user.
- Behavior clusters: rule-item candidate isolation; complete partial-rule validation and failure short-circuiting.
- Root-cause, blast-radius, test/runtime-path, and final convergence reviews were completed. No unresolved evidence could affect the result.
- This is a code-scope conclusion only; remote CI was not reviewed.

taojintianxia (approved):

PR #39922, opened 2026-09-22
Refactor SQL Federation into pluggable Calcite and NONE providers

terrymanu:


terrymanu:
### Result

**Review Result: Mergeable**

The three-argument `AlterSQLFederationRuleStatement` constructor is restored; the four-argument constructor remains available.

### Evidence

- The compatibility overload delegates with a null `providerType`. A new regression test passes.
- Compiled-class inspection confirms both constructor signatures.
- The focused test, Spotless, Checkstyle, and Apache RAT checks pass.

### Coverage

Standalone local candidate against `7e2dae148026`, including the new test file. This is a code-scope review; CI and downstream commercial adaptation were not reviewed.

